PR #23, opened 2023-09-16
Nyano walls

Colin-Tel:
## About the PR

Added some Nyano walls, as well as ported over their damage reductions and the ability to construct shuttle walls.

I put as much of this into Nyano folders as I could. Every change made to upstream files *should* have a comment attached to it.


Colin-Tel:
It seems that we would need to port over code changes from the following PR in order to take full advantage of the "minimumDamageToDamage" thing.
 https://github.com/Nyanotrasen/Nyanotrasen/pull/934

PR #57, opened 2023-09-28
Add orders access level

DebugOk:
## About the PR
Require a new access level for approving orders (anyone can still place them)

This is granted by default to the Head of Logistics, the captain. On extended access, the logistic technicians will also get it. It can also be granted via an ID card console to any ID.

## Why / Balance
Discussed in Discord

**Changelog**
:cl:
- tweak: Approving cargo orders now requires the Orders access level


Colin-Tel (approved):
Changed a line that looks like it was missed. Looks good now!

DebugOk:
> Changed a line that looks like it was missed. Looks good now!

Whoops, missed that one. Thanks.

Shouldn't really matter thought since it only affects doors (I believe), and we dont have doors for the orders access

PR #64, opened 2023-09-29
Missing map stuff part 1

Colin-Tel:
## About the PR

Added a bunch of stuff needed to load maps. Some things are present in `migration.yml` instead, like TechFabs.


FluffiestFloof:
Just wondering why you moved the stuff outside the Signs folder as that's where they are if we mirrors the base game files

Colin-Tel:
> Just wondering why you moved the stuff outside the Signs folder as that's where they are if we mirrors the base game files

image

Because I saw it like this in the Delta V original repo and made the assumption that `Signs` would've been a redundant folder.

Colin-Tel (commented):
I looked through the stuff again and it looks alright to me

PR #66, opened 2023-09-29
Reimplement cargo pallet selling

DebugOk:
## About the PR
Why isn't it like this by default..?

If we want stations without shuttles, we cant just have these thing spitting out cash, since theres nothing you can actually do with it


DEATHB4DEFEAT (approved):
amazing

PR #75, opened 2023-10-01
Update changelog system to be datetime based and merge all files

DebugOk:
## About the PR
Datetime is much better for this because we'll never catch up to wizdens high changelog ids. Also readd the changelog merging code because otherwise ours doesnt show

### DIFF
--- a/Content.Client/Changelog/ChangelogManager.cs
+++ b/Content.Client/Changelog/ChangelogManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


DEATHB4DEFEAT:
I've just replaced wizden changelogs with our own for Parkstation, but this is a good idea

DebugOk:
> I've just replaced wizden changelogs with our own for Parkstation, but this is a good idea

I'm not 100% sure if we'll keep wizden changelogs in or start leaving the file empty, but this is how it was on Nyano. (Minus the datetime thing, that was something I did for delta)

Colin-Tel:
🙃

PR #80, opened 2023-10-02
Adds PDA boxes and adjusts PTech stock

Colin-Tel:
## About the PR

A very MRP update.

Adds six different PDA boxes and places them in appropriate lockers. Each PDA box has a variety of six department PDAs (except for the prisoner PDA box, which is for the warden).

This PR also adjusts the PTech inventory so it makes more sense with what we're going for. The non-service encryption keys were commented out, while various service PDAs were added (with some having more in supply than others 🤡💦).

Also comments out the service headsets in the PietyVend since Chaplain isn't in the Service department.

## Why / Balance

This allows head roles more freedom to hire and promote crewmates within their department. Senior roles, while they are not available to join as, are instead able to be designated by the head of the department (granted, there's only *one* Senior PDA in the box where applicable).

All upstream file changes were commented appropriately.

## Media

image

**Changelog**

:cl:
- add: Added PDA boxes to department heads' lockers. Bug your manager for a promotion now!


Colin-Tel:
Holding off on merging until the mantis PR is in

PR #84, opened 2023-10-02
Prevent /suicide usage

DebugOk:
Cancels the event before it actually kills you, but still allows the event to go out in case any systems require it to stop other actions.

This'll just make it act like running /ghost 99% of the time.


DEATHB4DEFEAT (approved):
I wonder how Nyano did this, never tried finding it, is this what they did?

PR #99, opened 2023-10-04
A .38 special pistol for the mantis

FluffiestFloof:
## About the PR

A 5 round pistol that uses .38 special.
Fixes a bunch of small issues and missing stuff from previous .38 things.

## Why / Balance

There was a demand (by Colin) to have the Mantis have mindbreaker gun. Why not a small low ammo pistol with a marble grip that uses a very special kind of ammo that is hard to acquire :). 

## Technical details

buh gun
Mantis will probably need their own access since every item they usually had was under Detective access.
They'll probably also need all their old lockers and doors to be redone.

## Media

The one on the right. This is mostly to show the design compared to what we have.
ss (2023-10-04 at 04 00 27)

The ammo, which is the pistol one but more compact
ss (2023-10-04 at 04 00 10)

## Breaking changes

**Changelog**

:cl:
- add: Added a pistol in .38 for the Mantis.


DebugOk:
I'm sure it'll be fine

PR #104, opened 2023-10-05
Psionic Mantis access

FluffiestFloof:
## About the PR

Gives mantis their own access as their locker contains high risk items (a knife that is a steal objective, and probably a gun maybe)
I don't know if they'd need their own doors? Could easily be done if needed by inheriting normal science door and making it mantis only.

## Why / Balance

This also gives them back a flashlight, remove the spare clothes except the fez to mimic the detective closet. Probably outside this PR scope but eh.

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: Mantis gets their own access.


DebugOk:
Might have to merge master, which its not letting me do from the web interface

PR #113, opened 2023-10-05
Mediborg

FluffiestFloof:
## About the PR

So this re-adds mediborgs but extremely downgraded since it follows the same logic as the normal wizden borg.
This give them the bare minimum to do their job (health analyser) and they can create candies.

The PR additionally re-adds their old laws.

I'm sure people that enjoyed the old mediborg won't like having no tools on spawn but it give them something to bother epi for. The research is only Tier 1.
If required, they can always start with more modules if we want, maybe the treatment module. (Limits is 3 total for borgs)

## Media
ss (2023-10-05 at 11 55 11)

- [NO] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**


Colin-Tel (approved):
😄

PR #117, opened 2023-10-06
Remove upstreamed signs

FluffiestFloof:
Will undraft when they merge or close if they don't want the signs.
Mostly deletes the Law and Court sign and move stuff that wasn't correctly placed.

Yes this branch was meant to be the logidoor, I got sidetracked.


FluffiestFloof:
Was upstreamed successfully, merge whenever we get there so the mappers don't explode.

DebugOk:
This good to merge?

FluffiestFloof:
I don't think we caught up to that specific commit yet, so probably after/during the next sync

PR #119, opened 2023-10-06
Paralaxes

FluffiestFloof:
## About the PR

I think I got them all?


Resources/Prototypes/Nyanotrasen/Parallaxes/space.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- type: parallax
+  id: ShipwreckedTurbulence1
+  layers:
+    - texture:
+        !type:ImageParallaxTextureSource
+        path: "/Textures/Nyanotrasen/Parallaxes/Vaitarna.png"
+      slowness: 0.93
+      scale: 1,1
+      scrolling: "-0.006, -0.066"
```
DebugOk, line 67: What do we need this for? We won't implement shipwrecked

DebugOk (changes requested):

PR #120, opened 2023-10-06
Some nyano spawners

FluffiestFloof:
## About the PR

Velcro said it would need to be re-added but cleaned up a bit.

## Media

## Breaking changes

**Changelog**


IamVelcroboy:
> Velcro said it would need to be re-added but cleaned up a bit.

They might be fine. Just now is a good time to clean them up if they need to be. but otherwise, should be good.

Colin-Tel:
YAML Linter failing, seems to cite missing mob prototypes.

FluffiestFloof:
Right so the missing ones are NPC version of Xenos, aka non ghostroles versions. Do we want to re-add those or directly use the normal xenos similar to the ones used in the vent event?

IamVelcroboy:
We need the altered ones for Arena. I'll take a look in a min.

Colin-Tel (approved):

PR #122, opened 2023-10-07
Scrubs for Medical Interns

Colin-Tel:
## About the PR

I'm blue da ba dee da ba die

## Why / Balance

Interns look cute in their blue scrubs, don't tell me they don't.

## Media

image


FluffiestFloof:
Just rename them to "nurse" at this point

PR #137, opened 2023-10-09
Replace Salvage Jumpsuit sprite instead of adding a new item

FluffiestFloof:
## About the PR

Mostly to prevent two items with same name.


Colin-Tel (approved):
Salvage never looked so gritty

PR #139, opened 2023-10-09
Update HoP accesses

Colin-Tel:
## About the PR

Wouldn't it be funny if we didn't change this


FluffiestFloof:
hopcurity removed? smh

PR #158, opened 2023-10-11
Adds tilewindow cmd for mapping

IamVelcroboy:
- Ports command from Nyano
- I also alphabetized the tags


DebugOk:
If TileWindow is a Nyano command, it should be in that namespaces

IamVelcroboy:
I moved it. Do I actually need to add "Nyanotrasen" to the namespace as well?

Colin-Tel (approved):

PR #166, opened 2023-10-13
Paramedic Airlocks & SSU

rosieposieeee:
## Why / Balance

Because WD are fools who cannot handle the sheer power of the Paramedic having their own room.

Also adds Paramedic suit storage.

:cl:
- add: Added Paramedic airlocks.


Colin-Tel:
Reminder that the ID card console and access configurator handle accesses differently, so you need to make sure that you add the access in `access_configurator.yml` so that engineers can adjust door accesses.

I've got it, so don't worry about it.

PR #169, opened 2023-10-13
Remove Nyano deathgasps

DebugOk:
Doesn't re-add the glimmer-spesific gasps yet, and leaves the localization files intact for whenever someone adds that


Colin-Tel:
I'm not really knowledgeable about any of this so I'll defer to a coder's judgement.

DEATHB4DEFEAT (approved):

PR #172, opened 2023-10-13
Changed ghost role rules

Colin-Tel:
## About the PR

Changed some of the ghost role rules so they're more obvious to players. Players should be able to get a general idea of what is or isn't appropriate behavior.

## Why / Balance

Players will sometimes do stuff uncharacteristic of a role and get confused when an admin points it out. Players will sometimes think they are forced to behave in a certain way as roles and report players who do not do these things when playing as those roles.

## Media

image
image

## Breaking changes

**Changelog**


Colin-Tel:
Let me know what ghost roles need clarification with their rules.

Colin-Tel:
Gonna go ahead and merge since I don't think anyone's got any ideas.

PR #175, opened 2023-10-14
Harpy Customization Pack 1

VMSolidus:
## About the PR
This is the first of the promised content PRs for harpies, and also includes some bugfixes. 

## Why / Balance
-5 new wing patterns are added. Two Tone, Three Tone, Speckled, Undertone, and Feather Tips
-The HUD bug has been fixed
-Sparrow Tail has been removed for REASONS. 
-Harpies can now choose what instrument they want to imitate(From within a list of options). Presumably this means multiple harpies can start a band together! Right click yourself to select what instrument you'll mimic. 
## Media

New customization
swappableharpies


Colin-Tel:
image
The "Phoenix Tail" seems to be the default tail, the only reason a player would add it again is if they want to change the color of the tail. It comes across as a bug-at-first-sight, which I think could be addressed by just calling it "Basic Tail" much like the "Basic Wings"

Colin-Tel:
image
There's also missing localization strings when it comes to *all* harpy markings.

evilexecutive:
I've made the requested changes.

Content.Server/DeltaV/Harpy/HarpySingerSystem.cs
DebugOk, line 1: Why subscribe to the same event twice

DebugOk (changes requested):

Content.Server/DeltaV/Harpy/HarpySingerSystem.cs
evilexecutive: Let's hope the rogue semicolon doesn't strike again

DebugOk:
I still like how the customization code makes markings flesh colored by default. Flesh wings!

DebugOk:
> image There's also missing localization strings when it comes to _all_ harpy markings.

This is a bug:tm: unrelated to harpies (probably)

I also saw it on moths

evilexecutive:
>This is a bug™️ unrelated to harpies (probably)
>I also saw it on moths

Oh it's not, check the localization files on this PR for the fix. You can also apply that to moths

DebugOk (dismissed):
Code looks good enough, the rest Colin can review

Content.Server/DeltaV/Harpy/HarpySingerSystem.cs
```diff
@@ -0,0 +1,28 @@
...
+using Content.Shared.Mobs.Systems;
+using Robust.Server.GameObjects;
+
+namespace Content.Server.DeltaV.Harpy
+{
+    public sealed class HarpySingerSystem : EntitySystem
+    {
+        [Dependency] private readonly InstrumentSystem _instrument = default!;
+        [Dependency] private readonly MobStateSystem _mobState = default!;
+        public override void Initialize()
```
DEATHB4DEFEAT, line 12:
```suggestion
        [Dependency] private readonly InstrumentSystem _instrument = default!;
        [Dependency] private readonly MobStateSystem _mobState = default!;


        public override void Initialize()
```

Content.Server/DeltaV/Harpy/HarpySingerSystem.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            base.Initialize();
+
+            SubscribeLocalEvent<InstrumentComponent, MobStateChangedEvent>(HarpyStopSinging);
+        }
+        //Immediately closes the Midi UI window if a Singer is incapacitated
+        private void HarpyStopSinging(EntityUid uid, InstrumentComponent component, MobStateChangedEvent args)
+
+        {
```
DEATHB4DEFEAT, line 21:
```suggestion
        }


        //Immediately closes the Midi UI window if a Singer is incapacitated
        private void HarpyStopSinging(EntityUid uid, InstrumentComponent component, MobStateChangedEvent args)
        {
```

DEATHB4DEFEAT (commented):
muh formatting

Colin-Tel (approved):
🐦

PR #176, opened 2023-10-15
Update Autoborg.yml

VMSolidus:
## About the PR
Wizden gave us a list of random borg names that are 95% LRP/NRP. Through a colaborative effort with the DeltaV Discord voice chat, we have come up with a list of replacement names that fit the Culture naming conventions. As a rule, they should be short, whymsical, inside joke names. 

## Why / Balance
I Blame The Clown

### DIFF
--- a/Resources/Prototypes/Datasets/Names/autoborg.yml
+++ b/Resources/Prototypes/Datasets/Names/autoborg.yml
@@ -1,7 +1,7 @@
 - type: dataset
   id: names_autoborg
   values:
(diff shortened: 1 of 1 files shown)


TadJohnson00:
I dunno who was in the voice chat, who was consulted for approval on these names, or when this apparent meeting was held, but it looks like we're going from one list of LRP names to another list of LRP names.

Carolyn3114:
> Wizden gave us a list of random borg names that are 95% LRP/NRP

all of these names still seem similar in terms of "roleplay", just with different references

DebugOk:
Will not be accepted in it's current state, and no similar prs will be accepted without proper discussion

IamVelcroboy:
image

PR #182, opened 2023-10-16
Chemvend Jugs Adjustments (Again)

rosieposieeee:
## About the PR

Based on these calculations from Discord user Seafoam Green. Fixes my untested numbers in the previous PR that were too high.

I do worry this is too close to the original numbers, but the math is here. I would like to talk more about these numbers as a community before anything like this is merged.

I do have some concerns:

- Spaceacillin is left off of this.
- Tranex is left out of this. If you're wondering why I care about cillin or tranex, its because those are how you make medicated sutures. Not necessary to have in the calcs, but worth bringing up.
- Sigy is left off, but Seafoam said there was enough leftover for about 400u of it.
- I'm not sure this leaves enough for botany. Should probably calc that.
- Saline is left out. Not that important, but sodium is left out of the numbers entirely. I just set it back to 2.
- Tricord is left out of this calc, which is dylo and inapro. Nobody makes inapro for anything other than tricord anyway so that's accounted for at least.
- Dylo is also used in mindbreaker, which isn't calced here. Maybe not important, but Dylo is in high demand as a stepping stone for other drugs, maybe worth an extra jug for its supplies.
- All of this uses current-branch numbers. **_I have no idea what the rebase numbers or recipes are._** All of this needs testing.

image
image

Secondary concerns:
- Are the vends too slow? Can/should we speed them up?
- Some stations have two chemvends in the main medbay's chemistry lab, some only have one. Some have them in the extra medbays, which are often wrenched to the main one. Should we make mapping guidelines that account for that? What's the best practice?
- What should we ACTUALLY be aiming for medbays to start with stock for? 20 25u pills of each important drug? We want them to be restocking, but how often? Our goals should be clearly stated if we're going to make more tweaks.

## Media

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_vending.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_vending.yml
@@ -104,7 +104,7 @@
     sprite: Objects/Specific/Service/vending_machine_restock.rsi
     state: base
   product: CrateVendingMachineRestockChemVendFilled
-  cost: 6300 #Delta V - was 3820, see rebase PR #53
+  cost: 1 #Delta V - was 3820, see rebase PR #53
   category: Medical
   group: market
 
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
@@ -4,24 +4,24 @@
     Jug: 4
 ## Delta V - begin jug adjustments, see rebase PR #53
     JugAluminium: 2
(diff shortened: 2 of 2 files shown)


Carolyn3114:
could just be all 3 for simplicity? also note that pills are inefficient in the rebase due to moths being unable to eat them

Carolyn3114:
> could just be all 3 for simplicity? also note that pills are inefficient in the rebase due to moths being unable to eat them

also different overdose numbers, for example, 10 units of bicaridine is an overdose

DebugOk:
Closing as stale

PR #183, opened 2023-10-16
Increase default lobby duration

DebugOk:
Go from 2 minutes and 30 seconds to 4 minutes. We already use these values anyways, this just saves a line in the server config


Colin-Tel (approved):
😄 ⏱️

PR #185, opened 2023-10-16
Metempsychotic machine that definitely works.

PHCodes:
## Media

## Breaking changes

Makes my body scanners break on every server instance I run. Hoping to fix this by restarting. 
**Changelog**


PHCodes:
This is theoretically all of the code and it's working as intended when I actually clone someone. 

BUT.

The Physics in my local machine appear to be broken, even on clean builds. I don't think the damage is caused by anything in this PR. If someone else could try building this in their machine and letting me know what happens I would be very appreciative. Note that there is a chance this will break your physics forever like it did mine.

This PR might just be cursed.

DebugOk:
Still waiting on upstream to fix the scanners before this can be properly tested

DebugOk:
The scanners have been fixed, if you resolve the merge conflicts and make sure this works I can go test it

PR #191, opened 2023-10-17
Role whitelists

DebugOk:
## About the PR
1984 update real!

## Technical details
Instead of the way the old system did it via setting a "isWhitelisted" in the prototype, this instead uses a JobRequirement, meaning it can be added onto anything that requires them.

This does mean that anything using `TryRequirementsMet` will have to also pass in the players whitelist status. Clients can get it via the `IsWhitelisted` function from `JobRequirementsManager`, and the server can run
```cs
var isWhitelisted = player.ContentData()?.Whitelisted ?? false;
```
to get it, with player being the IPlayerSession belonging to the player.

## Breaking changes
This breaks any code that uses TryRequirementsMet it is updated to pass in the whitelist status, but this should always be possible with a one line change.

**Changelog**
:cl:
- add: Certain roles require whitelist now. These being the captain, head of security and nukie commander.


DebugOk:
The urge to say "Fuck it, ship it"

DEATHB4DEFEAT (approved):
Cool way to do that.

DEATHB4DEFEAT:
> The urge to say "Fuck it, ship it"

:shipit:

Issue #180, opened 2023-10-15
Harpy Bugs


evilexecutive:
I'm going to write this issue here because it's not really something I have the skill to address, and Harpies are spread across multiple PRs due to the scope anyway. In this case Major bugs are significantly gamebreaking, while their minor bugs are just inintended things that may annoy players(or me as their creator).

**Major Bugs**

- [x] - Harpies that are actively playing a Midi will keep the Midi player open upon death, stunning, or any other form of incapacitation.  A dead harpy with the midi player open can continue to play music, and presumably alert someone to their location. 

Therefore if you kill a harpy while they are singing, they will continue to sing even while dead. I've partially fixed this in #175, but my fix only addressed them being able to open the menu while dead. 

**Minor Bugs**
- [ ] - Harpies can't sing while handcuffed. 
- [ ] - Harpes can sing while muzzled. 
- [ ] - Harpy feather colors default to their skin tone. 
- [ ] - When Carry is added, Harpies can be played like a guitar by anyone who picks them up
- [ ] - Anyone can right click a harpy and tell them what instrument to mimic, not just the harpy herself

evilexecutive:
>  - Harpies that are actively playing a Midi will keep the Midi player open upon death, stunning, or any other form of incapacitation. A dead harpy with the midi player open can continue to play music, and presumably alert someone to their location.
Fixed by #175

PR #199, opened 2023-10-20
Another harpy bugfix

evilexecutive:
## About the PR
This PR provides a fix for #180 4th bug, concerning harpies defaulting their colors to skin tone when randomly generated(Such as via the Metempsychosis machine). I've accomplished this by setting the three default markings to use the same color as the character's hair, with a fallback of plain brown if the character is bald. Hopefully this will solve the problem of Metempsychosis harpies being "Very fucking ugly". 

## Media
image


DebugOk:
I better not see any more bugfixes >:(

~~For legal reasons thats a joke... Or is it?~~

PR #200, opened 2023-10-20
Deep-Fried Crispy Shader

PHCodes:
## About the PR

Added the Crispy shader back.

## Why / Balance

Because I missed it when readding the Deep-Fryer. 

## Technical details

I added an entry to Delta-V's shader.yml. I added a shader file. 

## Media

image

- [X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

None. 
**Changelog**


DebugOk:
I'll assume you tested it, otherwise we'll just see with the bug reports

PR #203, opened 2023-10-20
Update migration with temporary changes

IamVelcroboy:
## About the PR
Updates the Migrations file.

## Why / Balance
Most of these are temporary until proper entities are added. But will help mappers updating their maps.


Resources/migration.yml
```diff
@@ -145,6 +145,23 @@ CrateMedicalDefib: CrateMedical
 LeftLegSpider: null
 RightLegSpider: null
 WeaponShotgunEnforcerNonLethal: WeaponShotgunEnforcerRubber
+FloorPlanetAsphalt: FloorAsphalt
```
DebugOk, line 148: You cant migrate tiles afaik :clueless:

IamVelcroboy: Oh, I swear I'd seen that already. Let me dig...

IamVelcroboy: hey, look what I found...
image

PR #212, opened 2023-10-20
Update Edge

Colin-Tel:
## About the PR

Updated Edge with a Deep Fryer and an Electric Grill, and set the minimum players to 10. 

## Media

The map renderer won't give me a render.


Colin-Tel:
Note to self: check brig timer connections to doors

DebugOk:
> Note to self: check brig timer connections to doors

Do you still want to check this before its merged?

Colin-Tel:
> > Note to self: check brig timer connections to doors
> 
> Do you still want to check this before its merged?

yes :3

PR #226, opened 2023-10-23
Another harpy colors

evilexecutive:
## About the PR
Just a small change that makes it so that the default markings for harpies will also default to match their hair color. This prevents randomly generated(And Metempsychosis) Harpies from having skin color feathers. 

## Media

image


evilexecutive:
I'm so fucking confused how this got made. Ignore this PR

PR #297, opened 2023-10-25
Fixes issue #269 - MindSwap can now go back.

### LINKED ISSUE #269: One you mindswap, you cannot go back
Copy pasted from the bug reports channel.
## Description

There is a bug Adei discovered on one of the Rebase rounds, attached to a story in ⁠screenshots-and-stories. Basically, when you mindswap, you cannot mindswap back. Adei and Terry (both bartenders in a Revolutionaries round) found this out the hard way, Terry being the one who used the ability. They mindswapped, when they tried to Reverse Mindswap, it did not work, when they used normal mindswap, it did not work. 

**Reproduction**

- Have two people, one is the mindswapper.
- Walk up and mindswap.
- Test if you can mindswap back into your body, and if you cannot (using reverse mindswap and just the mindswap ability - both were applicable in this situation,) then that's the bug.

**Screenshots**

**Additional context**

PHCodes:
I guess I tested it only with a Urist McHands. Whoops. Fix is kinda jank, explained in comments.
fixes #269

## About the PR

Fixed MindSwap between players return action doing nothing. 

## Why / Balance

Defect. It was broken.

## Technical details

Basically... You have to unattach the minds before you can switch them, now. I did this by transferring the minds to nothing, which according to TransferTo's comments, is HOW you're supposed to unattach them. Doesn't make a lick of sense to me and doesn't feel like that code will be necessary eventually, but it is what it is. 

## Media

Ginyumemes

## Breaking changes

None.
**Changelog**

:cl: JJ
- fix: Fixed Mindswap return action between players.

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
@@ -165,7 +165,11 @@ namespace Content.Server.Abilities.Psionics
             if(!_mindSystem.TryGetMind(target, out var targetMindId, out targetMind)){
                 targetMind = null;
             };
-
+            //This is a terrible way to 'unattach' minds. I wanted to use UnVisit but in TransferTo's code they say
+            //To unnatch the minds, do it like this. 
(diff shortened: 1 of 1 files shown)


FluffiestFloof:
fixes #269

PR #303, opened 2023-10-25
Recipe Cookbook in the Guidebook.

PHCodes:
## About the PR

I added the Recipes entry to the Guidebook, so a player has a fighting chance at knowing what recipes are in the game without needing to deep dive the codebase. 

## Why / Balance

The wiki is not currently up to date with the recipes in the Rebase and this makes them easily available to players.

## Technical details

It's an XML file and some changes to some YML. That's it.

## Media

Cookbook

## Breaking changes

N/A
**Changelog**

:cl:
- add: Added Recipes to Guidebook.

### DIFF
--- a/Resources/Locale/en-US/deltav/guidebook/guidebook.ftl
+++ b/Resources/Locale/en-US/deltav/guidebook/guidebook.ftl
@@ -1,0 +1,1 @@
+guide-entry-recipes = Recipes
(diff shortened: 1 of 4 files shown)


DebugOk (changes requested):
This should be using the search bar, and would be a lot better if it used the auto generation like bartending and chemistry does

PHCodes:
It is auto-generated, but by the perl script from the previous PR. I'm not including that with this to keep this PR clean. I'd say both of those are 'nice to have' features and things we can iterate over. At the moment there is no way to access what half of these recipes even are without the codebase, so having literally anything available gives players a fighting chance at engaging with systems (like Butlertron, cargo bounties) that depend on these things right now.

deltanedas:
theres a proper autogeneration pr upstream

DebugOk:
https://github.com/space-wizards/space-station-14/pull/20978 does exactly what changes I requested, but its awaiting a fix/review

PR #316, opened 2023-10-26
Rename and refit brigmedic

TadJohnson00:
Renames brigmedic to corpsman, fits them out in new drip, everything should be commented a-okay and working properly.

## Media

## Breaking changes

**Changelog**


Resources/Locale/en-US/job/job-names.ftl
Colin-Tel, line 1: Do comments work like this on `.ftl` files? I thought they didn't.

TadJohnson00: Good catch

PR #318, opened 2023-10-26
Double bladed energy sword tweaks

Tryded:
Lower Desword's Price from 20 TC to 16 TC
Desword has 100% reflection chance for Energy Projectiles but none for ballistics
I was forced to not do this on Wizden due to them saying that the desword is too OP then saying it's too weak.
## Why / Balance

A lot of jobs couldn't afford double bladed energy swords as syndie and the main threats to a desword user is a disabler most of the time.
## Media

## Breaking changes

**Changelog**


Resources/Prototypes/Catalog/uplink_catalog.yml
```diff
@@ -92,7 +92,7 @@
   icon: { sprite: /Textures/Objects/Weapons/Melee/e_sword_double.rsi, state: icon }
   productEntity: EnergySwordDouble
   cost:
-    Telecrystal: 20 #(Originally 16)
+    Telecrystal: 16 #Originally Was 20 but no one bought it due to it's high cost
```
Colin-Tel:
```suggestion
    Telecrystal: 16 # DeltaV - Originally 20 but no one bought it due to it's high cost.
```
Remember to cite that the change is a DeltaV change in the comment.

PR #319, opened 2023-10-26
Cargoscrubs

VMSolidus:
Apparently cargo was never able to order Medical Scrub crates, thus making all the fancy new scrub colors unobtainable on any map that didn't have them mapped. 

:cl: VMSolidus
- add: Cargo can now order medical scrub crates


Colin-Tel (approved):
Tested, it works.

PR #324, opened 2023-10-26
Add access wire to airlocks

JustTNE:
## About the PR

Re-added access wires to airlocks

## Why / Balance

They were not re-added since rebase and are a very nice to have feature.

## Technical details

This is only a yaml change since upstream has all the code this used to use already included.

**Changelog**

:cl:
- add: Airlocks now have access wires

### DIFF
--- a/Resources/Prototypes/Wires/layouts.yml
+++ b/Resources/Prototypes/Wires/layouts.yml
@@ -4,6 +4,7 @@
   - !type:PowerWireAction
   - !type:PowerWireAction
     pulseTimeout: 15
+  - !type:AccessWireAction # Nyanotrasen
   - !type:DoorBoltWireAction
   - !type:DoorBoltLightWireAction
   - !type:DoorTimingWireAction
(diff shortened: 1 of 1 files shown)


Carolyn3114:
This makes hacking easier which is kinda lame

JustTNE:
This kind of implies that hacking is supposed to be difficult in the first place, which it's not. This is just how it was on pre-rebase delta-v, and I've personally made good use of that in some role play scenarios before as well as allowing maintenance drones to link things to doors they don't have access to for example.

Carolyn3114:
considering how insuls are supposed to be rare, and that wires were randomized because of this change *on nyanotrasen and delta v*, I'd say access wire is poopy

JustTNE:
The wires are still randomized on delta-v, that hasn't changed.

DebugOk:
At this time, I do not believe adding back access wires will be beneficial in any way. Doors can still be hacked open with ease, assuming you have a multitool and insulated gloves.

PR #336, opened 2023-10-27
Remove most syndicate maints loot

DebugOk:
## About the PR
Its just dumb, and shouldn't be spawning on its own


AitorLogedo:
why the ballon :(

PR #343, opened 2023-10-28
Forcefully reset invalid job preferences

DebugOk:
## About the PR
This'll cause all invalid preferences to be reset upon being loaded. This should prevent people from being able to ready up as a job they shouldn't have access to if it was set to high prior to being locked.

## Why / Balance
Apparently this was an issue that exists

**Changelog**

:cl:
- tweak: Invalid job preferences are now reset to never upon loading in.


DebugOk:
Fuck it, we ball

PR #359, opened 2023-10-29
Adjusted roles

Colin-Tel:
## About the PR

Command roles that have `canBeAntagonist: true` have it set to `false`. Janitor gains service access because for some godforsaken reason they didn't have it already.

## Why / Balance

We want to hold command to a higher standard but also allow traitor command that isn't held to this expectation. This results in *more* poor command members that neglect their staff, regardless of whether or not they have a gameplay-related reason to do it.

**Changelog**
They'll figure it out.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Cargo/quartermaster.yml
+++ b/Resources/Prototypes/Roles/Jobs/Cargo/quartermaster.yml
@@ -23,7 +23,7 @@
   startingGear: QuartermasterGear
   icon: "JobIconQuarterMaster"
   supervisors: job-supervisors-captain
-  canBeAntag: true # DeltaV - Can be antagonist
+  canBeAntag: false
   access:
   - Cargo
   - Salvage
--- a/Resources/Prototypes/Roles/Jobs/Civilian/janitor.yml
+++ b/Resources/Prototypes/Roles/Jobs/Civilian/janitor.yml
@@ -8,6 +8,7 @@
   supervisors: job-supervisors-hop
   access:
   - Janitor
+  - Service # DeltaV - quite honestly the most noble service occupation
   - Maintenance
   special:
   - !type:GiveItemOnHolidaySpecial
--- a/Resources/Prototypes/Roles/Jobs/Civilian/service_worker.yml
+++ b/Resources/Prototypes/Roles/Jobs/Civilian/service_worker.yml
@@ -10,6 +10,7 @@
   icon: "JobIconServiceWorker"
   supervisors: job-supervisors-service
   canBeAntag: true # DeltaV - Can be antagonist
+  antagAdvantage: 1 # DeltaV - Reduced TC: Accesses
   access:
   - Service
   - Maintenance
--- a/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
+++ b/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
@@ -24,7 +24,7 @@
   icon: "JobIconHeadOfPersonnel"
   requireAdminNotify: true
   supervisors: job-supervisors-captain
-  canBeAntag: true # DeltaV - Can be antagonist
+  canBeAntag: false
   access:
   - Command
   - HeadOfPersonnel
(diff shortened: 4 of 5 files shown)


FluffiestFloof:
Should Boxer get Theater instead of Service?
Probably same for Martial Artist

Colin-Tel:
> Should Boxer get Theater instead of Service? Probably same for Martial Artist

Doesn't the BoxingDrobe have a service access requirement?

FluffiestFloof:
> > Should Boxer get Theater instead of Service? Probably same for Martial Artist
> 
> Doesn't the BoxingDrobe have a service access requirement?

it does

DangerRevolution:
Spoke about this at length in the discord and BC I feel kind of strongly about this and some people agreed I just wanted to share my thoughts if you didn't mind.
To preface it: I think it makes no sense for Heads to be revs, but Syndies is a different case.

I really believe that having Syndie command staff has created really interesting scenarios, unique gameplay and really shook up the RP and Scenarios and possibilities; and had Sec actually investigating command staff and more likely to respond to calls about command staff doing illegal things (which they shouldn't anyways but you get the gist)
Just an abstract from a ramble off discord:
> - The thought that, you aren't immune as command staff was INTERESTING. In these shifts people had to INVESTIGATE COMMAND STAFF IF THEY WERE SUSPICIOUS. You can't meta that command can't be bad guys.
> - This allowed MORE damage and MORE skill from good syndie players and MORE RP and MORE unique scenarios, who could leverage their skills and their extra access & status to greater effect.
> - The MAIN ISSUE was that people AREN'T used to not trusting command staff. I still had to get to grasps with it like when Jamal tricked me and CE into early launching shuttle to secure his objectives.
> - I think there is a problem with the command double standard, but removing it outright limits risk, RP, interesting rounds and unique situations; i.e Jamal being a Syndie HoP with my backup ID. He could STILL be caught but it was a UNIQUE situation and interesting RP for Jamal, Sec and the Syndies he helped.

I grasp the concept that there's a general weariness towards giving Admins more things to police; I don't know the magic solutions to problems but I really feel like this hamstrings a lot of potential RP and unique situations and interactions, as well as giving more room for Syndies who already struggle; and given recently Sec and the server itself has not been understaffed it's not easy for them.

It's quite late; I hope this made sense but I really just wanted to voice this as some people shared the same sentiment:
I understand it creates a paradoxical situation but there should be some more thought to how to navigate around it, I would personally suggest extrapolating the rule of "command held to higher standards" into Syndie command too. Obvious Syndies are frowned upon as is but it's straight up LRP and bwoinkable for command staff; their goal should be to blend in as much as possible.
I think the issues of "abandoning their post so some roles are soft locked" and "being able to shout code words over comms at round start" are still problems but I really hope it's problems that can be negated or minimised in order to maximise the fun & uniqueness of the opportunity Syndie command provide.

Sorry to text dump, lots of love,
Danger! Revolution!

Colin-Tel:
I don't doubt there are interesting RP moments that can be had when Command roles can be antags, but the problem is that nobody likes command neglecting their duties and staff, which is exactly what letting them be antags does. 

Obviously not all traitor CMOs are going to leave the medbay for the doctors to figure out, but it makes it a helluva lot more likely. We want to expect more from Command on the station, and we want them to set a standard for the crewmates now more than ever (**especially** considering the huge influx of people in the server).

DangerRevolution:
Would reusing "higher standards for command staff" to further push for Syndie commands to blend in not alleviate this problem somewhat?

I appreciate that during the influx there needs to be teachers on staff but I *really* think this is a really interesting idea that shouldn't just be shut out. Kicking the bucket down the road and band-aiding it could be okay.

Colin-Tel:
> Would reusing "higher standards for command staff" to further push for Syndie commands to blend in not alleviate this problem somewhat?

Maybe it would, but I don't forsee that happening in most cases at the moment. From an admin perspective, it would be way easier to guarantee that command should focus on their department by not making them antags than to idealize and potentially enforce that traitorous command should do it too with rules potentially too vague to matter.

> I appreciate that during the influx there needs to be teachers on staff but I _really_ think this is a really interesting idea that shouldn't just be shut out. Kicking the bucket down the road and band-aiding it could be okay.

I still agree with you that it's an interesting idea; I've never said it wasn't. Don't interpret this as shutting it out. It's easy enough to decide on it again at a later time.

PR #361, opened 2023-10-30
Re-add Lean

IamVelcroboy:
## About the PR
- The reagent was added from Nyano but wasn't added to the actual drink.
- I tweaked the recipe for "realism" and moved to DV folder
- Also adds a recipe for Diphenhydramine, which is an ingredient for Lean

## Why / Balance
I was very disappointed the other day when I was given a cup of lean and it "tasted like water"

## Technical details
n/a

## Media
n/a

**Changelog**
n/a


Leander-0:
I like this PR.

IamVelcroboy:
Re-add Leander

PR #362, opened 2023-10-30
Tweaks Vector

Tryded:
Makes vector use .35 auto instead of .45 magnum
Makes Vector have 8 rounds per second
## Why / Balance

Wizden never updated the vector since they replaced it with the drozd. This fixes that by making vectors use .35 auto instead of .45 magnum since .45 magnum deals 35 damage per cartridge. Also it is a temporary fix until mappers remove vectors and replace them with drozds
## Media

## Breaking changes

**Changelog**


Tryded:
I am obviously willing to make the firerate lower

Colin-Tel (changes requested):
Ok so this thing SHREDS. 3 sheets of steel for a magazine for this unbelievable murder machine.
The Drozd is just this but way worse. There needs to be some kind of real reason to choose one over the other. And yeah, you'll need to lower the firerate; it's ridiculous.

Tryded:
> Ok so this thing SHREDS. 3 sheets of steel for a magazine for this unbelievable murder machine. The Drozd is just this but way worse. There needs to be some kind of real reason to choose one over the other. And yeah, you'll need to lower the firerate; it's ridiculous.

I'm going to lower the firerate to 8 rounds per second, however the problem about making these two guns different is that the drozd was meant to replace the vector however our maps still use them. This might just be a mapping issue.

FluffiestFloof:
It might be simpler to just replace all our mapped vectors with drozd. Except if we really want to keep them

Tryded:
> It might be simpler to just replace all our mapped vectors with drozd

This shall be a temporary fix I guess until drozd are mapped in

PR #363, opened 2023-10-31
Changes Head of Personal's job prioritization.

Adrian16199:
A change so that it doesnt prioritize everything as much as they would be a captain like one on job list selection before the round starts.

## About the PR

A single number

## Why / Balance

Because its weird to roll HoP over captain or any other head on High.
## Technical details

A number change

## Media

## Breaking changes

**Changelog**

-tweak: Changes prioritization on job list.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
+++ b/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
@@ -19,7 +19,7 @@
       time: 72000 # 20 hours
     - !type:OverallPlaytimeRequirement # DeltaV - Playtime requirement
       time: 90000 # 25 hours
-  weight: 10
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
```diff
@@ -19,7 +19,7 @@
       time: 72000 # 20 hours
     - !type:OverallPlaytimeRequirement # DeltaV - Playtime requirement
       time: 90000 # 25 hours
-  weight: 20
+  weight: 10
```
FluffiestFloof: Where comment? :trollface: 
The usual `# DeltaV - Changed HoP weight from 20 to 10 due to them not being more important than other Heads` or something like that

PR #370, opened 2023-10-31
Prevent labelling felinids

DebugOk:
## About the PR
Apparently this is a thing, which is really funny until people apply some questionable words.

### DIFF
--- a/Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
+++ b/Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
@@ -8,6 +8,7 @@ using Content.Shared.Item.PseudoItem;
 using Content.Shared.Storage;
 using Content.Server.Storage.EntitySystems;
 using Content.Server.DoAfter;
+using Content.Shared.Tag;
 using Robust.Shared.Containers;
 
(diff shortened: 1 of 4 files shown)


Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
```diff
@@ -122,19 +128,19 @@ public bool TryInsert(EntityUid storageUid, EntityUid toInsert, PseudoItemCompon
             return false;
 
         var item = EnsureComp<ItemComponent>(toInsert);
+        _tagSystem.TryAddTag(toInsert, PreventTag);
```
DEATHB4DEFEAT, line 131: Why is PseudoItem the one adding the tag?

DebugOk: Because it'd apply to anything else that is a pseudoitem aswell. Non-items shouldn't be labeled, and as pseudoitem adds an itemcomponent to allow inserting into storage, stuff that shouldn't be labeled can be labeled

DEATHB4DEFEAT: Alright

PR #401, opened 2023-11-05
Removes starvation damage from upstream species

### LINKED ISSUE #331: Starvation
## Description
Is it really worth keeping?

We never had it before, and it might be best to just get rid of it again

FluffiestFloof:
Fun fact, Vulpkanin never had any issue with starvation since they don't have starvationDamage :trollface: 

## About the PR

Removes the starvation damage.

## Why / Balance

While it's logical and realistic to literally starve, a lot of people seems to not like it.
I say skill issue, just eat but eh.

## Technical details

Comments out a couple things in basegame files.
closes #331 

## Media

## Breaking changes

**Changelog**

:cl:
- remove: Removed starvation!

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/arachnid.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/arachnid.yml
@@ -12,11 +12,11 @@
     species: Arachnid
   - type: Hunger
     baseDecayRate: 0.0125 # Spiders have slow metabolisms all things considered, so I decided to just make their hunger drain slower.
-    starvationDamage:
-      types:
-        Cold: 0.5
-        Bloodloss: 0.5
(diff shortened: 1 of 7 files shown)


Adrian16199:
It may be a skill issue, however I think the issue is the fact that the amount of ssd people can get so many and its hard to maintain all of em.

Adrian16199:
also felinid not takin damage? They do tho, i did almost die to starvation and i only play as felinid.

FluffiestFloof:
> also felinid not takin damage? They do tho, i did almost die to starvation and i only play as felinid.

oh ye right, they inherit from human so does harpy
Doesn't matter much for the goal of this PR

DebugOk:
I love it

PR #405, opened 2023-11-06
Move cat ears out of the Syndicate Uplink into AutoDrobe

FluffiestFloof:
## About the PR

Moves the cat ears out of the Syndicate Uplink and into AutoDrobe. Two units of cat ears. (Up to discussion)

## Why / Balance

Leo wants this.

## Technical details

N/A

## Media

N/A

## Breaking changes

**Changelog**

:cl:
- tweak: Cat Ears have been moved from the Syndicate Uplink into the AutoDrobe.


DebugOk:
I'm not sure about this, but I won't argue with Leo's decision. If its somehow abused it'll be reverted

PR #413, opened 2023-11-07
Set max head revs to 1

Colin-Tel:
## Why / Balance

After going over this idea in the Discord meeting a few days ago, it was met with unanimous agreement.

## Technical details

I changed an integer.

**Changelog**

:cl:
- tweak: Revolutionaries are now fewer in number.


Adrian16199:
I doubt its possible but you could technicly code it that it becomes two at like 40 people because i feel like in a server full of people (aka 100), I feel like one is not enough.

DebugOk (dismissed):
This is a gamerule variable, add it there instead.

Colin-Tel:
> This is a gamerule variable, add it there instead.

Should be fixed with this most recent commit, if I understood you correctly.

PR #426, opened 2023-11-08
Expand valid character names

Samsterious:
## About the PR
This PR brings over an updated version of [Delta-v-Legacy: Adjust character name validation #14](https://github.com/DeltaV-Station/Delta-v-Legacy/pull/14).

The following has been changed from the original PR:
- Numeric digits have _not_ been removed in order to accommodate things like IPCs and other silicon characters.
- `,` and `.` are permitted because some people's names necessarily include affixes such as `Name, Sr.` or `Name, Jr.`.
- `'` is permitted to accommodate certain Diona names such as `The Bird's Song` and similar.

## Why / Balance
This change was lost during the rebase, and several characters have had their names stripped of non-ASCII letters as a consequence.

## Technical details
The permitted Unicode blocks are documented below the regex string that controls the allowed glyphs in a character's name.

## Media
image

**Changelog**
:cl:
- tweak: Character name restrictions are more permissive.

### DIFF
--- a/Content.Shared/Preferences/HumanoidCharacterProfile.cs
+++ b/Content.Shared/Preferences/HumanoidCharacterProfile.cs
@@ -407,7 +407,16 @@ namespace Content.Shared.Preferences
             var configManager = IoCManager.Resolve<IConfigurationManager>();
             if (configManager.GetCVar(CCVars.RestrictedNames))
             {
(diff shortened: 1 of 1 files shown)


VMSolidus:
Oh thank god Farida Shi'Berud can finally have the apostrophe in her name.

PR #430, opened 2023-11-09
Laika is Smart Doggo

VMSolidus:
## About the PR
Have you ever seen a Labrador open a door? If Laika's smart enough to be a ghostrole, she'd probably be one of those smart dogs that can jump up and touch a button on the door to let herself out. This doesn't let her have any access, just that she can open powered airlocks that don't require an ID. Presumably she jumps up and bumps the door's push button

## Why / Balance
Oh god it's so painful not being able to go anywhere. 

**Changelog**
:cl: VMSolidus
- tweak: Laika has discovered how to press airlock buttons, now no evildoers are safe.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
@@ -48,3 +48,4 @@
     tags:
     - CannotSuicide
     - VimPilot
+    - DoorBumpOpener


TadJohnson00:
Uhhh fwiw as Laika's mum/dad I vote in favour of this change

Colin-Tel (approved):
Sounds good to me, I've seen Laika get stuck in public halls more often than not, so this should be a nice change.

PR #453, opened 2023-11-18
Salvage Map Rebase

IamVelcroboy:
## About the PR
- Ports and updates Nyano salvage maps
- Updates Wizden maps and re-saves in Delta V directory
- Removes all small maps
- Adds a few new maps 

## Why / Balance
Make salvage happy

## Technical details
n/a

## Media
n/a

## Breaking changes
n/a

**Changelog**
:cl:
- add: Salvage has been rebased. New but strangely familiar space debris will be pulled in by the salvage magnet. Caution is recommended.


VMSolidus:
Hey do you mind if we also make spawners for much more reasonable "Ultra high risk" salvage loot? Some of these maps have vaults with turrets, where all they guard is an empty ammo duffel bag, and a t3 part. The most extreme example being the Atlas maps that have a 2x turret vault behind 3 reinforced walls, which solely contains a single retro laser gun.

A lot of the Nyano maps had disproportionally low reward vs risk, which made them not worth doing.

IamVelcroboy:
I think it's very reasonable to continue to adjust these maps and balance risk/reward. I'd say that could be a follow up PR after these are merged, feel free to take a look at them and suggest/make changes.

VMSolidus:
Yea, I'll agree that we should, but as a separate PR. I also think we can run some experiments to find the average value of an Asteroid(assuming you sell all the refined ore), and get that as a comparison for a typical wreck. But obviously the wreck rewards should be proportional to risk.

On average, I think "Risky" wrecks should award more total loot value than asteroids, since thats the natural tradeoff for the risk of death. And that super high risk wrecks(assaulting a guarded vault, fighting any kind of boss battle), should have unique loot potential instead of just valuable loot

PR #467, opened 2023-11-21
Remove brig access for lawyer

IamVelcroboy:
## About the PR
-Completely removes brig access as it is redundant to lawyer
-Adds appropriate airlocks and windoors
-Sets migration to prevent map borks 

## Why / Balance
Avoid confusion/remove redundancy 

## Technical details
n/a

## Media
n/a

## Breaking changes
Migration should prevent any isues

**Changelog**
n/a


Colin-Tel:
Should have updated ~~all~~ some of the maps to have their lawyer rooms with lawyer access doors. Also made some minor adjustments to some other things

*changed because of the two pending map PRs

PR #471, opened 2023-11-23
Implementing Height Sliders

VMSolidus:
## About the PR
This PR adds the new Height Slider feature from Cosmatic Drift, coded by @LankLTE from Wizden.

## Why / Balance
I've spent about 4 hours playing in the new Cosmatic Drift server, and this feature was too awesome not to port. As an side, Gunnar McDorf now finally gets to be short. 

## Technical details
Height references:
Humans, Vulpakin, Diona, Arachnid, Moth, Reptillian, Slimes: 4'6" to 7'
Felinids: 3' 10" to 4' 4"
Harpies: 4'8" to 5' 2"
Oni: 7'3 to 8'

All of our DeltaV species have their own custom minimums and maximums. For the sanity of our playerbase I've kept the Min/Max values within reasonable ranges, and are much more restrictive than their implementation in Cosmatic Drift.

## Media
image

## Breaking changes
**Changelog**
:cl: LankLTE & VMSolidus
- add: Height Sliders have now been ported to DeltaV


DebugOk (changes requested):
Sounds like sprite squishing with extra steps. Doing this so it can't be merged until I've had a look at it

LankLTE:
Scale width / height is true by default, so you don't need to worry about defining it for each species.

Resources/Prototypes/Entities/Mobs/Species/dwarf.yml
```diff
@@ -5,13 +5,13 @@
   id: BaseMobDwarf
   abstract: true
   components:
-  - type: Hunger
-    # starvationDamage: # DeltaV - Disable starvation damage
-    #   types:
-    #     Cold: 0.5
-    #     Bloodloss: 0.5
+  #- type: Hunger # DeltaV - Disable starvation damage
```
DebugOk, line 8: Why is this being removed

Resources/Prototypes/Entities/Mobs/Species/dwarf.yml
```diff
@@ -62,6 +61,3 @@
   parent: BaseSpeciesDummy
   id: MobDwarfDummy
   noSpawn: true
-  components:
```
DebugOk, line 65: No comment

Resources/Prototypes/Entities/Mobs/Species/dwarf.yml
```diff
@@ -25,7 +25,6 @@
   - type: Sprite
     noRot: true
     drawdepth: Mobs
-    scale: 1, 0.8
   - type: Body
```
DebugOk, line 28: No comment

error.txt
DebugOk, line 1: Why does this exist

Content.Client/Humanoid/HumanoidAppearanceSystem.cs
```diff
@@ -30,6 +34,11 @@ private void UpdateSprite(HumanoidAppearanceComponent component, SpriteComponent
         UpdateLayers(component, sprite);
         ApplyMarkingSet(component, sprite);
 
+        var speciesPrototype = _prototypeManager.Index<SpeciesPrototype>(component.Species);
```
DebugOk, line 37: Too many changes without any comments to add a review on all

DebugOk (changes requested):

LankLTE:
> Sounds like sprite squishing with extra steps. Doing this so it can't be merged until I've had a look at it

Not sure exactly what you mean by this, but it's modified the same way as species are currently (uniformly modifying the scale on the SpriteComponent) so I don't see why it'd be a huge deal

DebugOk:
> > Sounds like sprite squishing with extra steps. Doing this so it can't be merged until I've had a look at it
> 
> Not sure exactly what you mean by this, but it's modified the same way as species are currently (uniformly modifying the scale on the SpriteComponent) so I don't see why it'd be a huge deal

Scaling one axis but not the other. I personally think it looks horrible and don't want to see it on player characters

LankLTE:
> Scaling one axis but not the other. I personally think it looks horrible and don't want to see it on player characters 

I agree it looks terrible, and for that reason it's disabled by default. It's only supported for species like dwarves to be able to exist as they do currently (upstream). This PR doesn't apply it to any species- they're all scaled uniformly.

Resources/Prototypes/Entities/Mobs/Species/dwarf.yml
deltanedas: branch troll probably

Adrian16199:
Im gonna give my cookie on it, I like the idea if the minimum height would be human size on all species besides harpies and felinid only because they are smaller by default. I would love to see bigger humans even if it would cost them bigger hitbox just for a roleplay thing, I would not accept this if it could allow a human to be even smaller in any sort of way for powergaming needs.

Adrian16199:
Which brings me a question..... due to size bein kinda of a factor regardin if you can easily shove someone or not, does this apply here? I would love to see a video regardin the highest human tryin to shove a normal size one.

error.txt
VMSolidus: Small mistake I chalk up to coding at 2 AM in the morning. When building, I run dotnet build > error.txt because I actually like being able to read and/or search the errors when it finishes.

VMSolidus:
So MassContest doesn't actually care about what a Species scale is, which is a common misconception because all three of the species we have with different mass also have a different scale. MassContests actually check a species FixturesComponent, which lists a Density stat. This PR doesn't change anyone's Density stat, so no matter what you scale an Oni or a Felinid to, they still have the same mass for the purpose of their shoving/dragging. 

Per discussion I'm going to open up the Maximums a little for some of the species, and tighten up the minimums for all the basic humanoids except for Human(who we want to keep solely because it'll get the dwarf fans to stop complaining that we don't have dwarfs)

VMSolidus:
Damnit there are so many scuffed changes to the core files due to how I merged them(manually moving files from one repo to another) that I basically need to start over with them from scratch and merge them CORRECTLY. That way I can see our changes and make sure we keep what we need to keep, and comment what came from @LankLTE

VMSolidus:
Yea so I'm going to close this, because I straight up don't have the skill needed to maintain a feature like this, and it touches entirely too much Wizden code for my or Debug's comfort. I am miserly with touching Wizden code, and I feel bad if I have to change one or two lines. This changes nearly 3000 lines of Wizden code.

PR #472, opened 2023-11-24
Allow nuclear operatives to buy deswords

Tryded:
## About the PR

allows nuclear operatives to buy deswords.
## Why / Balance

Our deswords can only reflect energy stuff, and the original reason desword was removed on nukie uplink upstream was because of it's original reflection. You'll still die to a shotgun if security decides to shoot you.

This is also more interesting than just using a fire axe to constantly pry up tiles and space stuff.
## Technical details

Just commented out the original code that disallowed it from nukie uplink
## Media

## Breaking changes

**Changelog**

:cl: tryded
- add: Added double bladed energy swords to Nuclear Operative Uplinks

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -95,11 +95,11 @@
     Telecrystal: 16 # DeltaV - Originally 20 but no one bought it due to it's high cost.
   categories:
   - UplinkWeapons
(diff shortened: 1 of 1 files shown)


Tryded:
I literally cannot upstream this because on wizden they still have the 75% reflection for all projectiles and hitscan

DebugOk:
I will revert this if this proves to be an issue or something that is abused

PR #493, opened 2023-11-30
Add Crusher Glaive to the Salvage Vendor

AeraAuling:
## About the PR
Added two Crusher Glaives to the salvage vendor

## Why / Balance
Added because Crusher Glaives are nice, basically. They are otherwise fairly identical to regular crushers. Readding them to the vendor should also allow for some visual variety and a bit of added flavor with salvage's choice of weaponry.

## Technical details
This touches the file at /Resources/Prototypes/Catalog/VendingMachines/Inventories/salvage.yml otherwise not much else is changed.

## Media

image

## Breaking changes

N/A
**Changelog**

-Adds Crusher Glaives to the salvage vendor, finally bringing proper polearms to the delta sector.


Resources/Prototypes/Catalog/VendingMachines/Inventories/salvage.yml
```diff
@@ -10,6 +10,7 @@
     HandheldGPSBasic: 2
     RadioHandheld: 2
     WeaponCrusher: 2
+    WeaponCrusherGlaive: 2
```
Colin-Tel:
```suggestion
    WeaponCrusherGlaive: 2 # DeltaV
```

Colin-Tel (changes requested):

Colin-Tel:
Why isn't this already in? Is it because of balance?

Adrian16199:
> Why isn't this already in? Is it because of balance?

I believe the reason is the fact that it is basicly the crusher but diferent look.
No diferent stats.

IamVelcroboy:
Are these items supposed to be research locked?

IamVelcroboy:
Or...should they be?

FluffiestFloof:
> > Why isn't this already in? Is it because of balance?
> 
> I believe the reason is the fact that it is basicly the crusher but diferent look. No diferent stats.

They heal more damage when you hit a marked enemy
Crusher:
https://github.com/DeltaV-Station/Delta-v/blob/2fa1a61913e4e819edde09973e00a6d2936fcf88/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml#L30-L33
Glaive:
https://github.com/DeltaV-Station/Delta-v/blob/2fa1a61913e4e819edde09973e00a6d2936fcf88/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml#L105-L108

They have a faster melee attack speed
Crusher:
https://github.com/DeltaV-Station/Delta-v/blob/2fa1a61913e4e819edde09973e00a6d2936fcf88/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml#L46-L51
Glaive:
https://github.com/DeltaV-Station/Delta-v/blob/2fa1a61913e4e819edde09973e00a6d2936fcf88/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml#L111-L113

Colin-Tel:
> Or...should they be?

I'd be more inclined to say that these should be research locked rather than have these available at the same time the objectively worse equipment is as well.

AeraAuling:
> > Or...should they be?
> 
> I'd be more inclined to say that these should be research locked rather than have these available at the same time the objectively worse equipment is as well.

I suppose they could be researched too yeah. I, did not know about the more health but I will say the slightly faster attack speed doesn't play out that much in practice. They attack at about the same rate when using the shot and strike.

DebugOk:
> Why isn't this already in? Is it because of balance?

I don't think these should be in until Wizden adds them, especially not as something available roundstart

deltanedas:
since its better in every way i think it should be in salvage tech, and probably upstreamed

PR #479, opened 2023-11-25
New Playable Species: Lamia

VMSolidus:
## About the PR
This is a draft PR for an upcoming new playable species, the Lamia. Lamia are an older species from the 2022 era of Nyanotrasen, and were previously abandoned code that was dropped from the game on October 13th, 2022. I was able to locate what I believe to be the last remaining branch containing Lamia, from a defunct server repository that ran an October 12th, 2022 build of Nyanotrasen. Thus I began a project to refurbish this code for use on modern SS14. 

The Lamia I wish to PR are only recognizable from the original Lamia in that they share the Tail Segment System. That is, they prominently feature a completely unique mechanic whereby their body is composed of multiple entities linked together in a chain. The original version of this code had a great many bugs related to the game's physics system, and it was severely limited by the Robust Toolbox engine at the time. In the year since Lamia's abandonment, Robust Toolbox has gone through many iterations and improvements, and has deprecated large parts of the physics system that old Lamia utilized. 

## Redesigns from the original 2022 Lamia, AKA Nyanotrasen Lamia. Vs. DeltaV Lamia
The original Nyanotrasen Lamia were made with several limitations in mind. Their size was heavily restricted by the physics engine at the time, we aren't beholden to that same limitation anymore. Thus DeltaV Lamia are vastly larger than the Nyanotrasen Lamia, featuring a tail that is 5 tiles in length, with fully functional physics collisions. They were also not able to wear Hardsuits due to limitations of the SpriteComponent, and thus were instead designed around having a "Barotrauma resistance". On DeltaV code, we can arbitrarily state that species use different optional sprites for items, therefore its possible to have for example a Nukie Hardsuit, with its equipped-outerwear state, equipped-outerwear-lamia, equipped-outerwear-lamiainitialsegment, equipped-outerwear-lamiasegment, and so on. The Lamia Segments can simply state that if they equip a hardsuit, they utilize the -lamiasegment sprite option. Therefore its no longer necessary to create an entire new item solely so that snakes can wear a hardsuit. 

**Positive Traits**
- Extreme Size. Lamia are 5 tiles long, and weigh as much as a car. They make a mockery of mass contests, and they can push physics objects around simply by slithering into them.
- Significantly larger health bar. A Lamia is put into critical condition at 200 damage, and dies at 300 damage. This is offset by the Lamia having a certain percentage of damage taken by the tail transferred to the main body.
- High resistance to forced movement. Space Wind at standard pressure cannot move them.
[...]

PR #499, opened 2023-12-02
Harpy Rebalance & Content Update 2

VMSolidus:
## About the PR
This is both a rebalance, and content update for Harpies. I made this PR primarily as a means of learning how to build Generic Visualizers, which is a skill I require in order to complete PR #479.

## Why / Balance
Harpies on-release were missing any mechanical positives. I had at one point intended to make their mechanical positive an increase in movement speed, but at the time did not have access to what I wanted to counterbalance it with. Since Wizden last updated Arachnids, they added in an organ datafield that would let me do this. 

In exchange for having the higher base speed, they're picking up a 2x multiplier on breath frequency. Harpies use their extremely high performance Avian lungs to both utilize a Syrinx for speaking, and to power their incredibly light body. The downside of this is quite noticeable for space station life, as a Harpy involved in a spacing needs to constantly monitor their oxygen supply, and will always be worried about where they're going to get their next air refill. Salvage Techs and Nukies are the most affected by this from what I can predict. In both cases, a Harpy Salvager or Nukie will need to strongly consider not spacing their "Workspace".

In testing, I had also determined that Piercing is actually a slight downgrade from Slashing. While resistances to it are a little less common, it also doesn't deal ongoing bleed damage. Security also starts with 50% Piercing resistance, making their armor quite effective against Harpy talons. Still, I have included this as a matter of flavor rather than function, since it helps distinguish the species more. And of course, we have entirely too many species that deal slashing damage melee. 

In my testing for their weight changes, I have determined that it takes a standard Human 2 to 4 shoves to push down a Harpy. I tried counting how many shoves it takes a Harpy to push down a Human, and gave up after 65 shoves didn't even budge the human. 

**Positive Traits**
- Somewhat smaller than humans, but larger than Felinids. 
- Uses their talons to deal Piercing damage
- Can imitate around 70% of the game's sound library through a huge list of voice emotes
- They can "Sing" midis by imitating instruments. Right click yourself to select an instrument
- Has a slightly higher base sprint speed than humans, 5.0 vs. 4.5(an 11% increase), walk speed is unaffected.

**Negative Traits**
- Extreme low density, weight set to 90. Harpies weigh less than even Felinids, and exactly half that of a Human.
- Takes 15% more Blunt, Slash, and Piercing damage
- Breathes **TWICE** as much air as a Human. All air tanks last half as long for a Harpy. Beware your 4 minute emergency tanks, and 10 minute full size tanks. 

## Technical details
This adds an additional pair of checks to the HarpySingerSystem to incorporate a generic Visualizer. Harpies have a sprite layer for their musical notes that is visible: false by default. When a Harpy opens their innate Midi Player, the generic visualizer toggles visibility on their musical notes layer to true. When they close their midi player(or if it is forced to close), it sets the layer visibility to false.

## Media
HarpyMusicNotes

**Changelog**
:cl:
- add: Harpies now have a visual indication when they're playing a Midi.
- tweak: Harpies have been re-balanced so that they now actually have a numerical positive.


Resources/Prototypes/DeltaV/Entities/Mobs/Species/harpy.yml
```diff
@@ -97,20 +100,28 @@
...
   - type: MeleeWeapon
     soundHit:
-      collection: Punch
+      collection: AlienClaw
     animation: WeaponArcClaw
     damage:
       types:
-        Blunt: 1
-        Slash: 5
+        Piercing: 5
```
DebugOk, line 107: Piercing is used for guns, not melee. It was also removed from Felinids for this reason and shouldn't be added as a melee damage type.

DebugOk:
> (Any midi player actually, not just their innate one, the game has no way of telling),

There should be? If not that might be something worth implementing.

> When they close their midi player, it sets the layer visibility to false.

When *they* close it? What about it being forced closed via a shove, going crit or whatever else forced it to close?

VMSolidus:
>There should be? If not that might be something worth implementing.

It turns out no, Harpies activate their midi player using the same event as every other source of midi players. I attempted to add a second event to their midi player, but it turns out that adding a second event to an action is impossible, the game simply won't trigger any events after the first one for a given action. Furthermore, there are no actions in the game that utilize more than one event. 

>When they close it? What about it being forced closed via a shove, going crit or whatever else forced it to close?

A misunderstanding, HarpySingerSystem works by forcing the midi player closed, AND preventing it from being re-opened if any of these stop conditions are met. So checking for when the midi player gets closed is appropriate here.

DebugOk:
You could do instead is add a new component thats added and removed by HarpySingerSystem, and check for that instead. You can subscribe to the ComponentStartup and ComponentShutdown events. It'll add a bit of complexity, but it wont trigger the singing sprites for anything that isn't the actual harpy sing action

DebugOk:
If you cant trust OnInstrumentOpen passing, that'd mean mime harpies cant use *any* instruments

VMSolidus:
>You could do instead is add a new component thats added and removed by HarpySingerSystem, and check for that instead. You can subscribe to the ComponentStartup and ComponentShutdown events. It'll add a bit of complexity, but it wont trigger the singing sprites for anything that isn't the actual harpy sing action

>If you cant trust OnInstrumentOpen passing, that'd mean mime harpies cant use any instruments

Okay I have no idea what you're talking about, Mime harpies can still use instruments, they just don't get to use their innate midi player. Their innate action is blocked, but they can still play instruments because they're using an instruments midi player and not their own. HarpySingerSystem is checking for an entity with the HarpySingerComponent that opens a midi player, all the HarpySingerComponent is doing is tricking the server into thinking Harpies are an instrument item, 
image

Also this is all now a moot point because I just now realized the mime isn't playing the musical notes with the guitar. Whelp, I guess we're in the clear on that one.

VMSolidus:
image

DebugOk:
> > You could do instead is add a new component thats added and removed by HarpySingerSystem, and check for that instead. You can subscribe to the ComponentStartup and ComponentShutdown events. It'll add a bit of complexity, but it wont trigger the singing sprites for anything that isn't the actual harpy sing action
> 
> > If you cant trust OnInstrumentOpen passing, that'd mean mime harpies cant use any instruments
> 
> Okay I have no idea what you're talking about, Mime harpies can still use instruments, they just don't get to use their innate midi player. Their innate action is blocked, but they can still play instruments because they're using an instruments midi player and not their own. HarpySingerSystem is checking for an entity with the HarpySingerComponent that opens a midi player, all the HarpySingerComponent is doing is tricking the server into thinking Harpies are an instrument item, image
> 
> Also this is all now a moot point because I just now realized the mime isn't playing the musical notes with the guitar. Whelp, I guess we're in the clear on that one.

If that doesn't catch, then cant you just add code to OnInstrumentOpen that adds a new component to the harpy when it has the ui opened via the action?

VMSolidus:
Untitled video - Made with Clipchamp (6)

No what I'm saying is that I was mistaken in assuming this would be a problem, and apparently the problem I thought would be a thing, doesn't actually exist, and that the feature works as intended. Here is a recording of the functionality.

Issue #328, opened 2023-10-26
breaching hammer isn't usable as a crowbar


Carolyn3114:
## Description
you cannot breach with breaching hammer

**Reproduction**
try to pry open door
cry

**Screenshots**

**Additional context**

Carolyn3114:
also, there is no cooldown for wielding/unwielding it

Tryded:
> also, there is no cooldown for wielding/unwielding it

Think that's intended now, however I probably could make it act like a crowbar

Carolyn3114:
> > also, there is no cooldown for wielding/unwielding it
> 
> Think that's intended now, however I probably could make it act like a crowbar

it's not intended, there is supposed to be a cooldown

Adrian16199:
This issue should be closed.

Colin-Tel:
bumping that this should be closed

PR #505, opened 2023-12-03
Allow breaching hammer to breach in like JoL.

Adrian16199:
## About the PR

I have allowed breaching hammer to enter departments.
## Why / Balance

To not make such a mess when a pesky syndicate is in the department that you cant access.
It was a nyano thing that it was capable of entering departments. I believe it was a good thing compared to waiting on an engineer to hack into the department or breaking every single window/wall/door with it when captain is in danger.
Also fixes the issue #328 
## Media

## Breaking changes

**Changelog**

-tweak: Breaching hammer now can breach into departments.


DebugOk:
Doesn't really make sense imo but eh

Adrian16199:
I mean, only reason why I though about it is because of it havin that function in nyano, it maybe be true that it indeed could make zero sense. Thats it. There is really no but to it.

PR #528, opened 2023-12-07
Prisoners require whitelist

DebugOk:
This is what happens when people cant behave with a certain role and only play it to break the rules. Might be temporary, or might be permanent. Who knows?

**Changelog**
:cl:
- tweak: Due to abuse, prisoners now require whitelist to play.


deltanedas:
wtf i thought prisoner means i be a shitter :trollface:

PR #541, opened 2023-12-11
Carp fillet's are now Sushi-Grade

ps3moira:
## About the PR

Removed the carpotoxin's from sashimi.

**Changelog**

:cl:
- remove: Removed Carpotoxin from Sashimi, now they are sushi-grade!
- tweaked: Tweaked Sashimi's description to reflect that it's no longer poisonous.


jzburda:
I thought it was pretty fun-e to see that carpotoxin was always in carp meat even when prepared, but sounds good to me

Colin-Tel (approved):
I made a similar change back on Nyano and it was lost, so this is fine.

PR #546, opened 2023-12-11
Bags of holding recipe tweak

UnicornOnLSD:
## About the PR
Modified the base cost of the three bag of holding, namely:
-Increased plasma cost from 750 to 1250
-Reduced bluespace crystals cost from 5 to 3

## Why / Balance
Five crystals is, in both my personnal opinion and the people I asked a bit too much, considering the only source is the oracle and even if made to be inconsistent to keep the bag's rarity, would be better off reduced, with of course a counter balance through the increased plasma cost.

Point is overall, to make the bags not a thing you unlock when playing EPI and at most keep to yourself but rather something that is, hard to make but not too hard, so namely salvagers could get their hands on it too as this is something I haven't seen once despite them being basically the back bone of EPI's fund, most of the time.

**Changelog**

:cl:

- tweak: Tweaked bags of holding cost to be more consistent.

### DIFF
--- a/devices.yml
+++ b/devices.yml
@@ -1,0 +1,189 @@
+- type: latheRecipe
+  id: TimerTrigger
+  result: TimerTrigger
+  completetime: 2
+  materials:
+    Steel: 300
+    Plastic: 200
+
+- type: latheRecipe
+  id: SignalTrigger
+  result: SignalTrigger
+  completetime: 2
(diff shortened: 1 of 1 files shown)


UnicornOnLSD:
fucked up code, boutta fix and make a new one :D

PR #549, opened 2023-12-12
Felinids are in your plants: The reckoning.

Adrian16199:
Due to implementation of not being able to shoot in any storage being extremely well implemented, it automaticly applies to this, allowin this to be an amazin gimmick like the duffelbag.

Code made by: Leander from Nyano.

## Why / Balance

To make the cat in bag mechanic be more than just haha cat in duffelbag.

## Media

## Breaking changes

**Changelog**

:cl:
- add: Felinids now can be in plants.
- tweak: Felinids are now size 100

### DIFF
--- a/Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemComponent.cs
+++ b/Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemComponent.cs
@@ -9,7 +9,7 @@ namespace Content.Server.Item.PseudoItem;
 public sealed partial class PseudoItemComponent : Component
 {
     [DataField("size")]
-    public int Size = 120;
+    public int Size = 100;
 
     public bool Active = false;
 }
--- a/Resources/Prototypes/Entities/Structures/Furniture/potted_plants.yml
+++ b/Resources/Prototypes/Entities/Structures/Furniture/potted_plants.yml
@@ -19,6 +19,11 @@
         density: 190
         mask:
         - Impassable
+  - type: Storage # Delta V - Felinid in plant
+    capacity: 150
+    whitelist:
+      components:
+        - Felinid # Delta V - Felinid in plant till here
   - type: Sprite
     drawdepth: Overdoors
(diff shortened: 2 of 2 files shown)


Leander-0:
Note felinids would still be able to shoot from inside those and bullets will get stuck inside the plant too leaving it impossible to get back inside again. This also would need for the potted plant prototype to be changed since you can go through doors and windoors with those.

ps3moira:
Just make it like ss13 where anyone can hide in plants, instead of decreasing the size of felinids.

Adrian16199:
> Just make it like ss13 where anyone can hide in plants, instead of decreasing the size of felinids.

The only reason why I decreased the size of felinids just so they can hide in the backpack aswelll, I dont think a normal human being would be capable of hiding in a bag. Even a duffelbag. Even tho technicly you can carry so many guns in a singular bag which doesnt make sense either but i guess felinids just simply have that as a unique thing.

Adrian16199:
> Note felinids would still be able to shoot from inside those and bullets will get stuck inside the plant too leaving it impossible to get back inside again. This also would need for the potted plant prototype to be changed since you can go through doors and windoors with those.

It was noted in the PR itself that this is not the case regarding being able to shoot. Not sure what you are refferencin regardin the "You can go through doors and windoors"

DebugOk:
Not only should they not fit inside backpack, them fitting inside plants makes no sense, and I don't want to know how that'll go with shooting. This wasn't merged on Nyano for a reason

Adrian16199:
Eh, I will just note that on the original Pr it wasnt merged because you cant shoot the felinid in the plant like with the duffelbag, which is currently a favourite mechanic on felinid, I will repeat it once more. You cant shoot inside the plant. That is all.

AitorLogedo:
> Not only should they not fit inside backpack, them fitting inside plants makes no sense, and I don't want to know how that'll go with shooting. This wasn't merged on Nyano for a reason

exactly why they shouldnt fit in a backpack ?

PR #555, opened 2023-12-13
Remove non-salvage testing map from pool

IamVelcroboy:
## About the PR
- Remove testing map from pool.

## Why / Balance
It's not meant for the salvage pool and contains several items that shouldn't be mapped in them.


Colin-Tel (approved):
lol whoops

PR #556, opened 2023-12-14
Make felinids scream in agony when sprayed with water.

Adrian16199:
Their screams will be heard thro walls and halls. You will hear a felinid scream in a shift atleast once. Their sheer agony.

## Why / Balance

Because thats how it should be

## Media

**Changelog**

:cl:
- add: Felinids now scream in agony from water.


DEATHB4DEFEAT (approved):

DebugOk:
Does this not have the ability to trigger several times from a single spray?

Adrian16199:
> Does this not have the ability to trigger several times from a single spray?

A probability was given and the 0.2 makes it so that it doesnt trigger (usually) the multiple screams, yes it makes the scream not always trigger but its better than to pretend that multiple screams constantly would be fine themselfs. Of course, I could not be a lazy idiot and make a cooldown system for it. However such thing should probably be done upstream.

DebugOk (approved):

PR #565, opened 2023-12-16
Wizmerge 16/12

DebugOk:
# For the love of god dont merge this yet lmao

The server doesn't have dotnet 8 installed yet and will shit itself if this is even partially published. Also I probably need to update the watchdog for the first time since we became delta :blunt:


DebugOk:
If tests pass locally, I'm just merging this and we'll deal with the shit that causes later

PR #585, opened 2023-12-21
Rebalance Changes For Desword

Tryded:
## About the PR

Lower Attack Rate to the same as Esword
Heightened the reflection chance for Energy shots to 90%
No more structural damage
## Why / Balance

Was originally way too OP to have high damage and high attack speed (Kill time is now 3 seconds instead of 2 seconds which is difference between countering with a gun/calling for help and dying
Can't upstream because upstream straight up removed it
## Media

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
@@ -183,12 +183,12 @@
         types:
             Slash: 17
             Heat: 17
-            Structural: 20
+ #          Structural: 20 DeltaV Rebalance
             Blunt: -4.5
     litDisarmMalus: 0.7
   - type: MeleeWeapon
     wideAnimationRotation: -135
-    attackRate: 1.5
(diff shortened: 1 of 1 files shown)


Tryded:
The reflection rate *can* be lowered

Tryded:
> The reflection rate _can_ be lowered

However again, this can be easily countered by a ballistic weapon of any sort or literally a bar of soap if they didn't buy no slips.

As for nukies, I'm unsure of exactly what caused it to be unbalanced due to me not knowing what round it is, however hearing from others I think it was because the nukie was either holding normal eswords in their pockets (Potentially being able to reflect even in pockets causing ballistics to be reflected), or there might just be some large oversight that I didn't think of. 

Another easy way to counter people with deswords is to use bolas, or to use explosives/chems

DangerRevolution:
> The reflection rate *can* be lowered

Why is the reflect going to 90%?

DebugOk:
This item is disabled in an upcoming update, and no balance changes will be accepted

PR #586, opened 2023-12-22
nerf loneop and ninja spawn rate

deltanedas:
## About the PR
youd see a loneop nearly every round, ninja not far behind
this makes them more tuned to 2h rounds, and less of a "when will it get here" but "may it get here"


ps3moira:
Ghost-main players in shambles

PR #587, opened 2023-12-22
Changes Reinforcement TC Cost

DangerRevolution:
## About the PR

- Drops the TC cost for Syndie Reinforcements to 10, Nukie to 15
## Why / Balance

It's never used, primarily because the cost is outrageous especially for a ghostrole. If C4 capable mice are 3TC, why is a human 16? A department-destroying, most-of-the-time round ending bomb is 11TC. 
### The cost is so high, you can literally buy nothing of value for yourself or the person you summon, there's 0 incentive to buy this, even for gimmicks.
What's even the value of *buying* another Operative, who's not on the manifest and starts in the most conspicuous clothing available?
- This makes the Reinforcement *more* valuable while still keeping him too expensive for conventional strategies and *too* expensive for mass-abuse by Nukies.
- The Nukie Cost does *not* lower dramatically because every NukeOp buying a reinforcement could lead to disastrous results.

**Changelog**

:cl: DangerRevolution
- tweak: Syndicate Reinforcement Cost has been lowered to 10TC, Nukie Reinforcement to 15.

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -431,7 +431,7 @@
   productEntity: ReinforcementRadioSyndicate
   icon: { sprite: Objects/Devices/communication.rsi, state: old-radio }
   cost:
-    Telecrystal: 16
+    Telecrystal: 10 # Delta-V - Lowers cost to 10, from 16.
   categories:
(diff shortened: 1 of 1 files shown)


DebugOk:
As it should be

DangerRevolution:
> As it should be

Any further context or explanation to that?

PR #596, opened 2023-12-24
Personalized HoP's locker

ps3moira:
## About the PR

Added a little picture of Ian taped inside HoP's locker.
## Why / Balance

It's hard to keep track of how much Ian's grown since he was a pup 😭 
## Media
image

- [X] I have added screenshots/videos to this PR showcasing its changes ingame

## Breaking changes

It breaks my heart to see how much he's grown...
**Changelog**
:cl:
- tweak: Changed the inside of HoP's locker.


leonardo-dabepis:
perfect 😃

Resources/Textures/Structures/Storage/closet.rsi/meta.json
DebugOk, line 1: Why is this file being touched?

VMSolidus (changes requested):
If we want to start making changes to upstream items like this, we generally should either submit a PR upstream, or use a migration to create a DV-LockerHeadofPersonnel. At which point by doing so, the contributor in question would be taking responsibility for maintaining the item. If you wish to do the latter, I would be willing to approve this PR in such a case, otherwise this should be closed and brought upstream. 

The advantage of an item migration is that it would give us full and complete control over everything we want to do with it, and we wouldn't have to worry about complicating upstream merges. The downside of course is that it adds a burden for maintainers here to keep items up to date with new upstream components.

PR #598, opened 2023-12-24
death to eorg

deltanedas:
## About the PR
- change from ExplodeOnTrigger to Explosive being removed, now even welderbombing should be impossible
- OnUseTimerTrigger is now nuked so you cant use flashbang/stingers (SHITSEC)
- Flash is nuked so you cant use Z with flash (SHITSEC)
- refactor to use proxy methods pro


DebugOk (approved):
Honestly still hate this code, but people still cant behave

PR #611, opened 2023-12-29
Remove potential psionic from Laika

FluffiestFloof:
## About the PR

Removes potential psionic from Laika.
Removes the starting headgear from Laika

## Why / Balance

No other pets have potential psionic and it doesn't make sense for this specific pet to get it.

## Technical details

N/A

## Media

## Breaking changes

**Changelog**

no cl since previous cl didn't mention this bit

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
@@ -53,4 +53,3 @@
     - CannotSuicide
     - VimPilot
     - DoorBumpOpener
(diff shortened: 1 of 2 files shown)


DebugOk:
Why does the dog have a whole ass headset and ear slot

FluffiestFloof:
> Why does the dog have a whole ass headset and ear slot

I have no idea, I wasn't the one that decided upon that.
Punpun can wear a headset and I think starts with one but they are monkey so _shrug_
I can get rid of it here too if you want

DebugOk:
The ear slot can stay, but the round start headset has to go. Honestly tempted to say the same for the id because the hos has more

PR #621, opened 2023-12-31
Make the soap smell like not condoms

leonardo-dabepis:
be the change you wish to see in the worl

## About the PR

made it so the soap smells like "ethyl methylphenylglycidate and cream" (ethyl asufihlafuiagife is the chemical used to create artificial strawberry flavoring) instead of condoms.
## Why / Balance

condoms are ICKY!!
## Media

## Breaking changes

**Changelog**

:cl:
- tweak: The heart soap no longer smells like condoms.


DebugOk:
Should be done upstream, and this just replaces a bad joke with something that most people won't even understand.

PR #622, opened 2023-12-31
Allows more markings for our species for arms and legs.

Adrian16199:
## Why / Balance

My reasoning to allow it is because the addition of bandages are great but the limit didnt allow much to be done so I personally think doubling the limit instead of makin it unlimited is better option but will see.
## Media

image
image

## Breaking changes

**Changelog**

:cl:
- tweak: Allows more markings limit on arms and legs for felinids, onis and vulpkanins.

### DIFF
--- a/Resources/Prototypes/DeltaV/Species/vulpkanin.yml
+++ b/Resources/Prototypes/DeltaV/Species/vulpkanin.yml
@@ -50,10 +50,10 @@
       points: 1
       required: false
     Legs:
-      points: 1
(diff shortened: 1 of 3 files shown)


DebugOk:
Oni and Felinids already have 2, which is plenty. Additionally, Vulpkanins have leg and arm markings that'd conflict if several could be applied. I don't see the point in merging this without the ability to disallow conflicting markings.

Adrian16199:
😔

PR #628, opened 2023-12-31
Update secret_weights.yml

Colin-Tel:
## About the PR

Adjusted weights for gamemodes governed by the Secret gamemode.

## Why / Balance

Survival rounds were quite frequent when the intent was for them to happen *sometimes*. The changes are as follows:

```
survival went from .40 to .27
Nukeops went from .12 to .15
zombie was unchanged
revolutionary went from .10 to .15
Traitor went from .35 to .40
```

**Changelog**

:cl:
- tweak: Adjusted the frequency of different gamemodes. Survival rounds are now somewhat less common.

### DIFF
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -1,10 +1,10 @@
 - type: weightedRandom
   id: Secret
   weights:
-    Survival: 0.40
-    Nukeops: 0.12
+    Survival: 0.27
+    Nukeops: 0.15
     Zombie: 0.03
-    Revolutionary: 0.10
-    Traitor: 0.35
+    Revolutionary: 0.15
+    Traitor: 0.40
     #Pirates: 0.15 #ahoy me bucko
 


DebugOk:
Survival is the MRP gamemode, not sure why you'd want more antag rounds

AitorLogedo:
> Survival is the MRP gamemode, not sure why you'd want more antag rounds

What exactly makes you think only a gamemode is "MRP" and how the gamemode which can have some kinda of cooperation and requires almost more rp to do certain gimmicks isn't the main one?MRP it's separated from the way the game its, and more about how each player treats the game modes. If you watched discord once in the past week you would see that people are tired of rounds of nothing and self antags because that's what survival encourages, because 7 rounds in a row of 2 hours of nothing then zombies isn't a good "MRP" loop is just awful.

DangerRevolution:
I wouldn't have gone in with the uh, energy of @AitorLogedo but
> If you watched discord once in the past week you would see that people are tired of rounds of nothing and self antags because that's what survival encourages

is real. people generally don't like event rounds, as often if you don't get some cool interesting RP you're left with nothing but RP-ing fixing meteor storms or fighting off bugs. I can say myself I've had rounds where if I don't get some cool gimmick, I'm kinda left a little bored with only predictable events coming in. I generally think the stats are fine, making Syndie rounds the main is probably fine and the changes to non-syndie/survival are minor. Always room to less drastically decrease the weighting if anything.

Might be missing something but, just thoughts off the top of my head and what we talked about in #general that helped @Colin-Tel make this

DebugOk:
Self antags wont get fixed by having antags in the round. The weights are currently fine as is, and with more midround antags being added survival should still be the default. Essentially halfing it is taking way too much.

PR #632, opened 2024-01-01
Tweak gamemode odds for secret.

BramvanZijp:
## About the PR
The weights for gamemodes were adjusted.
Survival: 40% > 30%
Nukeops: 12% > 15%
Zombies: 3% > 3%
Revs: 10% > 12%
Traitors: 35% > 40%

## Why / Balance
I know a comparable pr, 628, was denied, but me and several others disagree with that choice.

Survival as a gamemode, yes it leaves room for rp, but thats the only thing you can do, in an ideal world on MRP, this would be fine, but we need something to keep the round interesting and push rp forward, the game does this through antagonists. Survival, which pretty much nullifies this. should not be the most common gamemode. ESPECIALLY when you consider the fact that revs usually ends relatively early from what I have seen, leaving the end of the round as survival but worse. This means that right now, you flip a coin. Heads and the round gets boring after an hour to an hour and a half if you are lucky (survival/revs), tails and you either get excited and action packed gameplay (zombs/revs) or a long immersive story of constant tension and struggle (traitors). And before you say revs falls under the same category as zombs/revs/traitors, I disagree. A singular head rev on a station of 60-80+ people IMO is less of a threat then a loneop or ninja, but is then also the only antagonist force of that round, so I consider it a delayed survival.

So please, consider this request, and consider the fact that a lot of people are upset due to the game getting stale/boring pretty often. Yes this is MRP, but with the current state of delta V roleplay, we need something to keep rounds, that can often exceed 2 hours in length, from getting dry, which if you spectate/participate in a few rounds, you will notice happens far too often.

## Media

## Breaking changes

**Changelog**

:cl:
 - tweak: NanoTrasen stations have been observed coming under threat more often, stay sharp people!

### DIFF
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -1,10 +1,10 @@
 - type: weightedRandom
   id: Secret
   weights:
-    Survival: 0.40
-    Nukeops: 0.12
+    Survival: 0.30
+    Nukeops: 0.15
     Zombie: 0.03
-    Revolutionary: 0.10
-    Traitor: 0.35
+    Revolutionary: 0.12
+    Traitor: 0.40
     #Pirates: 0.15 #ahoy me bucko
 


DebugOk:
This should be done with admin discussion by Colin, not on a whim by a player.

AitorLogedo:
> This should be done with admin discussion by Colin, not on a whim by a player.

You closed the pr that collin made without any discussion, what else is to discuss?

PR #634, opened 2024-01-01
Listening Outpost Tweaks

VMSolidus:
## About the PR
These are all requested changes to the listening outpost, in response to feedback during initial testing. 

## Changes

- The map has been moved to a DeltaV directory.
- Event weights are adjusted to be the same as other midround antagonists. 

- Guards no longer show up as permanently SSD when examined.

image

- Guards now spawn with Death Acidifiers. 

image

- A disposals bin has been added that ejects trash into the southern caves.

image

- The number of EVA suits available to the outpost has been reduced to one. So as to prevent operators from taking guards onto any shuttle the station salvagers might accidentally give them.

image

**Changelog**
:cl: VMSolidus and DangerRevolution
- tweak: The Listening Outpost has received several tweaks and balance changes. It should spawn less often now.


Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/salvage.yml
```diff
@@ -6,6 +6,9 @@
     - type: NpcFactionMember
       factions:
         - Syndicate
+    - type: MindContainer
+      ghostOnShutdown: True
```
DebugOk: Redundant, ghostOnShutdown is true by default.
https://github.com/DeltaV-Station/Delta-v/blob/df92f0f1f6b188cf2ffc80f0e760706442e0080a/Content.Shared/Mind/Components/MindContainerComponent.cs#L33-L39

VMSolidus: Good catch, thanks!

DebugOk (approved):

PR #539, opened 2023-12-10
Salvage Wreck Overhaul & New Salvage Mobs

VMSolidus:
## About the PR
Uses #526 as a Dependency
TODO:

- [ ] Run every wreck in the entire game, both lists, through a quality check and balance pass
- [ ] Make new salvage mobs based around Hostile Humanoid NPCs
- [ ] Add Spawners for these new hostile humanoids, and make use of them in wrecks
- [ ] Somehow fix/get fixed the AI inability to bolt guns
- [ ] Stranded Nukie miniboss

## Why / Balance
A lot of new salvage wreckages have been added back into the game recently, however these were just done via porting the old Nyano era wrecks back to the list. The actual quality of wrecks varies wildly, with both the Wizden and Nyano lists having a few really terrible examples mixed in with some outstanding ones. A consistent issue seen is that wrecks do not often have their potential rewards balanced by risk. With some of the most egregious examples being a wreck that contains zero hazards of any kind, but rewards the players with two C-20r SMGs and two retro laser guns that are just sitting on a table, and the opposite having a reinforced vault guarded by multiple space bears and turrets, which contains solely a single retro laser with no other loot. 

I have a pretty decent amount of experience with design philosophy, and one such philosophy I intend to follow here is that the game needs to interact with the players. Instead of players simply getting loot for free because they know how to juke a carp, the loot needs to fight them. For a bloodred headset to exist on a wreckage, it needs to be worn by a hostile NPC. If salvage wants to loot a gun, that gun must first be fired at salvage. In the most extreme example possible, to get a C-20r, salvage must first fight someone who has a C-20r, kill him, and take his gun. Yes, that will introduce an extraordinary amount of risk to salvage, but the rewards must in turn match the risk. It may very well be that experienced salvagers could encounter a wreckage, and then turn away from it because they know they simply aren't equipped to handle fighting something like a stranded nukie.

## Technical details
TODO

## Media
The barge returns, but it won't be something cargo buys. 
image

## Breaking changes

**Changelog**

:cl: VMSolidus and DangerRevolution
- add: Added fun!
- remove: Removed fun!
- tweak: Changed fun!
- fix: Fixed fun!

PR #659, opened 2024-01-10
Randomized Listening Outpost Spawning

VMSolidus:
## About the PR
I have changed the listening outpost so that instead of FTL'ing to the station from null space, it instead spawns at a random location roughly 1 to 1.5 kilometers away, and at a random angle. 

## Why / Balance
This is the originally intended spawning system, but neither me nor DangerRevolution had yet worked out how to manage it. 

## Technical details
This code is very much shit and kind of rough, and it uses a deprecated function. It should probably be slightly rewritten to use the newer TryLoad system instead of LoadGrid. It also needs to be double checked for unnecessary "Using" statements, since I basically took the list from PirateSpawnRule while making this. Also it is shamelessly using code from said PirateSpawnRule

## Media
image

**Changelog**
:cl: VMSolidus and DangerRevolution
- tweak: The Listening Outpost now spawns a random large distance from the station, instead of null space.


deltanedas:
wont it be visible on mass scanners

VMSolidus:
> wont it be visible on mass scanners

Yes, and that's next on my list. I want to deprecate the IFF console and make it so that it instead shows up on the console missing a name entirely, and appearing as if its an asteroid.

deltanedas:
asteroids dont exist and dont do that so it will be metad outright or "hmm whats this unmarked object lets send out 5 suited up sec and the salv squad to investigate"

IamVelcroboy:
> asteroids dont exist and dont do that so it will be metad outright or "hmm whats this unmarked object lets send out 5 suited up sec and the salv squad to investigate"

That why we're going to start spawning in decoy asteroids and more random stuff as well.

DangerRevolution:
> asteroids dont exist and dont do that so it will be metad outright or "hmm whats this unmarked object lets send out 5 suited up sec and the salv squad to investigate"

intended goal of this is in #539; spawning a lot more in-map stuff for Salvage to encounter like this

>lets send out 5 suited up sec and Salv squad to investigate

You'd actually be surprised that it's been raided officially only once so far, and it's within 300 meters of the station

VMSolidus:
While looking into the IFF code and doing some experiments, I've concluded that it's not actually possible to have a grid like this spawn in with the IFFComponent, despite being able to simply shove it into the .yml file without issue. Instead I'm going to need to make a variant clone of the IFF Console that does the whole "Spoof an Asteroid" thing by default.

VMSolidus:
I'm closing this PR since it's being merged with PR639

PR #679, opened 2024-01-15
Prevent zombies from becoming psionic

DebugOk:
## About the PR
They're braindead, and people just use it to bypass chat restrictions

## Why / Balance
Some arguments got made against removing it, and while I do think they're good, people won't roleplay those out except in very rare cases. I already don't like zombies, and I don't think changing how people play them is very possible. So this is the next best thing

## Technical details
PsionicComponent now tracks which powers all got added. For whatever reason, you can have multiple powers, but PsionicAbility only accounts for one and wont resolve to the component that needs to be removed. 

## Breaking changes

**Changelog**

:cl:
- remove: Zombies can no longer be psionic


deltanedas:
zombies can still use telepathy i think which is still bad

it should be disabled outright or have the accent forced

DebugOk:
> zombies can still use telepathy i think which is still bad
> 
> it should be disabled outright or have the accent forced

Loosing psionic component means you loose telepath chat access

deltanedas:
oh good then no more telepathy becoming zombie snitch chat :trollface:

PR #661, opened 2024-01-10
Implement Ifriit

Adrian16199:
The mystagogue's guardian.

So as it stands, this can be merged upon innate tool fix gets merged into master and notin pyrokinesis until it gets reworked.
Converting it to draft and going to apply changes for it to use innate tool again.

## About the PR

YML
## Why / Balance

They used to be a thing, simply fixes them
## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase
image

https://github.com/DeltaV-Station/Delta-v/assets/144424013/5d1e7ab2-6804-4f03-b593-1d2bc838221e

## Breaking changes

**Changelog**

:cl:
- fix: Implements ifriit

PR #680, opened 2024-01-16
Fix Ifrit and make it a ghost role

Guess-My-Name:
## About the PR

Make ifrits like they used to be pre-rebase.

Short list of what needs to be done:
- [x] Make ifrit ghost role actually spawn an ifrit
- [x] Give them their hands back (InnateTool component doesn't work anymore)
- [ ] Figure out why they only get a welder in one of their hands and the second hand is empty (or just remove one of their hands and roll with it)
## Media

## Breaking changes

**Changelog**

:cl:
- add: Added ifrits, servants of the mystagogue and protectors of The Oracle.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/familiars.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/familiars.yml
@@ -33,8 +33,10 @@
     bloodReagent: Phlogiston
   - type: MeleeWeapon
     damage:
-      groups:
-        Burn: 10
+      types:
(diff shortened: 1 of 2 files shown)


DebugOk:
https://github.com/DeltaV-Station/Delta-v/pull/661 This is already being worked on. Please check existing pull requests in the future and talk with Adrian about which pr will stick

Adrian16199:
The pr almost did it but it wasnt spawning that second ifriit hand, unfortunate if i had to be honest, the only other way I can see it happening would require innatetool system to work which would be pain.

PR #681, opened 2024-01-16
Update warp point science->epi

IamVelcroboy:
## About the PR
Small thing. Was making warp point templates show up as "science" instead of "epistemics."


IamVelcroboy:
These were removed anyway. Will reopen with beacons.

PR #682, opened 2024-01-16
qol tweak: More context about how to access traitor uplink in character menu

Arkanic:
## About the PR
Add a bit more context about what the traitor codes actually mean to uplink menu

## Why / Balance
It's already displayed in the chat menu but from experience dumb people like me don't see the message or cannot read it because of reconnecting mid-round, or it is just so far up the chat log that it is sort of inconvenient to search for

My first traitor round I had to reconnect at the start of the game and so I spent the first 20 minutes of the round trying to figure out how to use the uplink code in the character menu; the PDA ringtone is, of course, easy to remember but not very intuitive for a new player - specifically adding it to the character menu clears a lot of confusion I feel, as it is the place that people go to to check antag status/tasks condensing useful information there seems to be the right thing to do

## Media

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Locale/en-US/game-ticking/game-presets/preset-traitor.ftl
+++ b/Resources/Locale/en-US/game-ticking/game-presets/preset-traitor.ftl
@@ -40,4 +40,4 @@ traitor-role-uplink-code =
 traitor-role-codewords-short =
     The codewords are:
     {$codewords}.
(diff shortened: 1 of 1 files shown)


DebugOk:
This should be done [upstream](https://github.com/space-wizards/space-station-14/)

PR #703, opened 2024-01-21
Implement gridinv, proper

DebugOk:
## About the PR
Implements gridinv, thats about it. (And also like 1 or 2 commits I apparently forgot)

## Why / Balance
I really couldn’t keep avoiding this change if I wanted to keep doing upstream merges easily. I know I made a promise I would avoid this, but it was always a “For as long as possible” thing. I don’t like having to do this either, and I know some people really hate gridinv, and I’m sorry for that. The only way we could keep using the old system is if someone rewrote a ton of code, which I’m not willing to do.

## Technical Details
Because I had to revert to October and repick all of the merges I did, I had like 1500 commits pending, which github didn’t like. Instead, I’ve squashed them all into 1 and applied that on top of the current master, fixing any conflicts that created. It seems to have worked, but I’ll still need time to review this pr just in case that resulted in anything weird.

## Breaking changes
Its gridinv, any prs adding items/storage that specify stuff using the old values will need to be modified. But it shouldn’t™ break anything beyond that.

**Changelog**
:cl:
- add: Added gridinv, sorry.


DebugOk:
I have read over all of the changes, and I believe there shouldn't be anything unintended in here except persistance being added. Oh well, fuck it.

Sorry to everyone who hates gridinv. I would've loved to keep avoiding this.

deltanedas:
persistence doesnt matter since you have to enable it in server config

DebugOk:
> persistence doesnt matter since you have to enable it in server config

Yeah, just saying cause it got squashed in with the rest and I was too lazy to undo it

PR #711, opened 2024-01-22
Update for preferred Delta V exterior lights

IamVelcroboy:
## About the PR
See title.

## Why / Balance
The WD exterior lights are way too bright and wash out too much color compared to the Nyano ones. Rather than edit every file that has them or that comes in from upstream, this just migrates them out. Easy peasy.


Resources/Prototypes/Entities/Structures/Lighting/base_lighting.yml
```diff
@@ -149,7 +149,7 @@
 - type: entity
   id: PoweredlightExterior
   description: "A light fixture. Draws power and produces light when equipped with a light tube."
-  suffix: Blue
+  suffix: Blue, Don't Map on Delta V #Delta V: Don't Map on Delta V
```
DebugOk:
```suggestion
  suffix: Blue
  noSpawn: true # DeltaV - Don't map these
```

IamVelcroboy: oh, good call. Forgot all about that.

DebugOk: If you're automatically migrating them anyways why bother even allowing them to be mapped at all

IamVelcroboy:
Wait, the tubes got lost somehow...oh nevermind. I see what it...going to add nospawn there as well.

Resources/Prototypes/Entities/Objects/Power/lights.yml
```diff
@@ -190,7 +190,8 @@
   parent: BaseLightTube
   name: exterior light tube
   description: A high power high energy bulb for the depths of space. May contain mercury.
-  id: ExteriorLightTube
+  id: ExteriorLightTubeBlue
```
DebugOk: Why is the id changing?

IamVelcroboy: I'm already pushing. No idea why it did that

DebugOk (dismissed):

PR #719, opened 2024-01-24
Update Salvage Vend

ps3moira:
## About the PR

Updated Salvage's vend machine to offer proper gear, as there's currently no in-game way of getting salvage gear other than it being mapped in. Gear added to salvage's vending machine is 3 of each bag type, 4 tactical gas masks, 3 salvage boots, and 5 salvage webbing.
## Why / Balance

Gives salvage proper gear that doesn't require map-ins.

**Changelog**

:cl:
- add: Added salvage gear to salvage vending machine.

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/salvage.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/salvage.yml
@@ -1,6 +1,13 @@
 ﻿- type: vendingMachineInventory
   id: SalvageEquipmentInventory
   startingInventory:
+    ClothingBackpackSalvage: 3 # DeltaV
+    ClothingBackpackSatchelSalvage: 3 # DeltaV
+    ClothingBackpackDuffelSalvage: 3 # DeltaV
+    ClothingUniformJumpsuitSalvageSpecialist: 3 # DeltaV
+    ClothingShoesBootsSalvage: 3 # DeltaV
+    ClothingMaskGasExplorer: 4 # DeltaV
+    ClothingBeltSalvageWebbing: 5 # DeltaV
     Crowbar: 2
     Pickaxe: 4
     OreBag: 4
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Catalog/VendingMachines/Inventories/salvage.yml
```diff
@@ -1,6 +1,13 @@
 ﻿- type: vendingMachineInventory
   id: SalvageEquipmentInventory
   startingInventory:
+    ClothingBackpackSalvage: 3 # DeltaV
+    ClothingBackpackSatchelSalvage: 3 # DeltaV
+    ClothingBackpackDuffelSalvage: 3 # DeltaV
+    ClothingUniformJumpsuitSalvageSpecialist: 3 # DeltaV
+    ClothingShoesBootsSalvage: 3 # DeltaV
+    ClothingMaskGasExplorer: 4 # DeltaV
+    ClothingBeltSalvageWebbing: 5 # DeltaV
```
VMSolidus, line 10: I actually think this should instead be empty tool belts instead of salvage webbing. I'm also not certain that it's a great idea giving people access to unlimited amounts of the explorer mask, seeing as it's a "Best in Slot" mask due to its armor values. People who aren't even salvage techs might try to hoard salvage restocks so that they can be broken open to get the armored masks. 

Keep in mind that salvage webbing is also a "Tier 2" salvage item, that is normally only obtainable from wrecks. It's seen as an upgrade item to replace your starting tool belt, since webbing can carry any item, while the belt can only carry tools. Other than that, perhaps these numbers should also be consistent as specific number of "kits" per refill. Judging by the rest of the items present, I think its meant to support "One refill to get two extra salvage techs".

DebugOk:
All these items are better versions of what they start with and can be obtained naturally

PR #730, opened 2024-01-29
Remove Insect Blood from Moths

ItzaLeaf:
## About the PR

Removed insect blood from moth, Wizden change Revert.
_(Note: insect blood can't be used for chemistry recipes which require "blood", along with bringing up the issue that standard blood packs shouldn't theoretically work either. Potentially leading to overcomplication of medical supplies and chemistry in the future... please no.)_
## Why / Balance

 - Possible future issue of technical implementation of features involving different chemistry reactions with various blood types.
 - Nerfs moth chemists as they can't use their own blood and need donations for recipes involving "Blood" reagent.
## Technical details

Just reverts the change made by the Wizden pull over, so should just work as it used to. (Normal moth blood)
## Media

## Breaking changes

Nothing.
**Changelog**

 - remove: Removed code giving moths insect blood. (Should auto-revert to standard now)

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/moth.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/moth.yml
@@ -34,8 +34,6 @@
     spawned:
     - id: FoodMeat
       amount: 5
-  - type: Bloodstream
-    bloodReagent: InsectBlood
   - type: DamageVisuals
     damageOverlayGroups:
       Brute:


Haltell:
Dionae and Slimes also don't have normal blood, so I personally don't see why this needs to be reverted, especially since it's changing upstream code downstream. As for possible future blood chemical reactions, those are also more likely than not going to be implemented upstream first.

DangerRevolution:
> Dionae and Slimes also don't have normal blood, so I personally don't see why this needs to be reverted, especially since it's changing upstream code downstream. As for possible future blood chemical reactions, those are also more likely than not going to be implemented upstream first.

+1 to this + this is just asking for issues down the line

DEATHB4DEFEAT:
I don't like this, complain to WizDen if you don't want the insect to have insect blood.

> Possible future issue of technical implementation of features involving different chemistry reactions with various blood types.

Using different blood types to make different things is fun and realistic, a great combination.

> Nerfs moth chemists as they can't use their own blood and need donations for recipes involving "Blood" reagent.

This is a plus, go interact with someone else, ask people for blood donations.

DebugOk:
Working as intended, no reason to change it

PR #740, opened 2024-01-30
Map Hotfix

IamVelcroboy:
## About the PR
Hammurabi/Arena/Tortuga/Asterisk
- Replaces all cryosleepers with standard


DebugOk:
I would appreciate if this stuff gets brought up to me first, instead of immediately reverting a ton of maps...

This is literally a one line fix

PR #745, opened 2024-01-31
Disables mind swap.

Adrian16199:
## Why / Balance

Okay so, this is a long one but this is somethin that I feel like just should be done until either a rework or just simply never, or at the least when the system is more advanced and not in its current state.
1. Reworking it could be a bad idea.
-Mass mindswap is a thing and it would be better to just simply make the ability only disabled from being rolled instead of reworking it due to the fact that it could easily fuck up something that feels fair.
"You got 1000 glimmer, the entire station gets mindswapped" Simple and you can prevent it, you know its coming, even if you dont want to. It is a consequence from a fuck up, not from you existing.
2.Why disable it.
-My first reasoning is the power itself is just unfun for everyone. Maybe at best for the guy who threw you into an old body of his before blowing it up or for the self-antager that threw an insulated cap at you, mindswapped and now is walkin around as if he owned the station.
-Second reasoning: 
Compare it to other abilitys. Even take somethin like a mass sleep, very similar when it comes to how one uses it.
Use it: sleep someone/mindswap someone to insta kill them.
However now the difference is: 
Mass sleep: someone is obvious and it is possible to find out that it happend by alot of people suddenly dropping and the target is always themselves.
Mind swap allows you to:  immediately leave the scene without looting the target because well, you are them now, nobody is going to believe a person who saw it because at the best, it looks like a suicide and the guy just shruged it off and left.
Usually the captain is the target, for the authority, objectives or even the fire power itself, so you became the captain and who is gonna question you? And even then, just mindswap with someone else and kill them again.
How does one find out?
Now thats the third reason: There is no way of finding out after it happend, at the best you can find out because you have been playing with that person for 100 hours and he stopped smoking suddenly.
Now you might say, well, why not just give a way to find out.
Now thats the neat part, reworking even in the first place takes time. So this could be even interpreted as revolutionary gamemode. Disabled till a rework at best.
Fourth reasoning: The consequences all goes to YOU, the target. People would start hating on you with the fact that you started to be a dick to them. They dont need to stay fully until the end of the shift and even then, they could not look at who played the character, they could leave with a sour taste towards you. All because a guy rolled this power as an antag.
Of course, there is an argument, there always will be against this change but tell me, what was the last time that you found this power to be giving someone a good rp experience?
Either spectating or being the target or even the one doing it?
Just triggering a minibomb in your hands, losing your body that you lived with this casually just because you have this power.
Why am I complainin with this power and none else?
Every other power doesnt feel cheap. Every other power doesnt need to paint you bad for somethin that you cant even control.
I dont recall a singular good experience with this, hell, even mass sleep can be funny or used in self defence.

## Media

## Breaking changes

**Changelog**

:cl:
- remove: Mind swap has been disabled.

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/psionicPowers.yml
+++ b/Resources/Prototypes/Nyanotrasen/psionicPowers.yml
@@ -7,4 +7,4 @@
     PsionicRegenerationPower: 1
     MassSleepPower: 0.3
     PsionicInvisibilityPower: 0.15
-    MindSwapPower: 0.15
+    # MindSwapPower: 0.15


AitorLogedo:
I don't like the premise of removing something without changing it or modifying only because 
someone doesn't like it

Adrian16199:
I will note that im not the only one who thinks that this power is a bit broken, if it was just only me then I would just assume that im just insane but when I attempted to remove thieving gloves, people spoke up and liked it, I do understand the fact that its eh to just remove something but im uncertain if I or someone would solve the issues of it, yeah, maybe the power itself could not be the issue but the lack of finding ways to actualy do somethin about it or such. People could prefer to disable it and get it later and even if nobody would want that, pr could be rejected, all up to people and debug.

DangerRevolution:
Eh, I rarely if ever see this power and it's generally led to interesting moments.

Plus uh, disabling it without any certainty about it ever coming back is eh.

Adrian16199:
Oh yeah uh, also, I will note that this only disables normal mindswap from being rolled, the event would still happen at high glimmer.

Adrian16199:
So, now Imma say this, the fact that I posted this and I have gotten alot of people talk to me, now I feel like I could get the exact solution of fixing it so im fully willingfull to get this either denied or the PR being fully changed so i will just put this on draft.

PR #752, opened 2024-01-31
Remove last remaining way of getting tesla generator

DebugOk:


deltanedas:
also remove the other parts from cargo since if you cant get generator someone wasting 10k on equipment gets scammed

DebugOk:
> also remove the other parts from cargo since if you cant get generator someone wasting 10k on equipment gets scammed

The other parts have their uses. You can harvest electric anomalies if you wanted to

deltanedas:
ok true

PR #767, opened 2024-02-03
Adds Delta V station beacons

IamVelcroboy:
## About the PR
- Adds DV specific station beacons for mapping
- Adjusts upstream beacons for DV specific changes

## Why / Balance
mapping goodness


Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
```diff
@@ -0,0 +1,28 @@
+station-beacon-epistemics = Epistemics
```
DebugOk, line 1: Science beacon has been renamed, redundant

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
```diff
@@ -0,0 +1,28 @@
...
+
+station-beacon-logistics = Logistics
+station-beacon-logistics-reception = Logistics Reception
+station-beacon-lo = LO's Room
```
DebugOk: Again redundant since there are already beacons for this

DebugOk (changes requested):
These names do not match the original style and are way too long, additionally several are redundant.

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
IamVelcroboy: This just moves it to our own DV file, so you can ignore the WD file

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
IamVelcroboy: I assumed you just did it this way to get through the big merge a little quicker. Would this not fit our file standards better and allow for keeping track of upstream file changes?

IamVelcroboy:
> These names do not match the original style and are way too long

Which ones? I could drop Glimmer Prober Room -> Glimmer Prober. I'm not sure which others don't fit though?

DebugOk:
> > These names do not match the original style and are way too long
> 
> Which ones? I could drop Glimmer Prober Room -> Glimmer Prober. I'm not sure which others don't fit though?

Fucking all of them outside of Virology, Cloning, Park and Mailroom.

Beacons for escape pods are also bloat, so is Cloning, Exam, the prober and the camera servers

IamVelcroboy:
> Beacons for escape pods are also bloat, so is Cloning, Exam, the prober and the camera servers

Escape pods on maps would make them easier for players to find. 

Epi, Logistics, Sec, Engi are all broken up into their different rooms, why would we not have one for the missing med rooms(cloning/exam)

Camera servers are meant for stations that have a server room, it would be analogous to the telecoms room.  

Every map is encouraged to have outposts, which are analogous to Sec Checkpoints

Boxing Ring, Reporter Office, Psych Office, etc are on almost every map. They will just be added manually if there isn't a pre-made beacon for them. Why wouldn't we make this simpler by just having them?

deltanedas:
> Escape pods on maps would make them easier for players to find.

no reason not to do it upstream

IamVelcroboy:
> no reason not to do it upstream

A couple of these could probably be upstreamed, sure. Some I don't think they'd use since they don't map the same as us, or use the RP roles as much.

deltanedas:
escape pods specifically every map has both here and usptream

IamVelcroboy:
Forgot about Mantis 😩. I'm at work now though.

IamVelcroboy:
Added mantis and shortened most of the names.

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
IamVelcroboy: Do you still want me to remove these changes in favor of the edits you made to the other file? I made them under the assumption this was the more favored method. But if you would rather not have them separated, it doesn't matter.

DebugOk: If you're gonna move them to their own files, revert what I did in the base files and I'll accept it

DebugOk (changes requested):

IamVelcroboy:
OK, made the changes. Shortened a few of the names.

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
```diff
@@ -0,0 +1,31 @@
+station-beacon-epistemics = Epistemics
+
+station-beacon-logistics = Logistics
+station-beacon-logistics-reception = Logistics Reception
+station-beacon-lo = LO
```
DangerRevolution, line 6: This is something I've mentioned in #940 and #942 , a new player (the target audience for this map) doesn't know what an LO is, at best they might know the abbreviation to "QM".
Something like:
```suggestion
station-beacon-lo = Logistics Officer's Quarters
```
is clearer for everyone and hopefully might get people to start picking up the Logi terminology over calling it Cargo

Resources/Prototypes/DeltaV/Entities/Objects/Devices/station_beacon.yml
```diff
@@ -0,0 +1,136 @@
...
+  - type: NavMapBeacon
+    text: station-beacon-virology
+
+- type: entity
+  parent: DefaultStationBeaconMedical
+  id: DefaultStationBeaconMetempsychosis
+  suffix: Cloning #Metempsychosis
+  components:
+  - type: NavMapBeacon
+    text: station-beacon-metempsychosis #change locale file when metem added
```
DangerRevolution:
```suggestion
    text: station-beacon-metempsychosis
```

DangerRevolution (commented):
Very exciting, very useful, very cool to stop having to rename all the beacons.

+ as in #940 and #942 would be very nice to avoid abbreviations that new players (the target audience for maps) and Wizdenners (the second audience for these) won't get

Resources/Locale/en-US/deltav/navmap-beacons/station-beacons.ftl
IamVelcroboy: Trying to keep the names short to prevent clogging up the map. All the other heads use abbreviations. So, this would be consistent.

DebugOk (dismissed):
As long as you aren't going to make the station maps extremely crowded, I'm fine with this

DebugOk (approved):

PR #789, opened 2024-02-07
Fix WT550 Safe Ammo

rosieposieeee:
image

One-line YML change. Tested and fires. Yippee!

**Changelog**

:cl:
- fix: Fixed the ammo in the WT550 safe not being top-mounted.


FluffiestFloof:
This should probably be upstreamed. If not, there should be a change marker since this touches wizden files

rosieposieeee:
eh fair enough

PR #792, opened 2024-02-08
Merge wizden up to 08/02

DebugOk:
I may have stopped maintaining after some internal drama, but I wont let this fork die because of it


IamVelcroboy:
Is it possible to skip the one that removes vehicles?

DebugOk:
They have a good reason, so no

PR #805, opened 2024-02-09
secret weight tweaks

JoeHammad1844:
## About the PR

new secret weights:
    Survival: 0.30
    Nukeops: 0.20
    Zombie: 0.10
    Traitor: 0.40
feel free to discuss these in the comments
## Why / Balance

survival is an overall boring round, antagonists are supposed to drive the round instead of it just degrading on its own, yeah theres a bunch of events but that happens in other game mods regardless. i get the point is to roleplay but traitors can also roleplay, and usually roleplay more interesting things like hostage taking

zombies is also just too rare, ive played over 1000 hours on wizden where zombies wight is also 10% and its still once in a blue moon, letting a rare game mode happen more often is fine

**Changelog**
:cl: JoeHammad
- tweak: Secret weights tweaked
- tweak: Traitor: 40%
- tweak: Survival: 30%
- tweak: Nukeops: 20%
- tweak: Zombie: 10%

### DIFF
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -1,9 +1,9 @@
 - type: weightedRandom
   id: Secret
   weights:
-    Survival: 0.44
-    Nukeops: 0.14
-    Zombie: 0.03
-    Traitor: 0.39
+    Survival: 0.30
+    Nukeops: 0.20
+    Zombie: 0.10
+    Traitor: 0.40
     #Pirates: 0.15 #ahoy me bucko
 


DangerRevolution:
There's currently a 30% chance (20+10) that the Station's main event in a round is a deathmatch (Nukes / Zombies)

JoeHammad1844:
would 15% nukeops and 45% tot be better?

rosieposieeee:
hard agree, survival is far too common for a gamemode that almost always lacks any conflict beyond self antags and shitters

UnicornOnLSD:
+1, now I see it as quite the opposite of the survival rounds we had right on rebase, with a lot of midround antag turning the shift into a high octane rush to patch the the station until evac
Now it's more of either a minor annoyance here and there (slimes, tarantulas, revenants, hell even listening post despite all the good work that was done on it), or a decent bit of self-antaggery going on because people are too bored (or shitters, but there's been a decent activity recently which doesn't really represent the whole issue with survival round from what I can see)

JoeHammad1844:
> +1, now I see it as quite the opposite of the survival rounds we had right on rebase, with a lot of midround antag turning the shift into a high octane rush to patch the the station until evac Now it's more of either a minor annoyance here and there (slimes, tarantulas, revenants, hell even listening post despite all the good work that was done on it), or a decent bit of self-antaggery going on because people are too bored (or shitters, but there's been a decent activity recently which doesn't really represent the whole issue with survival round from what I can see)

survival was removed on wizden for this very reason, and i agree with it

Guess-My-Name:
I dislike zombies as a gamemode, in my opinion it is worst of all, worse than revolutionaries ever were, so -1 for that reason alone.
And as far as decreasing chances for survival, I would be okay with that, but the reason most people state here is that there is no challenge at all during those rounds. Perhaps increasing chances for mid round antags would be a better solution?
In [events.yml](https://github.com/DeltaV-Station/Delta-v/blob/master/Resources/Prototypes/GameRules/events.yml#L100) we have decreased weighs of dragon, lone op and ninja compared to upstream, we also have no more rat kings from mouse migrations.

Resources/Prototypes/secret_weights.yml
```diff
@@ -1,9 +1,9 @@
 - type: weightedRandom
   id: Secret
   weights:
-    Survival: 0.44
-    Nukeops: 0.14
-    Zombie: 0.03
-    Traitor: 0.39
+    Survival: 0.30
```
DangerRevolution, line 4: `Survival: 0.35`

Resources/Prototypes/secret_weights.yml
```diff
@@ -1,9 +1,9 @@
...
+    Nukeops: 0.20
+    Zombie: 0.10
+    Traitor: 0.40
```
DangerRevolution, line 7: `Traitor: 0.35`

DangerRevolution (commented):
In my opinion :O

JoeHammad1844:
i disagree

DebugOk:
The entire weights will need to be put under a looking glass with admins before we change anything. Before anything is submitted, it'll need to be drafted and put through community feedback.

PR #817, opened 2024-02-11
uncomment terminator 2: judgement day

deltanedas:
## About the PR
if pro is not maintaining then :trollface:


VMSolidus:
I don't think we have any intention of re-enabling Terminators anytime soon, at least not in their current state. We feel that the current implementation of terminators is unsuitable for an MRP environment. Something similar, but more suitable perhaps, would be to rebase the Paradox Anomalies/Evil Twin. Since Debug has already stated that we will not be enabling Terminators, I will be closing this PR.

deltanedas:
ok bet

PR #818, opened 2024-02-11
Update server config

DebugOk:
This should hopefully fix shuttles causing a massive FPS drop while I investigate further


deltanedas:
rest in piss salv but at least we wont have 5 fps anymore

PR #821, opened 2024-02-12
Fix PseudoItems

DebugOk:
## About the PR
Really only used by felinids but now it properly works with gridinv

They wont fit into any bags outside of dufflebags, and leave a 2x2 and a 5x2 space available for other items

Didn't test further

## Why / Balance
Bag felinids

## Technical details
I wrote this code while *slightly* drunk, so who knows how horrible some of my shitcode is. 99% of it is just old pseudoitem code anyways

All I really did was fix the ability to set the grid size and pass through StoredOffset so you can avoid wacky offsets with wacky grids

## Media
image

## Breaking changes

**Changelog**

:cl:
- add: Felinids now fit in bags again!


Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
```diff
@@ -1,174 +1,43 @@
...
+        if (!args.CanInteract || !args.CanAccess)
+            return;
+
+        if (component.Active)
+            return;
+
+        if (!TryComp<StorageComponent>(args.Using, out _))
+            return;
+
+        // There *should* be a check here to see if we can fit, but I'm not aware of an easy way to do that, so eh, who cares
```
DEATHB4DEFEAT: `SharedStorageSystem.CanInsert` should work

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+
+public class SharedPseudoItemSystem : EntitySystem
+{
+    [Dependency] private readonly SharedStorageSystem _storageSystem = default!;
+    [Dependency] private readonly SharedItemSystem _itemSystem = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly TagSystem _tagSystem = default!;
+
+    [ValidatePrototypeId<TagPrototype>]
+    private const string PreventTag = "PreventLabel";
```
DEATHB4DEFEAT, line 23: What does this tag do?

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.cs
```diff
@@ -0,0 +1,155 @@
...
+using Robust.Shared.Containers;
+
+namespace Content.Shared.Nyanotrasen.Item.PseudoItem;
+
+public class SharedPseudoItemSystem : EntitySystem
+{
+    [Dependency] private readonly SharedStorageSystem _storageSystem = default!;
+    [Dependency] private readonly SharedItemSystem _itemSystem = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly TagSystem _tagSystem = default!;
```
DEATHB4DEFEAT:
```suggestion
    [Dependency] private readonly SharedStorageSystem _storage = default!;
    [Dependency] private readonly SharedItemSystem _item = default!;
    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
    [Dependency] private readonly TagSystem _tag = default!;
```

DEATHB4DEFEAT (dismissed):

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.cs
DebugOk: Stops you from applying labels using the hand labeler. Once people found this out people abused the crap out of it

Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
DebugOk: That requires you to have an item. Felinids aren't considered an item until past that. I could adapt the check, but I'll do that tomorrow

Content.Shared/Item/ItemComponent.cs
```diff
@@ -1,4 +1,5 @@
 using Content.Shared.Hands.Components;
+using Content.Shared.Nyanotrasen.Item.PseudoItem;
```
VMSolidus, line 2: This new using needs to be marked as //DeltaV

Content.Server/Nyanotrasen/Item/PseudoItem/PseudoItemSystem.cs
DebugOk: I rewrote the entire fucking check (or well, copied big chunks of it and replaced stuff here and there) to work with pseudoitem, not fun but it works

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.Checks.cs
```diff
@@ -0,0 +1,165 @@
+﻿using Content.Shared.Item;
+using Content.Shared.Storage;
+
+namespace Content.Shared.Nyanotrasen.Item.PseudoItem;
+
+/// <summary>
+/// Almost all of this is code taken from other systems, but adapted to use PseudoItem.
+/// I couldn't use the original functions because the resolve would fuck shit up, even if I passed a constructed itemcomp
```
deltanedas, line 8: creating a new component to pass into :
1. would fail resolve since its not actually added to the entity
2. extremely sus

im guessing this is just to make it so a felenid cant be picked up instantly, would be better to just have that be an event raised for ItemComponent that felenids cancel, so they still reuse all the logic but cant just left click to pick up like a mouse

DebugOk: Pseudoitems lack the itemcomponent when they're not inside a container because it brings a ton of unwanted features along with it. Even just adding it to run the check could result in weird shit happening if someone knew enough. This approach works "fine" for what its supposed to do

DebugOk: The entire reason I cant use the original functions as is is because they expect the item to have an itemcomponent, and theres no way to tell it to use anything else, even if it doesn't directly do anything with said component

Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
```diff
@@ -10,9 +10,8 @@
 using Content.Shared.Implants.Components;
 using Content.Shared.Interaction;
 using Content.Shared.Item;
-using Content.Shared.Item.PseudoItem;
 using Content.Shared.Lock;
-using Content.Shared.Nyanotrasen.Item.Components;
+using Content.Shared.Nyanotrasen.Item.PseudoItem;
```
VMSolidus, line 14: DeltaV usings need to be marked.

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.cs
```diff
@@ -0,0 +1,158 @@
...
+        var ev = new PseudoItemInsertDoAfterEvent();
+        var args = new DoAfterArgs(EntityManager, inserter, 5f, ev, toInsert, toInsert, storageEntity)
+        {
+            BreakOnTargetMove = true,
+            BreakOnUserMove = true,
+            NeedHand = true
+        };
+
+        _doAfter.TryStartDoAfter(args);
+    }
```
VMSolidus, line 158: We'll need to add a check for OnAttackAttempt, so that Felinids can't attack people near/holding the bag.

VMSolidus (changes requested):

Content.Shared/Item/ItemComponent.cs
DebugOk: Imports aren't marked

Content.Shared/Nyanotrasen/Item/PseudoItem/SharedPseudoItemSystem.cs
DebugOk: Fixed. The client will still predict the attack and play the sound/animation in the case of melee hits, but nothing will actually register

VMSolidus (approved):

PR #824, opened 2024-02-12
Prevent probers from being unachored by gorilla gauntlets

DebugOk:
## About the PR
Does what it says on the tin. If you try punting a prober, it'll shock you and some random shit around it

## Why / Balance
Dangerous probers should stay dangerous unless destroyed or if glimmer is lowered.

## Technical details
This makes it so `AttemptMeleeThrowOnHitEvent` is also raised on the entity thats getting thrown, and `GlimmerReactiveSystem` will cancel this event if glimmer is high enough, along with shocking the player and shooting (mostly harmless) lightning

**Changelog**
:cl:
- tweak: Glimmer probers now react violently to being hit by gorilla gauntlets if glimmer is above 500


DebugOk:
It would be funny if it also destroyed the anomaly core in the gauntlet, but thats a bit too much effort for me

VMSolidus (approved):
I've finished testing and reviewing this PR, it looks good to me.

PR #832, opened 2024-02-13
Makes mail teleporter craftable.

Adrian16199:
To-do
- [X] Figure out why it gives an error and not make it when you do all steps and screwdriver it.
- [X] Figure out how to make it drop the board upon being deconstructed since currently it doesnt drop ANYTHING.

## About the PR

Makes a mail teleporter craftable and only provides the board to the LO's locker.
## Why / Balance

The point of this PR is to allow courier's to perhaps still be able to keep goin, even if the mail teleporter that was mapped was nuked/destroyed/vanished.

## Media

## Breaking changes

**Changelog**


Resources/Prototypes/Nyanotrasen/Entities/Structures/Machines/mailTeleporter.yml
```diff
@@ -56,3 +56,5 @@
           True: {visible: true}
           False: {visible: false}
   - type: PowerSwitch
+  - type: Machine
```
deltanedas, line 59: inherit constructiblemachine and/or add Construction component

Resources/Prototypes/Nyanotrasen/Entities/Objects/Devices/CircuitBoards/production.yml
```diff
@@ -44,3 +44,22 @@
...
+      state: supply
+    - type: MachineBoard
+      prototype: MailTeleporter
+      requirements:
+        Capacitor: 1
+        MatterBin: 1
+      materialRequirements:
+        Steel: 4
+        Cable: 4
+    - type: ReverseEngineering
```
deltanedas: i can already see this being reverse engineered to make 20 mail teleporters for either epic money or to drain cargos bank account automatically

also it has no recipe anyway

Adrian16199: You are right on that one, should probably make it non reverser-able. Even tho, yeah, no recipe. Might aswell just make sure.

DebugOk (commented):
You should follow Deltanedas's advise and parent to the constructible thing, beyond that it looks fine to me.

Adrian16199:
Uh, if its not what I have given then that makes it very confusing but I will attempt to understand. It does work as it should currently in the build but if its neccesary.

IamVelcroboy (dismissed):
I think this was already approved by Debug. Should be good to merge.

Adrian16199:
> I think this was already approved by Debug. Should be good to merge.

I do believe the issue that was given was resolved because I was a dummy and simply didnt inherited the buildable parent like deltanedas suggested which would have saved me so much time.

IamVelcroboy:
> > I think this was already approved by Debug. Should be good to merge.
> 
> I do believe the issue that was given was resolved because I was a dummy and simply didnt inherited the buildable parent like deltanedas suggested which would have saved me so much time.

Yep, I saw the change

IamVelcroboy:
Imma merge this as the existence of the boards doesn't really affect anything. We can figure out what to do with it from there. Likely map it into board vaults/CE office.

Axionyxx:
> Imma merge this as the existence of the boards doesn't really affect anything. We can figure out what to do with it from there. Likely map it into board vaults/CE office.

Does the teleporter queue respect the fact there can be multiple mail teleporters? Has this been tested to make sure it wont spawn the same mail twice?

Adrian16199:
> > Imma merge this as the existence of the boards doesn't really affect anything. We can figure out what to do with it from there. Likely map it into board vaults/CE office.
> 
> Does the teleporter queue respect the fact there can be multiple mail teleporters? Has this been tested to make sure it wont spawn the same mail twice?

As it currently stands, the only board that exists is if either someone maps it (Maybe could have puted a DO NOT MAP on it) or you disassemble the mail teleporter itself.
And no, i do not believe it respects multiple mail teleporters so it can spawn multiple mails at the same time if there would be more than one.
The point of this PR was to really make it less inconvenient when somethin explodes or destroys it that isnt a syndicate bomb due to the fact that it would eject the board to allow it to still be repairable. It cannot be researched nor Reversed engineerd.

DangerRevolution:
> > Imma merge this as the existence of the boards doesn't really affect anything. We can figure out what to do with it from there. Likely map it into board vaults/CE office.
> 
> Does the teleporter queue respect the fact there can be multiple mail teleporters? Has this been tested to make sure it wont spawn the same mail twice?

It spawns double the mail from memory. We discussed this on Discord and so this will be put in the board vault & CE room to try negate whatever mailmaxxing Logistics gets up to. Still an issue though.

Axionyxx:
I don't think this should've been merged if it could allow for mail duplication

Adrian16199:
Im fine with it being unmerged if its neccesary.
It only happens extremely rarely that mail gets destroyed so this PR was a simple Hypothetically dot TM when it does happen to be able to repair it but if it gives an idea of abuse then thats a no bueno.
At the least I learnt on how constructions function and how to use it properly.

IamVelcroboy:
> I don't think this should've been merged if it could allow for mail duplication

It can't do anything on its own. These are only accessible if we make them accessible. That's what we were considering from here.

Axionyxx:
> > I don't think this should've been merged if it could allow for mail duplication
> 
> It can't do anything on its own. These are only accessible if we make them accessible. That's what we were considering from here.

Ah, my mistake. All good then

deltanedas:
it should be fixed to pick a random/alternating teleporter when multiple exist before mapping boards. cargo WILL use this to get double mail otherwise

PR #837, opened 2024-02-15
Vulpkanin's Deuteranopia

FluffiestFloof:
Everyone loves dog vision :trollface: 

## About the PR

Adds Deuteranopia as a selectable trait.
Adds Deuteranopia to Vulpkanin by default. It can be reverted using the Normal Vision trait similarly to the Harpy UltraVision.
Make Normal Vision and Ultraviolet Vision trait names into locale.

## Why / Balance

Why not, Harpy gets a default vision and it make sense for Vulpkanins.
I'm sure a lot of people won't like it, even less when the Metempsychotic machine gets readded and they roll into a Vulpkanin.

## Technical details

I don't think I *really* needed to remove all those funny lines but the UltraVision doesn't need em and since you can't really gain/lose Deuteranopia anyway outside of selecting the trait or being Vulpkanin...

## Media

ss (2024-02-15 at 11 41 31)

- [yes] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Added Deuteranopia trait.
- tweak: Vulpkanin now start with Deuteranopia by default which can be removed using the Normal Vision trait.

### DIFF
--- a/Content.Client/Nyanotrasen/Overlays/DogVisionSystem.cs
+++ b/Content.Client/Nyanotrasen/Overlays/DogVisionSystem.cs
@@ -1,13 +1,10 @@
 using Content.Shared.Abilities;
-using Robust.Client.GameObjects;
 using Robust.Client.Graphics;
-using Robust.Client.Player;
 
 namespace Content.Client.Nyanotrasen.Overlays;
 
 public sealed partial class DogVisionSystem : EntitySystem
 {
-    [Dependency] private readonly IPlayerManager _player = default!;
     [Dependency] private readonly IOverlayManager _overlayMan = default!;
 
     private DogVisionOverlay _overlay = default!;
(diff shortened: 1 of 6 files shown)


AitorLogedo:
the piss vision

VMSolidus:
Yea, removing those lines is fine. Ultravision originally copied all the Dogvision parts, but within seconds of having it up, I realized, "Half this code is unused, what the hell?", and cleaned it up by stripping out everything unneeded. I would have gone back and cleaned up dogvision too, had I not been concerned about scope creep.

This is fine.

VMSolidus:
In fact removing all those lines doesn't change the feature functionality in any way. They're totally vestigial. The removed code was in fact ADDING a bug by inherently tying component initialization/deinit to having a player session, thus making it so that taking over a creature with DogVision meant you wouldnt have it since it didnt initialize on your player session.

I am strongly considering adding an implant that just gives you the NormalVision component.

FluffiestFloof:
> In fact removing all those lines doesn't change the feature functionality in any way. They're totally vestigial. The removed code was in fact ADDING a bug by inherently tying component initialization/deinit to having a player session, thus making it so that taking over a creature with DogVision meant you wouldnt have it since it didnt initialize on your player session.
> 
> I am strongly considering adding an implant that just gives you the NormalVision component.

Nice good to know I pre-emptively fixed something.

DebugOk (approved):

PR #838, opened 2024-02-15
Glimmer Drain requires 5 crystal again

FluffiestFloof:
## About the PR

Uncommented the previous requirement for drainers so it goes from 1 to 5 crystals again.

## Why / Balance

Before Glimmer Mite it was impossible to actually get ectoplasm before it was too late, now you can literally get *almost* 25u Ectoplasm per Glimmer Mite (if you draw their blood + grind them) and they spawn between 50 to 900 glimmer.

This makes it so you cannot just create 4 drainer with one mite and instead requires you to extract 2 mites to create 1 Drainer.

The Glimmer Drain was never meant to be easy to make AFAIK so yeah.

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: Glimmer Drain now requires 5 Normality Crystals to craft.

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/Recipes/Construction/Graphs/structures/glimmerdevices.yml
+++ b/Resources/Prototypes/Nyanotrasen/Recipes/Construction/Graphs/structures/glimmerdevices.yml
@@ -30,36 +30,36 @@
           conditions:
             - !type:EntityAnchored {}
           steps:
-            - tag: NormalityCrystal # DeltaV - Temporarily makes Drain only need 1 Crystal until Golemnization is implemented otherwise it's just imposible to craft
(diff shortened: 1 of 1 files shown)


AeraAuling:
I've noticed that typically, you're only going to see one glimmer mite per shift, and it's only got enough blood for three crystals (supposing you grab it, or kill it without it bleeding) outside of using a very cheesy method to keep it alive when like, one bloodloss damage will just straight up kill it. Three crystals would be a bit more reasonable, that way you can't just get three drainers out of the mite, but it's still a reward for tracking it down and extracting the ectoplasm without losing any of it.

VMSolidus (approved):

DebugOk (approved):

PR #839, opened 2024-02-15
fix salvage strings

IamVelcroboy:
## About the PR
Fixes and obscures salvage debris names that show up in the magnet. 

## Why / Balance
Be best


DebugOk:
Too many changes, should be done via code

PR #840, opened 2024-02-15
*Properly* fix salvage strings

DebugOk:
Yes, I have confirmed via scsi the salvage system does respect abstracted prototypes


VMSolidus (approved):

DebugOk:
Bypassing velcro's review since yes

PR #841, opened 2024-02-15
New salv naming

IamVelcroboy:
## About the PR
Fixes and obscures salvage debris names that show up in the magnet.


DebugOk:
#840 takes priority because it seperates the files

PR #843, opened 2024-02-16
Adjust suffix for red light

IamVelcroboy:
## About the PR
Makes a small change for mapping QoL

## Why / Balance
couldn't tell the difference while mapping


VMSolidus (approved):
Technically we own this code, so you don't actually need to comment that this is a DeltaV change. We are Nyano for all intents and purposes now. Otherwise this is fine.

DebugOk:
> Technically we own this code, so you don't actually need to comment that this is a DeltaV change. We are Nyano for all intents and purposes now. Otherwise this is fine.

I've honestly considered moving everything from /Nyanotrasen into /DeltaV for whenever they start existing again, as it might make their life as a downstream easier. But I'll have to do some talks about that, obviously cant *just* do that as its technically not delta code by origin

PR #846, opened 2024-02-17
Energy Gun Rebalancing

Tryded:
## About the PR

Made it so energy guns charge 2 times as fast as they originally were.
## Why / Balance

Energy guns before took a really long time to recharge because of the high energy requirements. This PR makes it more like the pre-existing laser guns
## Technical details

YAML gaming
## Media

## Breaking changes

**Changelog**

:cl: Tryded
- tweak: Changed Energy Guns' charging time to half of what it originally was to be similar to normal laser guns.


DebugOk (commented):
Balance really goes way beyond me, @VMSolidus are you able to review this?

VMSolidus:
Yea, this works fine. It neither changes the DPS of the energy guns, nor their ammo supply since the ratio of battery cap to fire cost is preserved. This is just making it so that it takes the same amount of time to charge them as their lasgun equivalents. IMO that's fine, I've never actually felt like even if it charged as fast as a lasgun, it still wouldn't actually save me in any situation where that's actually going to matter(It's almost like a full minute to charge a lasgun.)

PR #847, opened 2024-02-17
Cap Gun Crate

dge21:
## About the PR

Added a new crate that cargo can purchase containing a cap gun, two speed loaders, and two boxes of cartridges, this time in the proper files
## Why / Balance

The only way to get more cap gun cartridges is to buy a toy box for 900 spesos and only get six cartridges and a bunch of other shit you might not want to deal with. It's a massive waste of resources.
## Technical details

It's a small yml addition, so it should be easy enough to edit if the price is too low.
## Media

## Breaking changes

**Changelog**

:cl:
- add: Added a new crate to delight all the clowns and mimes.


dge21:
I don't know what any of that means

PR #849, opened 2024-02-17
Psychologist & Lawyer stamp

Guess-My-Name:
## About the PR

two new stamps for better RP
added those to starting inventories as well as cmo locker (psychologist) and lawdrobe (lawyer), in case of in round promotions.
## Why / Balance

currently there's no way to make a paper impossible to change without a stamp, lawyer has to ask hop for one, psychologist has to ask hop or cmo

## Media

stampShowcase3
stampShowcase2
stampShowcase1

## Breaking changes

**Changelog**

- add: Added psychologist and lawyer stamps!


Resources/Locale/en-US/paper/stamp-component.ftl
```diff
@@ -15,3 +15,5 @@ stamp-component-stamped-name-warden = Warden
 stamp-component-stamped-name-trader = Trader
 stamp-component-stamped-name-syndicate = Syndicate
 stamp-component-stamped-name-ce = Chief Engineer
+stamp-component-stamped-name-lawyer = Lawyer
+stamp-component-stamped-name-psychologist = Psychologist
```
VMSolidus: We don't need to modify upstream localization files. 

These two lines need to be moved to a Resources/Locale/en-US/DeltaV/paper/ folder(You'll have to make a folder for this). You can basically copy the stamp-component.ftl file there, and delete everything but the lawyer and psychologist localizations.

VMSolidus (changes requested):

Resources/Locale/en-US/paper/stamp-component.ftl
Guess-My-Name: forgot we had our own locale folder as well

dvir001:
You can use the frontier sprites if you want
https://github.com/new-frontiers-14/frontier-station-14/pull/655
As we had the same PR not too long ago.

VMSolidus (approved):
This looks good now. If only stamps allowed setting a different RSI, so that we could avoid having to modify the upstream stamps. Eh that'll be something to PR to upstream anyway.

DebugOk (approved):
Its a shame this requires altering the upstream RSI, but it is what it is. It could be fixed at some point but eh

PR #856, opened 2024-02-18
Remove old syndicate cyborg

FluffiestFloof:
## About the PR

Remove old half reimplemented and commented out syndicate cyborg since upstream added one to the uplink already that serves the same purpose.
Mostly a suggestion but since it's literally just comments.


DebugOk:
Thank you

DebugOk (approved):

PR #858, opened 2024-02-19
Replace the new mentions of QM

FluffiestFloof:
QM isn't real.
Also replaces the two mention of RD I found with Mystagogue

Also now we have 2 LO beret.
Do we want both or yeet one of them? DV one spawns on the LO and the Wizden one is in the dresser

### DIFF
--- a/Resources/Prototypes/Datasets/ion_storm.yml
+++ b/Resources/Prototypes/Datasets/ion_storm.yml
@@ -342,8 +342,12 @@
   - MIMES
   - PARAMEDICS
   - PASSENGERS
-  - QUARTERMASTERS
-  - RESEARCH DIRECTORS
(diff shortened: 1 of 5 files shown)


Resources/Prototypes/Datasets/ion_storm.yml
```diff
@@ -342,8 +342,12 @@
...
   - PARAMEDICS
   - PASSENGERS
-  - QUARTERMASTERS
-  - RESEARCH DIRECTORS
+  # DeltaV - Logistics Department - Replace Quartermaster with Logistics Officer
+  - LOGISTICS OFFICERS
+  # End of modified code
+  # DeltaV - Epistemics Department - Replace Research Director with Mystagogue
+  - MYSTAGOGUES
+  # End of modified code
```
rosegayming, line 350: Could this just be one Delta-V block?

PR #863, opened 2024-02-19
Updates shoukou up to date.

Adrian16199:
## About the PR

Updates shoukou:
Redid captain's quarters to provide more space for the captain's antique laser's gun display. 
Replaced cloning with metem cloning.

Most importantly, fixes that one light in the substation's room behind medical to not be on the door.

## Why / Balance

Because wizden merge.
## Media

## Breaking changes

**Changelog**


IamVelcroboy:
This doesn't look like you pulled from the map directory, which would include the changes from [this](https://github.com/DeltaV-Station/Delta-v/pull/833)

Adrian16199:
Well thats just lovely then because I updated the master before doin mappin.

Adrian16199:
I could attempt again, just a flex that it just didnt though that it wouldnt update the map.

IamVelcroboy:
When you load a map in mapping it loads from the bin/data folder not the resources/maps (unless you direct it to do so specifically.) So, if changes were made on the map since last time you updated, you need to move the map from resources/maps over to bin/data. (or do like mapping ## /resources/maps/map.yml)

Adrian16199:
That just makes me think that many updates were neglected because i didnt though of that. Buh

IamVelcroboy:
The only other edits I can see that I made were adding the vault freezer and spare ID cabinet. As long as you still have those somewhere, should be good. Just duplicate the changes you made on the file after the light overhaul I did last week and should be ok.

Adrian16199:
I know that the ID cabinet is DEFINITELY there since I had to give a biscuit to it once, Freezer is still somethin I dont understand when it comes to its existance but its probably there aswell.
Will probably redo this tomorow tho.

IamVelcroboy:
Ok, no problem. Sorry for the confusion.

Adrian16199:
Need to redo it.

PR #864, opened 2024-02-19
Make loadacts more user friendly

DebugOk:
## About the PR
Helps Velcro out, and makes it easy for mappers to use his acts without having to do anything complicated.

To load from a file stream, the entries must be laid out like this
```yaml
- action: !type:WhateverActionComponent
  assignments:
  - 0: SLOT
  name: Action's Name
```
If an action lacks a name, it wont work through the open file dialogue, but it'll still work if the file name is passed directly. No additional checks are done to make sure the action is valid


IamVelcroboy:
Decent

DEATHB4DEFEAT (approved):

PR #871, opened 2024-02-20
Cap Gun Crate

dge21:
## About the PR

Added a new crate that cargo can purchase containing a cap gun, two speed loaders, and two boxes of cartridges, this time in the proper files and additionally not from the master branch
## Why / Balance

The only way to get more cap gun cartridges is to buy a toy box for 900 spesos and only get six cartridges and a bunch of other shit you might not want to deal with. It's a massive waste of resources.
## Technical details

It's a small yml addition, so it should be easy enough to edit if the price is too low.
## Media

## Breaking changes

**Changelog**

:cl:
- add: Added a new crate to delight all the clowns and mimes.

### DIFF
--- a/Resources/Prototypes/DeltaV/Catalog/Cargo/cargo_fun.yml
+++ b/Resources/Prototypes/DeltaV/Catalog/Cargo/cargo_fun.yml
@@ -8,17 +8,12 @@
   category: Fun
   group: market
 
-- type: entity
-  name: cap gun box set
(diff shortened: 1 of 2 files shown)


VMSolidus (approved):
This is probably fine.

VMSolidus:
Double checked to make sure the box wasn't more valuable than its contents. To my surprise the box + contents was worth 340, but cost 1000 to order. I then briefly considered making it cost only 500, then realized that would encourage people to buy a ton of them. 1000 is fine.

PR #874, opened 2024-02-20
Make periapsis only have 10 non-whitelisted slots, and make whitelist not shit

DebugOk:
## About the PR
Another server is coming to replace it

Inspired by Cosmatic drift's system, but in my own way

## Technical details
Whitelist now works based on slots instead of the hard player count. The `whitelist.max_players` is now ignored, and `whitelist.min_players` decides how many slots there are.

If it is set to 10, 10 non-whitelisted players will always be able to join, regardless of how many whitelisted players there are. With the execption being there is still space within the soft player cap

**Changelog**
:cl:
- tweak: Whitelist to join now works based on slots instead of player count
- tweak: Periapsis now has only 10 non-whitelisted slots


DebugOk:
Fuck it we merge

PR #876, opened 2024-02-20
Remove antique laser from captain's locker

IamVelcroboy:
## About the PR
This doesn't need to be in the locker.


deltanedas:
theres a nolaser variant from upstream why not use it

IamVelcroboy:
> theres a nolaser variant from upstream why not use it

Because then you are forced to map the hardsuit locker and I don't want to make a fourth captain locker w/ the hardsuit but w/out the laser. We're just going to make it a rule on DV to map the glass box and leave it mapper's choice if they want the hardsuit in the locker or the suit storage.

deltanedas:
:trollface: i love locker permutations

PR #886, opened 2024-02-22
Fix deep fryer divide by zero

VMSolidus:
## About the PR
Someday I'm going to rip all this code up and replace it with better code. For now, fixes a divide by zero. Verified by running the deep fryer all the way to empty, while listening to the debugger. 

## Why / Balance
Bad Math

image


DebugOk:
> Someday I'm going to rip all this code up and replace it with better code.

Why, do you not love the 1000 line file that causes github to hang for 10 seconds as the page loads

DebugOk (approved):

PR #905, opened 2024-02-26
Adds DV fluff to captain's nolaser locker

IamVelcroboy:
## About the PR
Title

## Why / Balance
Since this is the main locker we are using next to the suit filled locker, it needs the DV drip.


deltanedas:
why arent these in the dresser instead

IamVelcroboy:
> why arent these in the dresser instead

The beret could maybe go in the dresser. I figured the shoes/boots maybe made more sense in the locker 🤷‍♂️

I also wasn't quite sure how much room was left in the dresser because gridinv makes them finite. Could always increase the storage but I didn't have time to test earlier. So, this seemed the better route. For now anyway.

PR #906, opened 2024-02-26
Replaces romerol pill w/ ambuzol

IamVelcroboy:
## About the PR
Title

## Why / Balance
Another remnant form the pre-romerol switcheroo days.


deltanedas:
i always use that to mass produce ambuzol plus in secret salv lab :trollface:

PR #914, opened 2024-02-28
Port Nyano's FTL sound

DebugOk:
Ports https://github.com/Nyanotrasen/Nyanotrasen/commit/cbd7104306bc0b03004036f91459a7fa42bc5c0f

I like that sound better, so why not. I haven't checked the encoding but I *assume* its mono like its supposed to be. It seems to be fine in game


ps3moira:
I like the FTL sound from Nyano, builds up that feeling when you're arriving somewhere.

PR #937, opened 2024-03-07
2024/03/05 merge master

DebugOk:
Untested, unfinished. Theres other commits still up for grabs and the trading post still needs to be partially copied due to missing out on a lot of content because of it.

Most notably, the shuttle ui refactor is missing, due to trading post FTL changes. I may still take care of that, but ehh.

I also dont believe tests will pass, and thats fine. If they do, great, otherwise I'll fix them whenever I feel like it.


DebugOk:
Yaml linter passed locally, thats the only test I ran

deltanedas:
i thought deltav still uses brig access for lawyers

IamVelcroboy:
> i thought deltav still uses brig access for lawyers

No. we have a Lawyer access for that.

DebugOk:
All tests are now green locally
image

PR #945, opened 2024-03-08
German Accent

DangerRevolution:
## About the PR

This accent is a direct 1:1 from YogStation (linked in the .ftl), with minor tweaks. For future proofing and edge-case roleplay scenarios I've kept in references to Antags that aren't here in Station 14.

Gonna be honest, I haven't read the entire list yet.

This PR mostly relied on Bing AI to write this up for me. I'm not insane or a masochist so I didn't write up 400 word combinations for the accent and got Bing to do it for me.

## Why / Balance

I've seen a few people over the last week roleplaying Germans, with the accent and all. Thought it'd be cool to make them an accent. 

## Technical details

1:1 copy of the Scottish accent trait, just for German.

## Media

image

Please note: I'm not German, this is a north German-styled accent and I would appreciate some feedback for words that could be better etc etc.

**Changelog** DangerRevolution
:cl:
- add: Added the German Accent trait


VMSolidus:
Okay, so. I would really like to apologize for this. I had never actually read the code for Accent traits, and had mistakenly assumed that the Accent traits would have been coded in a sane and reasonable way. In the case of a German accent, it's most easily done by using exactly 7 lines of String.Replace. 

You can pretty easily replace Line 24 with about 6 or 7 lines of this specific Dotnet 8.0 function. Luckily Microsoft's own website for Dotnet 8.0 comes with very helpful code examples for how to use the function. In this context, the Variable "message" is a string that you can put directly into String.Replace. 

https://learn.microsoft.com/en-us/dotnet/api/system.string.replace?view=net-8.0

Doing so will make the entire combined 800 lines of accent word replacements obsolete. Which if you did so, could cut this PR down from +896 to only +44 to +50 lines. 

One more thing, please don't rely on AI to make this translation, in this context it is a genuinely awful idea as it makes the speaker sound like a stereotype of someone with a German accent, which can be seen as fairly offensive. I've went and found an article that describes specific sound shifting techniques for how to portray a realistic German accent, and conveniently enough most of it is done by replacing sounds(letters) with other sounds(letters). This is something that String.Replace can do quite handily. 

https://www.backstage.com/magazine/article/how-to-do-a-german-accent-75917/ 

I'm going to go ahead and draft this PR until these kind of changes are made.

Content.Server/DeltaV/Speech/EntitySystems/GermanAccentSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        SubscribeLocalEvent<GermanAccentComponent, AccentGetEvent>(OnAccentGet);
+    }
+
+    // converts left word when typed into the right word. For example typing you becomes ye.
+    public string Accentuate(string message, GermanAccentComponent component)
+    {
+        var msg = message;
+
+        msg = _replacement.ApplyReplacements(msg, "german");
```
VMSolidus, line 24: This line specifically can be replaced with a couple uses of String.Replace, which would make the Replacements unnecessary entirely.

VMSolidus (changes requested):

stellar-novas:
> One more thing, please don't rely on AI to make this translation, in this context it is a genuinely awful idea as it makes the speaker sound like a stereotype of someone with a German accent, 

AI was not used to generate the replacement list, it was used to convert the Yogstation json to ftl.

VMSolidus:
> > One more thing, please don't rely on AI to make this translation, in this context it is a genuinely awful idea as it makes the speaker sound like a stereotype of someone with a German accent,
> 
> AI was not used to generate the replacement list, it was used to convert the Yogstation json to ftl.

My mistake, however it well.. I'll be honest the SS13 forks aren't exactly known for having sane traits. I very distinctly remember Yoggstation having a Tourettes trait that was literally the worst possible portrayal of Tourettes. I wouldn't consider porting their code to generally be a good idea, as much of it is in extremely poor taste.

Content.Server/DeltaV/Speech/EntitySystems/GermanAccentSystem.cs
DebugOk: This how accent replacements should be done

DebugOk:
I'm making the decision that no more accents should be added, and I believe this to be too stereotypical. I dislike these sorts of accents in general and think they're of poor taste.

PR #950, opened 2024-03-10
Makes ifriit immune to zombie-fication.

Adrian16199:
Because uh.... higher being.... not really flesh.... pure fire.... gettin zombified? what type of horror is this when it comes to diseases? Now you gonna tell me you can zombify a computer.

**Changelog**

:cl: Adrian16199
- fix: Ifrits can no longer be zombified


DebugOk:
@VMSolidus stuff like this doesn't require a changelog entry

PR #952, opened 2024-03-11
Rules Update

GreggTheFather:
## About the PR

Updated rules 2.5, 2.10, 3.4, 4.8, 6.1, 8.4.
## Media

## Breaking changes

**Changelog**
EDIT: Forgot to add what exactly was changed/updated. Listing that now in order.
2.5: - Do not make yourself a major problem/annoyance/disruption for one specific crew member in a way that would actively detract from the other player's enjoyment of the shift.
2.10, Added onto EORG Definition:
 - End-of-round grief (EORG) is not allowed. This includes attacking, destroying, polluting, and injuring without reason both *at* and *on the way to* Central Command.
        * This also extends to the following:
            * Neutral parties, such as space dragons or sentient artifacts activating harmful nodes.
            * Attempts to block the docking ports or ports leading into Central Command
3.4: - Be respectful towards other players in LOOC and OOC channels, and avoid making others uncomfortable. This can range anywhere from starting arguments to personal attacks against others, depending on the context.
4.8: - Do not place players into cryosleep unless they have given consent to do so, they are fully catatonic, or they have been sentenced to preservative stasis. Always examine a character to double check if they are SSD or catatonic prior to placing them into cryosleep.
6.1, addition:     - Your master is held accountable if they order you to do something malicious.
        *This does not apply to breaking server rules. If you are ordered to do something that would break a server rule, disregard it and inform an admin.
8.4: - By picking prisoner, you have chosen to RP as a prisoner. You are still subject to rules pertaining to escalation, and should only seek to escape from the brig with good roleplay reasoning, such as abusive security, or a badly damaged perma. Escaping for no reason is considered a self-antag activity. If you are unsure whether your escape reason is valid, feel free to AHelp it first.

:cl:
- tweak: The server rules have been modified, be sure to read them!


therealDLondon:
Can you believe it guys, I helped. DLondon strikes again. Prisoners everywhere weep at this revelation

DebugOk (dismissed):
Colin would always edit the thing near the beginning with the summary, and this is a change that should absolutely include a changelog entry

Resources/ServerInfo/Rules.txt
```diff
@@ -40,12 +45,13 @@ If you have any questions about these rules, please use the admin help (ahelp) m
...
     - You are allowed to have knowledge of past experiences with someone else in prior shifts. This does not give you permission to hold a grudge against someone that results in you treating them differently in an unfair way (i.e. Meta-Grudging).
     - Specific players who were antagonists in previous rounds must not be treated differently because of it.
     - Do not "Antag Roll." This is the act of joining rounds for the purpose of seeing if you joined as an antagonist, and leaving soon after if not. Players who have a history of this behavior will have their whitelist revoked and/or face a ban.
     - [color=#ff0000]Do not stream the current round to the Delta-V Discord.[/color]
+    - Do not place players into cryosleep unless they have given consent to do so, they are fully catatonic, or they have been sentenced to preservative stasis. Always examine a character to double check if they are SSD or catatonic prior to placing them into cryosleep.
```
DangerRevolution, line 54: Not to be pedantic but given this will be read by new players this should probably explain the difference

i.e "... if they are SSD (yellow text when you inspect them) or catatonic (purple text when you inspect them)..." 

Or it should contain snippets of the associated texts. I see questions in #help sometimes of people asking the difference between the two or confusing the two.

PR #956, opened 2024-03-11
Fix some trading post oopsies

DebugOk:
Attempt number #_2_ at a station anchor without altering the map file, and readd a thing I removed while changing stuff


DebugOk:
Bounties are still fucked :/

PR #963, opened 2024-03-14
Removes a random `1` from SiliconPolice lawset

DangerRevolution:
## About the PR

Removed a rogue 1 from SiliconPolice ID

## Why / Balance

No other lawset has a 1


DangerRevolution:
This is the same branch as the one in https://github.com/DeltaV-Station/Delta-v/pull/953/files which was accepted, no other commits have changed since. Not sure why they're appearing again?

PR #965, opened 2024-03-14
Port frontier station 14's sizeattribute system

Mnemotechnician:
## About the PR
Frontier station has a sizeattribute system that allows **certain species** to change their size via two traits: small and tall.

This PR fully ports it, with some minor changes to adapt it for deltav.

Only 3 species are currently affected:
- Catgirls can become tall, becoming on par with humans, although losing the ability to fit in bags. Other than that, this change is cosmetic-only.
- Lizards can become tall, this is purely cosmetic
- Vulps can become small, which makes them on par with felinids. They can fit in bags, they get a smaller collision fixture, they get smaller mass (so harder to push things around).

## Why / Balance
This system, albeit being far from ideal, allows for some more additional character customization and should hopefully benefit roleplay. These changes should HOPEFULLY not have any noticable impact on the balance.

The tall trait is cosmetic-only for lizards and cats because it used to breed a lot of powergaming when it had effect on character's stats. Although mass contests are no longer a thing, it's still not a very good idea to give people a "make me stronger" button.

## Technical details
First off, the code sucks a lot (it was written by a rather newbie coder). But it worked for months on frontier, and I'd rather not touch something that works already. A few adjustments were made, though.

1. Every species that are supposed to be able to use the tall/small traits must have a SizeAttributeWhitelistComponent attached to them. The component specifies what changes are to be done to the character if they choose either of the two traits.
2. Choosing one of the traits adds a SizeAttributeComponent to your character. The SizeAttributeSystem then checks if the whitelist allows it, and does the neccessary changes, which currently can include:
    - Scaling the character (either just cosmetically, or, if shortCosmeticOnly/tallCosmeticOnly are false, then also physically - their fixture is also scaled, affecting their mass)
    - Changing the density of the character's collision fixture, affecting their mass (if non-cosmetic)
    - Adding or removing the pseudoitem comp to/from the character

Also this includes a new interface for components - ITransferredByCloning. Here, it's only used to transfer the SizeAttributeComponent when the person is cloned (because, for some reason, without this cloning doesn't preserve these traits correctly, causing, for example, small vulps to come out of the cloning pod normal-sized). However, it can be added to any other component to make it carry over when the person with it is cloned. This introduces a couple new lines to upstream's cloning system.

## Media
Github won't allow me to upload a 20 mb recording, so here's it uploaded to discord: https://cdn.discordapp.com/attachments/732665247302942730/1217934790528077925/weeee-2024-03-14_23.09.08.mp4?ex=6605d52a&is=65f3602a&hm=2b16795e3cae09aad16ec70a69742918a24efe5018822f207ec1a46db776ad12&

## Breaking changes
None? I hope.

**Changelog**

:cl:
- add: Two new traits were added: "short" and "tall". Currently, only the Vulpkanin can become shorter, and only Felinids and Reptillians can become taller.

### DIFF
--- a/Content.Server/Cloning/CloningSystem.cs
+++ b/Content.Server/Cloning/CloningSystem.cs
@@ -1,3 +1,4 @@
+using Content.Server._NF.Cloning;
 using Content.Server.Atmos.EntitySystems;
 using Content.Server.Chat.Systems;
 using Content.Server.Cloning.Components;
@@ -49,6 +50,8 @@ using Content.Server.EntityList;
 using Content.Shared.SSDIndicator;
 using Content.Shared.Damage.ForceSay;
 using Content.Server.Polymorph.Components;
+using Content.Server._NF.Cloning; //Frontier - needed for implementation of ITransferredByCloning
+using Robust.Shared.Serialization.Manager; // Frontier
 
 namespace Content.Server.Cloning
 {
@@ -78,6 +81,7 @@ namespace Content.Server.Cloning
         [Dependency] private readonly SharedJobSystem _jobs = default!;
         [Dependency] private readonly MetempsychoticMachineSystem _metem = default!; //DeltaV
         [Dependency] private readonly TagSystem _tag = default!; //DeltaV
+        [Dependency] private readonly ISerializationManager _serialization = default!; //Frontier -> deltav
 
         public readonly Dictionary<MindComponent, EntityUid> ClonesWaitingForMind = new();
         public const float EasyModeCloningCost = 0.7f;
@@ -248,10 +252,21 @@ namespace Content.Server.Cloning
             }
             // end of genetic damage checks
 
-            var mob = FetchAndSpawnMob(clonePod, pref, speciesPrototype, humanoid, bodyToClone, karmaBonus); //DeltaV Replaces CloneAppearance with Metem/Clone via FetchAndSpawnMob 
+            var mob = FetchAndSpawnMob(clonePod, pref, speciesPrototype, humanoid, bodyToClone, karmaBonus); //DeltaV Replaces CloneAppearance with Metem/Clone via FetchAndSpawnMob
 
             ///Nyano - Summary: adds the potential psionic trait to the reanimated mob.
             EnsureComp<PotentialPsionicComponent>(mob);
+
+            // Frontier - transfer of special components, e.g. small/big traits
+            foreach (var comp in EntityManager.GetComponents(bodyToClone))
+            {
+                if (comp is ITransferredByCloning)
+                {
+                    var copy = _serialization.CreateCopy(comp, notNullableOverride: true);
+                    copy.Owner = mob;
+                    EntityManager.AddComponent(mob, copy, overwrite: true);
+                }
+            }
 
             var ev = new CloningEvent(bodyToClone, mob);
             RaiseLocalEvent(bodyToClone, ref ev);
--- a/Content.Server/_NF/SizeAttribute/SizeAttributeSystem.cs
+++ b/Content.Server/_NF/SizeAttribute/SizeAttributeSystem.cs
@@ -1,0 +1,95 @@
+using System.Numerics;
+using Content.Server.SizeAttribute;
+using Robust.Server.GameObjects;
+using Robust.Shared.Physics;
+using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Physics.Systems;
+using Content.Shared.Item.PseudoItem;
+using Content.Shared.Nyanotrasen.Item.PseudoItem;
+
+namespace Content.Server._NF.SizeAttribute
+{
+    public sealed class SizeAttributeSystem : EntitySystem
+    {
+        [Dependency] private readonly IEntityManager _entityManager = default!;
+        [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+        [Dependency] private readonly AppearanceSystem _appearance = default!;
+        [Dependency] private readonly FixtureSystem _fixtures = default!;
+        public override void Initialize()
+        {
+            base.Initialize();
+            SubscribeLocalEvent<SizeAttributeComponent, ComponentInit>(OnComponentInit);
+        }
+
+        private void OnComponentInit(EntityUid uid, SizeAttributeComponent component, ComponentInit args)
+        {
+            if (!TryComp<SizeAttributeWhitelistComponent>(uid, out var whitelist))
+                return;
+
+            if (whitelist.Tall && component.Tall)
+            {
+                Scale(uid, component, whitelist.TallScale, whitelist.TallDensity, whitelist.TallCosmeticOnly);
+                PseudoItem(uid, component, whitelist.TallPseudoItem);
+            }
+            else if (whitelist.Short && component.Short)
+            {
+                Scale(uid, component, whitelist.ShortScale, whitelist.ShortDensity, whitelist.ShortCosmeticOnly);
+                // deltav: added PseudoItemShape
+                PseudoItem(uid, component, whitelist.ShortPseudoItem, whitelist.PseudoItemShape);
+            }
+        }
+
+        // deltav: added a shape argument and refactored the method
+        private void PseudoItem(EntityUid uid, SizeAttributeComponent component, bool active, List<Box2i>? shape = null)
+        {
+            if (active)
+            {
+                var pseudoItem = EnsureComp<PseudoItemComponent>(uid);
+                pseudoItem.Shape = shape;
+            }
+            else
+            {
+                if (!TryComp<PseudoItemComponent>(uid, out var pseudoI))
+                    return;
+                RemComp(uid, pseudoI);
+            }
+        }
+
+        private void Scale(EntityUid uid, SizeAttributeComponent component, float scale, float density, bool cosmeticOnly)
+        {
+            if (scale <= 0f && density <= 0f)
+                return;
+
+            _entityManager.EnsureComponent<ScaleVisualsComponent>(uid);
+
+            var appearanceComponent = _entityManager.EnsureComponent<AppearanceComponent>(uid);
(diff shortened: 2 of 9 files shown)


noctyrnal:
hope the other races get these traits in the future, would be cool

Adrian16199:
Im uncertain how it will be received when it comes to a vulpkanin being smaller like a felinid, a size change isnt only just mass contest but in combat aswel, smaller sprite always hides in bigger's sprite plus gaining the hiding in duffelbag due to the trait? hm.... like yeah, i guess there are small dogs irl that are just carried in bags but I suppose. Overall a cool feature but definitely needs to be looked at when it comes being smaller, being taller is definitely very cool. Sad that you would lose the funny duffelbag mechanic but oh well. At the least PR doesnt touch the main files almost at all.

DangerRevolution:
> Im uncertain how it will be received when it comes to a vulpkanin being smaller like a felinid

Like what people would think or mechanically?

Adrian16199:
> > Im uncertain how it will be received when it comes to a vulpkanin being smaller like a felinid
> 
> Like what people would think or mechanically?

Both, well, maybe not both, i guess the first would just be: Oh a small doggo, pretty cool, OH CARRIABLE ONE ASWELL.
Mechanic wise:
Felinids are payin for duffelbag plus bein smaller with 15% more damage received from blunt category.
Vulps would get it for free pretty much with no minuses. Im skeptical regardin that specific change.

Mnemotechnician:
> Both, well, maybe not both, i guess the first would just be: Oh a small doggo, pretty cool, OH CARRIABLE ONE ASWELL. Mechanic wise: Felinids are payin for duffelbag plus bein smaller with 15% more damage received from blunt category. Vulps would get it for free pretty much with no minuses. Im skeptical regardin that specific change.

I'd argue that being small is not *that much* of an advantage, but we certainly could add some debuff to counterweight it. Although if I have to add something, I might as well rewrite the system because the old code hurts my soul...

Adrian16199:
If there is room for improvement, go ahead, just keep the changes to main files as low as possible.
Regarding the debuff aspect, just a balance thing is nice to consider, who knows, maybe those few pixels could make vulps alot more robust/more threatening. Especially if they are gaining a duffelbag hiding aswell.
Probably not more damage received but soemthin else. Always time to think about it.

OCOtheOmega:
Check this out - https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/pull/45
Yeeeah a real height system 
You can control max and min height for all species and it's much more better than just have a two traits with scale.

DangerRevolution:
> Check this out - [Simple-Station/Parkstation-Friendly-Chainsaw#45](https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/pull/45) Yeeeah a real height system You can control max and min height for all species and it's much more better than just have a two traits with scale.

Pretty sure this isn't wanted due to sprite fucker-y

OCOtheOmega:
> > Check this out - [Simple-Station/Parkstation-Friendly-Chainsaw#45](https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/pull/45) Yeeeah a real height system You can control max and min height for all species and it's much more better than just have a two traits with scale.
> 
> Pretty sure this isn't wanted due to sprite fucker-y

The solution in pr is not better, I played on frontier and I'll be honest it doesn't work properly, for small races nothing changes and big races become normal size and it's not visible in the customization menu. To be honest this is a crutch and not a game innovation that will diversify the game. In any case you should check for yourself how it works before making such assumptions.

VMSolidus:
> > > Check this out - [Simple-Station/Parkstation-Friendly-Chainsaw#45](https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/pull/45) Yeeeah a real height system You can control max and min height for all species and it's much more better than just have a two traits with scale.
> > 
> > 
> > Pretty sure this isn't wanted due to sprite fucker-y
> 
> The solution in pr is not better, I played on frontier and I'll be honest it doesn't work properly, for small races nothing changes and big races become normal size and it's not visible in the customization menu. To be honest this is a crutch and not a game innovation that will diversify the game. In any case you should check for yourself how it works before making such assumptions.

If we were unconcerned about the volume of wizden code being touched, I'd merge parkstation's implementation of the height sliders in a heartbeat(Albeit after testing to make sure I'm not crashing things). However since right now we are concerned about maintaining compatibility with upstream, the frontier's approach that touches a minimum of upstream code is more desireable. However I also see a problem in that the TransferedByCloning function is incompatible with metempsychosis. We do not have wizden's cloning system. Since upstream cloning is unmaintained and deprecated, I opted to radically alter it with the Metempsychosis rebase.

Content.Server/Cloning/CloningSystem.cs
```diff
@@ -248,11 +252,22 @@ public bool TryCloning(EntityUid uid, EntityUid bodyToClone, Entity<MindComponen
             }
             // end of genetic damage checks
 
-            var mob = FetchAndSpawnMob(clonePod, pref, speciesPrototype, humanoid, bodyToClone, karmaBonus); //DeltaV Replaces CloneAppearance with Metem/Clone via FetchAndSpawnMob 
+            var mob = FetchAndSpawnMob(clonePod, pref, speciesPrototype, humanoid, bodyToClone, karmaBonus); //DeltaV Replaces CloneAppearance with Metem/Clone via FetchAndSpawnMob
 
             ///Nyano - Summary: adds the potential psionic trait to the reanimated mob.
             EnsureComp<PotentialPsionicComponent>(mob);
 
+            // Frontier - transfer of special components, e.g. small/big traits
```
VMSolidus, line 260: I need you to move "Transfer of Special Components" down into the FetchAndSpawnMob function, and then split it into two different conditions. We don't actually have Wizden cloning, we instead have a hybrid system that handles both Cloning and Metempsychosis based on checking for a bool. By placing this function up here, you are essentially making it so that when someone is reincarnated into a new body by the Metem machine, it transfers their tallness/shortness from old body to new body. Which is a big problem if for example their old body was a small vulpkanin and the new body is an Oni. 

This instead needs to be done only if the entity was cloned and not reincarnated. If they are instead reincarnated, an alternative behavior needs to be performed where it has a random chance of giving them the tallness/shortness(I would say at most 10% tall, 10% short, 80% normal).

Resources/Prototypes/DeltaV/Entities/Mobs/Species/vulpkanin.yml
```diff
@@ -97,6 +97,16 @@
       Female: FemaleVulpkanin
       Unsexed: MaleVulpkanin
   - type: DogVision
+  - type: SizeAttributeWhitelist # Frontier - tall/short traits
+    short: true
+    shortscale: 0.8
+    shortDensity: 140
+    shortPseudoItem: true
```
VMSolidus, line 104: Felinid mains are going to be upset when they find out Smol dogs have their gimmick too LOL. If you're going to do this to Vulps, you might as well do this to Harpies too.

Content.Server/_NF/SizeAttribute/SizeAttributeSystem.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                if (!TryComp<PseudoItemComponent>(uid, out var pseudoI))
+                    return;
+                RemComp(uid, pseudoI);
+            }
+        }
+
+        private void Scale(EntityUid uid, SizeAttributeComponent component, float scale, float density, bool cosmeticOnly)
+        {
+            if (scale <= 0f && density <= 0f)
```
VMSolidus, line 60: It would probably be better to do 
if (scale <= 0f | density <= 0f) return;
since && only returns if both scale and density aren't a natural number at the same time. Scale could be a negative number and density any natural number, and the function would continue as normal. Which would produce huge problems since we don't want to accidentally mirror people's character models in the case of a negative scale being put in somewhere. Keep in mind that the compiler doesn't actually know that having negative inputs for scale isn't acceptable, since it's technically possible and won't throw any errors. 

Alternatively you could put in a check for if scale is a negative number, and then set it to *= -1 to get a natural number again.

Content.Server/_NF/SizeAttribute/SizeAttributeSystem.cs
```diff
@@ -0,0 +1,95 @@
...
+        [Dependency] private readonly FixtureSystem _fixtures = default!;
+        public override void Initialize()
+        {
+            base.Initialize();
+            SubscribeLocalEvent<SizeAttributeComponent, ComponentInit>(OnComponentInit);
+        }
+
+        private void OnComponentInit(EntityUid uid, SizeAttributeComponent component, ComponentInit args)
+        {
+            if (!TryComp<SizeAttributeWhitelistComponent>(uid, out var whitelist))
```
VMSolidus, line 26: Since by introducing this system to Metempsychosis means having the possibility for entities to be handed a SizeAttribute after character creation, you now need to account for SizeAttribute being handed to an entity that potentially doesn't have SizeAttributeWhitelist. Since we're exiting out if a species isn't whitelisted for SizeAttribute, you might as well cleanup by RemComping the component if for whatever reason it was handed to a species lacking the whitelist. That will also be useful if for example someone hits AddComponent in VV without first doing so with SizeAttributeWhitelist.

VMSolidus (changes requested):

DebugOk:
We don't do suggestions on code that isn't ours. When these problems are addressed on Frontiers GitHub, this pr may be reopened.

Mnemotechnician:
> We don't do suggestions on code that isn't ours. When these problems are addressed on Frontiers GitHub, this pr may be reopened.

May I ask you what you mean by that, and which exact problems? Balance, or what was requested above? And, is it possible for me to just rewrite this from scratch and make it deltav code instead? (and then pr back to frontier, getting it merged there will take weeks or months)

And yea sorry for not responding, I didn't have time today

Adrian16199:
It would be an issue to adress this issues on frontier if half of the issues appear only on our side due to us having a metem cloning while any other server does not have it, especialy frontier so it technicly would mean that you would need to rewrite it from scratch. Atleast thats how I interpretated it.

PR #995, opened 2024-03-22
Ectoplasm changes

Guess-My-Name:
## About the PR

Separates normal ectoplasm and revenant's ectoplasm into two entities, revenant ectoplasm with significantly more reagent
## Why / Balance

Upstream has ectoplasm mapped on different salvage wrecks/debris and will probably have more of those in the future, because they serve no other purpose than making revenant plushies there.
On DeltaV this is problematic, because just one Ectoplasm is enough to make a glimmer drain (contains enough reagent for 6 normality crystals).
This PR makes changes the ratio: upstream Ectoplasm = 1 normality crystal (so that salvage's finds aren't completely useless) and the new  Revenant's Ectoplasm = 6 normality crystals.
## Media

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Revenant/Components/RevenantComponent.cs
+++ b/Content.Shared/Revenant/Components/RevenantComponent.cs
@@ -25,7 +25,7 @@ public sealed partial class RevenantComponent : Component
     /// Prototype to spawn when the entity dies.
     /// </summary>
     [DataField("spawnOnDeathPrototype", customTypeSerializer:typeof(PrototypeIdSerializer<EntityPrototype>))]
-    public string SpawnOnDeathPrototype = "Ectoplasm";
+    public string SpawnOnDeathPrototype = "EctoplasmRevenant"; //DeltaV - different prototype that contains more ectoplasm reagent for normality crystals
 
     /// <summary>
     /// The entity's current max amount of essence. Can be increased
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Specific/Medical/morgue.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Specific/Medical/morgue.yml
@@ -1,0 +1,28 @@
+- type: entity
+  parent: Ash
+  id: EctoplasmRevenant
+  suffix: DO NOT MAP
+  name: ectoplasm
+  description: Slimey, vaporous residue of an ancient spirit.
+  components:
+    - type: Sprite
+      sprite: DeltaV/Mobs/Ghosts/revenant.rsi
+      state: ectoplasm
+    - type: Tag
+      tags:
+        - Trash
+        - Ectoplasm
+    - type: SolutionContainerManager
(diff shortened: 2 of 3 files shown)


Bonktrauma:
I really don't like this, its hard enough to get a drainer as is and that wreck is relatively rare

Bonktrauma:
> I really don't like this, its hard enough to get a drainer as is and that wreck is relatively rare

Not to mention we can just change the wreck

Guess-My-Name:
> I really don't like this, its hard enough to get a drainer as is and that wreck is relatively rare

Screenshot from a round #10731, which lasted 2 hours 11 minutes.
image

Drainers were never intended to be something easy to make or something obtainable in every round. The only source for ectoplasm was revenant, salvage ectoplasms were added upstream and no one really noticed until recently.

> Not to mention we can just change the wreck

It's easier and better to do it like this moving forward, changing upstream wrecks is a worse solution.

Colin-Tel:
I think this PR is justified, however I think you could make the value `10` instead of `5` so only 3 of those wrecks would be the minimum instead of 6? It's an arbitrary amount, anyways.

Guess-My-Name:
> I think this PR is justified, however I think you could make the value `10` instead of `5` so only 3 of those wrecks would be the minimum instead of 6? It's an arbitrary amount, anyways.

I've seen many times in past month similar situations to screenshot above - salvagers bringing 5-8 ectoplasms from salvage to epi - which made me think about removing reagent entirely from non-revenant ectoplasm ([a very brief discussion on discord](https://discord.com/channels/968983104247185448/991506325316059257/1220191083980521492)), but then had second thoughts and decided to make it 5u, so there could be some more between departments interactions and epi wouldn't just default to blowing up themselves when glimmer gets too high.

I can make it 10u, but with 10u it should probably have do not map suffix, because just 3 of those would be enough to redtext all glimmer>500 objectives, if drainer gets built in mysta's room or in bridge.

Content.Shared/Revenant/Components/RevenantComponent.cs
```diff
@@ -25,7 +25,7 @@ public sealed partial class RevenantComponent : Component
     /// Prototype to spawn when the entity dies.
     /// </summary>
     [DataField("spawnOnDeathPrototype", customTypeSerializer:typeof(PrototypeIdSerializer<EntityPrototype>))]
-    public string SpawnOnDeathPrototype = "Ectoplasm";
+    public string SpawnOnDeathPrototype = "EctoplasmRevenant"; //DeltaV - different prototype that contains more ectoplasm reagent for normality crystals
```
deltanedas, line 28: this should be done in the revenant yaml

deltanedas (changes requested):
i agree with the idea

Colin-Tel:
It's been over a month, you can re-open this if you want to make the requested changes.

PR #939, opened 2024-03-08
Revert "Revert AME buff"

DebugOk:
Reverts DeltaV-Station/Delta-v#930

This was done with good reason upstream. @VMSolidus NEVER revert upstream PRs like that

PR #996, opened 2024-03-23
Revert "Revert "Revert AME buff""

IamVelcroboy:
Reverts DeltaV-Station/Delta-v#939

We might revert this again if everything burns but it's worth a shot. Let's gooo

**Changelog**
:cl: Velcroboy
- tweak: Reverted AME buff. AME will require more than 1 core again to function properly.

### DIFF
--- a/Content.Server/Ame/AmeNodeGroup.cs
+++ b/Content.Server/Ame/AmeNodeGroup.cs
@@ -182,7 +182,7 @@ public sealed class AmeNodeGroup : BaseNodeGroup
         // Fuel is squared so more fuel vastly increases power and efficiency
         // We divide by the number of cores so a larger AME is less efficient at the same fuel settings
         // this results in all AMEs having the same efficiency at the same fuel-per-core setting
-        return 2000000f * fuel * fuel / cores;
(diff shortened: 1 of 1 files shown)


Axionyxx (approved):
Let's do it!

DebugOk:
I've already said before this isn't a good idea without the pending fixed from upstream, but I can't stop you

IamVelcroboy:
> I've already said before this isn't a good idea without the pending fixed from upstream, but I can't stop you

We talked about it. We took into consideration what you said. Decided it's worth the risk/reward and we're prepared to undo it if it's too problematic.

DebugOk:
> > I've already said before this isn't a good idea without the pending fixed from upstream, but I can't stop you
> 
> We talked about it. We took into consideration what you said. Decided it's worth the risk/reward and we're prepared to undo it if it's too problematic.

If it breaks, I told you so ¯\\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

deltanedas:
is there any issue i havent seen any problem when not using 1 core ops

PR #998, opened 2024-03-23
Lunchbox size fix

FluffiestFloof:
Lunchbox now takes 8 space and only has 8 space of storage.

## Breaking changes

**Changelog**

:cl:
- tweak: Changed lunchbox size.


IamVelcroboy:
So, I think a better way to handle this would be to make the size of the box = the size of internal storage. If the standard small/normal/large sizes don't work, you can set a custom size in the `Item/shape` node. Maybe that with a slight reduction to how much it holds?

FluffiestFloof:
Sure. I will make the box only 4 slots then I suppose

IamVelcroboy:
> Sure. I will make the box only 4 slots then I suppose

No I mean, make the space the box takes up in another inventory bigger like 2 x8 or something. And then trim the extra tile in the last row.

IamVelcroboy:
Look at the cigar case for an example. It uses the AaBb system

FluffiestFloof:
done it takes 8 space and only has 8 space

rosieposieeee:
Personally I think 4-4 would be better than 8-8, if only because I can't imagine anyone carrying a lunchbox if it were that big.

IamVelcroboy:
> Personally I think 4-4 would be better than 8-8, if only because I can't imagine anyone carrying a lunchbox if it were that big.

If we make it smaller than 8, we'll have to start editing the inventory too. Otherwise, it will overfill

DangerRevolution:
Size fix > Food fix as Food requires alotta tags & upstreaming + the intention was to also allow people to carry contraband. 4x4 is a nice size for a lunchbox anyways.

IamVelcroboy:
How about 2x3?

DangerRevolution:
> How about 2x3?

What size is the toolbox in an inventory?

IamVelcroboy:
Toolbox is made so it doesn't fit in an inventory.. Other than bluespace maybe

IamVelcroboy:
Closest comparison, I can think of, is cigar case which is 2x3

FluffiestFloof:
Fine, I guess I'm creating new juice boxes because the orange carton is 2x2 for some reason

DangerRevolution:
I can do this at another time if you don't, but fixing the probabilities to be true probabilities rather than skewing towards healthy food would be nice. Station 14's "storageFill" system probabilities use the "Sequential" model rather than Simultaenous. The game rolls whether it puts the first item in the list, then the next, then the next, then the next. When using "or" probabilities this means that items that roll first; are more likely to be put in than items that roll second.

image

Item attached is related. That's how item picks work, I guess.

FluffiestFloof:
> How about 2x3?

6 slots could be doable if we ditch the note because biggest item for each item fill is 2 slots.
Except if we want to have it with more slots than space it takes, which is a big thonk personally. But it's not like I play :trollface:

IamVelcroboy:
Nah, Is good how it is. Juice boxes look great!

PR #1006, opened 2024-03-25
Low-Impact Defensive Landmines

rosieposieeee:
## About the PR

Putting landmines around areas like the armory and vault is fun and intimidating.

It also happens to be extremely counterproductive to the concept of protecting an area. They're basically free C4 already set and ready to go with little effort.

This PR adds a Low-Impact Landmine which only explodes a single tile. Point blank, on a naked moff, they do 75 damage. (Regular landmines do like, 150.) They are much more suitable for the intimidation factor - loud, flashy, but don't break everything they're intended to protect. They won't set each other off too! I recommend not sticking too many near windows, as reinforced walls can take like 5 of em before they break. 

The scaling on explosive damage is reaaaaaaaally weird; I tried a lot of different values for a number of effects (ie low damage but high range), but this seems to be the most consistent set. I kind of wish it didn't break tiles at ALL, but I can't seem to get the range any smaller.

They are also anchored by default and un-anchoring them has a delay of 30, making them SIGNIFICANTLY better as a deterrent. Not an unreasonable amount of time, but you still have to be careful if you want to be quiet.

image
ekM2vDZ4CK

No CL because this doesn't even do anything until mapped in.


Axionyxx (approved):

IamVelcroboy:
Part of me thinks they should still explode out to 2-3 tiles, even if they do dmg some walls slightly. If only to catch other players close by and...when step off is implemented.

IamVelcroboy:
Yeah, now that step-off was merged, you should test these again and I def think we'll want them to be 2-3 tiles of spread.

IamVelcroboy:
Another thing I'd like to know is...how much dmg will these do to players in full nukie armor? Will it still be enough of a deterrent?

rosieposieeee:
i learned a bit about explosions while looking at the breaching charges and i think i might be able to concoct a version that cant break ANYTHING but still deals serious damage, gonna look into it today

Axionyxx:
Are you still interested in finishing this?

rosieposieeee:
eh, nah

PR #1007, opened 2024-03-25
Reverts Nitrogen Tank Removal

DangerRevolution:
## About the PR

Reverts Wizden commit that removes Nitrogen Tanks from lockers
## Why / Balance

Slimes deserve to breathe too. Loadouts should be made rather than removing tanks. Why does a human get a tank in his locker, but someone has to go dive maints or beg Atmos/Engineering because they picked a Slime?

**Changelog**

:cl: DangerRevolution
- add: After a campaign from Slime Rights Activists, Nitrogen tanks have been readded to standard lockers.


DangerRevolution:
I really, really need to reclean my fork to fix the merge history :O

rosieposieeee:
i actually think this was a good change by WD and makes slimes more meaningfully different. if slimes can find nitrogen just as easily as every other species can find O2, why make them breathe nitrogen at all?

DangerRevolution:
>  good change by WD [...] makes slimes more meaningfully different [...] why make them breathe nitrogen at all?

1. Slimes breathe nitrogen because they couldn't port Ethereals; as there's no systems to support them so they ported something simpler for racial variety that they could do a minor tweak to change. 
2. Slimes are on their way out from Wizden, have no content, are the worst Wizden race and with this change; if you haven't gone to beg Atmos/Engi for a Nitrogen Tank or gone to dive through maints then they die when spaced or gassed, they already take extra damage from the Barotrauma.
3. this still impacts Salvage, who need to take an air canister rather than just O2 + Slimes don't get any relevant tank in their survival kits.

deltanedas:
(salv takes a n2 can rather, breathing air mix from a tank is bad for every species)

IamVelcroboy:
All the lines touching upstream files need DV notes.

IamVelcroboy:
Besides adding comments, this looks good other than I'd argue we don't need to add them to Engie/CE lockers as long as we have tank dispensers mapped in the dept. Same could probably go for Salv.

Resources/Prototypes/Catalog/Fills/Lockers/cargo.yml
```diff
@@ -6,6 +6,8 @@
   - type: StorageFill
     contents:
       - id: ClothingOuterHardsuitSpatio
+      - id: OxygenTankFilled
```
IamVelcroboy, line 9: Remove these. They can use tank dispensers. If dispensers aren't mapped, we will fix that.

Resources/Prototypes/Catalog/Fills/Lockers/engineer.yml
```diff
@@ -82,6 +82,8 @@
     contents:
       - id: ClothingOuterHardsuitAtmos
       - id: ClothingMaskGasAtmos
+      - id: OxygenTankFilled
```
IamVelcroboy, line 85: Remove these. They can use tank dispensers. If dispensers aren't mapped, we will fix that.

IamVelcroboy:
So, upstream just merged this [26752](https://github.com/space-wizards/space-station-14/pull/26752). I'm inclined to go with it. Seems reasonable. We could maybe still add a chance for them to spawn in emergency toolboxes though.

Adrian16199:
> So, upstream just merged this [26752](https://github.com/space-wizards/space-station-14/pull/26752). I'm inclined to go with it. Seems reasonable. We could maybe still add a chance for them to spawn in emergency toolboxes though.

that pr.... feels like such a bloat on mapping tbh.
also I do like the idea on the chance spawnin in emergency toolboxes.

IamVelcroboy:
> that pr.... feels like such a bloat on mapping tbh. 

I think the idea is that you replace some of the emergency lockers with these, not add more

PR #1012, opened 2024-03-27
Enforce CRLF removal early

Axionyxx:
Saving myself headaches down the line


Axionyxx:
CRLF, I've been typing it wrong the entire time

PR #1014, opened 2024-03-27
Increase the damage of rubber bullet projectiles

Samsterious:
## About the PR
Increases the blunt damage of rubber bullets from 3 to 5. Magnum revolver bullets were changed to 9 to match their greater stamina damage. The expected blunt damage of someone who has been stunned by rubber bullets is thus changed from 15 (9 for revolvers) to 25 (27 for revolvers). No beanbags were ~~harmed~~ altered in the making of this PR. They already deal 30 blunt damage per stun anyways.

None of the stamina damage values have been changed.

## Why / Balance
Upstream simply removed all rubber ammunition from the game, but after some discussion, it seems the balance issues are less severe for us. Raising the blunt damage of rubber bullets to exceed that of disablers — without modifying their stamina damage — is not only a more popular balance tweak, but it also ensures they exceed the automatic healing threshold, which should discourage security players from using rubber bullets by default when a disabler will do just fine.

## Media
image
image

**Changelog**
:cl: Fireheart&
- tweak: Rubber bullets are now 40% more lethal!


DangerRevolution:
Upstream: https://github.com/space-wizards/space-station-14/issues/26443

with the issue being that Rubber bullets are too strong as stun-weapons; how does this address the problem?

Samsterious:
It was discussion about that very PR that inspired this change. Our consensus was that the following things were true:
- Removing an entire category of weapons from the game is generally undesirable, if we can instead adjust the numbers to make them less unbalanced.
- Security in Delta-V is disadvantaged as compared to security in Wizden due to their stricter rules of engagement and higher standards of roleplay; therefore, the balance issue is less severe here than it is there.
- By increasing the damage of rubber bullets, we reduce the temptation to default to them simply because they are more effective for catching Syndies. They instead become an escalation above disablers and batons, a riot-control weapon, as intended.
- The corpsman provides security with the means to address the injuries that rubber bullets create, whereas before a single stun did not deal enough damage for security to bother taking their detainee to the medbay because they can be healed with nothing more than a little bedrest.
- Changes to the in-character policy should be made in order to further reinforce that rubber bullets are not a first-choice solution for forceful detention. They _are_ more powerful; they are also more harmful, and much more likely to cause an accidental death.

FluffiestFloof:
You forgot .38, except if you want to keep their 3 Blunt.

Samsterious:
Good catch! Thank you for that.

DangerRevolution:
> Changes to the in-character policy

If this materialises then this will be cool and make this all make sense.

> much more likely to cause an accidental death.

they deal 25 blunt

Samsterious:
They deal 25 blunt _at minimum_. My thinking is that rubber bullets are a lot less controlled than disablers or batons, and so are more likely to have more hits than necessary to stun. And if there is overlapping fire, or if the officers are too slow to cuff, it is possible that, in rare circumstances, someone gets put into crit when lethals were not justified.

But this is only a mechanical incentive. The other half, which is beyond the scope of this PR, is changing the in-character documentation to reinforce that rubbers are not a substitute for a disabler, but an escalation above and beyond, just as lethals are.

DangerRevolution:
So uh, what's the incentive to use the rubbers; which is behind research, requires more accuracy and more shots hit?

Samsterious:
They fire faster, have a lower time-to-stun, and have a longer endurance before you need to return to sec to rearm. They are the meta on Wizden (hence the aforementioned PR), and we have already seen such starting to creep in here.

UnicornOnLSD:
putting this out here, rubbers are gone upstream, and are already a powerful stun tool, at this point eh, worth keeping that up ?

Samsterious:
This PR's initial message and my earlier comment each explain my reasoning:

> ## Why / Balance
> Upstream simply removed all rubber ammunition from the game, but after some discussion, it seems the balance issues are less severe for us. Raising the blunt damage of rubber bullets to exceed that of disablers — without modifying their stamina damage — is not only a more popular balance tweak, but it also ensures they exceed the automatic healing threshold, which should discourage security players from using rubber bullets by default when a disabler will do just fine.
----
> It was discussion about that very PR that inspired this change. Our consensus was that the following things were true:
> 
> * Removing an entire category of weapons from the game is generally undesirable, if we can instead adjust the numbers to make them less unbalanced.
> * Security in Delta-V is disadvantaged as compared to security in Wizden due to their stricter rules of engagement and higher standards of roleplay; therefore, the balance issue is less severe here than it is there.
> * By increasing the damage of rubber bullets, we reduce the temptation to default to them simply because they are more effective for catching Syndies. They instead become an escalation above disablers and batons, a riot-control weapon, as intended.
> * The corpsman provides security with the means to address the injuries that rubber bullets create, whereas before a single stun did not deal enough damage for security to bother taking their detainee to the medbay because they can be healed with nothing more than a little bedrest.
> * Changes to the in-character policy should be made in order to further reinforce that rubber bullets are not a first-choice solution for forceful detention. They _are_ more powerful; they are also more harmful, and much more likely to cause an accidental death.

UnicornOnLSD:
fair point, I do believe that rubbers are the ultimate counter to most threat nonethless as security, not sure on if they should be buffed but considering we haven't added disablers to secoff lockers, yeah whynot

Samsterious:
This is only a buff in the strictest sense that their lethal damage is slightly increased. This PR is actually a *de*buff, because Security players are held to a higher standard and are subject to stricter rules of engagement. This reduces the attractiveness of rubbers because they are rendered more grisly. They remain vastly inferior to standard lethal weapons.

PR #1015, opened 2024-03-27
Increase moth color blending strength by 25% and replace color blending with the intended helper function

Axionyxx:
## About the PR
This would need to be properly tested to make sure it still achieves the desired effect. My hope is that by increasing the amount of color blending done, that it'd allow for greater freedom with the colors available

## Why / Balance
Bright purple moff :3

## Technical details
[Color.InterpolateBetween](https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared.Maths/Color.cs#L823) is what should be used here, instead of ripping apart the color and lerping everything manually. This shouldn't really affect anything too severely except making me happy by having a shorter function

## Breaking changes
It might cause players to have to reset the colors on their moth characters, but since its just a blending change it'd probably just come down to moving the sliders up slightly

**Changelog**
:cl:
- tweak: The color mixing applied to moths has been increased in strength by 25%.


A-z-z-y (approved):
Looks good, makes code more readable (self-documenting functions my beloved.)

PR #1018, opened 2024-03-28
Add revolutionaries

Samsterious:
## About the PR
Re-enables revolutionaries, updates their flavor-text to give revs more flexibility in supporting the cause and gives head revs a somewhat less murderous goal, adds a Syndicate encryption key to each head rev's backpack at round start, and makes cuffing an adequate means of defeating either head revolutionaries or Command.

The relative frequency of each game mode has been unchanged, aside from that they were all uniformly scaled down to give revolutionaries their default frequency of 20%. This PR is also split into several commits in order to more easily cherrypick from it in the event that it is desirable only in part.

## Why / Balance
Revolutionaries were removed from Wizden for two key reasons:

1. Players were failing to play the role they were given. Converted revs were voluntarily mindshielding themselves, ignoring orders from the head revs, and choosing not to pursue the objectives their role tasked them with. These concerns are not quite so potent on Delta-V, where players generally have a higher commitment to their roles rather than to victory or in-game projects from which they are distracted by becoming an antagonist.

2. Wizden's maintainers determined that revolutionaries required a more ground-up rework in order to become a more compelling game mode. Particularly, they felt that the forced round-removal and team deathmatch traits of revs rendered them an inadequate antagonist. I have had similar thoughts about nuclear operatives myself, but I think revolutionaries have greater potential than they do in terms of providing a meaningful roleplaying opportunity to the crew, especially if they are encouraged to take a less destructive approach.

The hope is that, by giving head revolutionaries a means to coordinate their plans, they are not forced into falling in line with the most violent of them all; instead, they can plan when to strike, coordinate the movement of supplies, and encourage one another to take a less overt approach to subversion in general.

By allowing both Command and revolutionaries to achieve victory without spilling any blood, we also allow for diplomatic roleplay between both factions. Perhaps the head revolutionaries stage a protest as a show of force, and Command makes the decision to step down to prevent a bloody skirmish that they cannot win; maybe Command discovers the names of the head revolutionaries and invites them to the negotiating table with the promise of fair treatment.

Removing the head revolutionaries' binding command over their converts allows players to choose a more supportive role instead of actively fighting and dying for the cause. This has the effect of reducing the actual physical power of revolutionaries, encouraging a more covert strategy, and helps alleviate the fact that it is not possible to opt out of being converted despite that most other antag roles can be declined. This prevents players from feeling coerced to roleplay something they would not like to roleplay.

This PR includes language that implies that being deconverted invokes the new life rule. That is, events that transpired between being flashed and being mindshielded are not remembered. This prevents one captured revolutionary from spoiling the entire plot, giving revolutionaries more wiggle room to enact their plans. We will need policy to reinforce this to ensure it is not missed in the fine print.

## Media
image

**Changelog**
:cl: Fireheart&
- add: Revolutionaries have been re-enabled.


UnicornOnLSD:
By doing a quick reading on the review, I want to point a few things out. Most of this is assuming we keep the chances made on DV regarding revies, namely lowering the amount of headrev to 1/2 based on pop

>adds a Syndicate encryption key to each head rev's backpack at round start

The goal would be to make organisation easier, understandable on one side, but would we have it as it used to be, the low amount of headrev would ultimately lead to not much communication, even so then the issue often proved to be the revies and not headrevs, regarding coordinations.
Moreover it certainly *is* a mean to charge a headrev but it is only possession charge.
I'd be on par to try this, it has potential, depending on how it is handled, but more organisation can sure be better, imo

>revolutionaries their default frequency of 20%

This is WAAAAY too high. As it stand Delta's secret weights are borked, being the following: 44% survival, 39% traitor, 14% Nukies and 3% zombies. Zombies and revies should have a similar weight, as ultimately it is a MAJOR event that implies ALL of the station directly, either for or against command namely, in the same vein zombies do with zombies against survivors. Traitors concern *some* crew but not all and permit for some mixed chaos among all of it, but such a large impacting game mode should not have a 1/5 change of happening, nukies are not AS bad as revies, and are sitting at 15%, there nukies are twice less likely as revies and god knows how hellish revies proved to be even compared to nukies

Would you follow my advice, I would recommend lowering survival, traitor and nukies to nudge some spot of revies at a nice 5%, or so. Would the PR be approved and revies turns out to be "successful", by that meaning it's not just a murder fest, it could be increased.

>Revolutionaries were removed from Wizden for two key reasons:

First, this is Delta-V, we first removed it on our own whim because of the poor RP it had, being a team deathmatch, often one sided because of the sheer power of multiple headrevs.
The counteract to this was making it a single headrev, meaning that instead o fbeing a massive TDM, cutting off the root would end it all unlike with multiple headrevs

Secondly, we lack informations regarding revolutions on delta, and would this PR be merged, I would propose to add more clear rules for revolution and revealing it, see https://forum.spacestation14.io/index.php?/topic/3450-revealing-a-revolution/ (Wizden's way to avoid just security doing things per law and getting slaughtered, among other things), along with them having the "expeditive execution", only appliable to headrevs which we don't have, here.

Ultimately, from what I know, revies were disabled on wizden for a "rework", causes being, unclear, maybe the arguments you mentionned. Said rework never came and revies were added with a 5% chance if I remember right.

>The hope is that, by giving head revolutionaries a means to coordinate their plans, they are not forced into falling in line with the most violent of them all; instead, they can plan when to strike, coordinate the movement of supplies, and encourage one another to take a less overt approach to subversion in general.
By allowing both Command and revolutionaries to achieve victory without spilling any blood, we also allow for diplomatic roleplay between both factions. Perhaps the head revolutionaries stage a protest as a show of force, and Command makes the decision to step down to prevent a bloody skirmish that they cannot win; maybe Command discovers the names of the head revolutionaries and invites them to the negotiating table with the promise of fair treatment.

Again, what proved to be an issue weren't the headrev, but rather the overall task the the revies themselves. I TOTALLY agree with changing their objective to detaining command instead of just round removing them, this is a WAY better option.

However, for the part on command of "achieving victory without spilling any blood", I highly doubt it is the case, to deal with a revolution as a loyalist is to cut it, to cut it means mindshielding all the masses or killing a headrev, and I doubt any headrev, or their follow will sit idly as they are on their way to execution for not being mindshieldable (they may not have to kill command but it does not change revolutionnary activities are heavily repressed and a capital crime, even more so in space law where YES it would need a trial, if station is not ablaze, but with all I've seen before I doubt just changing their goals will change the global behavior regarding dealing with revies)

Yet again you say it's HOPEFULLY the case, and I'd be on par with that, because frankly murder fest is not so fun, to me and many people, just changing their goal will not suffice, imo.

>Removing the head revolutionaries' binding command over their converts

I read the whole thing and haven't seen a statement where you said headrevs didn't order revies around anymore ? That's how I read it at least. I would like to remind you that first, only the headrev have their goal, and as of now a converted person simply follows their orders. The several blood fest we had, and I experienced on delta weren't the fault of the headrev but rather people not obeying them.

>This has the effect of reducing the actual physical power of revolutionaries

Would they not have to follow the headrev's orders, yes, they wouldn't have to be as aggressive, but a goal would remain of detaining command, the masses are always stronger and way I see it is:

If you are a group of revies, the headrev got jailed for well, sedition. You need to stop command, just beat up a secoff for his gear, not like he matters
We people are in groups and have an objective they'll take the easy route, and everywhere, including delta, often the easiest route to controlling someone is just brute force, either through stun or just lethality.
LOGICALLY if you are a command member and being talked down by a group of people demanding your surrender, with them having leverage, yeah, certainly you can agree if they don't aim to kill you, but this rarely works in similar cases.



TL:DR, I support the idea, making sure things are more orderly to ensure RP for revies is great, however:
-20% is way too likely, even so for such an impactful game mode
-we still lack directives on how to handle revolutions as security legally speaking, how it devolved every time I witnessed a revolution was just "kill the non minshielded until the you the right one", because EVERYONE was aggressive and the alternative, as security was death
-I am totally on par with making their objective to detain command, or exile them, whatever is just not round removing, because frankly it sucks.
-I feel like you see this change and hope people will act accordingly, but considering how some headrevs tried going pacifist and SOME revies just went haywire, the issue might be in the player base

Of course we can't tell without seeing results, thus why I propose some alternative to make it tamer, would it be approved.

DangerRevolution:
Very ambitious, putting a lot of faith in players; interesting to see how this works out

Bonktrauma:
One encryption key won't really do that much

Samsterious:
> One encryption key won't really do that much

My thinking is that giving only the head revs access to an encryption key means that they still have to take the time to reach out and communicate with their converts rather than simply handing them a key and directing them over the radio. This reduces the physical power of head revs and also means that one captured rev does not spoil the whole plot because Syndicate comms are compromised.

> Very ambitious, putting a lot of faith in players; interesting to see how this works out

Holding them to a higher standard, perhaps. I have trust in the players to search for a more interesting approach rather than simply breaking down the windows to the bridge and pulling Command out kicking and screaming. The incentives are also changed, here. There is some policy we need to write to codify what we actually want, but my opinion is that this is a vastly less deathmatchy game mode than nukies and yet players will riot if we remove the literal "kill everyone" game mode. :blunt:

> even so then the issue often proved to be the revies and not headrevs, regarding coordinations.

In the same vein, I am of the opinion that this is a playstyle and policy issue. We have a much more roleplay-oriented approach here than on Wizden, and most folks are interested in reinforcing that. Where folks are taking a powergamey approach, we can simply guide them towards more interesting playstyles, or otherwise stop trusting them with roles that have heightened roleplay standards if they are consistently taking the LRP route as a converted rev.

> nukies are not AS bad as revies

I do not concur with this in the slightest. Aside from the fact that nukies are literally tasked with killing everyone on the station, whereas head revs need only depose Command, revolutions offer far more opportunity for roleplay. They involve subterfuge and sabotage, infiltration and coordination. Nukie rounds begin and end with death. Revolutions involve hidden roles and building a network of collaborators and seeking the least dangerous means to achieve victory because Command and Security have a vastly superior arsenal. Nukies introduce themselves with a series of china lake explosions that space half the station and bloodred hardsuits that can tank an entire magazine from an SMG.

> This is WAAAAY too high.

If 15% is good enough for nukies, I do not see why 20% is so absurdly high for revolutionaries. I do agree that _both_ are too high, however, but I think adjusting nukies is beyond the scope of this PR and would better be accomplished in a general weights rebalance PR of its own. That said, the last time we had revolutionaries on Delta-V, they were at 10%. Perhaps that is a better target for this PR.

> The counteract to this was making it a single headrev, meaning that instead o fbeing a massive TDM, cutting off the root would end it all unlike with multiple headrevs

It seems that the consensus is that the converts are the source of power, not head revs themselves. Having more of them means the round does not end when one single person gets cuffed, and also offers an opportunity for head revs to be rescued by their comrades in the event they are captured. Moreover, it means they have accountability to one another; if we only have one head rev, they may simply choose to go loud, whereas if there are multiple head revs, they can communicate via the Syndicate comms they are granted at round start to strategize and discourage premature violence.

> Secondly, we lack informations regarding revolutions on delta

On this I agree. There is a great deal of policy that still needs to be written before this PR can be merged. Particularly, our current definition of powergaming is too open-ended, both leaving players unaware of the specific things they should avoid and causing our administrators to hand out some pretty strange bans of late. This seems to me to be the primary source of information for Security and revolutionaries alike in order to clarify what is a reasonable course of action and what is unfun powergaming.

> but with all I've seen before I doubt just changing their goals will change the global behavior regarding dealing with revies

You are absolutely correct that we have to do better than flavor-text. The primary change to game logic in this PR is that cuffing also counts for the purposes of defeating either Command _or_ head revolutionaries. That is, if Security manages to detain all of the head revs, the round ends. There is no need to expedite an execution, and requiring them to actually go through the motions of a trial means Security is incentivized to simply hold the head revs in detention, which offers the opportunity for them to be broken out by their comrades.

One of the things I would like to do in terms of overall game policy is see a less metagamey approach to antagonists. I _still_ occasionally stumble into "Syndie, therefore perma" folks, and this is not the approach Security should be taking with any of the hidden role antagonists; simply _being a Syndie_ is not a crime. Likewise, a "revolution" is an OOC concept that Security does not know about. They might know that flashes can be modified to influence others' minds, and that mindshield implants negate their effects, but if the head revs are taking a covert approach, then they are probably not actually guilty of anything other than _being a head rev_, which is not a crime.

There are some changes to space law that we will need to write as well before this can be merged, particularly how to address converted revs and how to hold head revs accountable for directly inciting unlawful behavior without actually participating in it. This, of course, would require Security to actually have evidence that the head rev is a head rev, instead of simply mindshielding everyone and executing anyone who cannot be mindshielded.

> I read the whole thing and haven't seen a statement where you said headrevs didn't order revies around anymore ?

That is changed in the flavor-text commit. I removed the language suggesting that converted revs must obey the head revs. I have also removed the language giving them an open-ended license to kill Command. They are simply tasked with supporting and protecting the head revs. Beyond that, they should probably just keep on doing what they were doing until they are asked to support the head revs somehow, perhaps by supplyling them with medicine from the hydroponics lab or acquiring a specific cargo order or handing over tools from engineering, and so on.

> We people are in groups and have an objective they'll take the easy route, and everywhere, including delta, often the easiest route to controlling someone is just brute force, either through stun or just lethality.

If converted revs immediately run off and go murderboning, they are acting beyond the scope of their role and thus run afoul of Rule 10, i.e. proportional damage for antagonists. Simply put, this is bannable behavior with the new converted rev objectives.

----

Thank you for your thorough design review, Unicorn; I know I disagreed with you on a lot of points here, but this has forced me to consider seriously the specifics of the implementation and to better recognize what actually needs to be changed in terms of policy to make this game mode work well.

PR #1025, opened 2024-03-28
Unlocked Booze-O-Mat

rosieposieeee:
## About the PR

Adds a Booze-O-Mat variant that has no access requirements. Mainly for the perma one in Sub, and it would be useful for the maints bars where it has to be hacked *every* time.

NOTE: I hereby release all changes made in this PR under MIT license.

**Changelog**
:cl:
- add: Added a variant Booze-O-Mat that starts unlocked.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Structures/Machines/vending_machines.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Structures/Machines/vending_machines.yml
@@ -48,3 +48,11 @@
       map: ["enum.WiresVisualLayers.MaintenancePanel"]
   - type: AccessReader
     access: [["Mail"]]
+
+- type: entity
+  parent: VendingMachineBooze
+  id: VendingMachineBoozeUnlocked
+  suffix: Unlocked
+  components:
+  - type: AccessReader
+    enabled: false


IamVelcroboy:
This is probably a good idea. I wonder if we should reduce some of the inventory in it? Stick to the drinkware and basic mixers? Like, if you want the good stuff, go to the proper bar.

rosieposieeee:
maybe, but eh, a bit out of scope for this

IamVelcroboy:
> maybe, but eh, a bit out of scope for this

I don't think so. I think it's perfectly relevant. Just copy over an inventory for it and edit in a Delta folder. But I'm not saying it's mandatory. Just putting it out there for consideration.

rosieposieeee:
fair, i really just meant this to be more of a bugfix though

Resources/Prototypes/DeltaV/Entities/Structures/Machines/vending_machines.yml
```diff
@@ -48,3 +48,11 @@
       map: ["enum.WiresVisualLayers.MaintenancePanel"]
   - type: AccessReader
     access: [["Mail"]]
+
+- type: entity
```
deltanedas, line 52: put in deltav namespace or add markers

deltanedas (changes requested):

Resources/Prototypes/DeltaV/Entities/Structures/Machines/vending_machines.yml
rosieposieeee: its already in dv...? Resources/Prototypes/DeltaV/Entities/Structures/Machines/vending_machines.yml ??

deltanedas: oh im blind

deltanedas (approved):

Axionyxx:
Only came across this when redoing all changelogs that had failed, but non-player facing changes don't need changelog entries

PR #1028, opened 2024-03-28
Prevent random slots from hiding character noses and hair

### LINKED ISSUE #982: Having masks in the pocket slot will hide snout markings
## Description
Having a mask in a pocket slot causes issues with snout markings and etc.

**Reproduction**
1. Start a round with a character that has snout markings or is a reptilian.
2. Get two masks, put one in your pocket and equip the other one. 
_(does **not** work with the breath mask, maybe only ones that hide identity?)_
3. Remove the equipped one and notice the missing snout markings.

**Screenshots**
_1: Equipped mask without mask in pocket
2: Unequipped mask without mask in pocket_

_3: Equipped mask with mask in pocket
4: Unequipped mask with mask in pocket (bug occurs here)_

WithoutMaskInPocketandWith

DinoWattz:
## Why / Balance
Small change that fixes #982

## Technical details
Changed the enumerator at line 97 to make it check only the "HEAD" and "MASK" slots.
Fixed formatting at line 167.

## Media
* [ ]  I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase


DEATHB4DEFEAT (commented):
It would probably be better to have clothing check if it's where it should be before doing things instead of adding yet another check against pockets.

DinoWattz:
> It would probably be better to have clothing check if it's where it should be before doing things instead of adding yet another check against pockets.

I don't think that's necessary, if it's not in the pocket slot then it's either equipped in the right place or somewhere in the inventory which doesn't count as clothing

DangerRevolution:
> > It would probably be better to have clothing check if it's where it should be before doing things instead of adding yet another check against pockets.
> 
> I don't think that's necessary, if it's not in the pocket slot then it's either equipped in the right place or somewhere in the inventory which doesn't count as clothing

Death is suggesting future-proofing clothing; rather than band-aiding this. i.e if we get more clothing slots in the future, or a character gets some new type of wacky clothing slot or feature. Less so about overdoing this simple fix and more future-prevention.

Pspritechologist:
> > It would probably be better to have clothing check if it's where it should be before doing things instead of adding yet another check against pockets.
> 
> I don't think that's necessary, if it's not in the pocket slot then it's either equipped in the right place or somewhere in the inventory which doesn't count as clothing

There are a variety of situations where it may be equipped to something that *isn't* a specific 'pocket' slot, and designing this in a way that any additions in YAML cause this system to break isn't a good design pattern.
Serialization needs to be considered as a variable when designing systems, they should never break because content was added.

DinoWattz:
> Death is suggesting future-proofing clothing; rather than band-aiding this. i.e if we get more clothing slots in the future, or a character gets some new type of wacky clothing slot or feature. Less so about overdoing this simple fix and more future-prevention.

If someone were to add a normal clothing slot this wouldn't affect much since this is only aimed towards masks and hats _(ones with "HidesNose" and "HidesHair" tag)_;
But if someone were to add a clothing slot that accepts almost anything like a pocket slot but isn't one ~~I'd hunt them down myself~~ it'd really cause an issue similar to the one this PR is trying to fix.

I'll see if I can improve it somehow.

Pspritechologist:
> > Death is suggesting future-proofing clothing; rather than band-aiding this. i.e if we get more clothing slots in the future, or a character gets some new type of wacky clothing slot or feature. Less so about overdoing this simple fix and more future-prevention.
> 
> If someone were to add a normal clothing slot this wouldn't affect much since this is only aimed towards masks and hats _(ones with "HidesNose" and "HidesHair" tag)_; But if someone were to add a clothing slot that accepts almost anything like a pocket slot but isn't one ~I'd hunt them down myself~ it'd really cause an issue similar to the one this PR is trying to fix.
> 
> I'll see if I can improve it somehow.

This already exists?? Don't spider people literally have a non-pocket free slot??
In a game where variables are free to be modified, it simply isn't your place to decide what way to use them is valid. If data is exposed, it muse be accounted for.

You also just literally don't gain anything by not doing it the proper way.

DinoWattz:
> This already exists?? Don't spider people literally have a non-pocket free slot?? In a game where variables are free to be modified, it simply isn't your place to decide what way to use them is valid. If data is exposed, it muse be accounted for.
> 
> You also just literally don't gain anything by not doing it the proper way.

Arachnids have two extra pocket slots, I checked and they really are flagged as pocket slots, I'm more afraid of whatever is called "suitstorage".
Either way I already have an idea of how to change the code now.

DinoWattz:
It now only checks for the necessary slots, masks would only hide your nose/hair if it's on your face and hats if it's on your head.

Axionyxx (approved):
Wouldn't this also affect our upstream?

DinoWattz:
> Wouldn't this also affect our upstream?

Just tested here and upstream also suffers from the same issue (#982), would fixing it here also fix it there? I'm kind of new to this

Axionyxx:
> > Wouldn't this also affect our upstream?
> 
> Just tested here and upstream also suffers from the same issue (#982), would fixing it here also fix it there? I'm kind of new to this

No, its the other way around. They cannot take fixes from here as we are under an incompatible license, however, changes they make will make their way down to us. It is better to contribute fixes to them, and have it make its way down to us.

Axionyxx:
Long story short, changes that affect files that aren't namespaced as DeltaV, Nyanotrasen or SimpleStation(I think) should be sent to https://github.com/space-wizards/space-station-14 instead, as that way it'll reach most server, instead of only benefiting us and our downstreams.

DinoWattz:
> No, its the other way around. They cannot take fixes from here as we are under an incompatible license, however, changes they make will make their way down to us. It is better to contribute fixes to them, and have it make its way down to us.

> Long story short, changes that affect files that aren't namespaced as DeltaV, Nyanotrasen or SimpleStation(I think) should be sent to https://github.com/space-wizards/space-station-14 instead, as that way it'll reach most server, instead of only benefiting us and our downstreams.

Alright, thanks for telling me

so I should send a PR to upstream instead, I'll close this one.

PR #1036, opened 2024-03-30
Barber

DangerRevolution:
## About the PR

Title.
## Why / Balance

Allows for Barbershop RP; which already exists, but in limited quantities as most people don't know it's mapped (because it's not mapped on every station, hence this is a station-specific role). and actually leverages the existence of the Barber system for something interesting. Again, use of this role is a mappers choice.

## Media

- [X] PR does not require an ingame showcase

**Changelog**

:cl:
- add: Barbers are now licensed to operate on station.


DangerRevolution:
Still need to think of something interesting for their job description & PDA but PR otherwise finito.

IamVelcroboy:
I personally like this. I think they should have a unique uniform if we add them though. 

I will add though, that this was initially voted down when I first suggested it. So, I'll let other maintainers/heads weigh in too.

DangerRevolution:
this and #1037 are covered in upcoming loadouts, when someone fixes the ID system :))

Timemaster99:
> this and #1037 are covered in upcoming loadouts, when someone fixes the ID system :))

#1218 allows for a custom ID inside of the PDA loadout slot to change the job name. You could reuse this system to make these roles now

PR #1041, opened 2024-03-31
PsychologistPDA can now be used as a medical scanner

blueDev2:
## About the PR

Made a psychologist able to use their PDA as a scanner
## Why / Balance

While a psychologist should focus on mental health, they should still be able to act as a doctor in the downtime (which is a fairly long amount of time)
The medical scanner for the psychologist PDA will be 2x slower than a standard doctor PDA
## Media

https://github.com/DeltaV-Station/Delta-v/assets/89804215/656bdf03-57fb-443c-b6fb-a87bf1be5d15

## Breaking changes

**Changelog**

:cl: blueDev2
- tweak: PsychologistPDA can now be used as a slow medical scanner

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/pda.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/pda.yml
@@ -903,7 +903,7 @@
     accentVColor: "#447987"
 
 - type: entity
-  parent: BasePDA 
+  parent: BasePDA
   id: PsychologistPDA
   name: psychologist PDA
   description: Looks immaculately cleaned.
--- a/Resources/Prototypes/Entities/Objects/Devices/pda.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/pda.yml
@@ -903,7 +903,7 @@
     accentVColor: "#447987"
 
 - type: entity
-  parent: BaseMedicalPDA #Delta-v Make PyschologistPDA able to use medical scan
+  parent: BasePDA 
   id: PsychologistPDA
   name: psychologist PDA
   description: Looks immaculately cleaned.
@@ -915,7 +915,11 @@
     borderColor: "#d7d7d0"
     accentVColor: "#447987"
   - type: Icon
-    state: pda-medical
+    state: pda-medical 
+  - type: HealthAnalyzer #Delta-v - (Start) Make PyschologistPDA able to use medical scan
+    scanDelay: 2 # Make the scan slower as psychologists should focus on mental health
+    scanningEndSound:
+      path: "/Audio/Items/Medical/healthscanner.ogg" #Delta-v - (End) Make PyschologistPDA able to use medical scan
 
 - type: entity
   parent: BasePDA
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Objects/Devices/pda.yml
```diff
@@ -903,7 +903,7 @@
     accentVColor: "#447987"
 
 - type: entity
-  parent: BasePDA
+  parent: BaseMedicalPDA #Delta-v Make PyschologistPDA able to use medical scan
```
DangerRevolution: because this is an upstream item; this could lead to a lot more future changes for the Psychologist PDA than just being a med scanner, if Wizden add more onto the MedScanner but specifically avoid the Psychologist Scanner; which would be undocumented & maybe unwanted. It'd probably be safer to mirror the 'medscanner' component and tag it as Delta-V.

DangerRevolution: i.e (indentation is dumb on github but)
```
- type: HealthAnalyzer
        scanDelay: 2-3 (make it a slightly worse med doctor PDA to signify that Psych shouldn't be doing this)

blueDev2: I get the issue in theory, but I really can't think of anything else that would be in BaseMedicalPDA besides the medical scanner
I do think making the analyzer take longer is a good idea, so I will mirror anyways

DangerRevolution (commented):
We have no maintainer but take this as an approving review

IamVelcroboy:
Hmm, I don't know if this is w/in the scope of what a psych does. They aren't medical docs.

blueDev2:
I don't personally have strong feelings either way, but I have seen psychologists complain about it in game.
If the psychologist is supposed to be a mostly-rp role, it probably should be under service, not medical.

Also, if it doesn't really need medical knowledge, the medical playtime requirement is a bit odd.

Adrian16199:
I believe it has medical playtime due to them having medical access.
I think its fine that they can *somewhat* do medical stuff on the side but they dont neccesarily need a pda with the scanner.
If its really neccesary then health scanner in the medivend or ask CMO for a new PDA.
It really is just a side thing that should be optional and not a thing that a psychologist should be focusing at.

blueDev2:
The only issue with health analyzers is the battery requirement, which I think is a reasonable downside.

As I said, I don't really have strong feelings about it. I would, however, like to see if any maintainer would like to accept or deny it.
More of a "put this issue to bed" kinda thing, so we have precedent over it.

IamVelcroboy:
I've weighed in. I'll wait to see what @Colin-Tel says for a second opinion.

Colin-Tel:
> I've weighed in. I'll wait to see what @Colin-Tel says for a second opinion.

I'm personally kind of against it. I do still remember when CPR was a thing, that someone took issue with the fact that the psychologist had CPR training.

dvir001:
Psychologists don't have actual medical training, and dont need the scanner based on RP standpoint.

IamVelcroboy:
Thanks for the contribution but I'm closing this as it seems the consensus is against it.

PR #1047, opened 2024-04-01
Change Survival Event Weights

rosieposieeee:
## About the PR

Changed some weights:

- Ninja is more likely (3 --> 7.5) because it is by far the most interesting survival antag and its a shame it doesn't show up very often. Still lower than upstream, which is 10.
- Dragon is slightly more likely (1 --> 2.5) because I think I've maybe seen a dragon spawn that wasn't a ninja call-in... oooonce??? Still lower than upstream, which is 5.
- Revenant is less likely (7.5 --> 3) as it is currently not very interesting and happens way too often. I have no idea why upstream would put this at 10, maybe because all the other interesting deadly shit is also 10.
- XenoVents is slightly more likely (1 --> 2.5) because, like Dragon, it seems to be *extremely* rare right now. I remember back when it was 4 it was WAY too often but now its a little too rare IMO. Even modified, its still half the weight of spiders & slimes.
- MouseMigration was commented out because it is Literally Nothing and wastes the slot of a potentially more interesting event.
- CockroachMigration, ditto above.
- MothroachSpawn, ditto above.
- VentCritters (??!) was commented out FOR BEING IDENTICAL TO MOUSEMIGRATION. God, no fucking wonder the mouse one seemed to be the most common event, it was DUPLICATED.
- MassHallucinations was commented out, also an uninteresting one. You hear a zombie groan or something and for a moment think this round might *actually* be interesting, but nope! Just pick the paracusia trait if you want this to happen.

**Changelog**

:cl:
- tweak: Tweaked survival event weights to be more fun.

### DIFF
--- a/Resources/Prototypes/DeltaV/GameRules/events.yml
+++ b/Resources/Prototypes/DeltaV/GameRules/events.yml
@@ -9,7 +9,7 @@
       path: /Audio/Announcements/aliens.ogg
     earliestStart: 20
     minimumPlayers: 15
-    weight: 1
+    weight: 2.5
     duration: 60
   - type: VentCrittersRule
     entries:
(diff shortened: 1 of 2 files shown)


Axionyxx:
This should be discussed in an issue prior to an actual pull request being made

rosieposieeee:
i was kind of hoping this pr could start the discussion...

stellar-novas:
Yeah, maybe reopen this and mark it as a draft? I think it makes more sense to keep discussion in one place.

Axionyxx:
Pull requests aren't the place for that. You use issues for the initial discussion.

AitorLogedo:
that decision should be made by the headadmin not the maintainer.

PR #1049, opened 2024-04-02
shoukou atmos rework and tweaks

deltanedas:
## About the PR
mostly started as getting a good gas recycler setup, but now has all atmos aspects of shoukou improved
- since thermomachines are really slow it both takes years for the co2 to heat up and the o2 to cool down. so i made it use teg burn chamber for heat and a radiator array for cooling the oxygen from epic temperatures. it makes lots of oxygen and is easy to control it
- teg is now good similar to submarine
- made teg and smes share smeses and renamed them main/aux to reflect this (with ame off 3 would be empty 2 would be full)
- removed the annoying catwalks, unnecessary pipes and mixers but still having outlets. nobody is going to use that for trit anyway. if you want to filter nitrogen etc (you should never do this) theres steel sheets to do that or whatever else you want.
- plasma chamber has an injector like all the others so you can filter that since its the only gas you should actually filter
- epi has proper atmos controls now
- chef freezer uses passive vent for cooling air vent for pressure
- med finally has exit button

## Media
teg

recycler + stuff

epi atmos

**Changelog**
:cl:
- tweak: Reworked Shoukou's atmospherics setup. The Gas Recycler is now usable among other things.


Adrian16199:
Honestly, I wasnt sure how to handle shoukou's atmos, there were many complaints but not sure how so uh... thanks for handling it.

PR #1069, opened 2024-04-09
Fix shuttle structure construction

IamVelcroboy:
## About the PR
Title + I removed some unnecessary Nyano references and comp-part duplications. 

## Why / Balance
System was pulled from Nyano but was incomplete and not updated with the fixes/edits from upstream. This created visual glitches and other wonkiness.

**Changelog**
:cl: Velcroboy
- fix: Fixed shuttle wall deconstruction causing visual glitches!


Axionyxx:
We should replace these with whats on upstream

IamVelcroboy:
> We should replace these with whats on upstream

That's exactly what this does

deltanedas (approved):
lgtm

PR #1076, opened 2024-04-10
Ifriit gets theirs power back

Adrian16199:
## About the PR

Brings back the power of ifriit.
## Why / Balance

People complained that ifriit weak, adjusted it to be as balanced as i can imagine/possible since i cant go 1,5 stacks of fire and more than 2 is just overkill on somethin like this.
## Media

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/PyrokinesisPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/PyrokinesisPowerSystem.cs
@@ -55,7 +55,7 @@ namespace Content.Server.Abilities.Psionics
             if (!TryComp<FlammableComponent>(args.Target, out var flammableComponent))
                 return;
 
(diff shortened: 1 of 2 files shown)


deltanedas:
should consider moths taking extra fire damage could get instacrit still

Axionyxx:
This is still going to deal too much damage in some cases. Ifriits could get another power temporarily but pyro is just not happening with how deadly fire is right now

Adrian16199:
> This is still going to deal too much damage in some cases. Ifriits could get another power temporarily but pyro is just not happening with how deadly fire is right now

Im fully willingfull to make it do only one stack but otherwise, fair.

stellar-novas:
Could pyro be made to deal some burn directly to get the ideal 1.5 fire stacks?

Adrian16199:
> Could pyro be made to deal some burn directly to get the ideal 1.5 fire stacks?

It depends really, the question is, is it ideal?
One stack on a moth is 28 damage
One stack on a human is 14.
What is ideal, less than 50 but not below 30?
Moths/dionas will be extremely countered just due to this being heat.

PR #1078, opened 2024-04-11
Spare ID cabinet will no longer drop a card on destruction

Axionyxx:
## About the PR
Also makes it take a lot of damage before it'll actually break

**Changelog**

:cl:
- fix: Fixed the spare ID safe being destructible


Mnemotechnician:
How will the heads choose an acting captain then (as required per SOP)?

Bonktrauma:
> How will the heads choose an acting captain then (as required per SOP)?

You are not meant to break the safe, your meant to ahelp and ask for a biscuit

PR #1086, opened 2024-04-12
lower use time on barber scissors

UnicornOnLSD:
## About the PR

Removed the huge use time on Barber scissors, it now takes 5 seconds and is audible, just as before but doesn't take years.
## Why / Balance

The use time was waaaay too long before. I don't have specific numbers but I am quite certain that giving someone a haircut and dying their hair took around three minutes. It may not seem that bad but try for yourself and see.
## Technical details

just snapped the lines that added time, it's just the same as upstream scissors now.
## Media

**Changelog**
:cl:

- tweak: Barber's scissors have been sharpened and can give haircuts properly.


Adrian16199:
First question: How long does it take approximately to do it with this PR in motion if its not instant.
If its instant then thats a massive no as if it will cause the same issue as people abusin it and makin everyone have a hairstyle they dont like as it had before.

pissdemon:
I'm not sure about this change. I think the reason the long doafter is in place is because people would abuse the barber scissors to give people unwanted forced haircuts, e.g. people buckled to beds in medbay. You can't even easily change your hair back because barber scissors are hard to get and mirrors currently don't work. The long doafter also makes it super obvious it's happening and who is doing it if they try.

I also just tried the scissors locally and the delays work as intended. Changing the hair takes 20s as in the yaml. Changing color also takes 20s, so does adding/removing a hair slot. I gave an Urist (with existing hair) a new haircut and hair color, and it took ~40s for me. That's not 3 minutes. Even in the worst case (bald and shaved man gets dyed hair and dyed beard) it's 120 seconds, and I don't think it's a situation that happens a lot.

UnicornOnLSD:
> First question: How long does it take approximately to do it with this PR in motion if its not instant. If its instant then thats a massive no as if it will cause the same issue as people abusin it and makin everyone have a hairstyle they dont like as it had before.

> it now takes 5 seconds and is audible

That's literally in the first sentence

UnicornOnLSD:
> I'm not sure about this change. I think the reason the long doafter is in place is because people would abuse the barber scissors to give people unwanted forced haircuts, e.g. people buckled to beds in medbay. You can't even easily change your hair back because barber scissors are hard to get and mirrors currently don't work. The long doafter also makes it super obvious it's happening and who is doing it if they try.
> 
> I also just tried the scissors locally and the delays work as intended. Changing the hair takes 20s as in the yaml. Changing color also takes 20s, so does adding/removing a hair slot. I gave an Urist (with existing hair) a new haircut and hair color, and it took ~40s for me. That's not 3 minutes. Even in the worst case (bald and shaved man gets dyed hair and dyed beard) it's 120 seconds, and I don't think it's a situation that happens a lot.

that is a fair point, I just find it excessively long, if for X reason it's not accepted for the delay to just be reverted to default, switching it to 10s is an alternative.

UnicornOnLSD:
Also, to add on the "it's not an issue", I find that it is, because sitting still for 40 seconds waiting for a bar to fill up (changing hair, and dying it) is not really uh, enjoyable. Thus why I reverted it to default, if not that making it 10 instead of 20 is still better.

Adrian16199:
Eh, 10 sounds more fittin than 5 to react to someone while you are typing. 5 seconds sounds like you typin and someone sends you thro disposals without you bein able to react unless you have ptsd from it.

UnicornOnLSD:
fair point

UnicornOnLSD:
now 10 seconds.

IamVelcroboy:
Honestly, what would be the best solution is if there was a consent window that popped up similar to the cloning one. `"Someone wants to give you a haircut. Do you consent?"`

pissdemon:
> Also, to add on the "it's not an issue", I find that it is, because sitting still for 40 seconds waiting for a bar to fill up (changing hair, and dying it) is not really uh, enjoyable. Thus why I reverted it to default, if not that making it 10 instead of 20 is still better.

I guess it's a fair point if you are cutting your own hair, but if you are cutting someone else's, you can always pass that time with roleplay. 40 seconds are over quick when you're talking with someone.

> Honestly, what would be the best solution is if there was a consent window that popped up similar to the cloning one. `"Someone wants to give you a haircut. Do you consent?"`

I think a better solution might be to keep the doafter at 20s, but much shorter for messing with your own hair.

IamVelcroboy:
> I think a better solution might be to keep the doafter at 20s, but much shorter for messing with your own hair.

When doing your own hair, should be like a 80% chance of getting a botched haircut

UnicornOnLSD:
>I guess it's a fair point if you are cutting your own hair, but if you are cutting someone else's, you can always pass that time with roleplay. 40 seconds are over quick when you're talking with someone.

I've lived it five times and no it isn't, main reason I'm doing this

IamVelcroboy:
> > I guess it's a fair point if you are cutting your own hair, but if you are cutting someone else's, you can always pass that time with roleplay. 40 seconds are over quick when you're talking with someone.
> 
> I've lived it five times and no it isn't, main reason I'm doing this

tbf, it is a good point. Barber shops/hairdressers are known for their social/gossip dynamic.

UnicornOnLSD:
>tbf, it is a good point. Barber shops/hairdressers are known for their social/gossip dynamic.

Out of all the times I've ever got a hair cut, only once it was in a barber room or by someone who was just acting as a barber
yes, in concept, but in practice if you just want to change your hair it likely isn't for a barber, and barbers can talk all they want and it shouldn't just be based on how long it takes to use the scissors, long use time is cumbersome at most.
If someone does take time going to whoever the barber is in let's say, Arena, they'll likely take time talking to them too, regardless of the use time of the scissors

PR #1097, opened 2024-04-15
Head of personnel's hardsuit

Adrian16199:
## About the PR

TO-DO:
- [ ] Make hardsuit fax real.
- [ ] Decide if to use simplestation sprite or use our own.
## Media

## Breaking changes

**Changelog**


ps3moira:
HoP doesn't need a hardsuit as there's no part of their job where they need a armored space suit to protect them from dangers unlike the CE, RD, CMO, HoS and Captain.

rosieposieeee:
if for nothing else at least its good for parity with the other heads all having suits of some kind, and the sprite is cute. i actually think this is really well balanced as you have it. maybe put 'only resistant to papercuts' in the description instead of in a yaml comment, because its funny, and consider removing or lowering the explosion resistance (so hop cant become the bomb squad) and making the slash resistance more minor. maybe also call it a softsuit instead of a hardsuit...?

BlitzTheSquishy:
I would say it should stay called a "hardsuit" due to well, the clown's being called a hardsuit. Despite it being literally made of an EVA and some crayons.

And to what ps3moira said.. CMO's is not armoured, it's basically just a fancy voidsuit for the CMO.

Adrian16199:
> HoP doesn't need a hardsuit as there's no part of their job where they need a armored space suit to protect them from dangers unlike the CE, RD, CMO, HoS and Captain.

How am I suppose to do paperwork when everything is spaced!
Outrages.

BlitzTheSquishy:
> > HoP doesn't need a hardsuit as there's no part of their job where they need a armored space suit to protect them from dangers unlike the CE, RD, CMO, HoS and Captain.
> 
> How am I suppose to do paperwork when everything is spaced! Outrages.

cant recover Ian's body from your office being bombed, if you cant survive in space yourself

ps3moira:
> I would say it should stay called a "hardsuit" due to well, the clown's being called a hardsuit. Despite it being literally made of an EVA and some crayons.
> 
> And to what ps3moira said.. CMO's is not armoured, it's basically just a fancy voidsuit for the CMO.

CMO's hardsuit is supposed to be a lot more protective over diseases, basically a better hazmat suit but we don't have much diseases content in 14 than 13.

BlitzTheSquishy:
I do believe the main purpose of this PR would be to have all of command to have recognizable hardsuits. Cause HoP is basically unrecognizable if an event that requires EVA to be used. (would likely be primarily for security or other commands to get easy visual resignation).

Adrian16199:
> if for nothing else at least its good for parity with the other heads all having suits of some kind, and the sprite is cute. i actually think this is really well balanced as you have it. maybe put 'only resistant to papercuts' in the description instead of in a yaml comment, because its funny, and consider removing or lowering the explosion resistance (so hop cant become the bomb squad) and making the slash resistance more minor. maybe also call it a softsuit instead of a hardsuit...?

Sure.
I would like it to be a hardsuit that is unique but isnt neccesarily clownsuit upgrade dot TM compared to EVAs
Especially if the constant argument is that they dont need it aka it would be used for anythin that isnt an office hence why I wanted to balance it to be just tiny bit better than an EVA but not good enough to be worn EVERYWHERE and have something unique to it that would scream office.
As far as im concerned tho, I dont think the fax gimmick will work which i had a second gimmick in mind and that bein havin camera function in the helmet.
Not sure if people would find that amusing or bad.

Adrian16199:
> I would say it should stay called a "hardsuit" due to well, the clown's being called a hardsuit. Despite it being literally made of an EVA and some crayons.

Could change that tbh.
Only gripe with the softsuit hardsuit argument is the sprite itself.
The clown's hardsuit doesnt look very soft.

UnicornOnLSD:
>Only gripe with the softsuit hardsuit argument is the sprite itself.
The clown's hardsuit doesnt look very soft.

If we're going full on terminology, hardsuit be the word that should be used, it's a suit that goes in space, hardsuit. If we are just going for name value calling it softsuit just for the lols is worth to, your call honestly

PR #1110, opened 2024-04-20
fixes harpie issues with tag inheritance

### LINKED ISSUE #1109: Harpies can't wear prison jumpskirt + you can't chameleon skirts
i'm fixing this in an hour when i get VSC working but just in case I give up, opening this as an issue

introduced in Harpy update, the skirts are missing chameleon tags and the prison jumpskirt is missing the skirt tag as tags aren't inherited

DangerRevolution:
## About the PR

fixes #1109
adds commented reminders to keep the skirts updated with whatever is added to the base
## Why / Balance

they're issues

**Changelog**
:cl: DangerRevolution
- fix: Chameleon clothing can now be set to skirts again
- fix: Harpies can now wear chameleon clothing


DangerRevolution:
this should also fix https://discord.com/channels/968983104247185448/1229863953560109196

issues were with syndicate jumpskirt? not sure why, it should inherit tags correctly, worth testing.
also fixes issue with Chameleon jumpsuit by overwriting Wizden null tag

DangerRevolution:
okay, after testing, the Syndicate jumpskirt is broken because of the lack of tag inheritance *again*. Fixed and tagged with Delta-V markers.

Resources/Prototypes/Entities/Clothing/Uniforms/base_clothinguniforms.yml
```diff
@@ -50,6 +50,11 @@
   - type: WirelessNetworkConnection
     range: 1200
   - type: StationLimitedNetwork
+  - type: Tag #Delta-V : Adds tags to fix the Syndicate Jumpskirt, which is parented to this? this needs to be updated with *every* new tag given to UnsensoredClothingUniformBase, as tags aren't inherited
```
Axionyxx: Why is tag code like this? Can't we just have a system that adds or removes tags on the fly

Axionyxx: I might try to create something basic to override tags without needing to change the base tag, because this seems like a nightmare to keep up with. And even if it isn't, its suboptimal :<

DangerRevolution: not sure if i'm misreading but the issue is that a child receiving tags overwrites the tags of the parent, this usually works however as Harpies need a "skirt" tag to wear their associated skirts, we're overwriting the parent tag to add the skirt tag in. In an ideal world, we should be able to just inherit the parent's tag and add the skirt tag over the top. 

Syndicate Jumpskirt parents off of this sensor-off version, to my understanding so the person making it can make a funny "haha they're evil so their sensor is off" code comment. Other solution aside from this gross tag nonsense is to just reparent the Syndicate Jumpskirts to ClothingUniformSkirtBase below this.

DangerRevolution: Tag code is pretty disgusting throughout the repo, inheritance would be nice but would instead cause problems when you *do* want to overwrite tags.

Axionyxx: So, you're only adding the skirt tag here, right?

DangerRevolution: i'm adding the skirt tag to: `UnsensoredClothingUniformSkirtBase` , in order to let Harpies wear the Syndicate Jumpskirt, I have to add the other tags in order to not overwrite the other tags needed.

To ClothingUniformSkirtBase, I'm adding the WhitelistChameleon tag, so the Chameleon jumpsuit can now disguise as skirts again.

To the Chameleon jumpsuit, I'm adding the skirt tag so Harpies can put it on first without having to disguise it as something else.

DangerRevolution:
- [X] TODO: Fix CC clothing

this is actually already fixed in this PR

DangerRevolution:
actually gonna reopen this PR, because issues keep cropping up and this PR should fix it, things can be redone when/if tag inheritance is made :O

A-z-z-y (approved):
Looks good!

PR #1111, opened 2024-04-20
survival tweaks

AitorLogedo:
## About the PR

Re-added immovable rod and made loneop start with 60 telecrystals
## Why / Balance

A fair ammount of people don´t like the current state of survival and since changing weights requieres a procces, Im hoping with this, survival can get a bit more interesting without any major changes
## Media

- [x] This PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Survival tweaks, loneops start with 60 telecrystals and there is some rods coming near the stations...


Axionyxx:
I've said this several times before, this sort of stuff needs discussion. This is also a change that needs to be discussed with admins, and if is doing several unrelated things at once

PR #1113, opened 2024-04-20
Basic tag override implementation

Axionyxx:
## About the PR
I have no idea if this plays nicely with tests, we'll see.

## Why / Balance
Tag inheritance hurts. Hopefully this makes it less of a merge conflict nightmare


deltanedas:
TagOverride is a bad name for this since its the opposite, redefining tags in `- type: Tag` is what overrides them

TadJohnson00:
Gonna assume this is abandonware unless this gets a response in, say, a week

PR #1115, opened 2024-04-21
Add paramedic winter coat to paramedic lockers that dont have the voidsuit

BlitzTheSquishy:
## About the PR

The locker that contained the voidsuit had the winter coat in it, but it was not present in the other locker.

## Why / Balance

Why have winter coat in one locker but not the other?

## Technical details

I just added a line from the same locker type from WizDen to Delta (WizDen only has one paramedic locker, which does have the coat, while Delta is missing it from one of their two lockers)

## Media

## Breaking changes

**Changelog**

:cl:
- add: The Non-Voidsuit Paramedic Lockers now have their Winter Coats.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/medical.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/medical.yml
@@ -177,6 +177,7 @@
     contents:
       - id: ClothingOuterCoatParamedicWB
       - id: ClothingHeadHatParamedicsoft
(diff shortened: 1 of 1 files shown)


DangerRevolution:
this should also be upstreamed

BlitzTheSquishy:
> this should also be upstreamed

This doesnt work on upstream
they have ONE paramedic locker while deltaV has two
The one in Wizden is equivilent to the Hardsuit variant we have here

I also fucked up this PR so i started again

PR #1116, opened 2024-04-21
Para Winter Coats in their Locker

BlitzTheSquishy:
## About the PR

Added parity between both paramedic locker types

## Why / Balance

The locker for paramedics that didnt have the voidsuit were lacking the void suit

## Technical details

It seems that the two objects under the same name ind D-V and Wizden dont contain the same items. As one of the lockers is missing the winter coat on the DV end but not WizDen (Wizden only has one paramedic locker)

## Media

## Breaking changes

**Changelog**

:cl:
- add: Winter Coats now in Paramedic Lockers that dont have a Void Suit


BlitzTheSquishy:
Fuckin I cant read *a*

BlitzTheSquishy:
JUST reclarified for myself

On wizden this locker looks like this 
on Delta V it is not the same (it is missing the coat) this is a Delta V merge, and cant be upstreamed as it is already present upstream

```yaml
- type: entity
  id: LockerParamedicFilled
  suffix: Filled
  parent: LockerParamedic
  components:
  - type: StorageFill
    contents:
      - id: ClothingOuterHardsuitVoidParamed
      - id: ClothingOuterCoatParamedicWB
      - id: ClothingHeadHatParamedicsoft
      - id: ClothingOuterWinterPara
      - id: ClothingUniformJumpsuitParamedic
      - id: ClothingUniformJumpskirtParamedic
      - id: ClothingEyesHudMedical
      - id: ClothingHandsGlovesLatex
      - id: ClothingHeadsetMedical
      - id: ClothingMaskSterile
      - id: HandheldGPSBasic
      - id: MedkitFilled
        prob: 0.3
```

IamVelcroboy:
The DV variants should probably just be moved to the DV directory anyway.

BlitzTheSquishy:
> The DV variants should probably just be moved to the DV directory anyway.

This is true, but I uh have a feeling that could brake a bunch of maps using maps that use one of the types of para locker

IamVelcroboy:
> > The DV variants should probably just be moved to the DV directory anyway.
> 
> This is true, but I uh have a feeling that could brake a bunch of maps using maps that use one of the types of para locker

As long as you don't change the ID name, it'll be fine.

BlitzTheSquishy:
Okay with the changes i just made the hardsuitlocker which is equivilent to the upstream version remains as is
but the variant locker has been moved into the Delta-V namespace 
image
without any damage to submarine's lockers. 
While also adding the change of the winter coat.
Can go with the first solution depending on which one is decided to be better as I am not too sure.

BlitzTheSquishy:
Just in case I myself wasnt clear, there is two ways to do this without it affecting maps as they are
Moving it to the DV namespace
Or simply just adding the winter coat

I am going to need some sort of idea what would be best,
And before anyone says upstream, para's already have the winter coats on their lockers upstream.

Axionyxx:
Imo, this'd be better as a loadout option, as it means less locker clutter.

PR #1118, opened 2024-04-21
Port carrying & pseudo-item tweaks & features from frontier

Mnemotechnician:
## About the PR
This ports the following 2 PRs of mine from Frontier Station: https://github.com/new-frontiers-14/frontier-station-14/pull/834, https://github.com/new-frontiers-14/frontier-station-14/pull/882

And introduces the following changes:
1. Fixes dropping the carried person when walking to a different grid (from station to shuttle or vice versa. Walking into space however will still make you drop them) and also makes sure that pressing shift while being carried does not make you escape.
2. Ensures that the carried person is always centered relative to the parent (under certain conditions, such as walking near a gravitational anomaly, their position can change, and that leads to really weird effects)
3. Fixes the mass contest in CarryingSystem that caused stronger entities to take longer to escape than weaker ones.
4. Adds popups for when you're getting picked up or being stuffed into a bag as a pseudo-item (e.g. a felinid)
5. Adds an action to stop escaping an inventory. This action gets added to your action bar when you attempt escaping and gets removed when you stop or escape. It applies both to carrying and items (hampsters, felinids, whatever else).
6. Adds a sleep action for pseudo-items stuffed inside a suitable bag. The bag must have a special component, which is added to the base backpack item and thus inherited by all soft bags (duffels, satchels, etc). Contrary to a popular belief, sleeping IS PURELY COSMETICAL and does not provide healing. (Beds provide healing when you buckle into them and that healing does not depend on whether or not you're sleeping)
7. Makes it so that when you try to take a pseudo-item out of the bag (e.g. a felinid), you automatically try to carry them (if you don't have enough free hands, they will be dropped on the floor like usually), and enables you to insert the carried person into a bag, but only if they're a pseudo-item (e.g. felinid). 

All of these changes are split into separate commits and are up for discussion (frontier has all of them). In case some of them are considered undesirable, they can be reverted. 

## Why / Balance
1, 2, and 3 are fixes, so there's not much to say about them.

4 makes it clear that you're being picked up and not stolen from, as it's really easy to confuse those things. 

5 is done because it's really easy to accidentally press a movement button and start escaping while being carried. It's undesirable and leads to awkward situations because there's no way to stop your escape.

6 is a purely cosmetic action that may be used for RP.

7 is also just an RP thing, besides it makes more sense that when you try to "take out" a felinid from a bag, you actually take them in your ~~paws~~ hands rather than drop them on the ground.

## Technical details
Most of the changes are done to delta-v (or nyanotrasen) specific code, in cases where my changes touched upstream code I tried to add comments clarifying it.

1 adds two checks to CarryingSystem. The first check is added to the reparent event handler and aborts dropping the person if the new parent of the carrier is the new grid, which means that the carrier just switched grids. The second check aborts escaping if MoveInputEvent.HasDirectionalMovement is false, which should be self-explanatory and is copied from EscapeInventorySystem.

2 adds an update loop to CarryingSystem, which iterates over all carrier-carriee pairs and makes sure the cariee's local position is always (0, 0), and also makes them fall on the ground if they got reparented (e.g. inserted into disposals).

3 basically changes `time = baseTime * (carrierMass/carrieeMass)` with `time = baseTime * (carrieeMass/carrierMass)` by swapping the arguments to the mass contest.

4 literally just calls SharedPopupSystem.PopupEntity which is only shown to the person being picked up.

5 introduces a new action which is added to the entity when it starts escaping an inventory, and gets removed regardless of the result. Activating that action aborts escaping. Not much to tell about it...

6 re-uses the existing sleep action. When a pseudo-item gets inserted into the storage of an entity with AllowsSleepInsideComponent, this action gets added to it, and when it gets out of an inventory, it gets removed. Activation of that action is handled by SleepingSystem, and is thus equivalent to sleeping in a bed. It also shows a popup when a pseudo-item tries to fall asleep while being inside an entity with ASIC.

7 adds a an InnateVerb for entities with CarryingComponent if the carried person has a PseudoItem component and the target is a container, and that person can fit inside that container. It also makes it so that OnGettingPickedUpAttempt in PseudoItemSystem tries to call CarryingSystem.TryCarry (which is a new method added to it) and returns upon succcess, or exerts the old behavior (dropping the pseudo-item on the ground) if it does not succeed. Doing this required making OnGettingPickedUpAttempt virtual in SharedPseudoItemSystem and overriding it in the server-side counterpart.

## Media
https://cdn.discordapp.com/attachments/732665247302942730/1231633173008613426/weeee-2024-04-21_18.42.53.mp4?ex=6637aac7&is=662535c7&hm=d9bdc984faf42e89f8b90a45f6f897cc7899981e130111060dd99365c140b9f2&

This showcases the changes 4, 5, 6 and 7. The first 3 changes are too subtle.

## Breaking changes
None I hope?

**Changelog**
:cl:
- fix: Carrying is less likely to behave erratically or suddenly interrupt now.
- add: You can now see when someone is trying to pick you up, and also you can interrupt your attempt at escaping from their hands or inventory.
- add: You can now properly take Felinids out of bags and place them inside.
- add: Scientists have discovered that Felinids can sleep in bags.


Content.Server/Resist/EscapeInventorySystem.cs
```diff
@@ -5,6 +5,7 @@
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Storage.Components;
 using Content.Shared.ActionBlocker;
+using Content.Shared.Actions;
```
Adrian16199, line 8: forgot a comment here

Content.Server/Resist/EscapeInventorySystem.cs
```diff
@@ -13,6 +14,7 @@
 using Content.Shared.Resist;
 using Content.Shared.Storage;
 using Robust.Shared.Containers;
+using Robust.Shared.Prototypes;
```
Adrian16199, line 17: forgot a comment here aswell.

Content.Shared/Resist/EscapeInventoryCancelEvent.cs
```diff
@@ -0,0 +1,6 @@
+using Content.Shared.Actions;
```
Adrian16199, line 1: shouldnt this entire file be done in delta V namespace aka done in delta V folder?

Adrian16199 (commented):
Otherwise looks good.

Adrian16199:
Without counting all the merge conflicts

Content.Server/Resist/EscapeInventorySystem.cs
Axionyxx: Imports don't require comments

Content.Shared/Resist/EscapeInventoryCancelEvent.cs
Axionyxx: Imports don't require comments

Mnemotechnician: I'm not sure it's worth giving just one single event that's only used in *changed upstream code* a dedicated deltav namespace. But if other people share the same opinion, I can move it.

Adrian16199:
I mean, I wasnt aware that imports dont need one, not sure why but if thats the case then I suppose thats fair.

Mnemotechnician:
> I mean, I wasnt aware that imports dont need one, not sure why but if thats the case then I suppose thats fair.

We don't usually handle imports ourselves, our IDEs do it for us.

Adrian16199:
> > I mean, I wasnt aware that imports dont need one, not sure why but if thats the case then I suppose thats fair.
> 
> We don't usually handle imports ourselves, our IDEs do it for us.

I dont wanna peanut post or make this unrelated comment but... IDEs?

Mnemotechnician:
Oh my god, I just realized I messed something up and some resources were not committed... I will fix it now.

A-z-z-y (approved):
Yeah I like it, fixes & QoL is always appreciated, and small feature additions are a nice bonus. :]

PR #804, opened 2024-02-09
Modify antagAdvantage values.

stellar-novas:
## About the PR

Reduced or removed antagAdvantage values for several roles.
## Why / Balance

For several roles, antagAdvantage is too high as their job advantages either aren't actually used, or are not true advantages.
<details>
  <summary>More details, originally posted on Discord, https://discord.com/channels/968983104247185448/1150555473699934370/1205366269851996250. </summary>
  So, going through the list
 Psionic Mantis has an advantage of 5. No other epi role has an antag, the special things the mantis gets are
- psi knife
 - 10 slash 5 holy, sure, but afaik you don't take holy if you aren't "evil" so it's only 10 slash, and you have a 30% chance to electrocute yourself on every attack if the target isn't psionic, making it useless in most pvp fights
- psibreaker gun 
 - it comes with mindbreaker rounds, which deal 3 blunt and inject mindbreaker toxin. Getting lethal rounds from sec feels about as easy as just getting a whole gun.
- permission to mindbreak, and potentially slip in other drugs too?
 - this feels moot, since doctors also have this and they have no advantage

Cargo tech has an advantage of 2, with the comment saying it's cause externals access.
This is completely moot since techs don't have externals. Mats and autolathe (free tools) is pretty good, but it's not great and I think the full 20tc is deserved.

Salv tech has an advantage of 3, "External Access + Free hardsuit and weapons". The salvage weapons aren't ideal for pvp imo, I don't think externals is much of an advantage, I don't remember any cases of ex. a body being hidden in space, however, the weapons and similar contraband they can get definitely are. Maybe reduce to 2 advantage?

Lawyer, advantage of 2, "Security Radio and Access", having many more opportunities to steal sec stuff is pretty fair, and sec comms is very helpful.

Service worker, advantage of 1, "access", they just get service, and no other service role has an advantage, they should get 20tc. Seems to be a hold over from nyano.

Station engineer/Tech assistant, advantage of 3, "External Access + Engineering", you're trusted in a lot of higher security areas now, like substations, which imo is the main advantage, as tools are fairly easy to acquire, barring insuls and engi goggles. Maybe drop it to 2 instead?

[...]

PR #1157, opened 2024-04-29
reduce atmos antag advantage

deltanedas:
## About the PR
#804 but atmos only because its very annoying

10 tc -> 18 tc, 4 bonus fish

## Why / Balance
because you only had 10 tc your stealth options were extremely limited, you can get 1 thing like a storage implant then a soap or some shit, for no good reason
- plasma is limited so plasmaflooding like old days is both not allowed and infeasible
- axe has no helpful use so running around with it makes you valid
- firesuits are all over the station

**Changelog**
:cl:
- tweak: Traitor atmos technicians now get 18 TC instead of 10 TC.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Engineering/atmospheric_technician.yml
+++ b/Resources/Prototypes/Roles/Jobs/Engineering/atmospheric_technician.yml
@@ -3,7 +3,7 @@
   name: job-name-atmostech
   description: job-description-atmostech
   playTimeTracker: JobAtmosphericTechnician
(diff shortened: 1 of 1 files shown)


Lyndomen:
Isn't this more then the normal engi gets- 17? They get a door remote, which is pretty decent for making getaways and bolting stuff. I certainly think they need more then TEN, but something like 16 makes a lot more sense to me.

therealDLondon:
This is going to be adressed by administration very soon.
Generally non-admins shouldnt make PRs relating to antag balance, or balance in general, unless you've like, talked to us

deltanedas:
hence added you as a reviewer

PR #1160, opened 2024-05-02
split the chest rigs inventory

MilonPL:
## About the PR

Split all of the chest rigs into 2x3 and 2x4. You can no longer fit shotguns in them.

## Why / Balance

People used to fit some shotguns in the chest rig/security webbing. After some admin discussion, we decided that instead of implementing a whitelist, it's better to split them. They also get two slots less now, meaning they aren't just a "better version" of the belt.

## Technical details

N/A

## Media

webbing

## Breaking changes

**Changelog**

:cl:
- tweak: The security webbing inventory is now split into two parts. You can no longer fit a shotgun in it.


TadJohnson00:
Personally recommend splitting into two 2x2s and a 2x3 to avoid the shotgun issue
Or even two 2x2s and three 1x2s

Oh and make it apply to the CMO webbing :jor:

MilonPL:
> Personally recommend splitting into two 2x2s and a 2x3 to avoid the shotgun issue 
> Or even two 2x2s and three 1x2s

I'm afraid it would make the inventory UI be awfully long because of all the spacing, and it doesn't really change anything, as you can't fit a shotgun in it either way, no matter if it's two 2x2s or one 2x4

TadJohnson00:
Fair call then :3

WarMechanic:
I think chest rigs were recently changes to function like belts, so this change might require further discussion.

edit: security chest rig parenting off the security belt and not any other chest rig is weird, i think this pr is a good call

ps3moira:
this looks like a good PR to do upstream

deltanedas:
if you dont want to store shotguns in them (which is a good idea) just limit max item size to normal

MilonPL:
> if you dont want to store shotguns in them (which is a good idea) just limit max item size to normal

But then the webbing would just straight up be an "upgrade" to the secbelt. Why have a belt at that point? By removing two slots, it doesn't provide the same mechanical advantage, while still being able to fit items that you realistically could put in there.

ghost:
Milon and co, was there any consensus on this? What should I do with it?

MilonPL:
> Milon and co, was there any consensus on this? What should I do with it?

i still think it's better than what we currently have :blunt: 
merge if you feel like it, feel free to close otherwise

TadJohnson00:
I'm in favour of this change.

ghost (approved):
YAML seems fine, does exactly as advertised. TJ approves so I guess I'll do it even if Moira says it might be a better as a PR to do upstream, and Deltanedas has a point that this is pointless to remove two slots, but also it's part of delta's balance at the same time, so like... Ehh.

If anyone wants to come after me for hitting the funny button, I just look at code man. Go against Milon and TJ, they're the real ones who wanted this change, this PR had little discussion and you should make a PR to revert it.

MilonPL:
pro

ghost:
(it was barely discussed)

TadJohnson00:
If WizDen does decide to split security webbing and belts in terms of purpose in future, we can reassess. As of right now, this change is fine to me and I find the removal of two slots a-okay for balance.

PR #1168, opened 2024-05-05
Increase Glimmer Threshold for Psionic Mute

Lyndomen:
## About the PR
The Psionic Mute that happens occasionally generally doesn't add much roleplay value. While it has its place, this will make it occur less - as the higher Glimmer threshold is met less frequently. 

## Why / Balance
Right now it triggers between 200-500 glimmer. I'm thinking if we move it to trigger between 500-900, it becomes less frequent, it becomes a signal that Epistemics is **causing issues**, and it also mildly contributes to lowering glimmer.

## Technical details
Changed a 200 to 500, and changed a 500 to 900 in the NyanoTrasen events.yml. 

## Media

## Breaking changes
Not removing anything, just changing when this event can occur, so I am hoping there are no risks here. 

**Changelog**
:cl:
- tweak: Psionic Mute now has a higher Glimmer threshold for when it may occur.

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/GameRules/events.yml
+++ b/Resources/Prototypes/Nyanotrasen/GameRules/events.yml
@@ -88,8 +88,8 @@
   noSpawn: true
   components:
     - type: GlimmerEvent
-      minimumGlimmer: 200
-      maximumGlimmer: 500
+      minimumGlimmer: 500
+      maximumGlimmer: 900
       glimmerBurnLower: 18
       glimmerBurnUpper: 40
     - type: PsionicCatGotYourTongueRule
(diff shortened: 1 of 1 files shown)


Lyndomen:
Fox has given me the thumbs up to submit this, but she should probably also review it.

Resources/Prototypes/Nyanotrasen/GameRules/events.yml
```diff
@@ -88,8 +88,8 @@
   noSpawn: true
   components:
     - type: GlimmerEvent
-      minimumGlimmer: 200
-      maximumGlimmer: 500
+      minimumGlimmer: 500
```
DangerRevolution: again, depends on convention but this may need Delta-V comments as this is modifying Nyano code..

Lyndomen: Updated! Thanks.

Axionyxx (approved):

PR #1169, opened 2024-05-06
glacier real

deltanedas:
## About the PR
the return of the king

- most of the busywork done
- salv moved from above engi to below logi
- atmos now only has n2 miner, cans instead of chambers with miners because its a planet or something.
  oxygen is filtered out of the air, does infact work
- teg replacing le experiment room, using its own air filtration oxygen source and a liquid oxygen can to get started fast
- space heaters thrown around maints for when atmos dies and something gets bombed
- justiceroid above the tyrant with justice staff rooms and courtroom
- psychologist + brigmed replacing old lawyer/courtroom meme

no render because it somehow OOMs :trollface:

planetmap is fucked so no surface yet


WarMechanic:
looking forward to more maps without o2 miners

deltanedas:
its still effectively infinite just you have to do more than 1 thing :trollface:

WarMechanic:
As someone who wasn't there to experience an earlier version of Glacier, this map has alotta personality so I'm excited to experience it for the first time. I also like how the cargo shuttle can still go to space via FTL to the ATS, and the potential for a planetary base is awesome.

Do you think it would be possible to drop the outside temperature lower than -60 degrees? I feel like the DoT of being cold is relatively low (as well as being completely nullified by winter coats), which makes hull breaches seem pointless. That being said, I do not have any experience in regards to how greater moles within colder atmospheres would pose different problems for Atmospheric technicians.

noctyrnal:
Finally I can do actual map votes with glacier and not bait people

deltanedas:
> Do you think it would be possible to drop the outside temperature lower than -60 degrees? I

the main issue with this long ago was going much lower would lead to pressure damage, i think it could go colder though with lower threshold we have today

TadJohnson00:
> Do you think it would be possible to drop the outside temperature lower than -60 degrees? I feel like the DoT of being cold is relatively low (as well as being completely nullified by winter coats), which makes hull breaches seem pointless.

Hull breaches were never supposed to be a major danger to the crew on this station, and the whole area was supposed to be completely accessible with a winter coat and nothing more. The point of hardsuits on this map was primarily for overpressurised rooms, not underpressurised ones. At least as far as I recall.

IamVelcroboy:
Funny planet map space trash 
image
image

deltanedas:
space litter

IamVelcroboy (approved):
Ice cold

UnicornOnLSD:
POG IN THE CHAT !!

PR #1171, opened 2024-05-07
shipyard rewrite/port

deltanedas:
## About the PR
port shipyard from pre rebase, but:
- a ship can have multiple categories, unused group is now a category
- console whitelist support added so you can have centcom shipyard, nukie shipyard, emagged normal shipyard console, etc.
- a good bit of the console code is in shared so access is predicted
- instead of managing a single shipyard map and doing fucky stuff to make sure spawned ships dont hit eachother, ships get their own temporary map similar to the loneop shuttle.
  once it enters ftl space the temporary map gets deleted
- RENAMED PROSPECTER TO PROSPECTOR, GRAMMAR!!!!

currently unobtainable i forgor how its meant to be gotten

also someone needs to look over the prices a lot of them seem insanely high, maybe copy frontier

# Media
temp map gets created then deleted after it ftls

ship automatically docks, no stations have the funny docks mapped so it will basically be random

ships have name and a random id similar to frontier

# Ship changes
all ships:
- made sure they all scrubbers in every room, adding waste lines if they didnt have
- for ones with distro, gave them an air pump for the canister
- ran fixgridatmos
- named apcs
- gave the generators fuel to start with, use 1 pacman and not 1-2 super pacmen (15kw uranium gen migration gaming)
- removed `invalid` etc from the yml

barge:
- made booze vendor AA
- split power into like 3 apcs

helix:
- replaced cryobeakers with electrolyser since its for healing not making homemade grenades
- proper chem access on chem, gave chem facing windoors
- split power into 4 apcs

pulse:
- made airlock bolt properly with logic gates
- added power monitor
- AA booze vendor instead of 2 (two) glasses on the table
- split power into 4 apcs


Axionyxx:
UI code got changed with the incoming merge, this will have to wait until after that is merged

deltanedas:
yep has a comment to update when merged

Axionyxx (changes requested):
I'm gonna request 2 things here.

An arbitrage test, and a test that makes sure shuttles are valid (if there isn't already a test that does this.)

deltanedas:
> a test that makes sure shuttles are valid

as in having != 1 grids? shuttles being loadable is tested already

Axionyxx:
> > a test that makes sure shuttles are valid
> 
> as in having != 1 grids? shuttles being loadable is tested already

Stuff like that, yes. Along with making sure there's a shuttle console, and at least 1 dock

Bonktrauma:
Shipyard isn't meant to be obtainable,  im pretty sure it's one of the consoles meant to be mapped into bridge, might add it to shuttle creation though

deltanedas:
yeah mapping being in locker or research whatevers best

Axionyxx (commented):
Fix all arbitrages and I'll approve

IamVelcroboy:
> yeah mapping being in locker or research whatevers best

Was previously mapped on the bridge

IamVelcroboy:
We should add warp points to all the shuttles too. Can be in a follow up PR but should be done eventually.

Axionyxx:
> We should add warp points to all the shuttles too. Can be in a follow up PR but should be done eventually.

When thats done, it should be added to the tests too. Ensure the shuttles always have a warp point

Axionyxx:
But should probably be a seperate pr lol

deltanedas:
test itself has dirty disposing but theres no arbitrage, can fix test now or later

Content.Shared/CCVar/CCVars.cs
Axionyxx, line 1: We have our own cvars file iirc

Axionyxx (changes requested):
One small nitpick, lgtm otherwise

Axionyxx (approved):

PR #1206, opened 2024-05-15
paradox anomaly antag rework

deltanedas:
## About the PR
- made it a lot better
- moved spawn from nyano events.yml to deltav
- made MidRoundAntagRule just give spawn locations to AntagSelection

only issue right now is the mind entity has your previous name and not the new one, probably an upstream issue idk


Axionyxx (approved):
Lgtm

PR #1262, opened 2024-05-26
Add Syndicate/generic radio implants, remove Syndicate encryption key from uplink

pissdemon:
## About the PR

- Adds a generic radio implant for traitors. Adds an action button for accessing encryption key storage (acts like an encryption key only 1x2 storage implant), and lets you use the key that's inside to communicate on the key's radio channels.
- Adds a Syndicate radio implant for traitors. Allows the implanted player to use :t to talk on the Syndicate radio channel. The key is not removable.

As with any other implant you can only have one of each implanted.

### Syndicate radio implant (2 TC, or two implanters for 3 TC)

The Syndicate encryption key is in a bad place right now and too risky to use. If just one traitor using it gets compromised (which can happen during random searches in the worst case), sec has access to a key - and as per admin policy, they are allowed to use it with the CO giving permission. Even without express permission sec players might just use it anyway, even though it's against the rules, and it's hard to catch that happening. 

Making it an implant enables Syndicate cooperation and makes it more viable because you can mostly stop worrying about Syndicate comms being compromised. Sec would need to have a good reason to implant search someone, and it's harder to do solo since an implanter is needed. They need an even better reason to implant Syndicate tech into themselves. Still, a risk remains: You now have a spent, DNA-laden implanter instead of a key to worry about, and you can easily be compromised if that is found. And, of course, a caught Syndie can still rat you out too if they think it will lessen their sentence.

Ultimately this will allow for some fun opportunities that didn't exist before, like an imprisoned Syndie coordinating with another Syndie outside to organize a break-out, disclosing a stash, etc.

Nukies still use headsets with Syndicate encryption keys. This rewards the crew with a key to use if they can capture a nukie alive (or a dead one without an acidifier implant).

### Generic radio implant (4 TC)

This one felt like a logical addition to the above. As before you can use a regular headset for your stolen encryption keys, but spending 4 TC nets you the ability to keep your stolen key mostly hidden from searches, at least as long as the implanter isn't found. I think the price is probably fair since it doesn't come with a key, and you still have to deal with disposing the implanter and actually getting a key which can be risky, especially for command or sec keys. If it turns out to be too unbalanced the cost can be tweaked or it can be removed completely.

Any inserted encryption key will fall to the floor when the implant is extracted.

## Technical details

The generic radio uses the inventory system for encryption keys instead of EncryptionKeyHolder because the latter doesn't really work for implants.

More channel specific implants can be easily added if desired.

### Message duplication 

If you use encryption keys with channels that overlap with the the channels of the keys in your headset, received messages on those overlapping channels will be duplicated. 

I think realistically this is only an issue when someone uses the station master encryption key in it. Usually you would just insert a key you do not have in your headset (e.g. to listen in on sec communications as a scientist). The station master key has the Common channel included in it though, as well as department channels. If someone runs into this issue they can take off their headset or pop out its keys. ("Officer, I just took it off because comms were too noisy, my head hurts...")

This ultimately is something that needs to be fixed by Wizden because it would probably require pretty big changes to the chat system.

### Storage still accessible after extraction

If the storage window for the generic radio implant is still open after implant extraction, it can still be manipulated, though it won't let you gain radio access again unless reimplanted. 

The storage implanter also has this issue, I filed this on WizDen: https://github.com/space-wizards/space-station-14/issues/28309

## Media
image
image

## Breaking changes

None, the Syndicate key still exists but just can't be bought through the uplink anymore.

**Changelog**
:cl:
- remove: Syndicate encryption keys can no longer be purchased.
- add: Syndicate radio implants are now available in uplinks. After implanting, you can use it to speak on the Syndicate frequency like you would on a headset (":t Hi!"). No headset needed!
- add: Generic radio implants are now available in uplinks. After implanting, you can insert an encryption key into its bluespace storage space, and use it like you would use a headset.

### DIFF
--- a/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
+++ b/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
@@ -10,6 +10,7 @@ using Robust.Shared.Player;
 
 namespace Content.Server.DeltaV.Implants.Radio;
 
+/// <inheritdoc />
 public sealed class RadioImplantSystem : SharedRadioImplantSystem
 {
     [Dependency] private readonly INetManager _netManager = default!;
--- a/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
+++ b/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
@@ -47,11 +47,11 @@ public sealed class RadioImplantSystem : SharedRadioImplantSystem
             return;
 
         // does the implant have access to the channel the implantee is trying to speak on?
-        if (ent.Comp.Implant is { Valid: true }
-            && TryComp<RadioImplantComponent>(ent.Comp.Implant, out var radioImplantComponent)
+        if (ent.Comp.Implant is {} implant
+            && TryComp<RadioImplantComponent>(implant, out var radioImplantComponent)
             && radioImplantComponent.Channels.Contains(args.Channel.ID))
         {
-            _radioSystem.SendRadioMessage(ent, args.Message, args.Channel.ID, ent.Comp.Implant.Value);
+            _radioSystem.SendRadioMessage(ent, args.Message, args.Channel.ID, implant);
             // prevent other radios they might be wearing from sending the message again
             args.Channel = null;
         }
--- a/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
+++ b/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
@@ -1,0 +1,115 @@
+﻿using Content.Server.Chat.Systems;
+using Content.Server.Radio;
+using Content.Server.Radio.Components;
+using Content.Server.Radio.EntitySystems;
+using Content.Shared.DeltaV.Implants.Radio;
+using Content.Shared.Radio.Components;
+using Robust.Shared.Containers;
+using Robust.Shared.Network;
+using Robust.Shared.Player;
+
+namespace Content.Server.DeltaV.Implants.Radio;
+
+public sealed class RadioImplantSystem : SharedRadioImplantSystem
+{
+    [Dependency] private readonly INetManager _netManager = default!;
+    [Dependency] private readonly RadioSystem _radioSystem = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<RadioImplantComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<RadioImplantComponent, EntInsertedIntoContainerMessage>(OnInsertEncryptionKey);
+        SubscribeLocalEvent<RadioImplantComponent, EntRemovedFromContainerMessage>(OnRemoveEncryptionKey);
+        SubscribeLocalEvent<RadioImplantComponent, RadioReceiveEvent>(OnRadioReceive);
+        SubscribeLocalEvent<HasRadioImplantComponent, EntitySpokeEvent>(OnSpeak);
+    }
+
+    /// <summary>
+    /// Ensures implants with fixed channels work.
+    /// </summary>
+    private void OnMapInit(EntityUid uid, RadioImplantComponent component, MapInitEvent args)
+    {
+        UpdateRadioReception(uid, component);
+    }
+
+    /// <summary>
+    /// Handles the implantee's speech being forwarded onto the radio channel of the implant.
+    /// </summary>
+    private void OnSpeak(EntityUid uid, HasRadioImplantComponent hasRadioImplantComponent, EntitySpokeEvent args)
+    {
+        // not a radio message, or already handled by another radio
+        if (args.Channel is null)
+            return;
+
+        // does the implant have access to the channel the implantee is trying to speak on?
+        if (hasRadioImplantComponent.Implant is { Valid: true }
+            && TryComp<RadioImplantComponent>(hasRadioImplantComponent.Implant, out var radioImplantComponent)
+            && radioImplantComponent.Channels.Contains(args.Channel.ID))
+        {
+            _radioSystem.SendRadioMessage(uid, args.Message, args.Channel.ID, hasRadioImplantComponent.Implant.Value);
+            // prevent other radios they might be wearing from sending the message again
+            args.Channel = null;
+        }
+    }
+
+    /// <summary>
+    /// Handles receiving radio messages and forwarding them to the implantee.
+    /// </summary>
+    private void OnRadioReceive(EntityUid uid, RadioImplantComponent component, ref RadioReceiveEvent args)
+    {
+        if (TryComp(component.Implantee, out ActorComponent? actorComponent))
+            _netManager.ServerSendMessage(args.ChatMsg, actorComponent.PlayerSession.Channel);
+    }
+
+    /// <summary>
+    /// Handles the addition of an encryption key to the implant's storage.
+    /// </summary>
+    private void OnInsertEncryptionKey(EntityUid uid, RadioImplantComponent component, EntInsertedIntoContainerMessage args)
+    {
+        // check if the insertion is actually something getting inserted into the radio implant storage, since
+        // this evt also fires when the radio implant is being inserted into a person.
+        if (uid != args.Container.Owner
+            || !TryComp<EncryptionKeyComponent>(args.Entity, out var encryptionKeyComponent))
+            return;
+
+        // copy over the radio channels that can be accessed
+        component.Channels.Clear();
+        component.Channels.UnionWith(encryptionKeyComponent.Channels);
+        Dirty(uid, component);
+        UpdateRadioReception(uid, component);
+    }
+
+    /// <summary>
+    /// Handles the removal of an encryption key from the implant's storage.
+    /// </summary>
+    private void OnRemoveEncryptionKey(EntityUid uid, RadioImplantComponent component, EntRemovedFromContainerMessage args)
+    {
+        // check if the insertion is actually something getting inserted into the radio implant storage, since
+        // this evt also fires when the radio implant is being inserted into a person.
+        if (uid != args.Container.Owner
+            || !HasComp<EncryptionKeyComponent>(args.Entity))
+            return;
+
+        // clear the radio channels since there's no encryption key inserted anymore.
+        component.Channels.Clear();
+        Dirty(uid, component);
+        UpdateRadioReception(uid, component);
+    }
+
+    /// <summary>
+    /// Ensures that this thing can actually hear radio messages from channels the key provides.
+    /// </summary>
+    private void UpdateRadioReception(EntityUid uid, RadioImplantComponent component)
+    {
+        if (component.Channels.Count != 0)
+        {
+            // we need to add this comp to actually receive radio events.
+            EnsureComp<ActiveRadioComponent>(uid).Channels = new(component.Channels);
+        }
+        else
+        {
+            RemComp<ActiveRadioComponent>(uid);
+        }
+    }
+}
--- a/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
+++ b/Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
@@ -16,6 +16,8 @@ public sealed class RadioImplantSystem : SharedRadioImplantSystem
     [Dependency] private readonly INetManager _netManager = default!;
     [Dependency] private readonly RadioSystem _radioSystem = default!;
 
+    private EntityQuery<ActorComponent> _actorQuery;
+
     public override void Initialize()
     {
         base.Initialize();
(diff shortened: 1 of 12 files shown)


Lyndomen:
Hello - headmins approved.

Resources/Prototypes/Catalog/uplink_catalog.yml
```diff
@@ -632,16 +632,17 @@
...
-  id: UplinkHeadsetEncryptionKey
-  name: uplink-encryption-key-name
-  description: uplink-encryption-key-desc
-  icon: { sprite: /Textures/Objects/Devices/encryption_keys.rsi, state: synd_label }
-  productEntity: BoxEncryptionKeySyndie # Two for the price of one
-  cost:
-    Telecrystal: 2
-  categories:
-  - UplinkDeception
+# Delta-V: replaced by syndicate radio implant
```
DangerRevolution, line 635: why not uh, drop the cost on this and keep this as an option? You can risk it all for a cheap option, or pay a premium for a hidden option?

Lyndomen: One person going for the cheap option obviates away the choice of everyone else. That's why the bulk purchase is still an option- you can get two for 3TC, which is only 1TC more then the original pair of keys.

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
```diff
@@ -0,0 +1,116 @@
...
+        SubscribeLocalEvent<RadioImplantComponent, EntInsertedIntoContainerMessage>(OnInsertEncryptionKey);
+        SubscribeLocalEvent<RadioImplantComponent, EntRemovedFromContainerMessage>(OnRemoveEncryptionKey);
+        SubscribeLocalEvent<RadioImplantComponent, RadioReceiveEvent>(OnRadioReceive);
+        SubscribeLocalEvent<HasRadioImplantComponent, EntitySpokeEvent>(OnSpeak);
+    }
+
+    /// <summary>
+    /// Ensures implants with fixed channels work.
+    /// </summary>
+    private void OnMapInit(EntityUid uid, RadioImplantComponent component, MapInitEvent args)
```
deltanedas: use `Entity<T>` and ref args for all these handlers

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
```diff
@@ -0,0 +1,116 @@
...
+            args.Channel = null;
+        }
+    }
+
+    /// <summary>
+    /// Handles receiving radio messages and forwarding them to the implantee.
+    /// </summary>
+    private void OnRadioReceive(EntityUid uid, RadioImplantComponent component, ref RadioReceiveEvent args)
+    {
+        if (TryComp(component.Implantee, out ActorComponent? actorComponent))
```
deltanedas: get an EntityQuery for this in initialize since this is run a lot
then _actorQuery.TryComp(ent.Comp.Implantee, out var actor))

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// <summary>
+    /// Ensures that this thing can actually hear radio messages from channels the key provides.
+    /// </summary>
+    private void UpdateRadioReception(EntityUid uid, RadioImplantComponent component)
+    {
+        if (component.Channels.Count != 0)
+        {
+            // we need to add this comp to actually receive radio events.
+            EnsureComp<ActiveRadioComponent>(uid).Channels = new(component.Channels);
```
deltanedas: borgs can use implants so this shouldnt bulldoze old channels

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// The entity this implant got added to.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? Implantee { get; set; }
+
+    /// <summary>
+    /// The channels this implant can talk on.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public HashSet<string> Channels { get; set; } = new();
```
deltanedas: use ProtoId not string and remove the {get; set; }

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
```diff
@@ -0,0 +1,22 @@
...
+/// This is for radio implants. Might be Syndie, might not be Syndie, but either way, it's an implant.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedRadioImplantSystem))]
+public sealed partial class RadioImplantComponent : Component
+{
+    /// <summary>
+    /// The entity this implant got added to.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? Implantee { get; set; }
```
deltanedas: remove get set

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+public abstract class SharedRadioImplantSystem : EntitySystem
+{
+    [Dependency] private readonly SharedStorageSystem _storage = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<RadioImplantComponent, ImplantImplantedEvent>(OnImplanted);
+        SubscribeLocalEvent<RadioImplantComponent, EntGotRemovedFromContainerMessage>(OnPossiblyUnimplanted);
+        SubscribeLocalEvent<StorageComponent, OpenRadioImplantEvent>(OnOpenStorage);
```
deltanedas: this should really just reuse storage implant code for this

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        _storage.OpenStorageUI(uid, args.Performer, component);
+        args.Handled = true;
+    }
+}
+
+/// <summary>
+/// Triggered when someone clicks the open radio implant action.
+/// </summary>
+public sealed partial class OpenRadioImplantEvent : InstantActionEvent
```
deltanedas: use storage implant action for this and the storage implant component

Resources/Prototypes/DeltaV/Actions/types.yml
```diff
@@ -0,0 +1,13 @@
+﻿- type: entity
+  id: ActionOpenRadioImplant
```
deltanedas, line 2: if made to reuse storage implant this action should inherit from it and just change the user facing fields

Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
```diff
@@ -15,3 +15,42 @@
       tags:
         - SubdermalImplant
         - BionicSyrinxImplant
+
+- type: entity
+  parent: BaseSubdermalImplant
```
deltanedas: inherit storage implanter

Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
```diff
@@ -15,3 +15,42 @@
...
+      storagebase: !type:Container
+        ents: [ ]
+  - type: UserInterface
+    interfaces:
+      enum.StorageUiKey.Key:
+        type: StorageBoundUserInterface
+  - type: RadioImplant
+
+- type: entity
+  parent: BaseSubdermalImplant
```
deltanedas, line 43: inherit storage implant

deltanedas (changes requested):

Resources/Prototypes/Catalog/uplink_catalog.yml
pissdemon: Because it would defy the point of making it an implant. It was discussed in general with Lyndomen. You weren't there for it, to be fair, so to fill you in I'm making this an implant is specifically so Syndicates can worry less about sec confiscating and compromising their comms, encouraging more Syndicate collaboration instead of everyone cooking their own meal.

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
pissdemon: Sure, but is that a code style change I missed out on or something?

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
pissdemon: The ActiveRadioComponent is on the radio implant itself, so it's not smashing anything. Unless you mean something else?

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
deltanedas: yeah a while ago, all new code should use it

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
pissdemon: encryptionKeyComp.Channels is a HashSet<string> which is why I chose that but since it's also used as a datafield for the syndie implant I'll do that

deltanedas: HashSet doesnt have AddRange anyway i think so it wouldnt matter what you use to add to it

Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
pissdemon: The syndie radio implant can't and won't because it's not supposed to have storage, by design

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
pissdemon: There is no storage implant component. However, I still removed OpenRadioImplantEvent and replaced it with OpenStorageImplantEvent since the code was indeed redundant.

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
pissdemon: [Removed this](https://github.com/DeltaV-Station/Delta-v/pull/1262/files/f95ba22f20175a3108e246c3483e5ddff4408286#r1616124068)

Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
deltanedas: it has storage thats whitelisted sounds like storage to me

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
pissdemon: I used UnionWith(), so...

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
pissdemon: Done

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
pissdemon: Done

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
pissdemon: Got it

Resources/Prototypes/DeltaV/Actions/types.yml
pissdemon: This will be kept but I swapped out OpenRadioImplantEvent with OpenStorageImplant. It still needs the changed icon, so I still need to apply an InstantAction comp to it which overrides the original, so there's not really anything left to inherit, unless wizden were to, for some reason, add more stuff to the action or something

deltanedas: you inherit the action event type and just set the differences
and would make it easier with action ecs if it already inherits BaseAction less work to do

Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
pissdemon: Changed to inherit it though I initially avoided it because of fear of breakage if the storage implant def changes, since this isn't a storage implant, it just happens to act similar to one

Content.Shared/DeltaV/Implants/Radio/RadioImplantComponent.cs
deltanedas: ok well upstream shitcode can be changed later

Content.Server/DeltaV/Implants/Radio/RadioImplantSystem.cs
```diff
@@ -24,31 +26,32 @@ public override void Initialize()
...
+    private void OnSpeak(Entity<HasRadioImplantComponent> ent, ref EntitySpokeEvent args)
     {
         // not a radio message, or already handled by another radio
         if (args.Channel is null)
             return;
 
         // does the implant have access to the channel the implantee is trying to speak on?
-        if (hasRadioImplantComponent.Implant is { Valid: true }
-            && TryComp<RadioImplantComponent>(hasRadioImplantComponent.Implant, out var radioImplantComponent)
+        if (ent.Comp.Implant is { Valid: true }
```
deltanedas: Valid: true isnt needed you can do `if (ent.Comp.Implant is {} implant && whatever(implant)`

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
```diff
@@ -24,7 +24,7 @@ public override void Initialize()
     /// <summary>
     /// Handles implantation of the implant.
     /// </summary>
-    private void OnImplanted(EntityUid uid, RadioImplantComponent component, ImplantImplantedEvent args)
+    protected virtual void OnImplanted(EntityUid uid, RadioImplantComponent component, ImplantImplantedEvent args)
```
deltanedas: why, nothing overrides it

deltanedas (changes requested):

Content.Shared/DeltaV/Implants/Radio/SharedRadioImplantSystem.cs
pissdemon: remnant from previous version, thanks for pointing it out

deltanedas (approved):

PR #1268, opened 2024-05-29
Micro quick fix

Colin-Tel:
## About the PR

Added physics dampening, added logi autolathe, added salvage rollerbed in case they get ouchies >w<

Also removed a little more than half of the placed rad collectors because of the RTG strategy and how good it is with 14 rad collectors on standby.

**Changelog**

too smol to cl


Colin-Tel:
image
new sec front !!1

IamVelcroboy (approved):

PR #1274, opened 2024-05-30
Update micro.yml

Colin-Tel:
## About the PR

Added southeast solar array, making it easier to obtain power, it's even got another hardsuit!

Changed the evac shuttle to the default, as it is more likely to dock at arrivals as an alternate evac dock (also altered layout of arrivals to facilitate this)

## Why / Balance

Power on Micro is a little tricky for players, and adding another solar array helps them while also giving something to build off of.

If Evac blows up, arrivals now actually functions as an alternate evac.

## Media

Micro-0
image
image
image
image

**Changelog**

:cl:
- tweak: Micro now has a second solar array in the southeast of the station, and its arrivals should function as an alternate evac.


Colin-Tel:
ok for real that's all the updates for now lol

unless...

IamVelcroboy (approved):

PR #1280, opened 2024-06-01
Quick Micro changes

Colin-Tel:
## About the PR

Upped max pop to 45
Added decals around cryo, added tiny fans for evac, gave mime a microphone (lol)

## Why / Balance

~54 job slots available, I may have lowballed the maxpop earlier 🙃 

**Changelog**

:cl:
- tweak: Micro can appear in lobbies up to 45 in population


Mnemotechnician:
maybe even 45 is underkill? We played it on something like 70 or 80 pop and it was quite nice.

Colin-Tel:
> maybe even 45 is underkill? We played it on something like 70 or 80 pop and it was quite nice.

It's tricky, because the map is so smol and I love doing disproportionate highpop events for maps like it, but it honestly is just too small for higher populations than 50, in my current opinion.

IamVelcroboy (approved):

PR #1287, opened 2024-06-02
Security Access Adjustments

Colin-Tel:
## About the PR

sec can no longer whack service workers in their job areas without breaking down the wall
also detective gains cryogenic access while cadets and secoffs lose it

## Why / Balance

There isn't much of a reason for sec to be able to barge into service-locked rooms willy nilly. If there's a reason, they can fetch someone with the access (i.e. command) or get assigned the access with an ID computer.

Detective should have cryogenics access to obtain things stored in cryosleep beds, a task that security cadets are already trusted with. Regular officers and cadets shouldn't be encouraged to sift through cryosleepers.

**Changelog**

:cl:
- remove: Security personnel no longer have service access by default.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Security/detective.yml
+++ b/Resources/Prototypes/Roles/Jobs/Security/detective.yml
@@ -17,6 +17,7 @@
   - Maintenance
   #- Service # DeltaV
   - Detective
+  - Cryogenics
   special:
   - !type:AddImplantSpecial
     implants: [ MindShieldImplant ]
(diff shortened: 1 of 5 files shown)


Resources/Prototypes/Roles/Jobs/Security/detective.yml
```diff
@@ -15,8 +15,9 @@
   - Security
   #- Brig # Delta V: Removed
   - Maintenance
-  - Service
+  #- Service # Delta V: Access adjustments 02June2024
```
Axionyxx: These don't need to be dated, it looks weird and doesn't match anything else in the codebase

Axionyxx (changes requested):

Resources/Prototypes/Roles/Jobs/Security/detective.yml
Colin-Tel: wilco

therealDLondon:
Headmin approved btw

Axionyxx (approved):

PR #1288, opened 2024-06-02
Make flatpacks mappable

IamVelcroboy:
## About the PR
Removes hideSpawnMenu from flatpacks so they can be mapped

## Why / Balance
Mainly for engine setups when we don't want to map an entire crate but simply add a piece or two. Could also be used in salvage wrecks down the line.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/flatpack.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/flatpack.yml
@@ -129,8 +129,8 @@
   id: TeslaGeneratorFlatpack
   name: tesla generator flatpack
   description: A flatpack used for constructing a tesla generator.
-  #categories:  #Delta V - No Tesla fo DV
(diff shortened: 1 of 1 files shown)


Axionyxx (changes requested):
A base item should never be mappable. You should override categories on the entities you want to map, or create crates for them.

Axionyxx:
You can also just use `spawn PrototypeId` in the console to spawn any item regardless of nospawn or hideSpawnMenu

IamVelcroboy:
> A base item should never be mappable. You should override categories on the entities you want to map, or create crates for them.

Should this just be abstracted then? Is there any case/use for a generic flatpack?

deltanedas:
> Is there any case/use for a generic flatpack?

this machine called flatpacker 9001 :trollface:

Colin-Tel:
Gonna close this for the time being.

PR #1296, opened 2024-06-03
Rebalance XenoVents event

Colin-Tel:
## About the PR

Pushes the earliest time of the xenovent event to 45 minutes into the round and requiring at least 20 crew.
Increases the spawn rate of xenos to make the event more exciting.

## Why / Balance

The probability for xeno spawns was broken a ways back, and the numbers we had for a while were a band-aid solution that never got adjusted. This meant that the event, combined with the xenos being kind of pushovers, resulted in a waste of time.

*Now*, the event will have more xenos spawn, and ideally it will cause an actual amount of havoc on the station as the crew repel them.

**Changelog**

:cl:
- tweak: The vents contain more scary xenos!


Lyndomen:
Approved by one headmin!

Axionyxx (approved):

PR #1302, opened 2024-06-04
Adjust Hunger rates

Colin-Tel:
## About the PR

Raised hunger rates for all species. Diona had their thirst rate raised to 1.5x the human thirst rate and their hunger is 0.4x the human hunger rate.

## Why / Balance

The current hunger rate has relegated chef and botany to not really being consequential in the rounds. Raising the hunger rate can aid against this problem if done carefully.

## Technical details

Adjusts the `baseDecayRate` in HungerComponent.

In a nutshell, hunger is reduced by .05 every second instead of by .001666666.

**Changelog**

:cl:
- tweak: The rate at which non-Diona species become hungry has slightly increased, while Diona get thirsty more often.


deltanedas:
vending machines should be more barren, both giving logi a reason to buy restocks and chef more than 0 customers

Colin-Tel:
> vending machines should be more barren, both giving logi a reason to buy restocks and chef more than 0 customers

Wouldn't it be cool if food&drink vending machines had inventory variation on round start so some will be as full as can be and some could be almost out?

deltanedas:
they can its just usually 1 type is empty thats it, vending machines are still the go to

Axionyxx (dismissed):
Instead of changing all these values on the prototype, future proof it by changing the value within HungerComponent.

Bonktrauma:
Resident diona main approves

Colin-Tel:
This one should be ready.

Resources/Prototypes/Entities/Mobs/Species/diona.yml
```diff
@@ -10,7 +10,7 @@
   - type: Hunger
     baseDecayRate: 0.0083
```
deltanedas: this should probably be increased a bit so its ~half instead of 1/6 which is a huge disparity

Colin-Tel: Sure thing.

deltanedas (approved):

deltanedas:
should probably be fine on master but if it fails fix :trollface:

PR #1317, opened 2024-06-07
Prescription hud fix

Timemaster99:
## About the PR
Fixed prescription HUD crafting, since I'm someone who uses it...

## Technical details
Made the construction graph check for VisionCorrection component instead of GlassesNearsight tag

## Media

## Breaking changes

**Changelog**
:cl: Timemaster99
- fix: Fixed prescription hud crafting


Axionyxx (changes requested):
No, this is a bad fix. The construction graph should be edited instead to reference something else unique about these sorts of glasses, like the vision correction component

Timemaster99:
> No, this is a bad fix. The construction graph should be edited instead to reference something else unique about these sorts of glasses, like the vision correction component

Resolved. I didn't know construction graphs could check for components

Axionyxx (approved):

PR #1383, opened 2024-06-14
add glimmer wisp

deltanedas:
## About the PR
- brought back glimmer wisp from legacy
- made its code a hell of a lot better than before and more reusable
- glimmer mite count doubles at 300(?) glimmer instead of 500(?) which wisp uses
  should give epi more of a threat than "oh a single mite cool"
  it should also have MobsPerPsionic looked at later on now that its not hardcoded, needing **20** psionics for 2 base mites is insane
- wisp has more health and takes less holy damage so you cant **2 shot** it with a bible
  it now takes 3 dispels to kill instead of 2 or 1 and a bible smack
- wisp drops 2-3 ectoplasm since you cant kill it in a second now

also added a Glimmer Creatures guidebook under Epistemics

## Why / Balance
gaming

## Technical details
instead of GlimmerWisp component its LifeDrainer and it can have everything set in yml now including whitelist for mobs so its not hardcoded to psionics only

the using zap power in combat code doesnt work idk why

## Media
guidebook entry in Glimmer Creatures

## Breaking changes
no

**Changelog**
:cl:
- add: Psionic-hunting Wisps can now appear when glimmer gets too high.


Axionyxx (commented):
Code looks good so far

Axionyxx (approved):
We can find other issues live

PR #1393, opened 2024-06-16
Pill bottles whitelist

BlitzTheSquishy:
## About the PR

Added a whitelist to Pill Cannisters to reduce their ability to be used for powergaming.

This PR is to attempt to push this change early (if at all) to Delta-V before upstream does

## Why / Balance

Due to the lack of a whitelist on pill cannisters it is possible to put 300u of reagents inside a single pill canister via the use of bottles, as well as effectively doubling that again with the use of combat medkits (though with those it's intentional). 

## Media

## Breaking changes

**Changelog**


Colin-Tel:
image

This won't affect my cigarette packet meta, will it?

IamVelcroboy:
This was rejected upstream. (And I sort of agree with the argument.) Ideally we avoid whitelist as much as possible. Another solution would be to increase the size of bottles to small. Which should probably be done upstream if it's going to be done.

deltanedas:
physically it makes no sense that you can fit 10 glass bottles in a ciggie pack

cdda inventory wyci

BlitzTheSquishy:
> image
> 
> This won't affect my cigarette packet meta, will it?

way less likely to happen but consider killing that too

BlitzTheSquishy:
> This was rejected upstream. (And I sort of agree with the argument.) Ideally we avoid whitelist as much as possible. Another solution would be to increase the size of bottles to small. Which should probably be done upstream if it's going to be done.

I mean *pill* canister

Also I'm not entirely sure if the tiles they take up is also tied to their "size" but theres a point where beakers or even shakers is preferable to bottles,

I personally dont understand the argument of *not* whitelisting them, without proposing another solution. Cause you can put LITERALLY any "tiny" item in a pill canister, which would be changing the size of a lot of different items (bullets, bottles, injectors, lighters)

IamVelcroboy:
> I personally dont understand the argument of _not_ whitelisting them, without proposing another solution. Cause you can put LITERALLY any "tiny" item in a pill canister, which would be changing the size of a lot of different items (bullets, bottles, injectors, lighters)

Bottles and injectors should not fit in a pill bottle. But like... 
20240619_110458_HDR

IamVelcroboy:
> I personally dont understand the argument of _not_ whitelisting them, without proposing another solution. Cause you can put LITERALLY any "tiny" item in a pill canister, which would be changing the size of a lot of different items (bullets, bottles, injectors, lighters)

Bottles and injectors should not fit in a pill bottle. But like...

BlitzTheSquishy:
> > I personally dont understand the argument of _not_ whitelisting them, without proposing another solution. Cause you can put LITERALLY any "tiny" item in a pill canister, which would be changing the size of a lot of different items (bullets, bottles, injectors, lighters)
> 
> Bottles and injectors should not fit in a pill bottle. But like... 

considering the fact pil lcans can only fit ***TEN*** pills, youre using the wrong size of pill canister

deltanedas:
literally just code cdda inventory in ss14...

pill canister holds 250ml of stuff and has max length of like 8 cm

IamVelcroboy:
> considering the fact pil lcans can only fit _**TEN**_ pills, youre using the wrong size of pill canister

I don't have a smaller one available but I'm certain both the lighter and bullets would still fit

IamVelcroboy:
Point is, whitelist is restrictive. A container is a container regardless of its intended function. I could put batteries, or bottle caps, or toothpicks, etc in a pill bottle. This does have gameplay value as players may be able to hide or stash various objects in something that wouldn't be the normal place to look for it. 

Is the size/inventory around a lot of objects currently abusable, yes. But I agree with upstream that whitelist isn't the best solution. Would be better to adjust the size of items that are currently formatted too small.

deltanedas:
pill bottle is 1x1 same as a pill so it should be made larger (and have its storage doubled to match) because how can a pill bottle hold 10 of itself

Axionyxx (approved):

Axionyxx:
There are better ways to do this yes, but I believe people keep abusing this so whatever

PR #1402, opened 2024-06-17
buff tank resistances

deltanedas:
## About the PR
remi and other common shitter/raider mobs can no longer break welding tanks, they need weapons now


Colin-Tel (approved):
The Remi community is *in shambles*

PR #1412, opened 2024-06-18
Every syndicate agent can get a free radio implant

WarMechanic:
## About the PR

Syndicate agents now get a singular radio implanter for free from the uplink rather than having to spend 2tc.

## Why / Balance

I was bribed
image
noctis approved by the way

## Technical details

YAML warrior

## Media

pending
image

bought
image

## Breaking changes

no

**Changelog**

:cl:
- tweak: Syndicate agents can get one radio implanter for free.


noctyrnal:
"noctis approved" means nothing sorgy :(

WarMechanic:
this pr is not for me :trollface:

Colin-Tel:
When this was talked about in admin channels, it was under the impression it would cost 1 less TC (aka cost 1 TC), not just free.

I disagree with the premise of it being free.

therealDLondon:
This was discussed briefly among admins and the agreement is that this is not a balance change we would want, unfortunately, it removes too much guesswork from syndicate agents and invalidates codewords. Sorry

Lyndomen:
I'll just chime in with more of an annecodatal idea: it might be cool to investigate this more broadly, perhaps by having various items on discount/free randomly per round, mirroring SS13's discount systems.

Issue #1346, opened 2024-06-10
Arena station Crew monitors display "No server"


Colin-Tel:
## Description

Crew monitors (both computers and handheld) display the map, but also display "No server" on the crew list. This is fixed by unwrenching, then re-wrenching the crew monitoring console.

**Reproduction**

1. Launch a round on Arena
2. Attempt to view any crew monitor while on the station.

**Screenshots**

I should have grabbed one while it was messed up T^T

**Additional context**

Arena station is the only station that is missing a `type` for the crew monitoring server in the map .yml, it's probably that.

![image](https://github.com/DeltaV-Station/Delta-v/assets/113523727/214c2660-0eaa-4d97-8f9d-98d7d1628c10)

PR #1414, opened 2024-06-18
Map Hotfix: The Hive

IamVelcroboy:
## About the PR
The Hive
- Adds CJ locker
- Adds singleton comp to crew monitor because you apparently have to do that manually now.
- Adds justice encryption key to telecomm server

### DIFF
--- a/Resources/Maps/hive.yml
+++ b/Resources/Maps/hive.yml
@@ -63741,6 +63741,9 @@ entities:
     - type: Transform
       pos: 144.5,4.5
       parent: 1
+    - type: SingletonDeviceNetServer
+      active: False
+      available: False
 - proto: Crowbar
   entities:
   - uid: 2132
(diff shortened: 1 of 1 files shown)


Axionyxx (dismissed):

Resources/Maps/hive.yml
```diff
@@ -63741,6 +63741,9 @@ entities:
     - type: Transform
       pos: 144.5,4.5
       parent: 1
+    - type: SingletonDeviceNetServer
+      active: False
```
deltanedas, line 63745: how does disabling it fix it not working what

IamVelcroboy: > how does disabling it fix it not working what

No idea. But all the other maps have that comp, including upstream. But for some reason when you map a new one, the comp isn't being added and it doesn't work.

IamVelcroboy: See also: [#1346](https://github.com/DeltaV-Station/Delta-v/issues/1346)

deltanedas (approved):

PR #1416, opened 2024-06-20
Pebble Update June 20th

therealDLondon:
## About the PR
- Removed Cloning
- Adds Cryogenics (With adjusted pump because the gate = a pump)
- Moves left evac pod and psych office more left
- Adjusts north evac hallway
- Adjusts medbay
## Why / Balance
Cloning 1984

**Changelog**

:cl: DLondon
- tweak: Adjusts pebble's medical

### DIFF
--- a/Resources/Maps/pebble.yml
+++ b/Resources/Maps/pebble.yml
@@ -38,6 +38,7 @@ tilemap:
   82: FloorRockVault
   83: FloorShowroom
   94: FloorSteel
+  4: FloorSteelBurnt
   2: FloorSteelCheckerDark
   98: FloorSteelDamaged
   99: FloorSteelDiagonal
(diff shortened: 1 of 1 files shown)


UnicornOnLSD:
not sure about the passive gate instead of regular gas pump, will it let a certain amount of pressure in and not out ? There was a volumetric pump back on tortuga once and it just overpressurised the system so y'know, making sure

UnicornOnLSD:
also I still see a plasma miner I MIGHT have forgotten while taking care of the map myself, so yep

deltanedas:
if someone sets distro to 600kpa or something it would overpressurize cryo, use a pressure pump
also ideally the filters would be reworked to be inline so they dont mess up pressure too but yeah

therealDLondon:
Ill be real I made a buncha silly errors ill fix this all properly another time

PR #1218, opened 2024-05-18
Senior ID cards get a custom job name

Axionyxx:
## About the PR
Makes it actually worth taking, as it isn't just a resprite in the loadout

**Changelog**
:cl:
- tweak: Senior PDA's now have an ID card with a custom job title

PR #1418, opened 2024-06-20
Senior ID cards get a custom job name: Attempt 2

WarMechanic:
## About the PR

So last time senior ID job titles tried to get implemented, it practically didn't work. I personally never saw senior IDs in effect, and they only applied to players at round-start without even influencing the crew manifest.
I'm not quite done yet, but right now my solution catches both roundstart and latejoining players. It also applies the correct ID icons, so security will finally see some fancy "M"s walking around. Senior ID job name and icon also display correctly in crew manifest now.

## Technical details

This solution reverts #1218 which relied on an event that was raised only after ready players spawned in at roundstart. Instead of a customJobName variable, we use job prototypes for senior roles and add logic before a player joins to use a 'virtualJob' based on your ID card. Senior ID cards now link to senior role prototypes instead of normal roles.

## Media

## Breaking changes

Not from what I've seen so far, although anything that relies on the redundant ID name change on spawn will have to be checked.

**Changelog**

:cl:
- fix: Fixed senior roles not displaying properly

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -851,6 +851,9 @@ namespace Content.Client.Lobby.UI
 
                 foreach (var job in jobs)
                 {
+                    if (job.EditorHidden)
+                        continue;
+
                     var jobContainer = new BoxContainer()
                     {
                         Orientation = LayoutOrientation.Horizontal,
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -851,6 +851,7 @@ namespace Content.Client.Lobby.UI
 
                 foreach (var job in jobs)
                 {
+                    // DeltaV #1418 - Allow hiding jobs to hide senior virtual jobs
                     if (job.EditorHidden)
                         continue;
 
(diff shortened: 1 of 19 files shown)


WarMechanic:
its in a working state now but the crew manifest ended up being bugged because senior job prototypes werent listed in their respective departments :blunt:

ill have to review the logic to cull anything thats not needed.

Axionyxx (changes requested):
This is adding additional data that was the entire reason we avoided doing this in the first place. It should be investigated why the code was faulty and fixed, not removed and replaced by the one thing that was being avoided

WarMechanic:
> This is adding additional data that was the entire reason we avoided doing this in the first place. It should be investigated why the code was faulty and fixed, not removed and replaced by the one thing that was being avoided

is it including unnecessary data in prototypes, or the fact that there are prototypes in the first place that you want to avoid? 

i added things like access to the senior job prototypes just as a formality although they will probably never be used so i can cull them if you want. i did initially try to parent the prototypes but that didn't work.

WarMechanic:
also i originally attempted to work from your fix, but i determined that id have to put bandaids in multiple different scripts and stuff. i didnt think the solution would be very readable and would be more annoying to maintain by extension

here most of the code takes place in one script, and it uses job prototypes for compatibility.

WarMechanic:
alright im going to assume the worst-case scenario in that you want me to remove the prototypes and replace them with components, ill start working on that now

WarMechanic:
actually ill upload it as a separate PR in the case that this solution would be preferred

PR #1468, opened 2024-07-07
remove research access from canisters

deltanedas:
## About the PR
title

## Why / Balance
- so ce can actually change the access of "his" departments canisters and not the captain
- every map has a port for gas input in epi
- there is never a reason for epi to open the valve of plasma cans
- make it so atmos locking cans is actually important because epi cant just immediately unlock it after getting it

**Changelog**
:cl:
- tweak: Epistemics can no longer unlock canisters.


Colin-Tel (approved):
Sounds good to me.

ps3moira:
isn't there artifacts that need plasma tanks to activate though?

ps3moira:
and also the basic gas mixing chambers in research departments

deltanedas:
cans dont need to be unlocked or open for ports to work

PR #823, opened 2024-02-12
Home run bat

musicmanvr:
## About the PR

Adds the Syndicate Home Run Bat, a highly unusual weapon designed only the craziest of operatives.

On hit, it launches a target like a gorilla gaunlet, but at triple the speed, doing 30 brute while also doing heavy structural damage.

## Why / Balance

Syndicate melee weaponry in the high TC territory has always been a problem, so I figured I'd throw my hat into the ring and give my own take on a high TC melee weapon a shot.

## Technical details

It's a new yaml file with some new sprites, that's all.

## Media

- [ X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showca

https://github.com/DeltaV-Station/Delta-v/assets/54725557/fcfb117f-fe57-487d-ab3c-7573ad8b18b1

## Breaking changes

Simple Yaml, happy times.

**Changelog**

:cl:
- add: Added the Home Run Bat for Syndicate operatives. Send your foes flying!

PR #1476, opened 2024-07-09
Homerun bat 2

deltanedas:
## About the PR
revival of #823 with small balance changes:
- nerfed its structural damage from 15 to 10
- made 16 TC since its not as op and so more roles can buy it not just passenger

also just inherits baseball bat now to not copy paste *everything*

## Why / Balance
it was too op for breaking down doors and shit now its more a funny spaceman launcher

## Media

## Breaking changes
no

**Changelog**
:cl: musicmanvr
- add: Added the Home Run Bat for Syndicate operatives. Send your foes flying!


dvir001:
Why do I feel this should have the ``G.O.R.I.L.L.A. gauntlet`` unanchorOnHit on it?

deltanedas:
oh real that was never added

ghost (approved):
From my look at this, this seems normal, and fine to merge. Didn't see any mistakes, everything's in DeltaV folders, all that good stuff. One day I should actually PR a baseball so the existence of baseball bats makes sense though that's a note to self 

Dvir poses a good point about that component, I'd love to fling Butlertron across the room with this but what's already here is alright.

deltanedas:
> baseball

already exists upstream just nothing uses it + you cant hit it with bat

deltanedas:
also the reason it cant unanchor on hit is you could use it to troll glimmer probers at 500

ghost:
> > baseball
> 
> already exists upstream just nothing uses it + you cant hit it with bat

Messed up.

dvir001:
> also the reason it cant unanchor on hit is you could use it to troll glimmer probers at 500

https://github.com/new-frontiers-14/frontier-station-14/pull/1368/files

Could consider adding a Whitelist / Blacklist on the ``unanchorOnHit`` so it can be added without too many issues.

musicmanvr:
this pleases me

PR #1481, opened 2024-07-11
Map Edits: Tortuga

### LINKED ISSUE #1475: tortuga issues
## Description
- entire teg setup is dogwater and hot loop doesnt even function because of how anchoring works
- there is no gas pump for limiting distro pressure
- salv dock should be flipped, if you dock something it basically blocks cargo shuttle from docking

- possibly overlapping lights, maybe someone built it midround

IamVelcroboy:
## About the PR
Tortuga
- Reconfigures TEG  a lil bit
- Moves plasma windows so port into burn chamber can function
- Fixes missing pump on distro
- Extends the storage/dock at salvage to add docking airlocks on the other side

## Why / Balance
Fixes [#1475](https://github.com/DeltaV-Station/Delta-v/issues/1475)


deltanedas (approved):
teg still so bad but at least it can in theory work

PR #1486, opened 2024-07-11
Re-enable "Glorious Death" objective

Colin-Tel:
## About the PR

Re-enables the die objective.

## Why / Balance

After admin discussion, to be more in-line with the rules pertaining to self-antag, we are willing to make this change to allow antagonists more freedom.

## Technical details

N/A

## Media

N/A

## Breaking changes

N/A

**Changelog**

:cl: Colin_Tel, DLondon
- add: Some traitors have seem to have nothing else to lose.


Colin-Tel:
Actually, it might be better to just revert the original PR lol

PR #1273, opened 2024-05-30
Security resprite and loadout revision

TadJohnson00:
## About the PR

Resprited detective gear as well as some other security and security-adjacent equipment, tweaked security's loadouts so they make sense. See #1197 for further details.
## Why / Balance

For consistency in uniform. Mild balance changes were made with regards to replacement of warden's and HoS' jackets with standard coats, see PR #1197.
## Technical details

n/a
## Media

Please see #1197 for further media.
330685316-d904513d-f158-48d1-9765-e99a7d6fb472

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: New security outer clothing for all security roles, as well as proper senior security officers' and cadets' jumpsuits.
- add: Jetpacks are now available in all security suit storage. Thank goodness for increased funding.
- tweak: Revamped security's clothing and equipment options; all gear is now up to code, and no further outdated equipment will be issued. This includes, most notably, the detective's uniform, which has received a whole makeover.
- fix: Fixed security hardsuits not allowing you to turn on internals without a mask.
- remove: Head of security's winter coat, warden's winter coat, security winter coat, security departmental winter boots, WizDen-pattern body armour, and some other minor changes necessary for respriting.

PR #1517, opened 2024-07-21
Revert "Security resprite and loadout revision"

Axionyxx:
Reverts DeltaV-Station/Delta-v#1273

After an admin discussion, this decision was made. This will be merged when we have a proper public statement.

:cl:
- remove: The security resprite has been reverted after admin discussion


Lyndomen:
Hello! The decision to revert this was made due to three primary factors. 

1. To be honest, we didn't fully review it before merging; and should have done so. Process improvements will be made to streamline this from a contributor standpoint. This sat for too long without proper feedback, sorry about that. 
2. There are a number of balancing adjustments that should be done, namely items should roughly match the damage reduction of their previous counterpart. We have noticed a number of items that provide almost nothing, which heavily tips things against security. 
3. Its our belief that the resprite here is pretty awesome, and I love the sprites individually. However,in a blind test, we were pretty much incapable of identifying at a glance who was in what role, and actively felt that many roles looked like the HOS. Roles should be recognizable and distinct from one another. 

I'd love to coordinate and collaborate on the next iteration of this, because it has a lot of promise!

PR #1526, opened 2024-07-23
Ports #576 from Einstein Engines : "Restore HoP as a functional job"

DangerRevolution:
## Why / Balance

Adds PDAs into HoPvend, while retaining their loss of access.

This lets HoP hire staff through the vend, without allowing them to roam freely (as the original concern in the revert).

While HoPs may be able to (with care) juggle IDs and give themselves access (although they risk wiping their ID if they misclick).  The fact that they do not innately have departmental accesses means they would be weirdly trespassing anywhere they walk in. They would be the same as a Scientist walking into Security with Security access on their ID, their only defence would be the meta-knowledge that Heads cannot be antags, which is outside the scope of this PR.

## Media

- [ X ] I have added screenshots/videos to this PR showcasing its changes ingame

image

**Changelog**

:cl: DangerRevolution, OldDanceJacket, VMSolidus
- add: Restored HoP's ability to hire departmental staff.

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/cart.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/cart.yml
@@ -3,8 +3,19 @@
   startingInventory:
     PassengerPDA: 5
     ClearPDA: 5
-# Begin Delta-V additions
-    ServiceWorkerPDA: 2
+  # Basic Department PDAs
+    SecurityPDA: 2
+    SecurityCadetPDA: 4
+    MedicalPDA: 2
+    MedicalInternPDA: 4
+    EngineerPDA: 2
+    TechnicalAssistantPDA: 4
+    SciencePDA: 2
+    ResearchAssistantPDA: 4
+    CargoPDA: 2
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Catalog/VendingMachines/Inventories/cart.yml
```diff
@@ -17,18 +28,18 @@
     ClownPDA: 1
     MimePDA: 1
     MusicianPDA: 1
-# End Delta-V additions
-    PassengerIDCard: 5
```
DangerRevolution, line 21: This is included in the PassengerPDAs at the top of this vendor

deltanedas:
> (although they risk wiping their ID if they misclick)

no, a sec id cant remove hop access etc only accesses the sec id has

this just gives hopcurity free aa and massively buffs antags raiding hops office

also this would conflict with every head locker having ids for their department, imo hop being a real job would focus more on sop nerdery of formally having sophonts join the crew with paperwork and what not, and give them more legal protection than just le having an id card

PR #1532, opened 2024-07-24
remove noobtraps from hive

deltanedas:
## About the PR
- removed the decals suggesting to have midpoints in singulo containment, this led to singuloose every round and there was no reason to have it anyway, cfgs span 7 tiles anyway
- bypass the shitty air chamber by default since noobs dont know its bad if you want to use it for something you still can
- remove completely useless mixer input from the recycler, what would you be trying to mix into the burn chamber from it exactly? instead it goes into waste so you can, in theory, replenish distro with it. of course its still impractical because its a bad setup but whatever
- add logic gates to salv airlocks so they dont space salv every round

### DIFF
--- a/Resources/Maps/hive.yml
+++ b/Resources/Maps/hive.yml
@@ -319,7 +319,7 @@ entities:
           version: 6
         -4,1:
           ind: -4,1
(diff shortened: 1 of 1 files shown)


IamVelcroboy:
First, I have to check if I have WIP on this map currently. I'm not sure off the top of my head.

1. Removing the decals is fine. 
2. Why would you bypass the chamber? It works as a buffer to spacing. If necessary it could be made larger I suppose. But it's there for a reason.
3. Mixer is for pulling CO2 in from the chamber. People will use No2 sometimes as well, which is what the port is for.
4. I was just waiting for the proper fix from upstream but 🤷‍♂️

deltanedas:
> Why would you bypass the chamber? It works as a buffer to spacing. If necessary it could be made larger I suppose. But it's there for a reason.

it does the opposite, it makes throughput so bad spacing takes hours to recover pressure from, it relies on diffusion which is very bad

it has 10000L of storage and assuming you fill it to 9mpa which realistically wont happen in any round, it will hold 36924.8 mols of air. assuming theres 1000 pipes distro would be 200kL. this entire buffer at max pressure which is never achieved will only be equivalent to 450kpa of distro pressure assuming 1000 pipes.

you can easily bring distro pressure well above that just by having mixer get pumped directly into distro. see a round a few hours ago where basically 1 spacing brought distro to 37kpa and it never recovered, leaving half the station unable to breathe.

> Mixer is for pulling CO2 in from the chamber. People will use No2 sometimes as well, which is what the port is for.

even if you bothered splitting n2o into nitrogen and oxygen, you dont want to send that few mols/s into the shitty burn chamber, it should go into waste (assuming of course you use recycler because theres no reason to). also because it used a mixer and not just a junction if either side is empty it just wont do anything. mixers are for getting a precise mix of two gases not joining two potentially empty pipenets

IamVelcroboy:
> > Mixer is for pulling CO2 in from the chamber. People will use No2 sometimes as well, which is what the port is for.
> 
> even if you bothered splitting n2o into nitrogen and oxygen, you dont want to send that few mols/s into the shitty burn chamber, it should go into waste (assuming of course you use recycler because theres no reason to). also because it used a mixer and not just a junction if either side is empty it just wont do anything. mixers are for getting a precise mix of two gases not joining two potentially empty pipenets

Maybe I'm misunderstanding which mixer is being referred to here. I thought it was the the one leading into the recycling loop.  As for the one I think you're talking about, that's fine. just send it to waste I guess. 

As for the chamber, let me do some tests on it. I feel like there's a way to make it more viable. 

I still have to look for WIP later though. Please check with map maintainers before doing PRs for them.

IamVelcroboy:
Closing for edits here [#1557](https://github.com/DeltaV-Station/Delta-v/pull/1557) See [convo ](https://discord.com/channels/968983104247185448/1267540825945014365) in Discord for why I don't believe the air chamber is the problem (but rather it was the mixer after the distro pump) 

All other edits were incorporated

PR #1541, opened 2024-07-26
Lowered the volume of the Femtanyl track

Colin-Tel:
## About the PR

Adjusted the gain by -10 dB.

## Why / Balance

It was loud.


deltanedas:
is it not possible to adjust this in yml

Colin-Tel:
> is it not possible to adjust this in yml

That's a damn good question.

ghost (approved):
Finally, Femtanyl at an acceptable volume instead of blasting it. Deltanedas poses a good point if that can be done in yaml but I'll accept the core file being lowered by like 10 decibels as a suitable fix for now.

Now only if Test Packaging would run...

ghost:
Thanks Null!!!

PR #1551, opened 2024-07-28
Update 5_RoleplayGuidelines.xml

Colin-Tel:
## About the PR

Made mention of the borg rule change in regards to "crew."

## Why / Balance

General confusion over what makes someone "crew."

**Changelog**

:cl:
- add: Added line to rule 5 in regards to what makes someone "crew" to a borg. Nukies and Zombies aren't crew, in case you were wondering.

### DIFF
--- a/Resources/ServerInfo/Guidebook/DeltaV/Rules/GameRules/5_RoleplayGuidelines.xml
+++ b/Resources/ServerInfo/Guidebook/DeltaV/Rules/GameRules/5_RoleplayGuidelines.xml
@@ -9,5 +9,5 @@
        - Some awakened pest roles (e.g. sentient mothroach) may be considered crew pets and fall under this rule.
   - [color=#ff0000]Non-security personnel should not seek to kill or detain unrelated threats.[/color] Crew may neutralize threats when they present themselves or if they are relevant to the crew member in question.
     - Certain severe emergencies may make it reasonable for normal crew to defend themselves and their station, such as nuclear operatives, a zombie outbreak, space dragon attack, or other critical situation.
-
+  - IF YOU ARE A BORG and your rules talk about "crew," know that "crew" means "person who works on the station." Someone can have an ID and NOT be crew, or they could die and be walking around harming people and NOT be crew (e.g. zombies); you can generally use common-sense to determine if someone is crew.
 </Document>


Radezolid:
I think there are two issues with this:
1- "Person who works on station" Then that means tha passengers aren't crew.
2- Cyborgs lack a way to check a manfiest, as such determining who is crew based on the definition will be hard and take time to memorize who works where and what's their look to know if they are truly crew.

In my opinion the simple "ID icon = crew" is a good (and intuitive) way to determine who is crew and who isn't since that is basically all the info the cyborg has.

Also think about this scenario:
X person wants to be borged, Y person has to gib them. Law 1 in a normal cyborg will now COMPLETELY make you prevent the borgification of X since they have to prevent crew damage at all costs. Previously it could be worked around by removing the ID out of sight of the cyborg.

Colin-Tel:
> 1- "Person who works on station" Then that means tha passengers aren't crew. 

I don’t really see the issue in that.

> 2- Cyborgs lack a way to check a manfiest, as such determining who is crew based on the definition will be hard and take time to memorize who works where and what's their look to know if they are truly crew.

Even with a manifest that’d be hard. You don’t need to memorize *everyone*, just people you interact with, usually within two or three departments.

> In my opinion the simple "ID icon = crew" is a good (and intuitive) way to determine who is crew and who isn't since that is basically all the info the cyborg has.

Borgs are smart, this just makes the determining of crew to utilize more common-sense and allow fewer instances of “nukies are crew because they have IDs”

> Also think about this scenario: X person wants to be borged, Y person has to gib them. Law 1 in a normal cyborg will now COMPLETELY make you prevent the borgification of X since they have to prevent crew damage at all costs. Previously it could be worked around by removing the ID out of sight of the cyborg.

This ultimately depends of what the Borg defines as “harm,” seeing as, so long as the crewmate’s brain remains intact and it makes it to an MMI, no meaningful harm is being experienced by the crewmate.

Also I don’t really see the problem in borgs having things they can’t really do, so it doesn’t really matter that much to me.

DangerRevolution:
> Also think about this scenario:
X person wants to be borged, Y person has to gib them. Law 1 in a normal cyborg will now COMPLETELY make you prevent the borgification of X since they have to prevent crew damage at all costs. Previously it could be worked around by removing the ID out of sight of the cyborg.

"Borg, Law 2, run four laps around the Station's Bridge. Don't come back until I tell you to"

PR #1558, opened 2024-07-30
(W.I.P) Lead, Lead Lined Crates, Lead Weaponry

Reene-Ramos:
## About the PR

Adds Lead Ore (Crabs), Lead Ingots, Lead Lined Crates, Lead Spears, and Lead Ammunition [This part is subject to change]

## Why / Balance

This adds a container that can be used to safely transport hazardous materials, such as Bananium.

## Technical details

New code adds: Lead spawning in asteroids
Not coded yet: Lead Ore Crabs, Lead Ore, Lead Ingots, Lead Crate

## Media
Will add media once I do the sprites.

## Breaking changes

**Changelog**

P.S. Coding will be slow, as I have to actually find what I need to code in the files first


Reene-Ramos:
Note to Self: remove the Basalt Lead spawner, and finally find the ore processor code

Samsterious:
I question whether adding this much content is actually justified solely by one key item. I think it would be much more reasonable to write a PR adding a hazmat container that has to be ordered from Logi, like artifact containers.

deltanedas:
artifact containers already block radiation

Samsterious:
This is true, although the catch is that they can only hold one stack of items. A crate variant dedicated specifically to hazmat would serve a function distinct from that. Regardless, I think this PR lacks the mechanical substance to justify an entire new ore material.

TadJohnson00:
Agreed. Besides, I don't think @Reene-Ramos is still working on this.

TadJohnson00:
Closing this PR due to inactivity.

PR #1563, opened 2024-07-30
Adjusts steal group for Shiva/Laika discrepancy on maps

IamVelcroboy:
## About the PR
Because we have 2 different sec animals and only 1 per map. This fixes the issue where one might not be available to secure the objective.

## Why / Balance
Fixes thief objective form being unobtainable. 

## Technical details
n/a

## Media
n/a

## Breaking changes
n/a

**Changelog**
n/a


deltanedas (approved):
shivaloose

PR #1580, opened 2024-08-04
glacier update

deltanedas:
## About the PR
- fix dorms closet facing a wall :trollface:
- add venting pump to teg incase someone opens it without a holofan and you need to get rid of the evil nitrogen
- add firelocks to main donut and air/fire alarms for them
- linked kitchen/bar vents to its air alarm so it can be controlled
- made air alarms in engi bridge etc not useless
- added built oreproc to logi so they dont need to build one every round


perryprog:
Regarding air alarms—there’s a few others out there that could use some tweaking. The air alarm in epi isn’t linked at all, and the scrubber and vent in botany aren’t linked up anywhere. There’s also a few cases where there’ll say just be a scrubber linked but not the vent linked when both are needed, if you want to cycle out air to reheat a room for instance. Finally there’s also a few cases of areas just missing air alarms, like a few parts of sec or atmos dorms. This is just from memory but I can get a more complete list if desired.

deltanedas:
yeah ill go over all of them

IamVelcroboy (approved):

PR #1593, opened 2024-08-10
give prisoners a criminal record roundstart

deltanedas:
## About the PR
title, and set them to detained automatically because they are

right now its rng but i want to change it to have a thing in profile for custom record lines

## Why / Balance
why are you in perma? erm i forgot

ideally players could add their own custom records but wyci

## Technical details
moved fugitive crimes data into `RandomCrimes` so it can be reused

## Media

**Changelog**
:cl:
- add: The Prisoner role now gets a random criminal record for every shift.


IamVelcroboy:
>ideally players could add their own custom records but wyci

Yeah, maybe in the future there could be a checkbox of potential crimes they could add/remove to their pool of potential crimes in chargen screen.

Love this idea overall though!

BlitzTheSquishy:
The system from CD would be better imo

pissdemon:
I do not think this is a good change. People should be free to choose what they are in prison for. This forces your character into having a criminal record that you might not agree with as a player which can put off people from playing the role. 

Those who remain are faced with the problem that they are not shown what they are in for. They cannot roleplay it until they are told, which ultimately means this does not really solve the issue of "why are you in perma? erm i forgot".

TadJohnson00:
Mixed on this change. I do like having them set to "detained", but there needs to be some direction discussion about whether this is a change we want for roundstart permabrig prisoners.

deltanedas:
yeah ill be redoing this at some point to be something you can pick in the character profile

Samsterious:
I would suggest taking a look at CD's employment/medical/security records. It is an unintrusive HRP mechanic that can be fairly easily ported to Delta-V without interfering with its MRP needs, and which gives players the ability to define their own IC crimes. Deciding FOR the players what their crimes are — especially because they cannot even see what their crimes _are_ — is not a good plan. Very often the players have an idea already in mind, and we should respect that.

deltanedas:
yes i know already

TadJohnson00:
Is this still getting worked on @deltanedas or will you implement this in another PR

deltanedas:
db stuff is fucked i doubt it

PR #1625, opened 2024-08-14
Adjusts AME rate of diminishing returns

IamVelcroboy:
## About the PR
Changes the AME fuel input to power output to follow a proper rate of diminishing returns. 

## Why / Balance
Previously we had a constant input to output increase:
Assuming a safe injection:core ratio we had,
`80000/160000/240000/320000/400000/480000/560000/640000`

The recent upstream merge created a bell curve drop off in returns:
`120.4/180.6/175.1/152.4/126.9/103.2/82.5/65.3`
The more cores you add, the less power you get.

This change would make it a proper diminishing return:
You still gain power with increased cores but less with each additional core
`90,300/180,600/233,445/270,915/300,000/323,745/343,830/361,230`

## Technical details
n/a

## Media
n/a

## Breaking changes
n/a

**Changelog**
:cl: Velcroboy
- tweak: Tweaked the AME to produce power with diminishing returns


IamVelcroboy:
If we wanted this to be closer to the values before such as 4-core=320k, we could increase the variable from .75 to something like .85 or .90. The diminishing return would be much more gradual/less impactful though.

PR #1631, opened 2024-08-14
make mindshields removable

deltanedas:
## About the PR
title

## Why / Balance
i thought they were for like a year but turns out no

## Technical details
improved version of https://github.com/space-wizards/space-station-14/pull/22225

**Changelog**
:cl:
- tweak: Mindshield implants can now be removed.


Axionyxx:
Unless it was fixed between then and now, removing mindshields does nothing

deltanedas:
debug had a comment on the old mindshield pr which is why i am doing that so it actually works

PR #1636, opened 2024-08-16
upstream merge

deltanedas:
## About the PR
notable changes:
- added back grappling hook
- added RE to diamond tipped drill like regular drill has
- shock collar recipe from nyano was unused, now its roundstart in the secfab
  upstream has its own delay mechanism so dont need to add anything
- infinite blood fix is merged
- updated gamerules to use new entity table stuff
- 50 ingredients in burger

excluded:
- meteor swarm scheduler changed to use old meteor event, still uses new code and scheduler
- mantle loadouts with timer
- contraband examine (disabled with cvar)
- german accent
- mining rebalance, magnet has its own balancing here

## Why / Balance
grappling hook is essential for spaceborgs and salv ever since space friction was reduced for mobs, and its not nearly as lrp as gravity guns (which are still commented)

no contraband examine since it would need a massive pr to make everything correct, and its doubtful that its even a good thing

**Changelog**
:cl:
- add: Added the grappling hook back to the research tree.
- tweak: Burgers can have 50 ingredients now.
- fix: Fixed Synthesis Specialists not showing up in the round end summary.


Resources/Prototypes/Entities/Structures/Wallmounts/shelfs.yml
```diff
@@ -35,7 +33,6 @@
       - !type:DoActsBehavior
         acts: ["Destruction"]
   - type: WallMount
-    arc: 175
```
IamVelcroboy, line 38: isn't this going to make them accessible from either side of the wall?

IamVelcroboy: Tested upstream. Apparently it defaults to 180

deltanedas: yes 180 is fine i think we had 175 for a while then it was upstreamed and changed to 180

IamVelcroboy: Really it should be 90, or you'll get occasional issues where you can access them through windoors/thindows but 🤷🏼‍♂️

deltanedas: pretty sure changing it to 90 would break maps, and be really annoying

IamVelcroboy: It wouldn't because no one has mapped them yet.

But also....I take it back...175 *is* what it should be [I've changed this before](https://github.com/space-wizards/space-station-14/pull/25451/files)

PR #1664, opened 2024-08-21
Rules: close MMI loophole in New Life Rules

pissdemon:
## About the PR
If you are put into an MMI, you now explicitly have to forget the last five minutes leading up to your death and cannot describe who or what killed you.

A player can be brought back without being cloned or borged if they are just put in an MMI. 

## Why / Balance
Currently the rule is too ambiguous and may be misinterpreted.

Your brain being put into an MMI may not be understood by players as "borging", especially since, as an MMI, you can refuse to be put in a cyborg body, or might simply not be placed into one because no bodies can be made at the time. I also get the impression that some Sec players want to put victims' brains into an MMI specifically to find out who killed them, which seems like  wanting to make use of that loophole or simply not being aware enough of the New Life Rules.

Pinging @Lyndomen since it's a rule change. 

**Changelog**
no cl no fun


Lyndomen (approved):
peak

deltanedas (approved):

PR #1699, opened 2024-08-28
add playtime requirement to disaster victim ghost roles

deltanedas:
## About the PR
20 hours so shitters dont get free AA and a fucking gun to play with
no whitelist so there still some variety from cap mains and who picks these roles

**Changelog**
:cl:
- tweak: Disaster victim ghost roles now have a playtime requirement.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
@@ -785,9 +785,9 @@
       rules: ghost-role-information-disaster-victim-rules
       raffle:
         settings: default
-      requirements:
+      requirements: # DeltaV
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
```diff
@@ -785,6 +785,9 @@
       rules: ghost-role-information-disaster-victim-rules
       raffle:
         settings: default
+      requirements:
```
MilonPL: There should be a DeltaV comment indicating the requirement has been added compared to the original code at the first line, not just one saying the amount of hours.

MilonPL (changes requested):

PR #1708, opened 2024-08-30
fix blanket whitelist not working in client

deltanedas:
## About the PR
server allowed it but client has copy paste code so it wasnt changed :trollface:
now it has a similar check in the copy pasted function

## Media

**Changelog**
:cl:
- fix: Fixed whitelists not working for some people.


Lyndomen:
pro we are doing sql stuff or something to fix  this, do we have to do the sql stuff?

MilonPL:
> pro we are doing sql stuff or something to fix this, do we have to do the sql stuff?

this will fix it but doing it through SQL will mark the currently whitelisted players as the 3 roles
it's a stupid fix and doing a migration would be better anyways

deltanedas:
serverside everythings fine its just an issue with opting in, you can keep or remove blanket whitelist from people at your discretion

PR #1721, opened 2024-08-31
mining re-rebalance

deltanedas:
## About the PR
- iron and quartz to 2, it's alright right now but you need to be mining full time + debris recycling makes a stupid amount of steel anyway
- gold uranium and silver to 3, its annoying how little you get. still not as stupid as 30 stacks of gold every shift but less unforgiving
- plasma is back to 5, even before rebalance it was fairly hard to get
- bananium still low cause 50 rad deathstack should take time

**Changelog**
:cl:
- tweak: Rebalanced mining again, it's a bit fairer.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Materials/ore.yml
+++ b/Resources/Prototypes/Entities/Objects/Materials/ore.yml
@@ -37,7 +37,7 @@
   - type: Material
   - type: PhysicalComposition
     materialComposition:
-      RawGold: 100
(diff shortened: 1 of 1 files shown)


jimmy12or:
Did Delta-V keep bananaium's radioactivity?

deltanedas:
yes

ghost (approved):
It's got delta-v comments, which means it should be fine. Hitting button now.

PR #1727, opened 2024-08-31
Map Edits: Tortuga

IamVelcroboy:
## About the PR
Tortuga
- Adds recycler in salvage
- Replaces tiny fans with directional fans
- Adds shelves in various places
- Envelopes!
- A couple very minor edits

## Why / Balance
🐢


therealDLondon (approved):
Looks good as far as my tired eyes can tell

PR #1736, opened 2024-09-01
fix

MilonPL:
## About the PR

overlooked it

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Species/harpy.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Species/harpy.yml
@@ -57,6 +57,7 @@
         visible: false
       - map: [ "id" ]
       - map: [ "gloves" ]
+      - map: [ "shoes" ]
       - map: [ "ears" ]
       - map: [ "outerClothing" ]
       - map: [ "eyes" ]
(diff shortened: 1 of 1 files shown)


ghost:
goes with the digitigrade deletion thing from the harpy fixes

If I ever find a way to implement harpies having no shoes without stepping on sharp objects hurting them in a modern context without some legacy code (since apparently they're intended to have no shoes as part of the species design, shoes don't fit their talons in the sprite) I might bring it back but the digitigrade thing being uncommented and in upstream files is bullshit. That's my mistake

ghost (approved):

jimmy12or:
> goes with the digitigrade deletion thing from the harpy fixes
> 
> If I ever find a way to implement harpies having no shoes without stepping on sharp objects hurting them in a modern context without some legacy code (since apparently they're intended to have no shoes as part of the species design, shoes don't fit their talons in the sprite) I might bring it back but the digitigrade thing being uncommented and in upstream files is bullshit. That's my mistake

I don't play as them often, but do Diona get hurt from sharp objects when stepped on? if not, their code might have something since they too can't wear shoes.

ghost:
Yeah. They get hurt by glass shards because stepping on them in real life hurts. I used to play an engineer diona and it was hellish because you'd go out to fix solars and trip on the glass shards

Most Nyano species used a tag that was called "ShoesRequiredStepTriggerImmune" (or something like that) and that somehow caused them to not be griefed by glass shards, I'm just guessing, and upstream this got converted to a comp that in the PR it's from, got converted into ClothingRequiredStepTriggerImmune, which works nowhere near the same way i think

PR #1737, opened 2024-09-01
lighthouse update

deltanedas:
## About the PR
- perma has air alarms and more than 1 apc now
- added a perma vox box with its own nitrogen supply
- add some more injectors to burn chamber so its faster
- atmosian combat gloves
- shocked the grilles outside atmos so its slightly harder to sabotage
- give salv a recycler for magnet debris
- unlocked the psychologists waiting room
- add shelves, the one in corpsman room has corpsman access


therealDLondon:
Do the atmos combat gloves exist to make combat gloves less valid?

deltanedas:
and because atmos inevitably need them, and tradition

IamVelcroboy:
Double dorms
image
Check for casing consistency on cameras and fax machines
image
image
image
Haha, ok....
image
Disconnected 
image
A couple comments from Discord. I def agree with the first one. 
image

Ideally engineering would also have access to the TEG. Maybe consider a hall through here...?
image
And thindows across here...?
image

I don't know, just a suggestion but, like I said, engi should have access here too, somehow.

IamVelcroboy (approved):
Again, going to merge, incorporate above into next PR, please.

PR #1738, opened 2024-09-01
Add ringbox and loadout rings!

beck-thompson:
## About the PR

Upstream didn't want any of this stuff but from what I see its wanted here. See (https://github.com/space-wizards/space-station-14/pull/31372) for a little more information.

I'm basically just adding the ring box and making rings obtainable in a few more places, specifically:
1.) Piety vend
2.) Loadouts
3.) Crafting

## Why / Balance

People keep getting married and they need rings!

## Requirements

image

**Changelog**

:cl: Beck Thompson
- add: The gold and silver ring and now available in the loadouts.
- add: Piety vend now has 2 silver rings and 2 ring boxes.
- add: Rings are now craftable on the lathe.


beck-thompson:
First delta commit, not sure if I put the files in the correct locations.

Resources/Prototypes/Loadouts/loadout_groups.yml
```diff
@@ -601,7 +603,7 @@
   loadouts:
   - MimeSuspendersRed
   - MimeSuspendersBlack
-
```
MilonPL: Should probably untouch it.

MilonPL (changes requested):

Resources/Prototypes/Loadouts/loadout_groups.yml
beck-thompson: I really have no idea what git is yapping about because I swear its exactly as it is now... I just copy pasted old file and just added those two lines and it still says that exists

MilonPL: Visual studio code, if that's what you're using, automatically removes trailing whitespaces, which isn't something that should be touched in this PR. You should be able to disable it in the user settings.

beck-thompson: Think I got it

MilonPL (approved):

therealDLondon (approved):
Milon's already given it the once over. This looks fine to me

PR #1743, opened 2024-09-02
Nukes AGPL headers

ghost:
## About the PR

This just nukes file comments. It's intended to be a minor PR but I fucked up the branch creation. That is an idiot maintainer moment.
I didn't realize I could make this in upstream and that it would appear here and this was meant to be a uh. Local-to-my-repo PR but oops. Deltanedas feel free to kill this one because I am Fucking Stupid.™️ Anyways, cutting to the chase in the why section.

## Why / Balance

This asks an important question - why do we have these on some files but not all delta files since the AGPL license change? It is inconsistent and confusing. If someone can answer that I'll revert the changes in this PR (which is currently removal) and then go back and paste this to every file from every PR that has been approved since, which is a lot of work.
## Technical details

Not player facing, just nukes comments.

**Changelog**

No CL no fun

### DIFF
--- a/Content.Server/DeltaV/StationEvents/Components/LoadFarGridRuleComponent.cs
+++ b/Content.Server/DeltaV/StationEvents/Components/LoadFarGridRuleComponent.cs
@@ -1,9 +1,3 @@
-/*
-* Delta-V - This file is licensed under AGPLv3
-* Copyright (c) 2024 Delta-V Contributors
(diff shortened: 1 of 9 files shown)


deltanedas:
putting it in every file is just a waste of disk when theres already the bit in readme about files after the commit

deltanedas (approved):

ghost:
Okay real

PR #1817, opened 2024-09-12
make fugitives anonymous

deltanedas:
## About the PR
the fugitive report no longer gives the name for humanoids

non-humanoid fugitives still have their name because otherwise you have literally nothing to go off of.

instead of a name you get 50 50 chance of dna or fingerprints, more things might be added in the future (like hair colour or whatever)

## Why / Balance
right now its completely trivial to find fugitives that dont fuck off to space or hide in a locker for the rest of the round since everyone can recognize their name. this should make it easier for them to blend in and make finding them among the crowd an actual challenge for sec (have to open crew manifest 99% impossible challenge) and create more rp

## Technical details
dont pass name

## Media
outdated im tired

## Requirements
- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion

## Breaking changes
no

**Changelog**
:cl:
- tweak: Fugitive reports no longer include the name of the fugitive. Use your detective skills!


Samsterious:
With the clerical error event occasionally removing crew from the manifest, and no certain way to confirm they have the right person, Security would be forced to start guessing. What is your thinking about the new gameplay here?

deltanedas:
searching them for suspicious items (like a gun :trollface:)

ideally there would be a mugshot generated or a description of hair added before this so they can be identified a bit easier

Samsterious:
Given that this would create cause for Security to start holding players in Perma only because they were profiled entirely off of morph, age category, and voice pitch, I think implementing something a bit more reliable is a must before we merge this.

Avalon-Proto:
> Given that this would create cause for Security to start holding players in Perma only because they were profiled entirely off of morph, age category, and voice pitch, I think implementing something a bit more reliable is a must before we merge this.

While that's one good point. Also note this will incentivize security to actually actively check the manifest, criminal records, and station record computers to background check who they think the fugi is

Samsterious:
I do not think ruining an innocent player's round because Security has zero hard evidence to prove they even have the right person — again, their only evidence, given the unreliability of these records, is a vague description — is an acceptable cost for encouraging Security to employ a slightly more diverse set of investigative tools. Remember that there are many characters on the station with no relevant records, if Fugitives can spawn. Security needs to be able to prove they have the right person somehow, and a vague description is not adequate.

ThataKat:
What if you gave sec DNA or prints? Still concrete info but also not as easily identified as a name. Detain suspects and check their info, should hopefully ensure the wrong person doesn't get permaed.
(It doesn't make a lot of sense for GALPOL to know DNA and not name, but maybe justify it as the fugitive operating under an unknown alias?).

Samsterious:
I am a fan of this suggestion, plus it gives the Detective an opportunity to track down the fugitive by keeping an eye on prints and such around the station. Perhaps randomly select one of DNA or Fingerprints, or something like that.

deltanedas:
i like randomly selecting info so you dont always have the same procedure to follow every time

MilonPL (approved):
Everything is clean code-wise, the helper methods do what it says on the tin, just needs the direction approval.

PR #1818, opened 2024-09-12
remove sentient artifact effect

deltanedas:
## About the PR
title

## Why / Balance
ive seen good rp from artifact players like 4 times total, the rest of the time they just spam activate and murderbone with trit/explosions/radiation/etc
it also does nothing on lowpop 
if the artifact also has structural phasing its literally impossible to do anything about a shitter using it to destroy everything

**Changelog**
:cl:
- remove: Artifacts can no longer become sentient.


Samsterious (approved):
Agreed with your sentiments here. This is a really low-quality bit to have such destructive power. Once I had sophont law RP because of it, but then Epi just crushed it because a lot of folks seem not to take sophont laws seriously lol

PR #1819, opened 2024-09-12
Limit "Die a glorious death" objective to one traitor per round.

beck-thompson:
## About the PR

See the title!

## Why / Balance

A lot of shifts (Especially with a lot of traitors) can end up having three or four people with the objective which ends up very chaotic and not fun. Limiting it to one per round still allows the objective to come up, just not dominate the round!

**Changelog**

:cl: Beck Thompson
- tweak: The "Die a glorious death" objective is now limited to one traitor per round maximum!

### DIFF
--- a/Resources/Prototypes/Objectives/traitor.yml
+++ b/Resources/Prototypes/Objectives/traitor.yml
@@ -66,6 +66,10 @@
       - EscapeShuttleCondition
       - StealCondition
   - type: DieCondition
+  # The game turns into a massive suicide pact if enough of the traitors have it. Not fun for traitors or security!
+  # Giving it a limit keeps the objective unique and interesting without being overdone.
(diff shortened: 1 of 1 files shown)


deltanedas:
i think that means 2 max

beck-thompson:
> i think that means 2 max

When I tested it only let one person have the objective so I think its OK? Looking at the code, it also appears to make sense and should work like I thought. Also your the one who wrote it so if it doesn't work I blame you :troll:

beck-thompson:
escape characters do not appear to work on github so that emoji is not correct 😭

deltanedas:
:trollface: `:trollface:`

deltanedas (approved):
@TadJohnson00

TadJohnson00 (approved):
cool and direction-approved

PR #1825, opened 2024-09-14
upstream merge 14/9

deltanedas:
## About the PR
- i forgor other stuff because trolled
- food sequence update, can look at any nyano/deltav recipes that can be changed to use the new thing
- new shuttle events
  visitor pda for passenger loadout had to be changed to not give tiders external access on lowpop
- made nyano circuit recipes consistent
- more plant mutations, holymelons give nyano HolyWater instead of upstream Holywater
- maxcaps can be limited with cvar if admins want to
- theres possibly a bug letting latejoins spawn on the nukie outpost
  i couldnt reproduce it no matter what i did so idk

excluded security trooper uniform/hat since it might not fit with the style @IamVelcroboy 

**Changelog**
:cl:
- add: Merged upstream changes for this week.


Samsterious:
Popping out of vacation mode to note that we had a couple concerns with this upstream merge (namely the shuttles, which are going to require some review from @IamVelcroboy in addition to an "Is this just a meme?" test, and the random sentience event, which should be tied to Glimmer).

PR #1837, opened 2024-09-16
Revert "Firelocks are no longer pryable by hand if they are powered

deltanedas:
## About the PR
you can now pry firelocks without a crowbar

## Why / Balance
- this did nothing in the event of spacing and power loss, which is extremely common
- this led to everything taking crowbars immediately because (brave statement) they dont want to get murdered by a powered firelock
- since everyone takes a crowbar, this had no effect on the powergamers but punished people that dont take them, in the case of fires by round removing them
- **nobody** enjoys getting killed by firelocks. not a single person

**Changelog**
:cl:
- tweak: You no longer need a crowbar to pry powered firelocks open.

### DIFF
--- a/Content.Shared/Doors/Systems/SharedFirelockSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedFirelockSystem.cs
@@ -21,7 +21,8 @@ public abstract class SharedFirelockSystem : EntitySystem
 
         // Access/Prying
         SubscribeLocalEvent<FirelockComponent, BeforeDoorOpenedEvent>(OnBeforeDoorOpened);
-        SubscribeLocalEvent<FirelockComponent, BeforePryEvent>(OnBeforePry);
+        // DeltaV: commented out to let people pry powered firelocks by hand
+        //SubscribeLocalEvent<FirelockComponent, BeforePryEvent>(OnBeforePry);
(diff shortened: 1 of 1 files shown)


MilonPL:
holy based

TadJohnson00:
Would it be possible to just comment out a line of C# here instead of fully reverting? If we do end up planning anything major with space wind or depressurisation, having the code would still be nice.

deltanedas:
could just add a return yeah :trollface:

Content.Shared/Doors/Systems/SharedFirelockSystem.cs
```diff
@@ -61,14 +60,6 @@ private void OnBeforeDoorOpened(EntityUid uid, FirelockComponent component, Befo
             WarnPlayer((uid, component), args.User.Value);
     }
 
-    private void OnBeforePry(EntityUid uid, FirelockComponent component, ref BeforePryEvent args)
-    {
-        if (args.Cancelled || !component.Powered || args.StrongPry || args.PryPowered)
-            return;
-
-        args.Cancelled = true;
```
TadJohnson00: as far as I can tell literally just comment out this one line and Everything Will Be Fine or some shit. no need to revert everything

deltanedas: yes ill just comment out the event subscription

TadJohnson00 (approved):
Direction approved 😎

Issue #1830, opened 2024-09-16
new shuttles need looking at


deltanedas:
## Description
- [ ] cruiser
- [ ] cryptid
- [ ] eternal
- [ ] flatline
- [ ] gym (this is the b-ball one i think its more of a meme)
- [ ] incorporation
- [ ] joe
- [ ] lambordeere
- [ ] meatzone
- [ ] microshuttle
- [ ] quark
- [ ] spacebus

Vapetastic-Gaming:
@IamVelcroboy is this good to close ?

IamVelcroboy:
Going to run through Direction again

Pharaz4:
@IamVelcroboy did direction ever get back to you on this one

PR #1839, opened 2024-09-16
disable new shuttle events

deltanedas:
## About the PR
content standards, #1830


ghost (approved):
Comments out things, I think this is smart, so I'm gonna do it. Considering we're gonna have to look at 'em.

PR #1840, opened 2024-09-17
make uncloneable popup more clear

deltanedas:
## About the PR
SOUL IS ABSENT -> INCOMPATIBLE GENOME

## Why / Balance
it sounds like someone left when no they did not its just a trait

**Changelog**
:cl:
- tweak: The Uncloneable trait's failure message is clearer now.


Avalon-Proto:
Oh gee I wonder why you made this one :p

PR #1841, opened 2024-09-17
lighthouse update

deltanedas:
## About the PR
all previous reviews addressed
- intercoms added to departments, perma, and a few halls
- downgrade a lot of reinforced windows/walls where there shouldnt be
- fixed the disposals below vault being disconnected
- fixed all camera and fax names
- made the non-command airlock for ai access external instead of engi locked so parameds can use it

- made top room of ai sat ai upload, move the console into it

- moved grav and comms/cameras into the back of engi (with a sneaky mineable path into it)
  engis can now go into the teg room and use the cooling loop access
  (old grav room is now a vacant room for tiders to make shops or whatever)

- atmos gloves replaced with insuls :(
- named a few air alarms
- added alerts computer to bridge, ce room, atmos


IamVelcroboy (approved):
I never said you had to remove the gloves. Just thought it was funny. I like the engie change though.

PR #1851, opened 2024-09-18
glacier update

deltanedas:
## About the PR
- moved fire axe cabinet to face the bridge instead of caps room
- add red phone to caps desk
- add robotech deluxe to robotics
- move analysis console so you can see the canister pump without right clicking
- add grinder and a chemmaster board to botany


IamVelcroboy (approved):
Interesting. Might try the autoclose: false thing on some perma cells

PR #1862, opened 2024-09-19
more yaml nospawn cleanup

MilonPL:
## About the PR

yaml cleanup mostly frontier files (ADEI!!!) and rodentia 
no comments since no one cares, idk if we even ported the latest mail pr and upstream doesn't have rubbers anyways

could probably move rubbers down to deltav projectiles section since you can't put it in a different file anyways :blunt:

## Why / Balance

nospawn finally removed after a year


ghost (approved):
Win. It's obsolete now!! YAY!!

PR #1893, opened 2024-09-22
upstream merge but real

deltanedas:
merge commit...


Content.Client/Commands/ActionsCommands.cs
```diff
@@ -50,7 +46,7 @@ public override void Execute(IConsoleShell shell, string argStr, string[] args)
     {
         if (args.Length != 1)
         {
-            LoadActs(); // DeltaV - Load from a file dialogue instead
```
IamVelcroboy, line 53: I need this back :(

deltanedas: idk i think it was breaking mapping

IamVelcroboy: Fwiw, I added it back so I could do some mapping last night w/ no issues. I don't know what will happen if we re-enable the mapping editor though.

deltanedas: make a pr then, if it breaks map editor again whenever map editor works it can be reverted

PR #1900, opened 2024-09-23
nukie mouse yaml cleanup

MilonPL:
## About the PR

cleanup
accidental, was trying to fix something else, but sure

## Technical details

it's all inherited anyways


MilonPL:
autotraitor is evil and should be removed too but that requires so much work for them to show at the EOR screen

deltanedas:
just remove it

MilonPL:
yeah good call

deltanedas (approved):

PR #1907, opened 2024-09-24
revert mapping mode

deltanedas:
## About the PR
it has pretty bad bugs still so back to caveman method of entity spawn menu :trollface:


MilonPL:
did you make sure it doesn't break anything this time :blunt:

deltanedas:
this just makes it use the old actions its not an actual revert

MilonPL (approved):

Issue #1849, opened 2024-09-17
Evac shuttles missing comp for screens to show ETA/ETD


IamVelcroboy:
## Description

- [x] NTES_Vertex
- [x] NTES_UCLB

Needs 

```
    - type: DeviceNetwork
      configurators: []
      deviceLists: []
      transmitFrequencyId: ShuttleTimer
```

**Additional context**
Easy peasy yaml copy/paste

FryOfDestiny:
I added a pull request, it's my first time so not sure if I did it right.

IamVelcroboy:
> I added a pull request, it's my first time so not sure if I did it right.

Looks good. Thank you

therealDLondon:
I know a PR was merged for the Titan but I believe the Vertex and Fishbowl share the same issue

IamVelcroboy:
> I know a PR was merged for the Titan but I believe the Vertex and Fishbowl share the same issue

Looks like the Vertex, yes. Fishbowl is good to go

IamVelcroboy:
Looks like Uncle Charlie's Luxury Boat needs it too. Which is weird, should have been fixed in first pass 🤔

PR #1945, opened 2024-09-30
Added DeviceNetwork to NTES_UCLB and NTES_Vertex

FryOfDestiny:
Added device network to the two shuttles still missing it as requested by https://github.com/DeltaV-Station/Delta-v/issues/1849


IamVelcroboy (approved):
Thanks again

PR #1949, opened 2024-09-30
add selfunremovable to headcages

### LINKED ISSUE #1386: headcages can be taken off by yourself
## Description
this gives sec no real reason to use them, it doesnt even need a doafter just take it off and youre ready to use your evil powers

if it could only be put on removed by other people then it would be effective, but right now theres no reason to because you cant use mental powers just because you are cuffed, so if there a dangerous psionic sec will just keep them cuffed 24/7 as if they're uncuffed they can just take the cage off and use powers.

if both are changed then sec has actual tools to keep psionics restrained properly, until johnny greytide helps them out

MilonPL:
## About the PR

title
fixes https://github.com/DeltaV-Station/Delta-v/issues/1386

## Technical details

yaml + deleted the 2 funny components that weren't referenced anywhere

## Breaking changes

**Changelog**

:cl:
- fix: You can no longer take headcages off by yourself.


deltanedas (approved):
headcage actually useful for sec now

PR #1965, opened 2024-10-05
Add new station: Byoin

FieldCommand:
## About the PR
Adds a small station that is about twice the size of Chibi. The station is made to be relatively easy to operate with a crew soft cap of 4. This map aims to fill a very niche spot in the map rotation when there are the fewest players on the server.

## Why / Balance
I made this map to add more variety when there are next to no players on the server. Making this map easy to operate with very few players. It is heavily inspired by the station "Chibi". Like the Chibi map, this map has a large reduction in many roles, including the omission of many roles that are quite RP havy.

The soft cap of this map is 4 players, but there are 26 total job slots, not including passenger
This station also has its own evacuation shuttle.

## Requirements

## Media
Byoin-1 92
Evac Basu

**Changelog**
:cl: Field Command 
- add: Byoin station, with soft population cap of 4.
- add: NTES Evac Basu shuttle


Resources/Prototypes/Maps/byoin.yml
```diff
@@ -0,0 +1,48 @@
+- type: gameMap
+  id: Byoin
+  mapName: 'Byouin'
```
IamVelcroboy: Change this back. Looking at Shoukou, accented characters here should be fine

IamVelcroboy:
@deltanedas I can't figure out what's tripping the test error. It says ` No user map found: /Maps/byoin.yml` but `byoin.yml` is clearly in `/Maps` 🤷‍♂️

MilonPL:
> @deltanedas I can't figure out what's tripping the test error. It says ` No user map found: /Maps/byoin.yml` but `byoin.yml` is clearly in `/Maps` 🤷‍♂️

I'm pretty sure it fails when *loading* the map, not because it can't find the map, as that's standard output. It's tripping because of Entity 0 which is uh, yeah. I've seen this error quite a few times in production too.

deltanedas:
remove all `invalid` references

IamVelcroboy:
> remove all `invalid` references

There aren't any

Monotheonist:
Will Byoin have AI?

FieldCommand:
> Will Byoin have AI?

Sorry for the late reply, Monotheonist. 

Originally I planned to make space for one, but this was before the AI was introduced, and as the progress of making this map went along, and what role this station is trying to fill, alongside what this station is based on, I soon went away with the idea of having a station AI. But if it's needed, I will add one, but if that is to happen, I will convert this PR to draft and do some more major rework across the board to take it from graveyard hours station to be more in line with a low- to mid-pop station.

IamVelcroboy (approved):

PR #1983, opened 2024-10-07
Adds the coordinates to the rattle message.

Radezolid:
## About the PR

Modified the rattle to show the coordinates next to the nearest beacon it appeared. It doesn't affect messages like the dragon's rift.

## Why / Balance

Better location data to be able to find the person, `west of Solars` doesn't help much sometimes.

## Technical details

Obtains the tracker's user coordinates using `GetMapCoordinates` and then adds the coordinates to the rattle message.

## Media

https://github.com/user-attachments/assets/d12e5010-2234-47a1-b33e-eb2a591152e4

It doesn't affect messages like the dragon's rift:
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

Hopefully none.

**Changelog**

:cl:
- tweak: Now tracking and death rattle implant messages show the coordinates at the same time as the nearest beacon.

PR #1984, opened 2024-10-07
Medical tracker implant

Radezolid:
## About the PR
This PR adds the medical tracker implant, pretty much a tracker implant with sensors on Vitals and not on Coordinates, but for medical and civilian use.
It can be printed at the medical techfab at the cost of 10 steel sheets and 10 glass sheets (Double the empty implanter).
Works best with https://github.com/DeltaV-Station/Delta-v/pull/1983

## Why / Balance
This tracker aims to reduce or at least help with the salvage deaths that end on round removal on the asteroid or wrecks, note that this doesn't help if the person is drifting and it doesn't have the coordinates configuration, so you can't track them on the crew monitor, encouraging antags to grab it.

## Technical details
Creates a new implant identical to tracker, but with `SensorVitals` instead of `SensorCords` and sends the rattle on the `Medical` radio channel.

## Media

https://github.com/user-attachments/assets/7a8196db-a77f-4002-9d7c-b86d22ea7889

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).

## Breaking changes

**Changelog**

:cl:
- add: Added a new medical tracker printable at the Medical TechFab which only relays Vitals sensors and it will notify medical if you go in critical state but it alone won't show the position on the crew monitor.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Misc/implanters.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Misc/implanters.yml
@@ -21,3 +21,11 @@
   components:
   - type: Implanter
     implant: SyndicateRadioImplant
+
+- type: entity
+  id: MedicalTrackingImplanter
+  suffix: MedTracking
+  parent: BaseImplantOnlyImplanter
+  components:
+    - type: Implanter
+      implant: MedicalTracker
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Misc/subdermal_implants.yml
@@ -65,7 +65,5 @@
     - type: DeviceNetwork
       deviceNetId: Wireless
       transmitFrequencyId: SuitSensor
-    - type: WirelessNetworkConnection
-      range: 10000
     - type: Rattle
       radioChannel: "Medical"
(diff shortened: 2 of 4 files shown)


Radezolid:
If any ideas on how to change the `Implanter` option from the medfab to show something like "medical tracker" please let me know.
Also if this should be a cargo crate let me know and I'll do changes to it, more than anything because of how this may affect kill targets on medical personnel.

Radezolid:
Been pointed out about possible powergaming with this so i'll hold off until i find a way to ensure it isn't powergamed.

Radezolid:
Me when forget that draft is a thing.

deltanedas:
command has always powergamed it with tracking implants (which tell sec, much more useful for catching the valids)

Radezolid:
> command has always powergamed it with tracking implants (which tell sec, much more useful for catching the valids)

Still, the main reason was to make new salvies not get RR due to mistakes (even if it's just skill issue) or lag in a wreck.

I've tought of:

1. Give it back into the lathe but make it stupid expensive.
2. Make it buyable on logi.
3. Make Salvage spawn with one in their bag (This might have issues as i've spawned as thief salvage and dropped the beacon on spawn).
4. Give the CMO two of them in their locker.
5. Lock it behind research, only printable at medfab but expensive.

But none of those really convinces me, since:

1. Logi can just buy mats and spam them.
2. Logi will powergame the shit out of it and only them have trackers.
3. Maybe? Feels a bit bad that in the end only salvage can get them.
4. They wont ever give them to people that needs them and they will give it to command.
5. Good fucking luck with this, by the time they research it salv got gibbed due to rot in space.

deltanedas:
if you add it as an orderable item people will just buy regular trackers instead since they have coords

Radezolid:
Can't find a viable way of making it so people don't powergame it.

PR #2002, opened 2024-10-13
blacklist nukedisk from slime storage

MilonPL:
## About the PR

what it says on the tin

## Why / Balance

bug fix, you can't access storage when the target cryos so if the captain has the nuke disk there it will get sent to nullspace

## Breaking changes

**Changelog**

no cl because i can't come up with a good one

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Misc/mouth_storage.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Misc/mouth_storage.yml
@@ -9,10 +9,11 @@
     grid:
     - 0,0,1,1
     maxItemSize: Small
-    blacklist: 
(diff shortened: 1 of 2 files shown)


deltanedas:
same applies to hypospray etc
it should just be listed like hands and inventory are

MilonPL:
> same applies to hypospray etc it should just be listed like hands and inventory are

that's the ideal long term solution but probably upstream, this is a quick fix so that it's not literally gamebreaking

deltanedas:
would also have to be done for rodentia storage i imagine

MilonPL:
> would also have to be done for rodentia storage i imagine

you're right yeah i forgot that was a thing

deltanedas (approved):

PR #706, opened 2024-01-21
More Borg Lawsets

DangerRevolution:
## About the PR

- Added several new lawsets to the Ion Storm pool
- Janitor Borg now naturally has a Janitorial Lawset
- Engineering Borg now naturally has an Engineering Lawset
- Medical Borg now naturally has a Medical Lawset
- Adds Potentially Harmful Lawsets to the Ion Storm pool
## Why / Balance

Station 13 has some cool lawsets + extra variety is always good.

Also, more lawsets = more opportunities for varied RP + adding borgs their own lawsets is good; i.e doctor one

## Media
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

image

image

image

**Changelog**
:cl: DangerRevolution
- add: Added several new potential Lawsets for Cyborgs.
- tweak: Janitor/Engineer/Medical Borgs spawn with their own lawsets.

PR #2005, opened 2024-10-14
Make all station borgs have Crewsimov

Tryded:
## About the PR

Made it so that the engineering, janitor, and medical borgs have crewsimov. Basically reverts #706
## Why / Balance

Makes the borgs laws consistent so that Station AI can easily work with borgs, also there were GIANT loopholes in the specialized job lawsets.
## Technical details

YAML only, literally just deleted stuff.
## Requirements

## Breaking changes

YAML
**Changelog**

:cl:
- tweak: Changed all station cyborgs lawsets to crewsimov.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml
+++ b/Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml
@@ -148,8 +148,6 @@
     interactFailureString: petting-failure-engineer-cyborg
     interactSuccessSound:
       path: /Audio/Ambience/Objects/periodic_beep.ogg
-  - type: SiliconLawProvider # Delta-V - Adds custom lawset for Engineering Cyborg
-    laws: Engineer
 
 - type: entity
   id: BorgChassisJanitor
@@ -211,8 +209,6 @@
     interactFailureString: petting-failure-janitor-cyborg
     interactSuccessSound:
       path: /Audio/Ambience/Objects/periodic_beep.ogg
-  - type: SiliconLawProvider # Delta-V Adds custom lawset for Janitor Cyborg
-    laws: Janitor
 
 - type: entity
   id: BorgChassisMedical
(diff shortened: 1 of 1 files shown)


deltanedas:
crewsimov doesnt have a law 0: you must always follow the ai
and if it was added it would still work for any lawset as it would be the most important law

Tryded:
> crewsimov doesnt have a law 0: you must always follow the ai

The Borgs are meant to follow the AI’s orders so long as it doesn’t violate their own laws.
> and if it was added it would still work for any lawset as it would be the most important law

Borg and AI law sets are going to be made so that they are synced on upstream as long as they aren’t individually ion lawed.

perryprog:
As a fairly frequent borg player, I don't think this is the right direction to take things. A huge part of playing borg to me is the differing lawsets that different chassis will have, as it can often have fairly significant gameplay implications. While there are issues with loopholes in some of them—namely engi borg comes to mind, though I have more to say on that in a bit—this can both be a good thing and also just something that is fixed on a case-by-case basis if it's needed.

Crewsimov _can_ end up being a fairly dry lawset fairly fast, especially if playing with people who aren't very considerate of the borg: you have to do every chore, every time, whenever someone says something. Having chassis that let you break out of just doing chores and being able to do your job can be really nice.

Engiborg's lawset is DEFINITELY problematic, but I think it has some interesting implications:
- Prioritizing the repair of the station over _everything_ else is an extremely wild choice for presumably NanoTrasen to have made. It speaks to the hyper-capitalist mindset that the station is more important than the crew, and it's cool to be able to interact with that.
- The engiborg using the term "station's inhabitants" is a little... not great. I would understand this being addressed, though it does have some merit. The main issue is when you get ion storm lawsets that you can't actually act on, such as "Only X is crew", but who is or isn't crew doesn't matter to this lawset.
- The "expand and upgrade the station" law is a really unique part of this lawset, as it actually gives you an explicit purpose to fulfill during downtime. While yes, you as a player could always choose to do this yourself, I find that it's more interesting game-play wise for this to be something you have to enforce on yourself.

I also think that janiborg's lawset is amazing, and I'm not aware of any issue with loopholes with it? It's also another situation of giving the borg a "purpose", and it can also result in interesting dialog as you may ask a crew member for "assistance in exterminating the enemy that resides within the LO's office".

deltanedas:
station inhabitants specifically is bad since that means mice and nukies (if they are on station) etc giving you priority repair targets

closing this since it wont matter for slaving, and AI can already say "janiborg there is a huge lube spill in epi", "engiborg sec blew up" "mediborg rescue this dead engineer", they are all perfectly valid with their laws.
law 2 wouldnt apply anyway as AI isnt crew so they wouldnt have to follow orders (janiborg), only prevent harm

PR #2024, opened 2024-10-20
Theobromine Tweaks (vulps + coffee)

ThataKat:
## About the PR
Tweaked theombromine stats to make its effects more logical. Small amounts will no longer cause vomiting, but large amounts are more likely to. 

Also made cocoa powder metabolize into theobromine.

## Why / Balance
At some point upstream changed theobromine such that any amount has a chance to cause vomiting. This sucks for any animal race that wants to enjoy coffee/chocolate products without vomiting everywhere and ruining the scene.

Cocoa powder is one of the most concentrated sources of theobromine irl, makes no sense that it didn't metabolize into it.

## Technical details
Modified the ingredients and toxins yaml files to tweak cocoa powder and theobromine. 

Theobromine metabolizes slightly slower now, allowing drinks like coffee to cause a slow buildup in the body (previously the amount of it being created perfectly matched the speed at which it metabolized). The toxin damage per second was slightly reduced to offset the slower metabolism.

Most notably, vomiting will no longer happen unless there is at least .5u, requiring the entity to consume a larger amount before vomiting can set in. Increased vomiting chance per second from 4% to 10% since vomiting should only occur at higher doses now.

Cocoa powder metabolizes into .2u theobromine per second (a high amount relative to the other reagents like coffee).

## Media
None

## Requirements

## Breaking changes
None

**Changelog**

:cl:
- tweak: Small doses of theobromine no longer cause vomiting in animal species.

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Food/ingredients.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Food/ingredients.yml
@@ -234,6 +234,9 @@
           reagent: Nutriment
           min: 0.1
         factor: 1
+      - !type:AdjustReagent # DeltaV - Make cocoa powder actually toxic to animal organs
+        reagent: Theobromine
+        amount: 0.2
   plantMetabolism:
   - !type:PlantAdjustNutrition
     amount: 0.1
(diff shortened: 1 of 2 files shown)


deltanedas (approved):
moth ops

PR #2026, opened 2024-10-20
remove 8 second role rule timer

deltanedas:
## About the PR
its now 3 seconds, the default

## Why / Balance
shitters dont care about rules anyway so the extra 5 seconds didnt matter
what it actually did is lock you out of a raffle if it had 8 seconds or less remaining, so if you are late seeing it you get just trolled

**Changelog**
:cl:
- tweak: Ghost role waiting time is down to 3 seconds, from 8.


MilonPL (approved):
"PR it away" real

PR #2028, opened 2024-10-20
add robotics access and mapping stuff

deltanedas:
## About the PR
adds the job spawn and these doors for robotics so they can be mapped before the job is added
- Standard
- Glass
- Maints
- Windoor

and lockable button/blast door. these should replace shutters since anyone can just pry shutters open regardless of power

right now the doors are research locked until the job is added, then they will be changed to Robotics. same for the spawner, its just a reskinned scientist spawner for now

there are unlocked variants incase they get a unique sprite in the future :trollface:

## Why / Balance
beep boop

## Requirements

## Breaking changes
no

**Changelog**
no cl no fun though technically it can be used ingame if you use access configurator to change doors then add access with cap/mysta/admin id :trollface:


deltanedas:
@IamVelcroboy lmk if there are any other things you think could be mapped early (except job spawn :trollface:)

IamVelcroboy:
> @IamVelcroboy lmk if there are any other things you think could be mapped early (except job spawn :trollface:)

Access on the vendor?

deltanedas:
access will be set once the job is actually added, it doesnt need any mapping changes

Colin-Tel (approved):

PR #2032, opened 2024-10-21
erase verb for all entities

MilonPL:
## About the PR

adds the erase verb to all entities

## Why / Balance

funny

## Technical details

the erase method is so fucking shit i need to refactor it upstream at some point but review hell
i commented it as best as I could but big code change so
split into two methods, one for NetUserId and another one for EntityUid
the NetUserId one calls the base EntityUid method and does some extra shit like nuking messages and spawning an observer
ideally this would all be in helper methods instead of whatever fever dream this is
AdminVerbSystem modified to add the erase verb for entities without a mind
**Contents of this entire PR are dual-licensed under MIT, forks that are not AGPL are free to use them.**

## Media

image
image
image

## Breaking changes

backwards compatible
trust 

**Changelog**

:cl:
DELTAVADMIN:
- tweak: The Erase verb is now available on all entities.

### DIFF
--- a/Content.Server/Administration/Systems/AdminSystem.cs
+++ b/Content.Server/Administration/Systems/AdminSystem.cs
@@ -385,78 +385,96 @@ public sealed class AdminSystem : EntitySystem
         }
     }
 
-        /// <summary>
-        ///     Erases a player from the round.
-        ///     This removes them and any trace of them from the round, deleting their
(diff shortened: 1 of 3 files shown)


MilonPL:
I hate this entire code, I might want to move the popup to a separate method and turn change the text based on if the target has a MindContainer or not. It's pissing me off.

PR #2033, opened 2024-10-22
disable the stupid fucking rate limit popup

MilonPL:
## About the PR

TITLE

## Why / Balance

some pro deleted the cvar to disable it and WE DO NOT FUCKING CARE

## Technical details

COMMENTS SHIT OUT

## Breaking changes

**Changelog**

no fun

### DIFF
--- a/Content.Server/Chat/Managers/ChatManager.RateLimit.cs
+++ b/Content.Server/Chat/Managers/ChatManager.RateLimit.cs
@@ -28,7 +28,7 @@ internal sealed partial class ChatManager
 
     private void RateLimitAlertAdmins(ICommonSession player)
     {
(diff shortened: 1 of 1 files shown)


deltanedas (approved):

perryprog:
For what it's worth, I believe the new intended way of disabling this announce was to set ChatRateLimitAnnounceAdminsDelay to be negative which is... kinda weird? I suppose it doesn't really matter though. 🤷

MilonPL:
> For what it's worth, I believe the new intended way of disabling this announce was to set ChatRateLimitAnnounceAdminsDelay to be negative which is... kinda weird? I suppose it doesn't really matter though. 🤷

That sounds like an awful way to handle it, but also this is WAY funnier.

PR #2060, opened 2024-10-27
fix mail jobs not being localized

deltanedas:
## About the PR
right now only happens for alt titles but if other jobs actually got localized it would matter too


MilonPL (approved):

MilonPL:
image
@deltanedas jinxed it

deltanedas:
?? "" ops

MilonPL:
> ?? "" ops

my original idea was to do `?? "Unknown"` ops but I decided to refactor the entire thing after seeing 70 warnings in the mailsystem.cs so that should be done at some point today

PR #2066, opened 2024-10-28
bring back old railing

MilonPL:
## About the PR

title

## Why / Balance

i like it

## Technical details

deltav `railing.yml` based on the upstream `railing.yml` with different sprites and suffixes
no you can't make it in game, why would you want to

## Media

image
image

## Breaking changes

**Changelog**

cant think of a good one and mostly affects mappers anyways

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Structures/Walls/railing.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Structures/Walls/railing.yml
@@ -1,0 +1,39 @@
+- type: entity
+  parent: Railing
+  id: RailingDeltaV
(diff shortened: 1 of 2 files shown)


Unkn0wnGh0st333:
So this lets them both co-exist? I like it

Colin-Tel (approved):
This PR is totally off-the-rails!

IamVelcroboy:
So, this will revert all current rails back to the previous sprite. If mappers wanted the "modern" ones they'll have to re-map them. Not a big deal. I'll add it to my mapping announcement at the end of the week. (I just don't want to do a bunch of little ones.)

PR #2067, opened 2024-10-28
mail refactor

MilonPL:
## About the PR

really painful, but gets rid of like 60 warnings
removed the unnecessary usings and dependencies from a couple of mail related files too
@deltanedas urgent

## Why / Balance

mostly bug fixes, because fucking mail doesn't work

## Breaking changes

the `TryGetMailTeleporterForReceiver` method now takes EntityUid of the receiver instead of the component and returns the `teleporterComponent` and `teleporterUid`
All of the namespaces got changed to `Content.Client.DeltaV.Mail`, `Content.Shared.DeltaV.Mail` and `Content.Server.DeltaV.Mail`.

**Changelog**

:cl:
- fix: Fixed not being able to unlock mail.


dvir001:
Please just move all the files to DeltaV from Nyanotrasen.
Also, the Frontier files to DeltaV name spaces and we can just update it on Frontier side to match after this PR.

This is like 3 forks deep now and will be nice to just be one.

Also we made some fixups for the Mail app code on our side, feel free to take them if you want.

https://github.com/new-frontiers-14/frontier-station-14/pull/1877
https://github.com/new-frontiers-14/frontier-station-14/pull/1885

Content.Server/Nyanotrasen/Mail/MailSystem.cs
```diff
@@ -194,26 +182,25 @@ private void OnAfterInteractUsing(EntityUid uid, MailComponent component, AfterI
...
+            if (HasComp<PdaComponent>(args.Used)) // Can we find it in a PDA if the user is using that?
             {
-                _idCardSystem.TryGetIdCard(args.Used, out var pdaID);
-                idCard = pdaID;
+                _idCardSystem.TryGetIdCard(args.Used, out var pdaId);
+                idCard = pdaId;
             }
-
-            if (HasComp<IdCardComponent>(args.Used)) /// Or are they using an id card directly?
+            else if (HasComp<IdCardComponent>(args.Used)) // Or are they using an id card directly?
```
whatston3: TryGetIdCard could fail - if you had an empty PDA that itself _was_ an ID card (somehow), you'd get left without an ID card with this.

Maybe replace L195 with this if you want to keep the priority the same?
`if (idCard == null && HasComp<IdCardComponent>(args.Used)) // Or are they using an id card directly?`

Alternatively, put the ID card check first, although the combination PDA/ID card's contents would never get used.

Content.Server/Nyanotrasen/Mail/MailSystem.cs
```diff
@@ -105,19 +95,20 @@ public override void Initialize()
...
             base.Update(frameTime);
-            foreach (var mailTeleporter in EntityQuery<MailTeleporterComponent>())
+
+            var query = EntityQueryEnumerator<MailTeleporterComponent, ApcPowerReceiverComponent>();
+            while (query.MoveNext(out var uid, out var mailTeleporter, out var power))
             {
-                if (TryComp<ApcPowerReceiverComponent>(mailTeleporter.Owner, out var power) && !power.Powered)
-                    return;
+                if (!power.Powered)
+                    continue;
```
whatston3: An unpowered mail teleporter should still be able to work - I think the old version here (_using continue rather than return holy shit_) is better.

whatston3 (commented):
Left a few minor points.  Looks good otherwise, thanks for the refactor, and what @dvir001 said makes sense to me.

MilonPL:
> Please just move all the files to DeltaV from Nyanotrasen. Also, the Frontier files to DeltaV name spaces and we can just update it on Frontier side to match after this PR.

Will do o7

> Also we made some fixups for the Mail app code on our side, feel free to take them if you want.

I'll look over them and in a separate PR just so it's less of a headache

Content.Server/Nyanotrasen/Mail/MailSystem.cs
MilonPL: I don't think I get it..? Mail teleporters *should* require power to work, so now if the teleporter is unpowered, it skips that iteration but keeps processing other teleporters.

whatston3: Imagine you have an entity that should function as a mail teleporter without having an ApcPowerReceiverComponent (maybe some admeme teleporter on a mail ship that FTLs in).  That is skipped by this.  The previous implementation checks "if this needs power but doesn't have it, stop everything".  The check makes sense, but you should still move on to the next mail teleporter.

whatston3: Put shortly, you should enumerate over all teleporters, not only those that require power.

MilonPL: OH yeah I get it.

whatston3: Thanks, sorry to pester.

deltanedas (approved):

PR #2069, opened 2024-10-28
Adds e-shotgun

Stop-Signs:
## About the PR
Energy shotgun has been nerfed significantly since its release, re-adds it to the HoS locker at round start.

## Why / Balance
Nukie win-rate has been through the roof recently, and HoS headache rate is at an all time high. Since it was added to upstream 3 months ago it has been nerfed significantly.

## Technical details
N/A

## Media
N/A

## Requirements

- [x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**


Radezolid:
Without auto recharge i doubt it will make a difference, also while at it it might be better to give it to the warden since HOS already has a gun.

Stop-Signs:
> Without auto recharge i doubt it will make a difference, also while at it it might be better to give it to the warden since HOS already has a gun.

how would we feel about moving the x-01 to warden locker and giving HoS the shotgun

looking at you @MilonPL

Radezolid:
> > Without auto recharge i doubt it will make a difference, also while at it it might be better to give it to the warden since HOS already has a gun.
> 
> how would we feel about moving the x-01 to warden locker and giving HoS the shotgun
> 
> looking at you @MilonPL

Shotgun feels more balanced for a warden than the X-01 IMO.

Stop-Signs:
After some extensive testing, I've compared the eshotgun to all of the main guns that security typically have access to, along with the weapons that sec can currently research. all of these tests were shooting the respective gun until it was completely empty, and every shot hit the respective targets.

Kammerer
image

Eshotgun
image

MK58
image

Drozd
image

Universal
image

IK-60
image

Laser Cannon
image

RPG
image


I believe the fact that you need a powered recharger to reload the e-shotgun, keeps it balanced. A kammerer which can easily be reloaded does more damage than the e-shotgun, this alone shows that there shouldn't be any major balance issues.

Radezolid:
> I believe the fact that you need a powered recharger to reload the e-shotgun, keeps it balanced. A kammerer which can easily be reloaded does more damage than the e-shotgun, this alone shows that there shouldn't be any major balance issues.

The whole point of the shotgun is that it SHOULD be stronger than a kammerer because if not why take the worse gun with you?.

MilonPL:
> After some extensive testing, I've compared the eshotgun to all of the main guns that security typically have access to, along with the weapons that sec can currently research. all of these tests were shooting the respective gun until it was completely empty, and every shot hit the respective targets.

These test are kind of inadequate because real people don't stand still during nuke ops. The benefit of e-shotgun over other energy based weapons is that you're able to hit at least *some* of your shots. That being said, I think discussing lasers game balance is a bit silly when an antag can just spend 12 TC to decide they don't want to deal with lasers.

Stop-Signs:
> These test are kind of inadequate because real people don't stand still during nuke ops. The benefit of e-shotgun over other energy based weapons is that you're able to hit at least _some_ of your shots. That being said, I think discussing lasers game balance is a bit silly when an antag can just spend 12 TC to decide they don't want to deal with lasers.

the testing was more to show that even at its max potential, the energy shotgun isn't any more powerful than the current options for sec. Lasers are somewhat of a beast to balance, as its almost the only thing sec get out of research, and they all do the same type of damage. Thankfully i am not here to balance lasers ;).

deltanedas:
the projectile needs to be changed to pass through windows like regular lasers do, otherwise its just a hard to recharge enforcer (which can use beanbags for stunmeta)

MilonPL (changes requested):
the laser bullets are broken and won't go through windows or hit holocarps, needs to have the `Opaque` mask
https://github.com/DeltaV-Station/Delta-v/blob/af2746cd30774f838eea31bbc48bac32d1e3d8af/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml#L1032-L1034

Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
```diff
@@ -1030,8 +1030,7 @@
           bounds: "-0.15,-0.3,0.15,0.3"
         hard: false
         mask:
-        - Impassable
-        - BulletImpassable
```
MilonPL, line 1034: comment these out rather than removing them

MilonPL (changes requested):

MilonPL (approved):

therealDLondon (commented):
I believe at this time we will not be adding the energy shotgun in its current state

therealDLondon:
An addendum to the above:
In its current state, the E-Shotgun is too powerful for being a free round-start weapon, from what I have observed on servers with it enabled and my own tests, being able to 2-Shot an unarmored target from almost 5 tiles away is far too strong. It feeds into the idea of the HOS Being "Just a Secoff but better" which I do not believe fits the vision of security on DeltaV. The X01 is still a good weapon, and the HOS still gets free pickings of the armoury should they need a bigger gun.

Avalon-Proto:
While fair on the whole 2 shots from 5 tiles. Can't *most* shotguns do that to unarmored targets?

PR #2078, opened 2024-10-31
Make powersinks more expensive

Stop-Signs:
## About the PR
Raised powersink price from 6 to 10

## Why / Balance
your essentially buying a blackout event, it should cost more than 6 tc. not to mention the fact that it ends the round 90% of the time cause the crew cant find it. for a round ending and incredibly useful tool such as this it should cost a little more
## Technical details
N/A
## Media
N/A

## Requirements

- [x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x ] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -978,23 +978,23 @@
   categories:
   - UplinkDisruption
 
-- type: listing
-  id: UplinkPowerSink
-  name: uplink-power-sink-name
-  description: uplink-power-sink-desc
-  productEntity: PowerSink
-  discountCategory: usualDiscounts
-  discountDownTo:
-    Telecrystal: 3 # DeltaV - was 4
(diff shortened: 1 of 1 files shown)


Radezolid:
I agree on the price going up, and seeing how easy it is to just wall off, maybe it should emitt a buzz.

deltanedas:
the counterplay to powersinks is extremely easy (unanchor the smeses and start a stationwide hunt for the sink) which completely prevents it from exploding, for 10 tc you can buy an AAC and some c4 to bomb the generator which is MUCH more useful

MilonPL:
I agree with deltanedas on this one. Realistically no one uses them anyways, and engineering gets a huge popup whenever there's a powersink. Plus it's easy to find walled off rooms with cables leading to them on the power monitor now.

PR #2091, opened 2024-11-02
Defib no longer revives dead people

Lyndomen:
## About the PR
Defib no longer revives dead people, but reduces a massive amount of asphyx damage when critical. 
Damage states for basemob (most things) changed to 120 critical, 300 dead. People will be in crit way longer. 

## Why / Balance
Cloning exists still here, and people treat death as nothing more than an inconvenience. Hostage situation? just kill the hostage and move on right? 

## Technical details
mostly YAML, i did butcher the defib C# tho 

## Media
Uhhh I think no?

## Requirements
- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
I mean this breaks the original intent of the defib.

**Changelog**
:cl:
- tweak: Defibrillator no longer revives dead people
- tweak: Defibrillator heals 100 asphyxiation on critical people
- tweak: Mobs with default health go critical at 120 damage (including players)
- tweak: Mobs with default health die at 300 damage (including players)

PR #2093, opened 2024-11-02
make people start rotting in 2 minutes

Lyndomen:
## About the PR
people now start to rot in 2 minutes. same reason as https://github.com/DeltaV-Station/Delta-v/pull/2091

pls dont repoban i have good intent
## Why / Balance
rotting is more fixable 

## Technical details
webedit 

## Requirements

**Changelog**

:cl:
- tweak: Rotting starts two minutes after death

### DIFF
--- a/Content.Shared/Atmos/Rotting/PerishableComponent.cs
+++ b/Content.Shared/Atmos/Rotting/PerishableComponent.cs
@@ -15,7 +15,7 @@ public sealed partial class PerishableComponent : Component
     /// How long it takes after death to start rotting.
     /// </summary>
     [DataField]
-    public TimeSpan RotAfter = TimeSpan.FromMinutes(10);
+    public TimeSpan RotAfter = TimeSpan.FromMinutes(2);
 
     /// <summary>
     /// How much rotting has occured


LunaLettiere:
For the same reasons as [#2091 ](https://github.com/DeltaV-Station/Delta-v/pull/2091) you should include an increase in biomass ratios so you still require sacrifice to correct the difficulty you're instilling, that or require less biomass for a cloning to happen.

Stop-Signs:
>salv refuses to get carp (even though its the most common mob on the most common debris)
>botany refuses to grow plants for chem
>LO cant fire salvs without a mats shortage

basically making oppo both incredibly sought after and a pain to get

Radezolid:
I agree that death is kinda meaningless mechanically (Today in a shift HOS charged into a hostage situation because "they can be revived later") but I don't know how to fix it neither, if you take 5x less time to rot it kinda sucks when you just started the shift for example, or made a simple mistake as a salvage. To put it into context a wreck despawns after around 5 minutes or so, if a salvie makes a dumb or lags and gets killed they are pretty much round removed since you kinda start worrying about a dead salvie after they don't come back from a wreck. Good luck getting EPI to research cloning and delivering to medbay unless you get competent people researching.

beck-thompson:
I feel like this won't actually change stuff like hostage situations, but will make stuff mass bombings much more deadly and "fun" for antags to carry out. If sec rushes in, the hostage situation will be over before two minutes so it wont help

deltanedas:
i agree this has no impact on hostage ops rp or nrp, it just means the spider that killed you in maints is much more likely to RR you

MilonPL:
I doubt this would change much. I agree that 10 minutes is a lot, yet you *still* see people rot. This doesn't actually address the issue of people not respecting life, because for hostage ops it's still plenty of time, and stasis bags stop rotting.

Lyndomen:
ok maybe next I make it so if you die you instantly explode into gibs

PR #2099, opened 2024-11-03
remove deep fryer test

deltanedas:
## About the PR
it doesnt actually test *anything*

## Why / Balance
makes test take some tiny amount longer and provides no benefit over yml linter along with an extra dummy prototype to maintain


MilonPL:
it tests that the deep fryer system exists, how will you notice that it's gone now!! smh

deltanedas:
help i deleted deep frier and the test didnt stop me

PR #2102, opened 2024-11-03
disable eating brains

deltanedas:
## About the PR
title

## Why / Balance
borgs arent allowed to remember what led to their death, so this isnt an issue
as such eating brains is only done by shitters or people that are used to wizden and think they have to destroy the brain to avoid being snitched on

## Technical details
requires that you have 20 stomachs to stomach eating a brain :trollface::trollface::trollface:

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: You can no longer eat brains, they aren't allowed to remember who killed them anyway!

### DIFF
--- a/Resources/Prototypes/Body/Organs/human.yml
+++ b/Resources/Prototypes/Body/Organs/human.yml
@@ -50,6 +50,8 @@
   - type: Examiner
   - type: BlockMovement
   - type: BadFood
+  - type: Food
+    requiredStomachs: 20 # DeltaV: Prevent eating brains as borgs don't remember what happened to them
   - type: Tag
     tags:
       - Meat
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Body/Organs/human.yml
```diff
@@ -50,6 +50,8 @@
   - type: Examiner
   - type: BlockMovement
   - type: BadFood
+  - type: Food
+    requiredStomachs: 20 # DeltaV: Prevent eating brains as borgs don't remember what happened to them
```
MilonPL: the comment should be on line 53 not 54 because we're adding the component rather than changing a value
```suggestion
  - type: Food # DeltaV: Prevent eating brains as borgs don't remember what happened to them
    requiredStomachs: 20 
```

MilonPL (changes requested):

Resources/Prototypes/Body/Organs/human.yml
deltanedas: real

MilonPL (approved):

PR #2125, opened 2024-11-05
Raise the execution damage multiplier.

Radezolid:
## About the PR
Raised the execution damage to x18 (a x2 increase over current multiplier) of the projectile.
Requested by a [WYCI](https://discord.com/channels/968983104247185448/1011259887759659049/1303476684132978788), While I don't fully agree with the PR, just putting their suggestion here.

## Why / Balance
It really doesn't change much for regular executions but the fact that you instantly die instead of going into critical state.
There is a problem with balance when used with beanbags though, if they do it twice on a healthy person it leaves them at 360 Blunt damage, 40 short of gibbing.

## Technical details
Modified a value in `ExecutionSystem.cs`.

## Media

None.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Executions now do x18 damage instead of x9.

### DIFF
--- a/Content.Server/Execution/ExecutionSystem.cs
+++ b/Content.Server/Execution/ExecutionSystem.cs
@@ -41,7 +41,7 @@ public sealed class ExecutionSystem : EntitySystem
     [Dependency] private readonly GunSystem _gunSystem = default!;
 
     private const float GunExecutionTime = 6.0f;
-    private const float DamageModifier = 9.0f;
+    private const float DamageModifier = 18.0f; // DeltaV - Raised damage x2 over the original 9.0f
 
     /// <inheritdoc/>
     public override void Initialize()


MilonPL:
I don't think this is going to change much. If the whole reason behind it is "I don't want to be shot twice", you can just ghost.
Going straight from healthy to dead isn't going to play the deathgasp too, which is arguably a bad thing?

ThataKat:
Med will cry when the secoff comes in with >600 slash because the ninja caught them in maints and executed them when they went crit.
I personally think it's fine that most weapons don't go straight to dead even if it might go against the perception of how it "should" end.

Radezolid:
I know, that's why I stated that I disagree with the PR, but it was easy to make and someone requested it, so i just voiced their opinion.

PR #2141, opened 2024-11-07
stock market refactor

deltanedas:
## About the PR
- renamed StockCompanyStruct to StockCompany, it didnt need Struct in its name. classes dont have Class in their name etc
- made MarketChange a struct
- instead of 2 copy paste functions to buy and sell, just TryChangeStocks that handles negatives to sell
- log the user instead of loader since you can steal pdas


MilonPL:
FUCK it only took 2 days (why didn't you stop me)

MilonPL:
@deltanedas while you're at it, something that probably should've been done but I didn't think about, is also playing the deny sound if TryChangeStocks fails, to give player feedback in case we can't afford it or we don't have enough stocks to sell

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
MilonPL, line 1: Shared.Audio and Shared.Player unused

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -22,13 +22,13 @@ namespace Content.Server.DeltaV.Cargo.Systems;
...
+    [Dependency] private readonly AccessReaderSystem _access = default!;
     [Dependency] private readonly CargoSystem _cargo = default!;
     [Dependency] private readonly IAdminLogManager _adminLogger = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
     [Dependency] private readonly ILogManager _log = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
-    [Dependency] private readonly IdCardSystem _idCardSystem = default!;
+    [Dependency] private readonly IdCardSystem _idCard = default!;
     [Dependency] private readonly SharedAudioSystem _audio = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
```
MilonPL: remove

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -64,38 +64,30 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
         if (args is not StockTradingUiMessageEvent message)
             return;
 
+        var user = args.Actor;
         var companyIndex = message.CompanyIndex;
-        var amount = (int)message.Amount;
-        var station = ent.Comp.Station;
+        var amount = message.Amount;
         var loader = GetEntity(args.LoaderUid);
         var xform = Transform(loader);
```
MilonPL: remove

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -64,38 +64,30 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
...
 
         // Attempt to retrieve ID card from loader
-        IdCardComponent? idCard = null;
-        if (_idCardSystem.TryGetIdCard(loader, out var pdaId))
-            idCard = pdaId;
+        _idCard.TryGetIdCard(loader, out var idCard);
 
         // Play deny sound and exit if access is not allowed
-        if (idCard == null || !_accessSystem.IsAllowed(pdaId.Owner, ent.Owner, access))
+        if (idCard == null || !_access.IsAllowed(idCard.Owner, ent.Owner, access))
```
MilonPL: idcard can never be null here, no need to check
no need for .Owner either

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -64,38 +64,30 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
...
 
-        if (!TryComp<AccessReaderComponent>(ent.Owner, out var access))
+        if (!TryComp<AccessReaderComponent>(ent, out var access))
             return;
 
         // Attempt to retrieve ID card from loader
-        IdCardComponent? idCard = null;
-        if (_idCardSystem.TryGetIdCard(loader, out var pdaId))
-            idCard = pdaId;
+        _idCard.TryGetIdCard(loader, out var idCard);
```
MilonPL: trygetidcard can still fail so it should probably just return then

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -110,109 +102,85 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
...
+        if (amount > 0)
+        {
+            // Buying: see if we can afford it
+            if (bank.Balance < totalValue)
+                return false;
+        }
+        else
+        {
+            // Selling: see if we have enough stocks to sell
+            var selling = -amount;
```
MilonPL: remove

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -110,109 +102,85 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
...
+            if (currentOwned < amount)
```
MilonPL: wont this not work for selling? you're comparing a negative number, should do Math.Abs

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
```diff
@@ -110,109 +102,85 @@ private void OnStockTradingMessage(Entity<StockTradingCartridgeComponent> ent, r
...
         if (!TryComp<StationBankAccountComponent>(station, out var bank))
             return false;
 
         var company = stockMarket.Companies[companyIndex];
-        var totalValue = (int)Math.Round(company.CurrentPrice * amount);
-
-        // See if we can afford it
-        if (bank.Balance < totalValue)
-            return false;
+        var totalValue = (int)Math.Round(company.CurrentPrice * (float) amount);
```
MilonPL: redundant cast to float, remove

MilonPL (changes requested):

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
deltanedas: its meant to be used just wrong var name :trollface:

Content.Server/DeltaV/Cargo/Systems/StockMarketSystem.cs
deltanedas: meant to use selling

PR #2147, opened 2024-11-08
A Ninja Adjustment

Avalon-Proto:
## About the PR
Ninjas have always had trouble dealing with psionics, so, they are now insulated. Also their gloves are now slightly better than standard thieving gloves at stealing, as it can make interesting scenarios allowing them to cloak and steal

## Why / Balance
Everyone and their mother with mass sleep likes to valid hunt, so lets remove their option to ninja hunt. And thieving gloves so that ninjas can pull more interesting shenanigans

## Technical details
Changed a number and added a component
## Requirements

## Breaking changes

**Changelog**

🆑 
 - Tweak: Ninja helmets have been imbued with anti psionic sigils, and they are sneakier than ever with the ability to steal silently

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Hands/gloves.yml
+++ b/Resources/Prototypes/Entities/Clothing/Hands/gloves.yml
@@ -226,7 +226,7 @@
     fiberMaterial: fibers-nanomachines
   - type: FingerprintMask
   - type: Thieving
-    stripTimeReduction: 1
+    stripTimeReduction: 3 #DeltaV- Made them better for thieving activities
     stealthy: true
   - type: ToggleClothing
     action: ActionToggleNinjaGloves
--- a/Resources/Prototypes/Entities/Clothing/Head/helmets.yml
+++ b/Resources/Prototypes/Entities/Clothing/Head/helmets.yml
@@ -175,6 +175,7 @@
     sprite: Clothing/Head/Helmets/spaceninja.rsi
   - type: Clothing
     sprite: Clothing/Head/Helmets/spaceninja.rsi
+  - type: TinfoilHat #DeltaV- No more sleeping ninjas
   - type: Tag
     tags:
     - WhitelistChameleon


Radezolid:
Question, as far as i know tin foil hats combust once after someone tries to use powers on the person. Would the ninja's helmet combust too?

Avalon-Proto:
I am honestly not sure! I couldnt find the comp to modify it anywhere

Stop-Signs:
Ninja is already really powerful, so i dont necessarily agree with yet another buff, i understand that mass sleep validhunting exists but if sec is getting tortured by a stun and steal gun ninja it might be worth it to have those people.

Avalon-Proto:
While true, also consider this- secoff in pairs will destroy a ninja if they are out to kill since the stun has a cool down and ninja has 0 armor. This is just to stop those solo players from getting a free kill by sleeping the ninja or mind swapping them etc

Lyndomen:
Idk it seems to me like the problem here is with those psionics, not with the ninja

Avalon-Proto:
I mean, just seems like an easier solution to protect the ninja from psionics rather than try and mess with psychic code

Lyndomen:
At some point tho this problem exists for every solo antagonist. If it's a problem, it is much better to address it at the root rather than give every solo antag free insulation.

Avalon-Proto:
What do you suggest instead then? Id rather not take away the possible tools for antags or sec to get useful psionics like sleep or zap, since both can be used to help with goals and the like

Lyndomen:
> What do you suggest instead then? Id rather not take away the possible tools for antags or sec to get useful psionics like sleep or zap, since both can be used to help with goals and the like

I guess I don't see it as an issue myself. If security is using it, okay fair game. Ninja also has a one button stunlock, and could steal psionic insulation if they really want. If crew does it, they get bwoinked for validhunting, if it's applicable.

Radezolid:
I'm against this, as a ninja enjoyer, they are already absolutely broken having to fear psionics is cool but I do agree on the no counter play and window for valid hunting it gives. This does look like a community issue rather than mechanical.
Edit: My suggestion (WYCI) is making psionic powers be at least a small do after, let people react.

deltanedas (changes requested):
even nukies can get mass slept, doesnt make sense that ninja of all antags gets immunity

and would need psionic insulation refactor anyway so ninja helmet doesnt burn up at high glimmer

Avalon-Proto:
Yeah just closing this one tbh I need tah figure it out

PR #2156, opened 2024-11-09
fix metempsychosis

### LINKED ISSUE #2153: metem machine sucks
## Description
Dsymorphia machine supposed to give consequences to people dying, but leading consequences is /ghost 

**Reproduction**
Die, forget to tick uncloneable, come out as Oni, /ghost

**Screenshots**

**Additional context**

MilonPL:
## About the PR

somewhat fixes https://github.com/DeltaV-Station/Delta-v/issues/2153
it will always use your player character on the first time then work like it did before with karma
if you have to be cloned twice during one shift that's a skill issue 

## Technical details

holy fucking shitcode
fixed metem not respecting the nonhumanoid pool but it was never working in the first place apparently
moved everything to deltav, file scoped namespaces, fixed the test being shit, fixed the components being shit too, use protoid
instead of having a separate system and a huge ass method inside of the upstream cloning system with like 20 usings and no comments, moved everything to CloningSystem.Metempsychosis.cs

## Breaking changes

**Changelog**

:cl:
- tweak: The metempsychosis will always use your player character on the first cloning attempt.
- fix: Fixed metempsychosis never selecting non-humanoid characters.


Content.Server/DeltaV/Cloning/MetempsychoticMachineComponent.cs
```diff
@@ -0,0 +1,21 @@
...
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.DeltaV.Cloning;
+
+[RegisterComponent]
+public sealed partial class MetempsychoticMachineComponent : Component
+{
+    /// <summary>
+    ///     Chance you will spawn as a humanoid instead of a non humanoid.
+    /// </summary>
```
deltanedas, line 11: add bit about no karma

Content.Server/Cloning/CloningSystem.cs
```diff
@@ -86,6 +73,8 @@ public override void Initialize()
         {
             base.Initialize();
 
+            _sawmill = _log.GetSawmill("cloning"); // DeltaV
```
deltanedas: why do this instead of Log.Error which uses system.cloning

deltanedas (changes requested):

Content.Server/Cloning/CloningSystem.cs
MilonPL: fair, leftover from another implementation before i had the genius idea to do this
(i also originally wanted to use cloning.metem but whatever)

deltanedas (approved):

PR #2171, opened 2024-11-11
bluespace fire extinguisher 

MilonPL:
## About the PR

adds a self-refilling bluespace fire extinguisher that can hold 200u of water (twice the standard one)
refills at a rate of 10u every 10 seconds
researched in T2 industrial, same as holofans, because it's meant to be used by atmos ig
takes 10 steel, 2.5 silver, 5 plasma and 2 bluespace to make

## Why / Balance

fire extinguishers are underused and this is fun

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl: MilonPL, dotCatshark
- add: Added self-refilling bluespace fire extinguishers! They can be unlocked in the T2 industrial research!


perryprog:
Would recommend giving this as a researched module to borgs as well given the current borg propulsion method is just a fire extinguisher itself, which tends to run out very quickly. (The better solution is that borgs get a proper jetpack module, but this is presumably difficult to implement as they'd have no way to refill the jetpack.)

dvir001:
@MilonPL Can you empty it into a chem master for infinite water right now?

MilonPL:
Good point it probably shouldn't be drainable. But I'm not sure how much of an issue that would be considering how all chemlabs have a sink anyways, and it's easier to get a water tank from maints, than make an expensive T2 bluespace tech just to use it as a portable sink.

Avalon-Proto:
Do hope to see this soon. Though making a Borg version would also be very nice

deltanedas (approved):

PR #2190, opened 2024-11-13
stamina damage resist

MilonPL:
## About the PR

adds a resist to incoming projectile stamina damage
most hardsuits provide a 10% protection, sec and syndicate hardsuits provide 70% protection, with exceptions for raid suits, jugg, elite, hos, warden, etc, whatever makes sense to be higher
melee damage like batongs either takes the blunt damage coefficient or the melee stamina damage coefficient on the armor, whichever one provides more resist

## Why / Balance

stumeta bad and i'm cooking

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Hardsuits and armor vests now provide protection against incoming stamina damage.

### DIFF
--- a/Content.Shared/DeltaV/Damage/StaminaSystem.Resist.cs
+++ b/Content.Shared/DeltaV/Damage/StaminaSystem.Resist.cs
@@ -60,7 +60,13 @@ public sealed partial class StaminaSystem
     /// <summary>
     /// Applies stamina damage from melee attacks with armor resistance calculations
     /// </summary>
-    public void TakeStaminaDamageWithMeleeCoefficient(EntityUid target, float damage, StaminaComponent? stamina = null, EntityUid? source = null, EntityUid? with = null, bool visual = true, SoundSpecifier? sound = null)
+    public void TakeMeleeStaminaDamage(EntityUid target,
+        float damage,
+        StaminaComponent? stamina = null,
+        EntityUid? source = null,
+        EntityUid? with = null,
+        bool visual = true,
+        SoundSpecifier? sound = null)
     {
         if (!Resolve(target, ref stamina))
             return;
@@ -74,7 +80,13 @@ public sealed partial class StaminaSystem
     /// <summary>
     /// Applies stamina damage from projectiles with armor resistance calculations
     /// </summary>
-    public void TakeStaminaDamageWithProjectileCoefficient(EntityUid target, float damage, StaminaComponent? stamina = null, EntityUid? source = null, EntityUid? with = null, bool visual = true, SoundSpecifier? sound = null)
+    public void TakeProjectileStaminaDamage(EntityUid target,
+        float damage,
+        StaminaComponent? stamina = null,
+        EntityUid? source = null,
+        EntityUid? with = null,
+        bool visual = true,
+        SoundSpecifier? sound = null)
     {
         if (!Resolve(target, ref stamina))
             return;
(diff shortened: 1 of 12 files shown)


deltanedas (changes requested):
i like an old prs stamina damage specifiers better, allows for more control over the different types of stuns

also the logic for not having an examine is terrible, if theres no examine the average player wont know it exists and the pro gamers will just grep the source anyway giving them an advantage

MilonPL:
> also the logic for not having an examine is terrible, if theres no examine the average player wont know it exists and the pro gamers will just grep the source anyway giving them an advantage

yes real i'm just coping with stamina not being a damage type 

i think that melee weapons should use the stamina resist based on blunt damage coefficient rather than having a separate value, and then the StaminaDamageCoefficent only works for projectiles (or maybe takes the higher value out of the two?)

Lyndomen:
I had started work on this yesterday to do a similar thing, by changing the YAML of all stamina weps to use bluntdamagestamina, which is then reduced by armor. Was advised this was maints hell 

I also made shockdamagestamina exist, so shock damage can similarly do stam damage. I moved baton to shock, and the goal is blunt = good at stunning unarmored and shock = good at stunning armored but still slightly worse than it is now. 

https://github.com/DeltaV-Station/Delta-v/commit/3749fcadc9cf4d962ed41b1b4d1fb02e835d0fab

I guess some guidance about what is best would be great

ThataKat:
Particulars might have to be tweaked but holy based. No more stunning nukies with rubbers?

Stop-Signs:
I really appreciate someone taking the time to do this. I do agree with delta in that you should be able to examine to see the stamina res, else it's only going to be abused by the "pros" and noobs will have a harder time understanding.

deltanedas (dismissed):
stunmeta is dead piemeta is new king

deltanedas:
change cl to reflect batong change

MilonPL:
> change cl to reflect batong change

hold on i was cooking

MilonPL:
conflicts no longer that evil

Content.Shared/DeltaV/Damage/StaminaSystem.Resist.cs
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+
+        return coefficient;
+    }
+
+    /// <summary>
+    /// Applies stamina damage from melee attacks with armor resistance calculations
+    /// </summary>
+    public void TakeStaminaDamageWithMeleeCoefficient(EntityUid target, float damage, StaminaComponent? stamina = null, EntityUid? source = null, EntityUid? with = null, bool visual = true, SoundSpecifier? sound = null)
```
deltanedas: TakeMeleeStaminaDamage for the love of god

Content.Shared/DeltaV/Damage/StaminaSystem.Resist.cs
```diff
@@ -0,0 +1,87 @@
...
+        var coefficient = GetMeleeCoefficient(target);
+        var finalDamage = damage * coefficient;
+
+        TakeStaminaDamage(target, finalDamage, stamina, source, with, visual, sound);
+    }
+
+    /// <summary>
+    /// Applies stamina damage from projectiles with armor resistance calculations
+    /// </summary>
+    public void TakeStaminaDamageWithProjectileCoefficient(EntityUid target, float damage, StaminaComponent? stamina = null, EntityUid? source = null, EntityUid? with = null, bool visual = true, SoundSpecifier? sound = null)
```
deltanedas: .

deltanedas (changes requested):

Content.Shared/DeltaV/Damage/StaminaSystem.Resist.cs
MilonPL: :trollface:

deltanedas (approved):

deltanedas:
nukies are back

PR #2191, opened 2024-11-13
make disablers hitscan

MilonPL:
## About the PR

made the disabler use hitscan
reduced the damage from 30 to 26, still needs 4 hits, but slightly weaker
meant to be merged with https://github.com/DeltaV-Station/Delta-v/pull/2190
mostly checking if people like it, we can revert it anyways, but if it turns out to be good, could probably give it to other disabler guns like the x-01

## Why / Balance

disablers and projectiles in general are a massive joke when you can just avoid them
by buffing the disablers security might prioritize them over less lethal means
for now only affects the practice and regular disablers, disabler SMG stays the same because of it's firerate

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Disablers are now hitscan.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Weapons/Guns/Projectiles/hitscan.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Weapons/Guns/Projectiles/hitscan.yml
@@ -13,3 +13,35 @@
   impactFlash:
     sprite: Objects/Weapons/Guns/Projectiles/projectiles.rsi
     state: impact_omni
+
+- type: hitscan
+  id: Disabler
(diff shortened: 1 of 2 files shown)


Unkn0wnGh0st333:
I mean disablers being projectiles kinda gives the person an advantage. Plus security also has flashers which will just stop a person in their tracks. (Counterargument sunglasses)

PubliclyExecutedPig:
disabler smg's are unchanged, is that intentional?

MilonPL:
> for now only affects the practice and regular disablers, disabler SMG stays the same because of it's firerate

MilonPL:
fuck

PR #2207, opened 2024-11-15
The Removal Of War

Radezolid:
## About the PR
Removed the war declarator from the commander's starting gear.

## Why / Balance
It makes a LRP gamemode kill whatever it had left of RP and it just becomes a big NRP TDM round.

## Technical details
Commented out 2 lines in a YAML.

## Media
None.

## Requirements

## Breaking changes

Hopefully none.

**Changelog**

:cl:
- remove: Removed the war declarator from the nuclear operatives.

### DIFF
--- a/Resources/Prototypes/Roles/Antags/nukeops.yml
+++ b/Resources/Prototypes/Roles/Antags/nukeops.yml
@@ -85,8 +85,8 @@
   equipment:
     neck: SyndicateWhistle
     outerClothing: ClothingOuterHardsuitSyndieCommander
-  inhand:
-  - NukeOpsDeclarationOfWar
+#  inhand: # DeltaV - Removal Of War.
+#  - NukeOpsDeclarationOfWar
 
 #Nuclear Operative Medic Gear
 - type: startingGear


Stop-Signs:
While I agree war needs some adjustment to prevent every tider passenger from immediately running into sec and taking all the firearms. I also disagree with removing it entirely, war is just such a fun game-mode once you get past the tide of "give me a weapon" people.

Lyndomen:
Warops was an experiment, and it's ok to say it doesn't really work on mrp. If there are anydments suggested/, implemented that would be sick but wyci demon

Monotheonist:
I'm going to agree with Lyndomen - it was an experiment, and if it doesn't work out, then we have a conclusive reason as to why. If there's amendments, suggest, and perhaps implement.
I am of the opinion that war should not last forever.

ThataKat:
I agree, it was absent for a long time for a reason. People have fun with it, but primarily for mechanical reasons, and combat. Not for the good rp. As an MRP fork, I think we tried it and it doesn't fit

CybersunOperative:
The original Contrib that added War Ops to Delta V has voiced that it was never meant to stay.

Direction has echoed similar sentiments, and with this PR, we wish to see War Ops removed.

Direction Approved :)

deltanedas (approved):

PR #2219, opened 2024-11-17
Buff beanbags shells

Radezolid:
## About the PR
Made the beanbag stamina damage as it was before the upstream nerf. My goal is to bring back the usefulness of the double barrel shotgun as a tool to stop bar fights and it being a better weapon for syndie bartenders.

## Why / Balance
Armored targets now have resistance to stamina damage, so this change makes the bartender be able to actually use their shotgun on disruptive people in the bar.

Balance tested with a enforcer without missing a shot:
Blood red took 70 stamina damage.
Juggernaut took 35 stamina damage.
Durathread vest goes down in 3 shots.
Plate carrier goes down in 4 shots.

## Technical details
1 line change of YAML.

## Media
None.

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Beanbags hits hard again, bartenders rejoice!

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Ammunition/Projectiles/shotgun.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Ammunition/Projectiles/shotgun.yml
@@ -26,7 +26,7 @@
       types:
         Blunt: 10
   - type: StaminaDamageOnCollide
-    damage: 40 # 3 hits to stun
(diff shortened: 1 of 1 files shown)


deltanedas (changes requested):
this requires you to land both shots with no delay, should be changed to 55 or so to be more forgiving

deltanedas (approved):

PR #2237, opened 2024-11-19
Slight SLP-57 nerf

BlitzTheSquishy:
## About the PR

Reduced the firerate of the SLP-57 to 5.5

## Why / Balance

Bring it in line with both the Pollock and MK58, despite it being less accurate, it can still put bullets down range faster and still pretty accurately, over its counterparts

## Requirements

## Breaking changes

**Changelog**


MilonPL:
After testing it in game, I don't think it's accurate enough for it to matter, especially considering 6 is the base fire rate value for pistols.

BlitzTheSquishy:
> After testing it in game, I don't think it's accurate enough for it to matter, especially considering 6 is the base fire rate value for pistols.

SLP-57 theres a pretty big difference between how fast they put bullets down range
because Pollock and MK58 use 5, not 6

PR #2244, opened 2024-11-20
Security magboot change

Radezolid:
## About the PR
Removed the ability for them to be toggled off, but gave them a slot for knives or anything with the `ConcealCarry` tag.

## Why / Balance
People were complaining about sec using them all the time, now they will be slowed constantly if wearing them.

## Technical details
Modified the RSI and the YAML.

## Media

https://github.com/user-attachments/assets/a020add3-c9c7-4a16-8df3-9805acbd4fde

## Requirements

## Breaking changes

Hopefully nothing.

**Changelog**

:cl:
- tweak: NT added a slot for sidearms and knives for the security magboots, to squeeze that in the boots now lack the ability to be turned off.


MilonPL:
Walking is still slower than just wearing magboots, and you still can't walk over space lube, so mechanically there's pretty much no reason not to have them on all the time. Officers will just keep them in their bag anyways and switch when needed, so this doesn't achieve anything, and I don't think this is really an issue either way.

image

PR #2245, opened 2024-11-20
fix deck the halls in jukebox

MilonPL:
## About the PR

mixed down to mono, removed the one random extra line
i shouldn't have to do this yet here we are


Lyndomen:
Built different

therealDLondon (approved):
Oh the humanity

PR #2253, opened 2024-11-21
Remove the storage implant from the uplink

Radezolid:
## About the PR
This PR removes the storage implant from the uplink.

## Why / Balance
It started as simply changes to make it unable to store high risk items, but then containers happend and my smooth brain couldn't figure that out.
Storage implant with high risk item = boring, no RP.

## Technical details
YAML changes

## Media
It gone.
image

## Requirements

## Breaking changes

None.

**Changelog**
:cl:
- remove: Removed the storage implant from the uplink!

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -1261,24 +1261,24 @@
 
 # Implants
 
-- type: listing
-  id: UplinkStorageImplanter
-  name: uplink-storage-implanter-name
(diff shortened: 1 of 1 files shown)


JustAnOrng:
i think storage is a very valid implant, only due to its cost. AT LEAST half of all of your TC is a big ask.
Plus, it's like. 3x3? you can store, like, barely anything.

Lyndomen:
i personally feel that if we can blacklist objective items and storage items from being in the implant, this is fine to keep with a significant reduction to its cost

PR #2261, opened 2024-11-22
disable funny stack price test

deltanedas:
## About the PR
remove the test that has never had a true positive in deltav history, only false positives

## Technical details
1 line ops

## Requirements
:trollface:


MilonPL (approved):
real
DeleteAllThenGhost next

PR #2263, opened 2024-11-22
add debug logging to entitywhitelist

MilonPL:
## About the PR

title

## Why / Balance

image

### DIFF
--- a/Content.Shared/Whitelist/EntityWhitelistSystem.cs
+++ b/Content.Shared/Whitelist/EntityWhitelistSystem.cs
@@ -115,7 +115,6 @@ public sealed class EntityWhitelistSystem : EntitySystem
             return false;
 
         // Begin DeltaV
-        Log.Debug($"Checking whitelist pass for entity {ToPrettyString(uid)}");
         var isValid = IsValid(whitelist, uid);
         Log.Debug($"Whitelist validation result for entity {ToPrettyString(uid)}: {isValid}");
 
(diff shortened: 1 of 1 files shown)


Content.Shared/Whitelist/EntityWhitelistSystem.cs
```diff
@@ -11,12 +12,14 @@ public sealed class EntityWhitelistSystem : EntitySystem
     [Dependency] private readonly SharedRoleSystem _roles = default!;
     [Dependency] private readonly TagSystem _tag = default!;
 
+    private ISawmill _sawmill = default!; // DeltaV
```
deltanedas: its a system just use Log.Debug

MilonPL: won't Log.Debug kill performance when doing a ton of terminal IO? i'm not sure if it's actually more expensive compared to a single sawmill, but that is what's been happening whenever we got invalid component or tag names

MilonPL: update: it kills TPS just as much :trollface:

Content.Shared/Whitelist/EntityWhitelistSystem.cs
```diff
@@ -113,6 +114,20 @@ public bool IsWhitelistPass(EntityWhitelist? whitelist, EntityUid uid)
         if (whitelist == null)
             return false;
 
+        // Begin DeltaV
+        Log.Debug($"Checking whitelist pass for entity {ToPrettyString(uid)}");
```
deltanedas: redundant because of the log right after it

MilonPL: real

deltanedas (approved):

PR #2264, opened 2024-11-22
Revert "Ports Bloodstains And Footprints from EE"

deltanedas:
who couldve seen this coming


Emily9031:
Guess since I made the comment about performance in the first place, I wanna say a couple things. I do think that while performance from this was definitely impacted quite a bit, I still think it was at least _usable._ I'd guess it was somewhere in the range of feeling like playing at 20-30fps, which while it sucks for high intensity combat moments, isn't a complete dealbreaker for me. And I kinda worry that if this feature is just reverted and removed, it will end up not being added back at all.

I'd personally prefer if the feature is kept in, but maybe just have an issue made about it needing some optimization. At least unless more players than just me start reporting performance issues from it.

deltanedas:
> it will end up not being added back at all.

it can probably be fixed by checking for stains on the same tile then not making more, but effort

MilonPL:
>it can probably be fixed by checking for stains on the same tile then not making more

i'll look at it later, but if it modifies too much code or takes too much effort to implement, we'll just revert

dotCatshark:
> Guess since I made the comment about performance in the first place, I wanna say a couple things. I do think that while performance from this was definitely impacted quite a bit, I still think it was at least _usable._ I'd guess it was somewhere in the range of feeling like playing at 20-30fps, which while it sucks for high intensity combat moments, isn't a complete dealbreaker for me. And I kinda worry that if this feature is just reverted and removed, it will end up not being added back at all.
> 
> I'd personally prefer if the feature is kept in, but maybe just have an issue made about it needing some optimization. At least unless more players than just me start reporting performance issues from it.

I agree that it should be kept in rather than removed. I don't think it's enough of a big deal for a feature to cause lag that is clearly not gamebreaking outside of what is seemingly just edge cases for it to be completely removed. If it gets worse later or if more people complain about it actually **making the game unplayable** we can remove it but in the meantime it's kind of just removing a cool feature that has apparently not caused very many problems up to this point. And yeah, if it gets removed it probably won't be added back in, going based on the track record of things that are """pending rework.""" Also, the coders themselves have said themselves that optimization IS possible.

Please do not merge this PR.

Unkn0wnGh0st333:
Please optimize... don't 1984 this gangsta addition my super amazing awesome team of cool people.

Proxysseia:
runs fine on my machine even with high amount of decals and my pretty iffy specs. makes combat spaces look very pretty and was overall a great addition to DV. would be sad to see it go.

HTMLSystem:
I'd say my computer isn't the best either, hell I can't even run replays, but I've had no problems with the bloodstains or footprints at all.

deltanedas:
there is a fix pr open itll be used instead when its done

PR #2270, opened 2024-11-22
limit footprints to 2 per tile

MilonPL:
## About the PR

title
adds logic to the FootPrintsSystem.cs to track the number of footprints and limit it to 2 per tile
doesn't seem like a lot but still looks good and 1 per tile looks shit

image

## Why / Balance

5 tps ops

## Breaking changes

**Changelog**

:cl:
- tweak: Limited footprints to two per tile.


Lyndomen:
Is it worth removing the timed expiration of these prints?

batydoo:
> Is it worth removing the timed expiration of these prints?

As a regular Janitor player I really really don't like the current state of footprints. I think its a really cool idea to have messes spread about the place, but I don't like how they disappear by themselves and remove liquid from puddles making it so if people are walking on messes the puddle evaporates without being cleaned up!! its quite annoying to play as janitor with this. it doesn't help that you have to click on each footprint individually to clean them up combined with the fact they disappear automatically means i never clean them up! I just ignore them which as someone trying to keep this place tidy is really frustrating.

Lyndomen:
> > Is it worth removing the timed expiration of these prints?
> 
> As a regular Janitor player I really really don't like the current state of footprints. I think its a really cool idea to have messes spread about the place, but I don't like how they disappear by themselves and remove liquid from puddles making it so if people are walking on messes the puddle evaporates without being cleaned up!! its quite annoying to play as janitor with this. it doesn't help that you have to click on each footprint individually to clean them up combined with the fact they disappear automatically means i never clean them up! I just ignore them which as someone trying to keep this place tidy is really frustrating.

So from what I hear, we could either:
Make footprints not expire OR
Make footprints not take reagents from puddles (and keep them expiring)

Content.Server/_EE/FootPrint/FootPrintsSystem.cs
```diff
@@ -21,6 +25,16 @@ public sealed class FootPrintsSystem : EntitySystem
...
+    [Dependency] private readonly SharedMapSystem _mapSystem = default!; // DeltaV
+
+    // DeltaV - Max amount of footprints per tile
+    // Not stored on the component because its convenient here
+    private const int MaxFootprintsPerTile = 2;
+
+    /// <summary>
+    ///     DeltaV: Dictionary tracking footprints per tile using tile coordinates as key
+    /// </summary>
+    private readonly Dictionary<Vector2i, Queue<EntityUid>> _footprintsPerTile = new();
```
metalgearsloth, line 38: If you don't store this in a component it gets nuked on any downstreams trying persistence.

Also obviously not storing the grid itself so breaks on multi-grid.

MilonPL: fuck you're right, i forget that persistence is even a thing

PR #2276, opened 2024-11-24
fix borgs not having names

deltanedas:
## About the PR
job borgs now have names
newly built borgs are unaffected

**Changelog**
:cl:
- fix: Fixed borgs not having names.


MilonPL (approved):
real

PR #2294, opened 2024-11-27
Home run bat discount

Stop-Signs:
## About the PR
Added it to the common discounts category, reduced its discounted value from to to 6

## Why / Balance
It had a discount price but not a discount rarity, price reduced as the weapon is not used, it should encourage a reason to buy this over arguably better alternatives

## Requirements

## Breaking changes

**Changelog**


therealDLondon:
I would like something written in the "Why/Balance" section, as opposed to "why not"

Stop-Signs:
> I would like something written in the "Why/Balance" section, as opposed to "why not"

there

deltanedas (approved):

PR #2300, opened 2024-11-27
Tranquilizer adjustment

Stop-Signs:
## About the PR
Adjusted tranquilizers so they can actually sleep people, raised the amount of shots it takes to sleep people with them.

## Why / Balance
Why: @MilonPL  requested it, also tranqs have been the objectively worst ammo type for a hot minute

Balance: Being honest i went into my dev env thinking this would be super OP. however i was pleasantly surprised by how fair it is. unless you hit 3 shots within >4 seconds then the nocturine will metabolize quick enough to mitigate any effects. and even if you do hit all 3 shots, the sleep does not last very long at all, only about 2 seconds.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Tranquilizer rounds once again can sleep targets, however you must hit more shots to do so.


deltanedas:
https://youtu.be/cpwM5MTNSgc

therealDLondon:
Tranqs dont need a buff, they're fine as is.
We arent returning to a new stun meta we JUST left one

deltanedas:
as-is this will just be the go-to nukie hunting ammo, should be using the same system as syringe gun so theres some counter to it, or not piercing hardsuits

PR #2305, opened 2024-11-28
make doorjack 4 tc

deltanedas:
## About the PR
4tc and 2 on discount

## Why / Balance
jaws of life and susbox:
- arent controlled items
- have no charges
- make much less suspicious sounds
- are 2 tc

currently if you hear a door getting emagged you first think ninja because of how rare doorjack is from being outclassed, now it has a good chance of being seen again

at 4 tc doorjack is vaguely competitive, do you want the sustainability of prying doors open or the quick burst of 3 doors in succession

## Technical details
no

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: The doorjack now costs 4 TC instead of 6.


therealDLondon:
How would you feel about 4TC with 2TC at discount?

deltanedas:
still very little reason to do it when you can get:
- a radio
- 2 red soaps
- a susbox to complement jaws and vice versa
- pistol mags

its rare that you need a quick burst of 3 doors, still dont think people would use it when you can get both generally superior options for the same price

JustTNE:
I feel like the doorjack should be a little more expensive. It has the ability to keep doors permanently opened until manually fixed, which is an advantage the jaws don't have.

deltanedas:
with AI that wont matter since it can instantly unbolt them

JustTNE:
That won't fix the access on the door being broken, or has that behavior changed?

JustTNE:
Also, I do believe the doorjack is a smaller item than the jaws, which makes it easier to store in a storage implant. That is probably another reason it should be a little more expensive than 2 tc.

deltanedas:
no but ai can then bolt it closed or shock you because you didnt buy the insuls

ThataKat:
Also need to consider skill and speed. Susbox can effectively do the same, but a lot of players are bad at hacking, etc. Both susbox and jaws take significant time as well. AO is much quicker and easier to use. Price reduction isn't a bad idea but I think the 4tc suggestion is more reasonable.

Kr8art:
I think two TC is honestly fine. The doorjack has the advantage of speed, but is limited in charges, whereas the jaws take longer at unlimited charges. 

Personally I do not consider the doors staying open to be always advantageous. If the doorjack is for example used to quickly burst into the bridge to get a steal objective, it's very likely that the alert level gets raised once people discover the jacked doors.

Resources/Prototypes/_DV/Catalog/uplink_catalog.yml
```diff
@@ -82,9 +82,9 @@
   productEntity: Doorjack
   discountCategory: rareDiscounts
   discountDownTo:
-    Telecrystal: 3
+    Telecrystal: 1
   cost:
-    Telecrystal: 6
+    Telecrystal: 2
```
Lyndomen, line 85: 4

Resources/Prototypes/_DV/Catalog/uplink_catalog.yml
```diff
@@ -82,9 +82,9 @@
   productEntity: Doorjack
   discountCategory: rareDiscounts
   discountDownTo:
-    Telecrystal: 3
+    Telecrystal: 1
```
Lyndomen: 2

Lyndomen (changes requested):

Lyndomen:
then good to go

PR #2308, opened 2024-11-28
Give Captain's PDA Useful Apps

Emily9031:
This is my first PR, please let me know if I did anything wrong or stuff I can do better.

## About the PR
I added the SecWatch, AstroNav, StockTrading, And Glimmer Monitor apps to the captain's PDA.

## Why / Balance
The captain should have these apps so they can keep an eye on how each department is doing, and in the case of the Astro Nav, get back to the station if they end up lost in space with their jetpack. (This has happened to me once)

Feedback from the Discord led me to exclude the MedTek app from being added, which would let the captain use their PDA as a health analyzer. If you think it should be added, let me know I guess.

## Technical details
Modifies `Resources/Prototypes/Entities/Objects/Devices/pda.yml`

## Media
Screenshot 2024-11-28 085137

## Requirements

## Breaking changes
None. Hopefully.

**Changelog**

:cl:
- tweak: The Captain's PDA now comes with more apps.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/pda.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/pda.yml
@@ -548,6 +548,16 @@
     borderColor: "#7C5D00"
   - type: Icon
     state: pda-captain
+  - type: CartridgeLoader
+    uiKey: enum.PdaUiKey.Key
+    preinstalled:
+    - CrewManifestCartridge
+    - NotekeeperCartridge
+    - NewsReaderCartridge
+    - SecWatchCartridge # DeltaV: SecWatch replaces WantedList
+    - AstroNavCartridge
+    - StockTradingCartridge # Delta-V
+    - GlimmerMonitorCartridge
 
 - type: entity
   parent: BasePDA
--- a/Resources/Prototypes/Entities/Objects/Devices/pda.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/pda.yml
@@ -548,6 +548,7 @@
     borderColor: "#7C5D00"
   - type: Icon
     state: pda-captain
+  # Begin DeltaV additions
   - type: CartridgeLoader
     uiKey: enum.PdaUiKey.Key
     preinstalled:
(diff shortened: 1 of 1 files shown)


Emily9031:
Just noticed my commit names aren't right. Oh well, I'll do it properly next time.

MilonPL (changes requested):
I don't think this is a good change: it removes an aspect from the game when you can just remotely view everything that's happening around you. The captain should have to talk and interact with department heads to know how their department is doing. They can get updates over the comms, even. Upstream removed the WantedList app from captain's PDA for similar reasons, see https://github.com/space-wizards/space-station-14/pull/33084. 
> The Captain might sometime need to fill in for Medical, yet there is no push for the Captain PDA to have the Med Scanner.
>You're a Captain and need to work a department? You have access to the cartridge in the head locker by default, go install the software there.

Emily9031:
I don't fully disagree that captains should have to go around and interact with department heads to see how the departments are doing. However, keeping up with every department on the station is pretty difficult, and sometimes those departments heads simply don't exist or are unavailable. These apps just let the captain have a little convenience in being able to deal with multiple things at once, and have the information available when they need it, instead of having to scroll through an hour of chat messages to find something that was said on the radio.

Also, I've been told by staff that the captain using the medical scanner cartridge or other cartridges like that can be a bit powergamey, but I suspect that may be a case by case kind of thing. Regardless, if any of the included apps should specifically be excluded, feel free to let me know.

dvir001:
The captain job isn't to remotely babysit other departments from the bridge, they should not have the apps to allow it.

ThataKat:
I will also note that if the Captain feels they need these apps they should be able to go to the respective head lockers and use the cartridges to install them themselves. I believe all of these have cartridges (except maybe glimmer monitor?). Id also rather not push captain behavior in this direction by making it round start

deltanedas:
glimmer is regularly reported on radio even

Emily9031:
General consensus seems against this, sooooo I'll just close it for now.

PR #2316, opened 2024-11-29
Un-Overtunes CHIMP (still overall buff)

BlitzTheSquishy:
## About the PR

Reduced the Chimp down 6 points of damage

## Why / Balance

having the CHIMP be 30 damage with omega particles allows it to 2 shot diona, and nearly 3 shot moths, while 4 shotting everyone, with 24 it reduces to 5 instead, diona probably still easily 2 shot and moth a more clean 4.

Largely due to the TC cost of the experimental upgrade (which is rather cheap for its damage output) this should  bring it more in line with how other syndicate weapons are done

which is generally *slightly better* than weapons security can access.

## Requirements

## Breaking changes

**Changelog**


sheepism1:
the CHIMP upgr chip is NOT a weapon on its own and PAINFULLY OBVIOUS

deltanedas:
iirc its cheaper to emag a protolathe and mass produce actual lasers, this should have a niche as a very hard hitting upgrade

Avalon-Proto:
> iirc its cheaper to emag a protolathe and mass produce actual lasers, this should have a niche as a very hard hitting upgrade

It's either cheaper or same price to Emag a protolathe yeah. But by emagging a protolathe you can now get a porta charger and a 2nd chimp. Now you have 2 chimps you swap in the charger all the time to keep the fight up

sheepism1:
Simple easier strategy:

Weaponised laser manipulation (0tc)

Emag (like 4 or 6 idk)

2 lasers and a port recharger (bribe cargo)

Get silly with it

Lyndomen:
Imo diona being 2 shot is a diona problem too, not really a weapon problem. If anything I would have lived to see a ROF decrease over damage decrease

BlitzTheSquishy:
> > iirc its cheaper to emag a protolathe and mass produce actual lasers, this should have a niche as a very hard hitting upgrade
> 
> It's either cheaper or same price to Emag a protolathe yeah. But by emagging a protolathe you can now get a porta charger and a 2nd chimp. Now you have 2 chimps you swap in the charger all the time to keep the fight up

this is also generally cheaper than the other method

Lyndomen:
Direction taking this in for review, eta 3 days probably

deltanedas:
@Lyndomen true

Avalon-Proto:
I'll be honest even with the mega damage buff I've *rarely* seen this thing in use just because even if it is a syndie energy weapon. It's just. Not used. Due to the fact it's super niche

Lyndomen:
damn i had feedback and consensus but fell asleep

Lyndomen (changes requested):
Damage is fine as is . Fairly mediocre ttk already at 2.66 secs or something. the bullet travel times are also miserably slow- hitting 4 shots is hard

PR #2324, opened 2024-11-30
Add the gas analyzer to the tools module

Radezolid:
## About the PR
This PR adds a gas analyzer to the tools module

## Why / Balance
Engineering cyborg suffers being unable of measuring the atmos status around them and distro, currently the only way to obtain a gas analyzer is waiting for EPI to research advanced tools if they don't explode first.

## Technical details
YAML change.

## Media

image

## Requirements

## Breaking changes

Just the UI like the Advanced Tools module, I don't know how to fix that.

**Changelog**

:cl:
- tweak: The tools module now comes with a gas analyzer installed.


JustTNE:
As a cyborg player with about 340 hours of playtime, this *is* a nice addition, I'm just worried about all the tools cycling they have to do getting even worse. To my knowledge, there is no way to "swap hands" in any direction other than forward down the list.
This doesn't really make sense from a logic perspective, but it might genuinely be better to put this into like the construction module, which will let Engi borgs use it (which makes sense) while keeping the tools module from having too many tools to cycle through.

Radezolid:
Gas analyzer in the construction module makes no sense as it doesn't aid in construction also it's one extra press, it's the same length as the advanced tools module.

JustTNE:
It is one extra press, but one extra press adds up *real* quick if you have to cycle around constantly to deconstruct something like a reinforced wall. That's my only concern with this. If there was a way to "go back" in the tools list that wouldn't be nearly as bad.

deltanedas (approved):

deltanedas:
shift X or something can be added in the future
and whenever rpd is added it will go in an atmos module

PR #2337, opened 2024-12-01
Makes cyborgs subservient to the Station AI

deltanedas:
## About the PR
borgs are now enslaved to the ai and emagging makes the funny law -1 instead of 0 to not get outed easily

[ported from goob](https://github.com/Goob-Station/Goob-Station/pull/919), but instead of adding it before every law it has a system to keep it synced properly with borg type switching and such

also fix funny `{$name}` for emag law 4, no comment because it should be moved into deltav file at some point anyway along with the law -1

## Why / Balance
silicons are probably not crew so borgs have no real obligation to listen to ai, now ai is the true mastermind of borgs (except if you emag a borg then you can tell it to ignore ai)

## Technical details
`SlavedBorgSystem` adds the law and funny borg switching partial system keeps it when switching borg type

## Media
some funny

## Requirements

## Breaking changes
no

**Changelog**
:cl: deltanedas, Ilya246
- add: Borgs have a new law that makes them obey the AI.

### DIFF
--- a/Content.Client/DeltaV/Silicons/Laws/SlavedBorgSystem.cs
+++ b/Content.Client/DeltaV/Silicons/Laws/SlavedBorgSystem.cs
@@ -1,0 +1,5 @@
+using Content.Shared.DeltaV.Silicons.Laws;
+
+namespace Content.Client.DeltaV.Silicons.Laws;
+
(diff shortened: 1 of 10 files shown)


MilonPL (approved):
🗣️ 🔥

PR #2339, opened 2024-12-02
Make medical borgs able to fly in space

Emily9031:
## About the PR

This PR makes it so medical borgs can fly in space as if they were using a jetpack.
## Why / Balance

The medical borg sprites show it as using some sort of technology to float around. It only makes sense that this would allow them to float around in space too. Additionally, this gives them the opportunity to do some paramedic style work and space rescues.

Regarding balance, with this implementation, the borg will move slower than a person with a jetpack would, but the borg does not need to refuel a jetpack. Additionally the borg won't create the jetpack particle trails behind it, but that could be added with some more code.
## Technical details

Super simple one line change. Just adds the JetpackUser component to the medical borg.
## Media

https://github.com/user-attachments/assets/6a08f829-dc03-4315-9dbc-743d7c04f9b4

## Requirements

**Changelog**

:cl:
- add: Cyborgs are now able to slowly fly through space.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
+++ b/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
@@ -248,6 +248,7 @@
     visMask:
     - PsionicInvisibility
     - Normal
+  - type: JetpackUser # DeltaV: Lets medical cyborgs fly in space
 
 - type: entity
   abstract: true
--- a/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
+++ b/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
@@ -248,7 +248,7 @@
     visMask:
     - PsionicInvisibility
     - Normal
-  - type: JetpackUser # DeltaV: Lets medical cyborgs fly in space
+  - type: JetpackUser # DeltaV: Lets cyborgs fly in space
 
 - type: entity
   abstract: true
--- a/Resources/Prototypes/borg_types.yml
+++ b/Resources/Prototypes/borg_types.yml
@@ -172,7 +172,7 @@
     damageContainers:
     - Biological
   - type: FabricateCandy # Nyanotrasen - The medical cyborg can generate candies filled with medicine.
-  - type: JetpackUser #DeltaV: lets them fly in space
+  - type: JetpackUser # DeltaV: Lets medical cyborgs fly in space
 
   # Visual
   inventoryTemplateId: borgDutch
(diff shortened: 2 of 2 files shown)


deltanedas (dismissed):

Emily9031:
Does this need another review? Or is it ready to be merged? Sorry I'm new to this stuff.

deltanedas:
waiting for directions input since no other borg gets completely free flight like this

IamVelcroboy:
Suggestion: Make this a module that could be use by any borg, and maybe add to the default setup for med borg if that's what direction decides we want.

Emily9031:
> Suggestion: Make this a module that could be use by any borg, and maybe add to the default setup for med borg if that's what direction decides we want.

Gotcha. I think making a whole borg module miiiight be a little bit beyond my ability, but I could maybe do it. Though borgs can also just use the fire extinguisher module to be able to fly around in space or zero gravity.

gluesniffler:
> > Suggestion: Make this a module that could be use by any borg, and maybe add to the default setup for med borg if that's what direction decides we want.
> 
> Gotcha. I think making a whole borg module miiiight be a little bit beyond my ability, but I could maybe do it. Though borgs can also just use the fire extinguisher module to be able to fly around in space or zero gravity.

https://github.com/Simple-Station/Einstein-Engines/pull/550 could probably use this as a starting point, though its kinda shitcode.

ThataKat:
Heya, direction here. We're chatting about this so we'll get back with recommendations at some point in the future, but here's what I'm thinking currently:

I like this, and it makes total sense that you chose to just add it to the mediborg since it's the only one visually flying. But for gameplay reasons, it would be nice for other borgs to have similar abilities (ignoring fire extinguisher). *But* an inherent jetpack on every borg is weird because... it just doesn't make sense visually they're not all hovering. So it's a slight dilemma. 

I'll try and get back with a solidified response in the next day or two.

ThataKat:
Direction here. Our recommendation is going to be giving this to all borgs. It's just a nice QoL for them, and there's no mechanical reason to restrict it to only the mediborg. No module needed.

A direction member has suggested making it consume extra power to move in space, but if that's not feasible it's all good, we'd still like to see this implemented.

Emily9031:
I think I've done all I can here. I looked into it a lot and I'm not able to implement the extra power usage with my current knowledge. If a more experienced coder is able to do it, please go ahead. For now though, I've updated the changelog, applied the flight to all borgs, and this is ready for review again.

One concern I had is if syndicate borgs should be able to do this too. If you think they should be then everything is fine. If you would like it restricted from them let me know, it's an extremely simple change, just one line of code still.

ThataKat:
I was honestly under the impression that the assault borgs already have some limited mobility in space. It's perfectly fine for them to have it too, it's practical and makes sense of basic borgs have it.

deltanedas (approved):

PR #2364, opened 2024-12-05
fix tidy_map.py

MilonPL:
## About the PR

refactored the script because it was a hot mess 
updated to work with the latest version of ruamel.yaml, which, shockingly, changed things (actually loads now) 
moved to deltav namespace, made the copyright clearer
removed deprecated garbage, cleaned up imports, and maybe made it usable

## Why / Balance

hive before running the tool 👎 
image

hive after running the tool 👍 
image

image

## Breaking changes

none, unless you’re emotionally attached to the broken version
@IamVelcroboy should probably make a contrib announcement about it


Tools/DeltaV/tidy_map.py
```diff
@@ -0,0 +1,205 @@
...
+        "EmitSoundOnCollide",
+        "Fixtures",
+        "GravityShake",
+        "HandheldLight",  # Floodlights are serializing these?
+        "PlaySoundBehaviour",
+    ]
+
+    # Components that will have specific fields removed and may be removed entirely if empty
+    REMOVE_COMPONENT_DATA: Dict[str, List[str]] = {
+        "Airtight": ["airBlocked"],
```
deltanedas, line 33: these components should be checked for which are TimeOffsetSerializer and can be removed

deltanedas (changes requested):

deltanedas (approved):

PR #2383, opened 2024-12-07
Randomizes the appearance of Fugitives

Monotheonist:
## About the PR

Allowed Fugitives to not be your character. 1984, no more Okelte Valid. It turned out it was a one line change for both.
## Why / Balance

Me and Lyndo were not fans of how Fugitives spawn as your character that you already picked. This fixes it by spawning you as a random character that is the morphotype that you were, which helps with continuity from a story standpoint of a round, because it's weird to have John Tider die permanently from dying to a dragon and then have a dude that looks exactly like him be an escaped fugitive or an LP operator with a different name. 'Course, that can add to the story of a round, but I feel as if it makes more sense for it to be a random fucking guy.
## Technical details

Adds ` - type: RandomHumanoidAppearance
        randomizeName: false` to the fugitive and therefore causes them to spawn with a random appearance of the morphotype you were. Random name is kept intact so you'll still end up as Sleve McDichael.
## Media

image
Technically is part of this because the player character was a harpy.
## Requirements

## Breaking changes

Not a breaking change in the sense of technicality, but it might be difficult for Vox Players, since the morphotype is inherited, to which I say:
- You get masks and oxygen tanks by default in the Listening Post, just go into the closet, grab an oxytank and then breathe through your voice mask. Easy fix.
- I think leaving Fugitive like it is helps add to the survival factor, for them.

**Changelog**

Yes CL, yes fun.
:cl:
- tweak: Fugitives are no longer clones of people on station.

### DIFF
--- a/Resources/Prototypes/DeltaV/GameRules/events.yml
+++ b/Resources/Prototypes/DeltaV/GameRules/events.yml
@@ -110,8 +110,6 @@
       - type: RandomMetadata
         nameSegments:
         - names_death_commando
-      - type: RandomHumanoidAppearance
-        randomizeName: false
       - type: AutoImplant
         implants:
         - DeathAcidifierImplant
--- a/Resources/Prototypes/DeltaV/GameRules/events.yml
+++ b/Resources/Prototypes/DeltaV/GameRules/events.yml
@@ -110,6 +110,8 @@
       - type: RandomMetadata
         nameSegments:
         - names_death_commando
+      - type: RandomHumanoidAppearance
+        randomizeName: false
       - type: AutoImplant
         implants:
         - DeathAcidifierImplant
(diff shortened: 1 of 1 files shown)


Lyndomen:
Direction review pending

Lyndomen:
also CL might be nice so people know it is a thing

Monotheonist:
Alright, yes CL, yes fun.

deltanedas (dismissed):

Lyndomen (changes requested):
Direction requesting update: LPO to stay the player character. Our preference is that this would be toggleable by character preference, but thats a big wyci it so whatever.

Monotheonist:
That's legit, because the LPO is barely seen in rounds if at all, and it is less weird that they are clones if they aren't on station.

Monotheonist:
Pretty much ready to go if the tests don't fail.

deltanedas (approved):

PR #2394, opened 2024-12-08
Reduce RE time

Stop-Signs:
## About the PR
5 seconds per percent instead of 30, maybe now this machine will actually be useful
## Why / Balance
Reverse engineering machines, despite reworks, have always been kind of, well, bad. this should hopefully bring them to relevancy.

takes ~8 minutes to RE something, 5 seconds may be a bit over tuned, but im willing to adjust based on direction

## Technical details
1 line C#

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Reverse engineering machines now function much faster!

### DIFF
--- a/Content.Shared/Nyanotrasen/ReverseEngineering/ReverseEngineeringMachineComponent.cs
+++ b/Content.Shared/Nyanotrasen/ReverseEngineering/ReverseEngineeringMachineComponent.cs
@@ -80,7 +80,7 @@ public sealed partial class ReverseEngineeringMachineComponent : Component
     /// How long to wait between analysis rolls.
     /// </summary>
     [DataField]
-    public TimeSpan AnalysisDuration = TimeSpan.FromSeconds(30);
+    public TimeSpan AnalysisDuration = TimeSpan.FromSeconds(5);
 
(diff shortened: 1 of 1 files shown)


Radezolid:
This machine lets you get T3 almost for free and no glimmer, so at least it costs time. A toggle to allow faster research but at the cost of generating glimmer would be better.

Content.Shared/Nyanotrasen/ReverseEngineering/ReverseEngineeringMachineComponent.cs
```diff
@@ -80,7 +80,7 @@ public sealed partial class ReverseEngineeringMachineComponent : Component
     /// How long to wait between analysis rolls.
     /// </summary>
     [DataField]
-    public TimeSpan AnalysisDuration = TimeSpan.FromSeconds(30);
+    public TimeSpan AnalysisDuration = TimeSpan.FromSeconds(5);
```
MilonPL: make it 15

MilonPL (changes requested):

MilonPL (approved):

PR #2395, opened 2024-12-08
Add Machine Incompatible Trait

Emily9031:
## About the PR

This adds a new trait that makes characters unable to have their brain put into a man-machine interface.
## Why / Balance

I think people wanted it.
## Technical details

Added new Unborgable trait, blacklisted any brains with it from being put into MMIs, and caused a message to appear when inspecting brains with this trait.
## Media

image
Indicator of a brain that is machine incompatible
## Requirements

## Breaking changes

I don't think I broke anything.
**Changelog**

:cl:
- add: New Machine Incompatible trait, makes characters unable to have their brain put into a man-machine interface.


Emily9031:
I wanted to add a popup when you attempt to put a brain with this trait into an MMI, but I wasn't able to get the code working. If someone knows more and is able to do it, go for it.

Content.Server/Body/Systems/BrainSystem.cs
MilonPL, line 1: put this in a separate shared system so that the popup doesn't take a second to show, no reason for it to be in here

Content.Server/Body/Systems/BrainSystem.cs
```diff
@@ -33,12 +39,23 @@ private void HandleMind(EntityUid newEntity, EntityUid oldEntity)
...
             if (!_mindSystem.TryGetMind(oldEntity, out var mindId, out var mind))
                 return;
 
             _mindSystem.TransferTo(mindId, newEntity, mind: mind);
         }
 
+        private void OnExamined(Entity<UnborgableComponent> ent, ref ExaminedEvent args) //DeltaV
+        {
+            var msg = new FormattedMessage();
+            msg.AddMarkupPermissive("[color=red]This brain is damaged beyond use.[/color]");
```
MilonPL: this should be localized and use AddMarkupOrThrow then
also not sure why you need to create a new message

Content.Server/Traits/Assorted/UnborgableComponent.cs
```diff
@@ -0,0 +1,10 @@
+namespace Content.Server.Traits.Assorted;
+
+/// <summary>
+/// This is used for the unborgable trait.
+/// </summary>
+[RegisterComponent]
+public sealed partial class UnborgableComponent : Component
+{
+
+}
```
MilonPL: doesn't need a body

Resources/Prototypes/Entities/Objects/Specific/Robotics/mmi.yml
```diff
@@ -41,6 +41,9 @@
         whitelist:
           components:
           - Brain
+        blacklist:
+          components:
+          - Unborgable # DeltaV
```
MilonPL: this comment should be on the first line

MilonPL (changes requested):

Content.Server/Body/Systems/BrainSystem.cs
Emily9031: I spent half an hour moving everything over just for it to change nothing and the popup still took a second, so the best I can do is like, ask the Discord people to give me a tutorial on how to do this.

Content.Server/Traits/Assorted/UnborgableComponent.cs
Emily9031: I literally just copy pasted the unclonable component and changed it to unborgable. They all look like this.

MilonPL: still doesnt need a body, you can just use a semicolon instead

Content.Server/Body/Systems/BrainSystem.cs
deltanedas: you can use pushmarkup yes

PubliclyExecutedPig:
Is like the brother from better call saul

Content.Server/Body/Systems/BrainSystem.cs
Emily9031: I can't figure out how to do this, and every time I think I have it, there's still a delay, so I give up. Someone else can figure it out. Otherwise, the code works fine.

Content.Server/Traits/Assorted/UnborgableComponent.cs
```diff
@@ -0,0 +1,7 @@
+namespace Content.Server.Traits.Assorted;
+
+/// <summary>
+/// This is used for the unborgable trait.
+/// </summary>
+[RegisterComponent]
```
deltanedas: add NetworkedComponent, move to shared + in DeltaV folder

Content.Server/Body/Systems/BrainSystem.cs
```diff
@@ -33,12 +39,20 @@ private void HandleMind(EntityUid newEntity, EntityUid oldEntity)
...
+            if (HasComp<UnborgableComponent>(oldEntity)) // DeltaV
+                EnsureComp<UnborgableComponent>(newEntity);
+
             if (!_mindSystem.TryGetMind(oldEntity, out var mindId, out var mind))
                 return;
 
             _mindSystem.TransferTo(mindId, newEntity, mind: mind);
         }
 
+        private void OnExamined(Entity<UnborgableComponent> ent, ref ExaminedEvent args) //DeltaV
```
deltanedas: put this in an UnborgableSystem also this makes no sense to just look at someone and know they cant be borged, return early if the entity has BodyComponent

deltanedas (changes requested):

Content.Server/Body/Systems/BrainSystem.cs
Lyndomen: imo its slightly better to not have people waste time shoving brain into an MMI if it was never going to work. Maybe it could be "This brain is incompatible with being digitized due to its oversized Frontal Lobe"

deltanedas: that should be on health analyzer like unreviveable, you cant just look at someone and know that

deltanedas: made it only show up on the brain itself and health analyzer

deltanedas (dismissed):

deltanedas:
you can now only see the examine for the brain itself, to see it for a person you need a health analyzer (will also properly work with surgery and stuff now)

Lyndomen (commented):
Direction good with this still displaying on inspect of brain.

deltanedas (approved):

PR #2397, opened 2024-12-08
Differentiate warden and HOS suit

Stop-Signs:
## About the PR
Warden suit is slower but much tankier, comparible to jugsuit but not quite as absurdly protective

## Why / Balance
Why was the warden suit just the HOS suit but slower, we may never know. now its essentially a nukie commander suit with a 40% slow

## Technical details
YML

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Security warden suit has been made tankier but slower to differentiate it from the HOS


Radezolid:
Security is meant to be slow in a hardsuit but faster than an engineer, while they are meant for combat their gear should be worse than the syndicate's after all you have guns and coordination with the rest of security.

Setting the speed at this value means that any future syndicate hardsuit would need to be at maximum 10% slower than normal speed to be better than the security hardsuit.
Also I say it constantly but I feel a power creep happening with damages going up and armor now being more protective too.

Stop-Signs:
> Security is meant to be slow in a hardsuit but faster than an engineer, while they are meant for combat their gear should be worse than the syndicate's after all you have guns and coordination with the rest of security.
> 
> Setting the speed at this value means that any future syndicate hardsuit would need to be at maximum 10% slower than normal speed to be better than the security hardsuit. Also I say it constantly but I feel a power creep happening with damages going up and armor now being more protective too.

the only syndie suit that is slower than sec hardsuits is the juggersuit, even with these changes. syndie hardsuits are still only a 10% movespeed reduction. 

furthermore we really shouldn't be balancing around nukies specifically. the antag sec is going to be fighting most in hardsuits is the ninja, who gets speed magboots and a 0% slow hardsuit. a 5% increase in speed isnt going to break the power budget that sec has, especially since they are now wearing magboots that slow them a furhter 30(?)%

Radezolid:
They don't need the magboots to hunt a ninja, you need coordination for that. Sec gets all the tools required to hunt the ninja already. Only scenarios where you would need the magboots are no gravity or lube bombs.
The syndicate suit can be bought by any syndicate agent too + juggernaut is meant to be slow due to their resistance. What i said was for any future syndicate hardsuit, or any antag which uses one that gets added has to be at maximum 10% slower than regular speed which is a low reduction
Security uses hardsuits to hunt dragons, ninjas, nuclear operatives and salvage. On every of those examples the ninja and salvage won't really affected by the speed change since they are either super fast or you will confront them in space. Speed is a lot in this game, dragons will able to punish less dumb pushes from a secoff and nuclear operatives can get flanked harder now that sec is faster.

MattFright:
I will say i really like the warden's hardsuit changes at least. The warden's hardsuit already looks a whole lot bulkier, and having them be more distinct because really in practice it just feels like a pointless reskin, it's not that different from regular ones.

I don't know about making regular sec hardsuits faster but corps' being faster would keep it more in line with the paramedic role i guess. Even if it takes a small defense nerf to compensate, i do think the corpsman's hardsuit should be at least a little bit speed over overwhelming combat protection.

That's just my two cents on it as someone who's played a ton of sec officer, warden and corpsman, as well as the antags that they fight.

Lyndomen:
Direction taking this in for review, sorry for delay

Lyndomen:
Howdy, Direction review complete. Short story, we feel that bumping the general/corpsman security suit speed is not needed at this time. We do like the Warden suit changes, and would like to publish those.

Lyndomen (changes requested):
remove sec speed bump, keep all changes to warden

Lyndomen (dismissed):

deltanedas (approved):

PR #2412, opened 2024-12-10
fix "teach person a lesson"

MilonPL:
## About the PR

now it just subscribes to the MobStateChangedEvent, then checks if the state change is to Dead, and if the mind of the entity that died matches the mind we have as the target
also sets it in the comp rather than the funny list in the system because we have the means to do ECS properly

## Why / Balance

no more noobtrap redtext because you didnt know to open your objectives after killing someone

## Breaking changes

**Changelog**

:cl:
- fix: The Teach Person a Lesson objective will now register properly when your target dies.

### DIFF
--- a/Content.Server/DeltaV/Objectives/Components/TeachLessonConditionComponent.cs
+++ b/Content.Server/DeltaV/Objectives/Components/TeachLessonConditionComponent.cs
@@ -1,17 +1,11 @@
-﻿using Content.Server.Objectives.Systems;
+﻿using Content.Server.DeltaV.Objectives.Systems;
+using Content.Server.Objectives.Components;
 
-namespace Content.Server.Objectives.Components;
+namespace Content.Server.DeltaV.Objectives.Components;
 
 /// <summary>
 /// Requires that a target dies once and only once.
 /// Depends on <see cref="TargetObjectiveComponent"/> to function.
 /// </summary>
 [RegisterComponent, Access(typeof(TeachLessonConditionSystem))]
-public sealed partial class TeachLessonConditionComponent : Component
-{
-    /// <summary>
-    /// Whether the target has been killed
-    /// </summary>
-    [DataField]
-    public bool WasKilled;
-}
+public sealed partial class TeachLessonConditionComponent : Component;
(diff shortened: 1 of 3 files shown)


Content.Server/DeltaV/Objectives/Components/TeachLessonConditionComponent.cs
```diff
@@ -7,4 +7,11 @@ namespace Content.Server.Objectives.Components;
...
 [RegisterComponent, Access(typeof(TeachLessonConditionSystem))]
-public sealed partial class TeachLessonConditionComponent : Component;
+public sealed partial class TeachLessonConditionComponent : Component
+{
+    /// <summary>
+    /// Whether the target has been killed
+    /// </summary>
+    [DataField]
+    public bool WasKilled;
+}
```
deltanedas: use CodeConditionComponent and set it to true when the target dies
then remove OnGetProgress

deltanedas (changes requested):

Content.Server/DeltaV/Objectives/Systems/TeachLessonConditionSystem.cs
```diff
@@ -1,48 +1,46 @@
...
+            // Check if this objective's target matches the entity that died
+            if (targetObjective.Target != mindId)
+                continue;
 
-    // Clear the wasKilled list on round end
-    private void OnRoundEnd(RoundRestartCleanupEvent  ev)
-    {
-        _wasKilled.Clear();
+            _codeCondition.SetCompleted((uid, null));
+        }
```
deltanedas, line 44: this will NRE, add it to the query

deltanedas (changes requested):

Content.Server/DeltaV/Objectives/Systems/TeachLessonConditionSystem.cs
MilonPL: the comp is nullable and it resolves it anyways, tested in debug and works fine?

deltanedas: then just pass uid if it resolves, not funny `(uid, null)`

MilonPL: real

deltanedas (approved):

PR #2414, opened 2024-12-11
glacier surface update

deltanedas:
## About the PR
- remade surface entirely so its a bit more cosy like the original concept
- atmos engi salv and the med supply room are all locked

- surface shuttle can be boarded at arrivals
- the surface shuttle console is cargo-locked so cargo techs (valet!!!) can ferry people to and from the surface
- had to remove random rotation so not have rotation rng prevent the shuttle from redocking to the station. i dont like it anyway (extremely annoying for cargo gaming when its far away or if you get lost) so rip bozo

misc changes:
- make med more sop friendly
- public cargo ordering console

## Media

**Changelog**
:cl:
- add: You can now visit the surface of Glacier's planet, get a cargo tech to fly you down at arrivals!


IamVelcroboy:
I don't know how the atmos works here. Does it not need scrubbers?

deltanedas:
> Does it not need scrubbers?

the passive vent network combines distro + heating + scrubbing (inline filters)

the filters do need turning on but that gives atmosians a reason to pay the surface a visit after they have teg up and have nothing to do for 2 hours

Colin-Tel (commented):
I can't seem to FTL from the surface to the station dock; it instead will put the shuttle somewhere on glacier station's map randomly. Is this intended?
image

Avalon-Proto:
> I can't seem to FTL from the surface to the station dock; it instead will put the shuttle somewhere on glacier station's map randomly. Is this intended? image

It's the station random rotation doing it. I saw this happening the other day when I spawned a mining airlock with the shuttle. It fails to FTL back to its dock due to the fact it's not in a fixed position

Colin-Tel:
> It's the station random rotation doing it. I saw this happening the other day when I spawned a mining airlock with the shuttle. It fails to FTL back to its dock due to the fact it's not in a fixed position

Well I guess that means I haven't been able to replicate it being fixed, then.

deltanedas:
ill look into it tomorrow but the prototype rotation removal should have fixed that

Colin-Tel:
It still appears to not dock correctly. Can you confirm?
image

deltanedas:
womm, are you sure you recompiled after updating (actually fixed now :trollface:)

Colin-Tel (approved):

PR #2424, opened 2024-12-12
Auto ACO Procedure bug fixes

SolStar2:
## About the PR
Fixed CVar to use float minutes instead of TimeSpan so it can be modified safely.
Made debug mode run the announcements faster for easier testing.
Make sure that there is no persistence with the components on init by assigning values on init event.

## Why / Balance
People reporting that the ACO announcements are sometimes not working. I haven't been able to replicate the bug but these fixes cant hurt and may fix the issue.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/DeltaV/Station/Systems/CaptainStateSystem.cs
+++ b/Content.Server/DeltaV/Station/Systems/CaptainStateSystem.cs
@@ -11,7 +11,6 @@ using Content.Shared.DeltaV.CCVars;
 using Robust.Shared.Configuration;
 using Robust.Shared.Prototypes;
 using System.Linq;
-using JetBrains.FormatRipper.Elf;
 
 namespace Content.Server.DeltaV.Station.Systems;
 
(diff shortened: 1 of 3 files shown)


Content.Server/DeltaV/Station/Systems/CaptainStateSystem.cs
```diff
@@ -140,6 +129,25 @@ private void HandleNoCaptain(Entity<CaptainStateComponent?> station, CaptainStat
...
+    private void SendAnnouncement(EntityUid ent, string message)
+    {
+        if (_station.GetOwningStation(ent) != null)
+            _chat.DispatchStationAnnouncement(
+                ent,
+                message,
+                colorOverride: Color.Gold);
+        else
+        {
+            Log.Error($"{ent} does not belong to a station.");
```
deltanedas: CaptainStateComponent is only added to stations so theres no way this would happen

deltanedas (approved):

SolStar2:
Thanks for reviewing this so quickly!

PR #2429, opened 2024-12-12
Aco fix round 2

SolStar2:
## About the PR
After more testing I found that the StandardNanotrasenStation does not shutdown and reinit between rounds in release environment. Thats why I was unable to find the bug before but this should fix it now.

## Why / Balance
Millions condemned

## Requirements

## Breaking changes

**Changelog**


deltanedas:
the station entity is deleted when a round restarts so this still wont do anything
its more likely to be a timing issue with checking jobs a tick or so early before its registered

PR #2432, opened 2024-12-13
Change tracking implants radio rattle from security to medical

Radezolid:
## About the PR
This PR modifies the tracking implant to send the message to the medical channel instead of the security one.

## Why / Balance
Theoretically they are meant for prisoners, but it's mainly used to know when command dies. So this adds a extra layer of communication that delays the response from security. This can also be used by an evil paramedic or doctor as an opportunity to do their objectives.

## Technical details
YAML change.

## Media
image

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Tracking implants now rattle on the medical channel instead of the security one.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
@@ -101,7 +101,7 @@
       mobState:
       - Critical
     - type: Rattle
-      radioChannel: "Security"
+      radioChannel: "Medical" # DeltaV - Extra layer against sec just spamming them for all of command. Was "Security"
 
 #Traitor implants
 


FieldCommand:
Why not both channels? Commands usually don't die out of nowhere, usually someone made them dead, so having security knowing too would be good. But letting medical know is also good, as paramedics can get there to assist and/or medical prepare for a high-value patient arriving to medical.

Radezolid:
Two reasons:
1. `RadioSystem` only allows to use `SendRadioMessage` on a single channel, so a rattle with multiple channels is not possible and i can't change that due to my smooth brain on C#.
2. This is another discouragement for sec to stop powergaming trackers, they are meant for prisoners and you already have the AI on crew monitor + paramedic + corpsmen so you don't really need it and removing the trackers is a bad idea.

PR #2450, opened 2024-12-15
Add error logging for Auto ACO procedure.

SolStar2:
## About the PR
Add Error logging for Auto ACO Procedure for easier debugging.

## Why / Balance
People still reporting issues with Auto ACO Procedure and I'm still unable to reproduce. Adding Error logging to better understand the cause.

## Technical details
I also explicitly initialized values to false that are supposed to be false so that there is no chance for fuckery in deserialization.

## Requirements

## Breaking changes

**Changelog**


Content.Server/DeltaV/Station/Components/CaptainStateComponent.cs
```diff
@@ -18,7 +18,7 @@ public sealed partial class CaptainStateComponent : Component
     /// Assume no captain unless specified
     /// </remarks>
     [DataField]
-    public bool HasCaptain;
+    public bool HasCaptain = false;
```
deltanedas, line 21: false is the default this just wont do anything

deltanedas (approved):

PR #2466, opened 2024-12-18
cleanup and carrying refactor

deltanedas:
## About the PR
- cleanup of 2 recent prs
- surface level refactor of deep fryer component
- full refactor of carrying, moved it to shared too

## Why / Balance
good

## Technical details
begone shitcode

removed the explicit escape from inventory thing since nothing uses it + escape from inventory *should* be able to just work
removed redundant checks and fleshed out CanCarry, it was checking if the user was alive when you cant start carrying when crit anyway

## Media
felenid chucking amusement park...

## Requirements

## Breaking changes
moved everything from Content.Server.Carrying and Content.Shared.Carrying to Content.Shared.DeltaV.Carrying
CarrySlowdownComponent now only has 1 field, Modifier, that is applied for both sprint and walk speed

**Changelog**
:cl:
- fix: Fixed not being able to throw felenids...


Lyndomen:
🔥🔥🔥🔥

PR #2468, opened 2024-12-18
buff ore box

deltanedas:
## About the PR
made it the same size as ore bag of holding instead of literally 1 and a half ore bags

## Why / Balance
theres no reason to use it with such terrible storage, holding a second ore bag shouldnt completely outclass it. not to mention to smelt it you have to take the ore out 1 by 1

in ss13 it has no limit so this is still "underkill"

## Technical details
no

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Made the ore box hold much more ore.


Lyndomen (approved):
fastest direction vote ever

PR #2476, opened 2024-12-18
Add External access to the prison guard role

Radezolid:
## About the PR
Title.

## Why / Balance
More in line with the rest of security + being unable to use the airlocks to the mining rocks and unable to rescue prisoners on Hammurabi using the airlocks to space.

## Technical details
YAML change.

## Media

No need.

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Prison guards now have External access.

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/Roles/Jobs/Security/prisonguard.yml
+++ b/Resources/Prototypes/Nyanotrasen/Roles/Jobs/Security/prisonguard.yml
@@ -20,6 +20,7 @@
   - Security
   #- Brig #Delta V: Removed brig access
   - Maintenance
+  - External # DeltaV - Added external access
   special:
   - !type:AddImplantSpecial
     implants: [ MindShieldImplant ]


Lyndomen:
Direction review not needed, but no brainer approve anyway

IamVelcroboy:
Alternatively, we could add sec-locked external airlocks. We might need those anyway at some point/Hammurabi would probably benefit from them.

Radezolid:
That would require mapping on all maps, opposed to just giving them external access.
Not saying I'm against the airlock existing, it just makes more sense to getting them the access to be more like the rest of security.

IamVelcroboy:
> That would require mapping on all maps, opposed to just giving them external access. Not saying I'm against the airlock existing, it just makes more sense to getting them the access to be more like the rest of security.

How many maps do guards need external access? They shouldn't really be roaming outside of perma much as they aren't full officers. Really only pertains to maps prison mines and maybe in some cases if they get lavaland access

Radezolid:
I don't want them to roam in space, they do need it if a prisoner escapes to space or, the case that made me do this PR, try to get prisoners to safety during a nuke ops round.
If a prison guard is doing security's duty instead of keeping perma confined and safe they they need to be fired. 
On the map thing, only maps that have a mining area are Hammurabi, Tortuga.
Being able to move in and out of the station seems like a basic security need and it should be standard all around the department.

IamVelcroboy:
> I don't want them to roam in space, they do need it if a prisoner escapes to space or, the case that made me do this PR, try to get prisoners to safety during a nuke ops round. 

OK, I'll give it to you for chasing escaped prisoners/Nuke Ops

alterae:
> I don't want them to roam in space, they do need it if a prisoner escapes to space or, the case that made me do this PR, try to get prisoners to safety during a nuke ops round. If a prison guard is doing security's duty instead of keeping perma confined and safe they they need to be fired. On the map thing, only maps that have a mining area are Hammurabi, Tortuga. Being able to move in and out of the station seems like a basic security need and it should be standard all around the department.

yeah as a regular perma player i 100% concur. this is an essential thing for prisoner safety on many maps, as well as prisoner security

deltanedas (approved):

PR #2479, opened 2024-12-19
Map Edits: Temporary Mining Shuttle Edits

IamVelcroboy:
## About the PR
Arena/Hammurabi/TheHive/Tortuga
- Adds mining shuttle dock
- Adds Psych locker

Arena
- Adds DetDrobe

Hammurabi
- Adds WinterDrobe

## Why / Balance
⛏

## Technical details
This is a temporary update to last for the duration of the X-Mas maps. I will have better placements when those are reverted.


therealDLondon (approved):
Surely this wont break anything 🔥

PR #2484, opened 2024-12-20
Return of war

Stop-Signs:
## About the PR
War is possible again

## Why / Balance
oh boy, i could type a whole thesis on why removing this was questionable, but ill keep it simple.
warops was fun, a large amount of the community liked it, it was removed without much input from the community and I have gotten many requests to PR it back in. if it ends up being bad then it can easily be removed once again, but I and many others are begging for this to be given another chance
## Technical details
yml and basically reverts #2207 

## Requirements

## Breaking changes

:cl:
- add: WarOps returns

### DIFF
--- a/Resources/Prototypes/Roles/Antags/nukeops.yml
+++ b/Resources/Prototypes/Roles/Antags/nukeops.yml
@@ -85,8 +85,8 @@
   equipment:
     neck: SyndicateWhistle
     outerClothing: ClothingOuterHardsuitSyndieCommander
-#  inhand: # DeltaV - Removal Of War.
-#  - NukeOpsDeclarationOfWar
+    inhand:
+    - NukeOpsDeclarationOfWar
 
 #Nuclear Operative Medic Gear
 - type: startingGear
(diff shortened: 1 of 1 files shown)


Lyndomen:
Taking in for direction review, gonna say a week but maybe less

jink23:
I agree with Gates overall that WarOps gives normal crew a bit more 'input' into a nukie round because on average a normal nukie round consists of them raiding armory and by that point all non-sec/cap can do is run and flee. Meanwhile on warops it gives people time to 'react' to the fact the station is about to come under attack and choose whether they're gonna bunker down somewhere or arm up and fight with sec, sec willing that is.

Generally I see it as a positive and gives people the 'chance' to interact with a nukie round more than a normal nukie round.

MattFright:
The removal of Warops felt like a strawman pretending like it was widely accepted that Warops is incompatible with MRP when most people actually disagreed.

I'd hope we can add it back and if it's to be removed again, there needs to be actual discussion about it unlike last time where only one person went "we did the science, it definitely does not work, we're all in agreement on this" and that was enough for the PR to be accepted.

rosieposieeee:
im personally against it as i do think it is design-wise antithetical to mrp, but if we want to do another round of testing and see how the community feels, i wouldnt mind.

deltanedas:
theres nothing to test as theres nothing changed from last time

Sifb26:
Sorry for the delay, direction consensus was to deny this PR.

deltanedas:
also nothing changed at all to improve rp

far cry 3 vaas

PR #2488, opened 2024-12-20
Add kill fellow traitor objective

beck-thompson:
## About the PR

Added a new traitor objective: kill fellow traitor. It should be a little less uncommon than kill target head objective (If someone has a better weight let me know. Its hard to choose 😆!)'

## Why / Balance

Right now traitors are all really buddy buddy. There is very little reason not to fully trust each other and work together! This leads to some really fun traitor RP, but also some really annoying and (In my opinion) "metagamey" team ups where traitors just spam codewords in the general radio and buy loot boxes. I hope this PR makes them at least think a little bit before instantly trusting each other.
 
## Technical details

Just added a new system that only targets traitors for kill objectives. Also ensures that you don't have any other objectives that target them (Like if you have a keep alive objective).

## Media

Screenshot 2024-12-20 123000

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: New traitor objective: Kill fellow traitor. It is rare!

### DIFF
--- a/Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
+++ b/Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
@@ -5,7 +5,5 @@ namespace Content.Server.DeltaV.Objectives.Components;
 /// <summary>
 /// Sets the target for <see cref="TargetObjectiveComponent"/> to a random traitor.
 /// </summary>
-[RegisterComponent, Access(typeof(KillPersonConditionSystem))]
-public sealed partial class PickRandomTraitorComponent : Component
-{
-}
+[RegisterComponent]
+public sealed partial class PickRandomTraitorComponent : Component;
--- a/Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
+++ b/Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
@@ -1,0 +1,11 @@
+﻿using Content.Server.Objectives.Systems;
+
+namespace Content.Server.DeltaV.Objectives.Components;
+
+/// <summary>
+/// Sets the target for <see cref="TargetObjectiveComponent"/> to a random traitor.
+/// </summary>
+[RegisterComponent, Access(typeof(KillPersonConditionSystem))]
+public sealed partial class PickRandomTraitorComponent : Component
+{
+}
--- a/Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
+++ b/Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
@@ -46,16 +46,19 @@ public sealed class KillFellowTraitorObjectiveSystem : EntitySystem
         // Going through each OTHER traitor
         foreach (var traitor in traitors)
         {
-            // Assume it will be a valid traitor.
-            validTraitorMinds.Add(traitor.Id);
-
+            var valid = true;
             // Going through each of OUR objectives.
             foreach (var objective in args.Mind.Objectives)
             {
-                // If one of OUR objectives already targets a traitor, remove them from the vaild list.
+                // If one of OUR objectives already targets a traitor, don't add it to the list.
                 if (TryComp<TargetObjectiveComponent>(objective, out var targetComp) && targetComp.Target == traitor.Id)
-                    validTraitorMinds.RemoveAt(validTraitorMinds.Count - 1);
+                {
+                    valid = false;
+                    break;
+                }
             }
+            if (valid)
+                validTraitorMinds.Add(traitor.Id);
         }
 
         // No other traitors
--- a/Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
+++ b/Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
@@ -1,0 +1,70 @@
+﻿using Content.Server.DeltaV.Objectives.Components;
+using Content.Server.Objectives.Components;
+using Content.Server.GameTicking.Rules;
+using Content.Server.Objectives.Systems;
+using Content.Shared.Objectives.Components;
+using Robust.Shared.Random;
+
+namespace Content.Server.DeltaV.Objectives.Systems;
+
+/// <summary>
+///     Handles the kill fellow traitor objective.
+/// </summary>
+public sealed class KillFellowTraitorObjectiveSystem : EntitySystem
+{
+    [Dependency] private readonly IRobustRandom _random = default!;
+    [Dependency] private readonly TargetObjectiveSystem _target = default!;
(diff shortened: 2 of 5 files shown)


Lyndomen:
Woah! Taking in for direction review, 3 days :)

Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Content.Server.Objectives.Systems;
+
+namespace Content.Server.DeltaV.Objectives.Components;
+
+/// <summary>
+/// Sets the target for <see cref="TargetObjectiveComponent"/> to a random traitor.
+/// </summary>
+[RegisterComponent, Access(typeof(KillPersonConditionSystem))]
```
deltanedas: remove access, it has no fields

Content.Server/DeltaV/Objectives/Components/PickRandomTraitorComponent.cs
```diff
@@ -0,0 +1,11 @@
...
+public sealed partial class PickRandomTraitorComponent : Component
```
deltanedas: `: Component;` and remove {}

Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            // Assume it will be a valid traitor.
+            validTraitorMinds.Add(traitor.Id);
+
+            // Going through each of OUR objectives.
+            foreach (var objective in args.Mind.Objectives)
+            {
+                // If one of OUR objectives already targets a traitor, remove them from the vaild list.
+                if (TryComp<TargetObjectiveComponent>(objective, out var targetComp) && targetComp.Target == traitor.Id)
+                    validTraitorMinds.RemoveAt(validTraitorMinds.Count - 1);
```
deltanedas: this could double-remove, have it just not add if there are duplicates

deltanedas (changes requested):

Content.Server/DeltaV/Objectives/Systems/KillFellowTraitorObjectiveSystem.cs
beck-thompson: Yeah that makes a lot more sense! Let me know if that looks good

beck-thompson:
Also, forgot to mention, but there issue with the standard way the traitor objectives are assigned (If there are no valid targets it crashes) so you gotta either fix that (Just add an if statement to check if its empty and if it is, cancel the objective) or like be smart when doing the yaml values so it never tries to assign a kill objective when there are no targets

deltanedas (dismissed):

Lyndomen (dismissed):
Direction approved, awesome objective. Our only **suggestion** is to bump the weight a little, we would like to see it slightly more. (Up to .2 maybe)

deltanedas:
if its too common on 0.2 it can be lowered to 0.15 or 0.1 later

jink23:
Meh I don't think this is a good thing, before when you had an RNG chance to roll another syndie in a kill objective it could ICly be viewed as 'oh syndies accidentally targetted a fellow aligned agent since the syndicate is a conglomerate and doesn't 'always' work in unison' Now it's specifically 'lmao syndies kill each other because why not' which seems really dumb imo

Lyndomen:
> Meh I don't think this is a good thing, before when you had an RNG chance to roll another syndie in a kill objective it could ICly be viewed as 'oh syndies accidentally targetted a fellow aligned agent since the syndicate is a conglomerate and doesn't 'always' work in unison' Now it's specifically 'lmao syndies kill each other because why not' which seems really dumb imo

Thank you for the feedback, but per what little lore this is, the syndicate is a very loose group of corporations both working for and against one another. They have always been allowed and encouraged to steal/kill one another, this is nothing terribly new. In terms of the rationale for why they have to die, that's up to the players to roleplay.

jink23:
> > Meh I don't think this is a good thing, before when you had an RNG chance to roll another syndie in a kill objective it could ICly be viewed as 'oh syndies accidentally targetted a fellow aligned agent since the syndicate is a conglomerate and doesn't 'always' work in unison' Now it's specifically 'lmao syndies kill each other because why not' which seems really dumb imo
> 
> Thank you for the feedback, but per what little lore this is, the syndicate is a very loose group of corporations both working for and against one another. They have always been allowed and encouraged to steal/kill one another, this is nothing terribly new. In terms of the rationale for why they have to die, that's up to the players to roleplay.



Eh all this does is generate more antag infighting specifically which really doesn't help with the antagonists goals of 'driving' the round as is, if they're just busy murdering each other then I don't see how they will impact the round as intended if one is just dead and the others objectives are just 'kill fellow traitor and escape'

Like at that point just make an "NT Counter agent" whose whole purpose is to secretly hunt syndicate agents in a more 'vigilante' sense

Kr8art:
Personally I also got kinda mixed feelings about this. Teaming up with other traitors was already decently rare (at least in my experience) and it's also one of the things that made it arguably more fun. 

> some really annoying and (In my opinion) "metagamey" team ups where traitors just spam codewords in the general radio

I don't even think this necessarily disincentives codeword spam on the radio, unless you have this objective and want to keep your status as traitor secret from your target. For everyone else there really isn't a reason to not spam your codewords since any rival agents who have you as target already know your name and the fact you're an agent. 

The things its gonna dissuade are secretive meetups which can get you killed easily. The help/keep alive objectives are also gonna be harder as people could pretend to have those objectives when they're really here to kill you.

Also, players who are more on the powergamey side of things are probably going try to RR traitor targets as quickly as possible to get a higher chance of cashing in on a bunch of unspent TC. Imagine finally rolling traitor for the first time in ten or so rounds only to get gibbed fifteen minutes in because you ended up as a kill target and they wanted your TC.

beck-thompson:
> Teaming up with other traitors was already decently rare (at least in my experience) and it's also one of the things that made it arguably more fun.

I personally am seeing the opposite! Almost every round I play traitors are teaming up like crazy. Like I said, some of the time this is great for RP and fun, but there are also a lot really lame meet ups where its just buy the big loot box and do nothing else. The goal of the objective is to just make it so that traitors think a little before teaming up and don't have free trust with each other.

> if they're just busy murdering each other then I don't see how they will impact the round as intended if one is just dead and the others objectives are just 'kill fellow traitor and escape'

This shouldn't be a thing because the objective is supposed to be rare. It shouldn't be coming up that much, and if this is happening a lot, then we should decrease the probability.

Finally, if you guys have any better solutions to the issue of traitors teaming up constantly, I'm all ears! Its a hard problem and this is just one solution and if you have more I'll totally to add them.

deltanedas:
before this theres no reason not to have "apes together strong" every traitor on the server teaming up, now there can be some distrust when you find a traitor, because theres a small chance they might be there to kill you (after uplink locking was added nobody hunted other traitors for tc)

PR #2496, opened 2024-12-21
upstream merge 21/12

deltanedas:
## Notable changes
- man o war shuttle with 4 free agents @DeltaV-Station/direction
- **holopads real** including the bugfixes
- **atmospheric pipe monitor computer real**
- advanced SMESes for mapping
- inverted lizard plushie that makes you !heW
- godawful buttons in top left are fixed
- author of the clearly nuclear music yoinked it, rip
- using new holy water and holy damage from upstream, had to migrate every HolyWater reagent to Holywater (thats the 2 lines of map changes)
- vox detectives dont get their vest on the floor

christmas things:
- present anomaly for farming presents
- smite cranberry

## Excluded changes
- ame power change since we have our own ame curve
- revenant damage modifier is the same as before

**Changelog**
:cl: Santa
- add: Merged the Christmas cheer from upstream. Ho ho ho!


ThataKat:
Direction reviewing, 1-2 days

Lyndomen (approved):
Good, man o war we're gonna keep an eye on but should be fun. Velcro wants us to keep an eye out for any barricades if bariccades are already mapped over a window, i think the only place that might be the case is pebble engi maints

PR #2518, opened 2024-12-23
more harmless critters from events

deltanedas:
## About the PR
there can now be a *lot* of mice, especially devastating to the kitchen
mothroaches and cockroaches get less of a boost since theyre a bit more special

mice can go up to 14-21 on high pop and 10-15 on lowpop, others are lower

## Why / Balance
:mouse: Piep!

## Media
look at them go... (spawned 2 each, outdated)

## Breaking changes
**Changelog**
:cl:
- tweak: More mice and mothroaches can come out of vents, a lot more...


Kr8art:
Chef mains in shambles, if there were any in the first place.

deltanedas:
john chef scrambling to lock away all the food only for it to be even hungrier xenos:

ThataKat:
Direction looking this over, response forthcoming

Stop-Signs:
i think with all the recent chef buffs that this is fine

ThataKat:
Initial direction request is to reduce the cap to around 20ish mice, but otherwise looks good to merge! Numbers can always be tweaked later if the mouse tide needs more or less power

deltanedas:
new numbers are 10-15 mice at lowpop and 14-21 at 80p
mothroaches slimes etc have lower numbers and scale less too

at lowpop instead of getting only 1 mouse its now capped to the base min and max, so you always get minimum 10-15 mice
also minimum 2-3 of other mobs like slimes so its not just 1 slime on deadpop staring at you menacingly, and the player scaling only kicks in when it goes past the base min/max

deltanedas:
cap was adjusted to ~20 :trollface:

PR #2519, opened 2024-12-23
Precog fixes

SolStar2:
## About the PR
Fixed precog results not being cached on init, resulting in result messages not being found. Empty messages are no longer sent when a result message is not found.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Precognition results not appearing when power is used.


deltanedas (approved):

deltanedas:
not having results was also a fix pushed to wrong branch :trollface:

PR #2543, opened 2024-12-27
Diona immune to FTL Knockdown

dvir001:
## About the PR
Give diona a stronger grip when getting hit by FTL jumps, allowing them to just stand normally without the need to sit.

## Why / Balance
Diona already slow, might as well give them an extra edge.

## Technical details
Added a new comp from Frontier that make entity skip the FTL knockdown event.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- add: Diona are now immune to the force of an FTL jump.


deltanedas (approved):

Lyndomen (approved):
Fantastic change, thanks for the port

PR #2554, opened 2024-12-29
Add missing roles to Secure Cabinet

FieldCommand:
## About the PR

A tweak to add Psionic Mantis and Psychologist

## Why / Balance

I simply forgot to add them when I made the original PR.
And I think Psychologist should have one, and Mantis was added due to request.

## Requirements

- [x] It does not require an ingame showcase.

**Changelog**


Radezolid:
I don't think a changelog is required as the change isn't user facing.

FieldCommand:
> I don't think a changelog is required as the change isn't user facing.

Fair, makes sense. I will change that then.

IamVelcroboy (approved):

PR #2562, opened 2024-12-30
Colin Map Update - Holopads

Colin-Tel:
## About the PR

Adds holopads and cargo ass
## Media

2lazy
## Requirements

**Changelog**

nah I'm good

### DIFF
--- a/Resources/Maps/asterisk.yml
+++ b/Resources/Maps/asterisk.yml
@@ -82,11 +82,11 @@ entities:
           version: 6
         -1,-1:
           ind: -1,-1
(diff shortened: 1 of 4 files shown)


IamVelcroboy (approved):
- Asterisk needs long range bridge and cargo holopads
- You might want to move your TEG sensors to just before they enter the TEG, after they've been cooled/heated (both maps)
- I guess it's optional but you might want consider holos on Chibi too simply because they can be used to make calls by players too. Long range Cargo and Bridge for communicating with shuttles maybe
image
image

PR #2575, opened 2025-01-01
fix config being fake

deltanedas:
## About the PR
hardcoded nonsense that i cant find so it stays DeltaV for now


Axionyxx:
> ## About the PR
> hardcoded nonsense that i cant find so it stays DeltaV for now

They aren't hard coded, config presets are specified in the server config

deltanedas:
yes i found that out minutes later, still annoying that theres no test for it

PR #2579, opened 2025-01-01
Moved HOPs access configurator to the administrative assistant starting gear

Radezolid:
## About the PR
Removed the access configurator from the HOP's locker and gave one to the admin assistant starting gear.

## Why / Balance
HOP being a service head of staff doesn't really use the access configurator, on the other hand, the administrative assistant would make more sense to be able to assist with changing accesses on doors or lockers should it be needed.

## Technical details
2 YAML line change.
Touches upstream `heads.yml`.

## Media
N/A

## Requirements

## Breaking changes

None.

**Changelog**

No CL as the administrative assistant isn't public to the players yet.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
@@ -118,7 +118,7 @@
     children:
     - !type:NestedSelector # DeltaV
       tableId: LockerFillHeadOfPersonnelDeltaV
-    - id: AccessConfigurator
+    #- id: AccessConfigurator # DeltaV - They don't really need it, moved to the administrative assistant starting gear.
     - id: BoxEncryptionKeyPassenger
     - id: BoxEncryptionKeyService
     - id: BoxHeadset
--- a/Resources/Prototypes/_DV/Roles/Jobs/Command/administrative_assistant.yml
+++ b/Resources/Prototypes/_DV/Roles/Jobs/Command/administrative_assistant.yml
@@ -27,6 +27,7 @@
     belt: BoxFolderClipboard
     id: AdminAssistantPDA
     pocket1: RubberStampAdminAssistant
+    pocket2: AccessConfigurator # To assist with secure closets access configurations / airlock configs should a head need it.
   storage:
     back:
     - Flash


Lyndomen:
Do CEs get one? I forgor

Radezolid:
> Do CEs get one? I forgor

CE and HOP have one each in their lockers, with this it would be only CE and the admin assistant only if they spawn since they don't have a guaranteed locker.

Lyndomen (approved):

FieldCommand:
To me, the hop having an access configurator seems to be a remnant from the time that they were considered second in command. Since HoP are now only service-related and head of said service department, it does not make sense for them to have one of two access configurations to me. I have never needed it as a hop, and more often than not when it has been needed, it has been the captain coming and taking it anyway. The CE having one does, as they are the ones most likely to need to reconfigure new airlocks. So moving it to the new administrative assistant to lend out to heads that need it does make sense, but that is also very dependent on mappers mapping the locker for the administrative assistant. But I would also like to see the captain getting one. And I don't even think the access configurator is a steal objective.

Avalon-Proto:
>  So moving it to the new administrative assistant to lend out to heads that need it does make sense, but that is also very dependent on mappers mapping the locker for the administrative assistant. 

This is moving it to spawn in their inventory round start. The locker is an optional thing that is not needed w/ extra things

HTMLSystem:
As a HoP main I would like for the access config to stay with the HoP and the assistant could just get another one of their own? I've needed to use it a few times as HoP and it feels like we're taking away the very few things HoP CAN help with atp :(.

deltanedas (approved):

Radezolid:
> To me, the hop having an access configurator seems to be a remnant from the time that they were considered second in command.

It was.
>  very dependent on mappers mapping the locker for the administrative assistant.

Administrative assistants spawn with it on their pocket, not on locker exactly for that reason.

>But I would also like to see the captain getting one.

No, the captain shouldn't have one, they should talk with the rest of the station to accomplish what they want as that's their role on station, to delegate.
>And I don't even think the access configurator is a steal objective.

It isn't.

> As a HoP main I would like for the access config to stay with the HoP and the assistant could just get another one of their own? I've needed to use it a few times as HoP and it feels like we're taking away the very few things HoP CAN help with atp :(.

HoP is head of service, it doesn't make sense for them to have that tool.

PR #2581, opened 2025-01-02
Harpy shoes 1984

Monotheonist:
## About the PR

This simple two line change nukes any semblance of there having been shoes on harpies, by giving them the diona inventory template. Before you ask, no, this wasn't my idea, _it was originally like this where they're being maintained currently (EE,) it just got changed because of a technical issue when it came here._
## Why / Balance

One thing I've always noticed about Harpies here is that they were never intended to wear shoes. There are 2 to three pixels sticking out of the sprite when you look at it on the server, and I've been unable to play harpy after noticing this. It pisses me off so much. Of course, this is the nuclear option; there's another solution, like making a displacement map, which might make more sense, but I feel like this also adds a unique and kind of evil downside of having to make them be like dionae and suffer with glass shards and other things, so I am doing this first to provoke conversation of if this is too much.
## Technical details

Two line yaml change, gives them the template id of Diona, which is shoeless. This is to prevent the extra bloat of bringing back the digitigrade template yaml, but as a side effect, anything that happens to that template will reflect on harpies, too. Just something for Deltanedas to note.

## Requirements

- [X] I have added media to this PR or **it does not require an ingame showcase.**

**Changelog**

:cl:
- tweak: Harpies can no longer wear shoes. The moths ate them all.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/Species/harpy.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/Species/harpy.yml
@@ -128,6 +128,7 @@
     weightlessAcceleration: 2.5
   - type: Inventory
     speciesId: harpy
+    templateId: diona # Reusing this instead of digitigrade
     displacements:
       jumpsuit:
         sizeMaps:
@@ -157,6 +158,7 @@
     species: Harpy
   - type: Inventory
     speciesId: harpy
+    templateId: diona # Reusing this instead of digitigrade
     displacements:
       jumpsuit:
         sizeMaps:


gluesniffler:
If you're gonna take away their ability to wear shoes, you need to also give them a buff in return. The reason the shoes thing got reverted by Milon was twofold:

1. It broke a few things because of how Adei ported it.
2. There were a lot of complaints from what little harpy players we have since its a nerf to an already weak species.

Additionally, not being able to wear shoes entails a hard weakness to anything involving slips, with not much to show for that drawback.

Monotheonist:
I'm fairly sure this need for a buff was negated by EE by making them immune to steptrigger things like glass (thereby giving them noslips,) which would suddenly make them absurdly OP for security players.
I'm willing to give them a buff, but I don't have any ideas, so if anyone could suggest something to go along with this that would actually improve the harpy experience (cold resistance, if they don't have that, or like, faster punch speed) I will gladly buff them (or, once again get the displacement map thing done.)

gluesniffler:
The EE steptrigger "buff" was accompanied with a selective refactor for the system, which has them set to be immune only to landmines, glass, and mice traps. 

So you have a couple routes from here, either:

- Give them minor easy buffs in YML
- Port the rest of the changes from EE as in this PR https://github.com/Goob-Station/Goob-Station/pull/1052/
- Implement the displacement map for harpies made by @AstrodogeDX. [This port PR for Avali/Resomi](https://github.com/Fansana/floofstation1/pull/372) seems to have examples of how to use them. 
image

Lyndomen:
I haven't consulted Direction, but my personal opinion is that a displacement map would be best for this situation. 

I am not the biggest fan of steptrigger immune, it doesn't make the most in game sense.

Monotheonist:
This is GREAT feedback, thank you guys for contributing to it (as well as people I asked on discord about this!!) From my understanding displacement maps does seem to be the way to go, so I'm gonna close this and open up a seperate and more comprehensive PR at some point.

PR #2595, opened 2025-01-03
Admin assistant feedback popups

beck-thompson:
## About the PR

Add feedback popups for the new job!

## Technical details

Added a new system to watch for the end of round and send the popups to the assistants themselves, and command.

## Requirements


beck-thompson:
I don't think the discord channel is set up yet so that needs to be done first obviously!

Axionyxx (commented):
Hardcoded bad, you should make this a generic system configured fully through yaml

deltanedas (dismissed):

deltanedas:
it will be reverted after some time anyway doesnt matter

beck-thompson:
Looks like the relay is set up on the servers and discord now! Should be good to merge whenever they are added to all the maps and stuff

deltanedas:
wrt antimov see https://github.com/DeltaV-Station/Delta-v/pull/2329, looking for headmin input here

beck-thompson:
> wrt antimov see https://github.com/DeltaV-Station/Delta-v/pull/2329, looking for headmin input here

I think this is for #2643 right? And yeah, if the board needs to be changed let me know I don't really know the borg laws that well

deltanedas:
real and true dementiaposted

beck-thompson:
Checked it again with the maps (Now that they actually have assistants) and it seems to work properly!

beck-thompson:
People seem to like the job so this isn't necessary anymore

PR #2598, opened 2025-01-03
Remove "keep alive" traitor objective

beck-thompson:
## About the PR

Removed the keep alive traitor objective.

## Why / Balance

I think this one of the least fun traitor objectives. To complete it:
1.) Your target can't go to cryo.
2.) Your target can't have DAGD or an objective that is super risky. A lot of traitors like to go out in a ball of flame!

If your target falls into one of those categories, it is basically impossible to complete.

However, if they don't, its basically a free objective that requires no input from you! Removing it will just force the progress objective to be chosen for social which I think is OK. Its basically keep alive but much more interactive and fun!

**Changelog**

:cl:

- remove: Removed keep fellow traitor alive traitor objective.

### DIFF
--- a/Resources/Prototypes/Objectives/objectiveGroups.yml
+++ b/Resources/Prototypes/Objectives/objectiveGroups.yml
@@ -50,7 +50,7 @@
 - type: weightedRandom
   id: TraitorObjectiveGroupSocial
   weights:
-    RandomTraitorAliveObjective: 1
+    # RandomTraitorAliveObjective: 1 DeltaV removed for being a less fun version of progress objective.
     RandomTraitorProgressObjective: 1
 
 #Thief groups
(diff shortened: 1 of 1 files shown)


beck-thompson:
One more note, if this is decided a good idea by direction, it might be a good idea to decrease the social objective chance slightly. About 1/4 of of the objectives will be for traitors to help each other in this PR as is.

Lyndomen:
Under direction review, 24 hours

Lyndomen (changes requested):
Requesting TraitorObjectiveGroupSocial weight to be 0.5, then all set!

Lyndomen (approved):

deltanedas:
traitor hiveminds from everyone having 3 keep alive objectives in shambles

PR #2629, opened 2025-01-06
Remove glue requirement from prescription HUDs and departamental glasses

sowelipililimute:
## About the PR

This PR removes the requirement for space glue in order to craft the prescription HUDs and departamental glasses, such as the prescription security glasses and the prescription medical hud.

## Why / Balance

Space glue as-is is a somewhat rare item in the game, and locking one of the basic visual cues and protective equipment for security or medical behind obtaining this rare item is somewhat annoying, as in the best case scenario, it amounts to a chore to get a cheap purchase from logi, and in the worst case scenario, you aren't going to have that important equipment without sacrificing your vision for the entire shift.

This doesn't produce particularly engaging gameplay in either scenario.

Via the principle of less is more, it follows that removing glue from the recipe construction makes sense, instead of adding more ways to get glue in order to handle this one specific scenario.

## Technical details
- Removed space glue drink requirement from the crafting graphs

## Media
grafik

## Requirements

**Changelog**

:cl:
- tweak: Prescription departamental glasses and HUDs no longer require glue to obtain

### DIFF
--- a/Resources/Prototypes/_DV/Recipes/Construction/Graphs/clothing/prescription_departamental_glasses.yml
+++ b/Resources/Prototypes/_DV/Recipes/Construction/Graphs/clothing/prescription_departamental_glasses.yml
@@ -21,12 +21,6 @@
             - material: Cable
               amount: 5
               doAfter: 5
-            - tag: DrinkSpaceGlue
-              name: space glue
-              icon:
-                sprite: Objects/Consumable/Drinks/glue-tube.rsi
-                state: icon
-              doAfter: 5
     - node: prescsecglasses
       entity: ClothingEyesPrescriptionSecurityGlasses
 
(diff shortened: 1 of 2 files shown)


FieldCommand:
Maybe an idea would be to add space glue to the tech vendor, like say, 5 bottles of them. I do agree that space glue is somewhat rare, but I also think it should be a part of the recipe.

Lyndomen:
meowdy, just a thought here, but what if we made space glue craftable via chem?

Lyndomen:
I generally agree with "Less is More", but if we are deepening the web of connections between departments, that's pretty good in my book. I'm also not adverse to this change, but would ask that the construction description be updated to omit the usage of glue

deltanedas:
it just shouldnt use the space glue tube item (which can be empty and work just fine), it should use reagent construction graph conditions like IEDs do

deltanedas (approved):
visually impaired rise up

PR #2631, opened 2025-01-06
Speech is relayed by holopad holograms (#33978)

deltanedas:
## About the PR
fixes holopads, its the only thing missing

## Media

**Changelog**
:cl:
- fix: Fixed holopads not letting the AI hear anything.


chromiumboy:
I don't think this PR will fix the problem your having. It only changes which entity speech is played from (from the holopad to the the holopad's hologram), it doesn't touch how the AI hears things

deltanedas:
it still worked, and was the only thing missing between here and upstream

chromiumboy:
Well, I can't argue with results, but I have no idea why it worked :P

PR #932, opened 2024-03-06
Psionic Refactor/Rework Part 1: All The Powers, Latent Psychics.

VMSolidus:
## About the PR
This is Part 1 of 3 for my project to completely rework the entire Psionic systems, and everything else adjacent to them. Since the Psionic Systems are so all encompassing, I have elected to split the Refactor into three distinct refactors, each focused on one specific slice of Psionics. 

Part 1: Powers
Part 2: Golemancy & Psychic Threats
Part 3: Glimmer Systems

This PR addresses a vast number of issues with nearly every psychic power in the game, making improvements to player experience across the board. I am also making system improvements in the code behind the scenes, with new additions that allow contributors to create all new types of items, entities, and powers. These systems are being made more interactive with each other. 

Some of the powers are getting new use-cases being added, while others are getting substantial reworks. Additionally, Pyromancy & Noospheric Zap are being completely reworked into powers that can actually be used by players. There will no longer be such a thing as a power intended solely to be used by monsters. If a new psychic power is to be added, it must be obtainable by players through the same means as all other powers. Therefore it must be designed to be used by players. 

That being said, with the addition of some new systems, it is still possible to have monsters cast powers far more powerfully than players, or unlock use-cases for powers that are otherwise difficult for players to obtain. This is accomplished through the new system of Psychic Amplification & Dampening.

## Latent Psychic
Latent Psychic is a new trait that is intended to serve as an opt-in to Psionics during character creation. This essentially allows players to decide in advance if they want to interact with the game's Psionics system, effectively letting players roleplay around psionics without needing to be concerned about random chance, or to select a role that starts psionic by default. Latent Psychic takes the place of the original PotentialPsionicComponent, and has a few changes to it from the original system:

- [x] Latent Psychics can hear partial snippets of Telepathy, effectively having an always-on "psychic dreams" mechanic. They cannot however speak telepathically until they obtain their first power.
- [ ] Latent Psychics can(pending implementation of Golemancy in Part 2) be sacrificed by the Chaplain, albeit for a significantly more minor amount of glimmer reduction, and without any chance at obtaining Items of Power.
- [ ] Latent Psychics are hunted by glimmer intruders such as Wisps. 
- [x] Latent Psychic is a full prerequisite for obtaining psionic powers.
- [x] If the trait is lost, a character can never again become psionic unless they die, and are reincarnated by Metempsychosis.
- [x] For system keeping purposes, roles that start psionic by default are also Latent Psychic
[...]

PR #2667, opened 2025-01-09
Revert "allow cloning living people (#932)"

deltanedas:
## About the PR
accidentally included it so reverted it now

## Why / Balance
it was breaking cloning and not suitable for mrp anyway

**Changelog**
:cl:
- fix: Fixed cloning not working.


Avalon-Proto:
I mean I feel like we should have this as an option for those people who are alive, but are so heavily damaged it's more efficient to get them a new body and transplant their brain into it

deltanedas:
if theyre alive you can print new limbs and just attach them

and i imagine direction wasnt aware of it to begin with

Lyndomen:
correct was not aware and would have killed it, cloning is for deads

PR #2668, opened 2025-01-09
make limb severing bleed instead of bloodloss damage

deltanedas:
## About the PR
title

applies to damage-caused severing (which probably wont matter as youll already be bleeding) and surgical amputation

## Why / Balance
bloodloss damage makes no sense when its heavy arterial **bleeding**
so now it does that, 6 bleeding stacks for limbs and 10 (max) for the head

now when delimbing people you need gauze to keep them from bleeding out, crazy

next step is making incision etc in surgeries make you bleed instead of doing bloodloss damage like lyndo wanted :trollface:

## Technical details
simple enough

## Media
evil surgery and evil execution vs urists

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Losing a limb makes you bleed profusely instead of dealing bloodloss damage.

### DIFF
--- a/Content.Server/Body/Systems/BodySystem.cs
+++ b/Content.Server/Body/Systems/BodySystem.cs
@@ -23,6 +23,7 @@ namespace Content.Server.Body.Systems;
 
 public sealed class BodySystem : SharedBodySystem
 {
+    [Dependency] private readonly BloodstreamSystem _bloodstream = default!; // Shitmed Change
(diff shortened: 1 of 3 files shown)


Avalon-Proto:
Now here's the question- if the limb is severed by an e-sword would they still bleed out

deltanedas:
yes they would :trollface: artery gaming

PR #2676, opened 2025-01-10
Depression RP Trait

LioValkyria:
## About the PR
First ever PR but here we go, added the Depression trait for round start depression meds

## Why / Balance
Although this trait has no effect on gameplay for those that do RP a character having depression, having to go to the CMO or Phych round start for your meds is interesting the first time not every time, and it would not make sense for said employee to arrive on station without at least some meds

## Media
PillCanisterAntidepressants
TraitScreenAntidepressant
## Requirements

## Breaking changes

**Changelog**

:cl:
- add: RP Depression trait

### DIFF
--- a/Resources/Prototypes/_DV/Traits/disabilities.yml
+++ b/Resources/Prototypes/_DV/Traits/disabilities.yml
@@ -30,3 +30,11 @@
   category: Disabilities
   components:
   - type: Unborgable # Automatically gets moved to the brain
+
+- type: trait
+  id: Depression
+  name: trait-depression-name
+  description: trait-depression-desc
+  traitGear: PillCanisterNeurozenium
+  category: Disabilities
+  components: []
--- a/Resources/Locale/en-US/_DV/traits/traits.ftl
+++ b/Resources/Locale/en-US/_DV/traits/traits.ftl
@@ -35,3 +35,6 @@ trait-addicted-desc = You crave the substance, and your thoughts keep drifting b
 
 trait-unborgable-name = Machine Incompatible
 trait-unborgable-desc = Your brain cannot be put into a man-machine interface.
+
+trait-depression-name = Depression
+trait-depression-desc = No mechanical effect. The world is dark but there is a light somewhere, calling to you.


Resources/Prototypes/_DV/Traits/disabilities.yml
```diff
@@ -22,3 +22,12 @@
...
   - type: Pain
+
+- type: trait
+  id: Depression
+  name: trait-depression-name
+  description: trait-depression-desc
+  traitGear: PillCanisterNeurozenium
+  category: Disabilities
+  components:
+  - type:
```
deltanedas: remove these lines

deltanedas (changes requested):

Lyndomen:
maybe good idea to add the low on oxy visual filter, steal from chronic pain

LioValkyria:
> maybe good idea to add the low on oxy visual filter, steal from chronic pain

that would require writing C# which I do not know how at all

Lyndomen:
heard! If you're interested in figuring it out, there is a similar system for Chronic Pain that causes a visual effect as well, it would _probably_  just be a lotta copy pasting. If not- totally okay!

LioValkyria:
Somebody else in the future (maybe even me) will have to tackle that, because after looking at the pain system it's a couple of steps too complicated for a first time doing any of this.

Resources/Prototypes/_DV/Traits/disabilities.yml
LioValkyria: Removing those two lines absolutely breaks everything and it refuses to build, unless I am somehow doing something wrong
I will continue trying to look for a solution but it does work with those two still there.

Lyndomen:
> Somebody else in the future (maybe even me) will have to tackle that, because after looking at the pain system it's a couple of steps too complicated for a first time doing any of this.

no worries!! Thank you for the contribution, I think its great.

Resources/Prototypes/_DV/Traits/disabilities.yml
LioValkyria: fuck it apparently it works to just remove those VSCode just lies to me

deltanedas (dismissed):
rp gaming speedmerge incoming

Colin-Tel:
Tests seem to keep failing, doesn't look like an accident

Lyndomen:
Caught exception while writing TraitPrototype prototype Depression. Exception:
Robust.Shared.Serialization.Manager.Exceptions.NullNotAllowedException: Null value provided for reading but type was not nullable!
Real

deltanedas:
`components: []` gaming then

deltanedas (approved):

PR #2677, opened 2025-01-10
Add Chronic Pain to People Upon Death

Lyndomen:
## About the PR
adding pain on death that persists until you get a new body. painkiller stonks go up

## Why / Balance
value to life, dont die, treat your character as if they want to not die

## Technical details
c# and fucked if i know. draft because its a nightmare 

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Dying now causes your character to suffer from chronic pain, get prescribed painkillers!

### DIFF
--- a/Content.Server/_DV/Pain/PainSystem.cs
+++ b/Content.Server/_DV/Pain/PainSystem.cs
@@ -19,7 +19,7 @@ public sealed class PainSystem : SharedPainSystem
     {
         base.Initialize();
         SubscribeLocalEvent<PainComponent, MapInitEvent>(OnMapInit);
-        SubscribeLocalEvent<PainComponent, MobStateChangedEvent >(HandlePainTrigger);
+        SubscribeLocalEvent<PainComponent, MobStateChangedEvent>(HandlePainDeath);
     }
 
     private void OnMapInit(Entity<PainComponent> ent, ref MapInitEvent args)
@@ -91,9 +91,11 @@ public sealed class PainSystem : SharedPainSystem
         }
     }
 
-    private void HandlePainTrigger(EntityUid uid, PainComponent component, ref MobStateChangedEvent  args)
+    private void HandlePainDeath(EntityUid ent, PainComponent component, ref MobStateChangedEvent args)
     {
-        if (args.NewMobState == MobState.Dead && HasComp<HumanoidAppearanceComponent>(uid))
-            EnsureComp<PainComponent>(uid);
+        if (args.NewMobState == MobState.Dead && args.OldMobState != MobState.Dead && HasComp<HumanoidAppearanceComponent>(ent))
+        {
+            EnsureComp<PainComponent>(ent);
+        }
     }
 }
--- a/Content.Server/_DV/Pain/PainSystem.cs
+++ b/Content.Server/_DV/Pain/PainSystem.cs
@@ -19,7 +19,7 @@ public sealed class PainSystem : SharedPainSystem
     {
         base.Initialize();
         SubscribeLocalEvent<PainComponent, MapInitEvent>(OnMapInit);
-        SubscribeLocalEvent<PainComponent, MobStateChangedEvent>(HandlePainDeath);
+        SubscribeLocalEvent<HumanoidAppearanceComponent, MobStateChangedEvent>(HandlePainDeath);
     }
 
     private void OnMapInit(Entity<PainComponent> ent, ref MapInitEvent args)
@@ -91,11 +91,9 @@ public sealed class PainSystem : SharedPainSystem
         }
     }
 
-    private void HandlePainDeath(EntityUid ent, PainComponent component, ref MobStateChangedEvent args)
+    private void HandlePainDeath(Entity<HumanoidAppearanceComponent> ent, ref MobStateChangedEvent args)
     {
-        if (args.NewMobState == MobState.Dead && args.OldMobState != MobState.Dead && HasComp<HumanoidAppearanceComponent>(ent))
-        {
-            EnsureComp<PainComponent>(ent);
-        }
+        if (args.NewMobState == MobState.Dead)
+            EnsureComp<PainComponent>(args.Target);
     }
 }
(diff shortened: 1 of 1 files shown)


Bonktrauma:
I would argue excluding diona from this,  considering they can reform themselves at will and i believe in ss13 servers with pain have them not feeling pain?

Lyndomen:
> I would argue excluding diona from this, considering they can reform themselves at will and i believe in ss13 servers with pain have them not feeling pain?

new body no pain, same as cloning

Bonktrauma:
> > I would argue excluding diona from this, considering they can reform themselves at will and i believe in ss13 servers with pain have them not feeling pain?
> 
> new body no pain, same as cloning

What about the second point

Lyndomen:
> > > I would argue excluding diona from this, considering they can reform themselves at will and i believe in ss13 servers with pain have them not feeling pain?
> > 
> > 
> > new body no pain, same as cloning
> 
> What about the second point

second point that some ss13 severs have it? their lore isnt ours :blunt:

Content.Server/_DV/Pain/PainSystem.cs
```diff
@@ -87,4 +94,13 @@ public override void Update(float frameTime)
             component.NextUpdateTime = curTime + TimeSpan.FromSeconds(1);
         }
     }
+
+    private void HandlePainTrigger(EntityUid uid, PainComponent component, TriggerEvent args)
```
deltanedas: subscribe to broadcast `MobStateChangedEvent` and if the new state is dead + the mob is a humanoid add pain

Lyndomen:
i hear complaints that people that dont opt in to pain might not like the blurred screen effect, does anyone have any ideas about similar visual effects that might not cause issues?

FieldCommand:
An idea, but not sure how it would look, but a combination of the nearsighted trait visual and a pulsing white fog ring (like the red one when you are wounded, just white). Or one of those two.

Content.Server/_DV/Pain/PainSystem.cs
Lyndomen: ok this has been changed, it isnt firing off (or erroring out :trollface: ) ty for help

Smugman:
While I like the idea of making death have some negative consequence, I think this is a bit too punishing, will be severely upsetting to many players and will make chemists want to die.

Currently, during a 2 hour shift i can expect to need 200-300u of Soretizone to provide 2-3 refills. With this, I would reasonably have to expect to make Painkillers for around a Third of the Station, who all dont have a "buffer" of pills and will want to get rid of the obnoxious effect immediatly.

Plus, in the case of mass crew-death like during particularly loud nukies, I think this would get even worse. 

However, I do like how this would make "teach a lesson" more punishing to be the victim of.

I think that adding some kind of penalty to needing defibrilation is good, but having it be immediatly this bad and only really resolvable via perpetual upkeep from chem or cloning, which is generally only available sometimes (epi 💤) and in limited amounts (biomass, but that might be a skill issue).

Plus, the Idea of people demanding to be swapped into a full new body because they died once seems a bit too out-there for my tastes.

Lyndomen:
Painkiller crate in logi real

HTMLSystem:
Perhaps there should be a way to turn the chronic pain filter off like you can with colorblindness, as the effect can cause some people migraines and worse. Yes, you should value life, but there are some times where you just die even though you were being careful. And ir would suck to have a migraine inducing filter over something you couldn't control.

HTMLSystem:
That or you can make it temporary, which I feel would make more sense. Whatever you wanna do.

Lyndomen:
Ok so I'm hearing two issues. 

One, the visual effect can be an accommodation issue. I can check on mitigating that a bit. 

Two, it might be too penalizing and should have a permanent solution that isn't just "get new body lol". 


How do people feel about something like a pain inhibitor surgical implant? Rimworld joywire time

Content.Server/_DV/Pain/PainSystem.cs
```diff
@@ -17,6 +19,7 @@ public override void Initialize()
     {
         base.Initialize();
         SubscribeLocalEvent<PainComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<PainComponent, MobStateChangedEvent>(HandlePainDeath);
```
deltanedas: remove component

Content.Server/_DV/Pain/PainSystem.cs
```diff
@@ -87,4 +90,12 @@ public override void Update(float frameTime)
             component.NextUpdateTime = curTime + TimeSpan.FromSeconds(1);
         }
     }
+
+    private void HandlePainDeath(EntityUid ent, PainComponent component, ref MobStateChangedEvent args)
```
deltanedas: remove ent and component

Content.Server/_DV/Pain/PainSystem.cs
```diff
@@ -87,4 +90,12 @@ public override void Update(float frameTime)
...
+    {
+        if (args.NewMobState == MobState.Dead && args.OldMobState != MobState.Dead && HasComp<HumanoidAppearanceComponent>(ent))
```
deltanedas: dont need to check old state, replace ent with the event's entity

deltanedas (changes requested):

Content.Server/_DV/Pain/PainSystem.cs
Lyndomen: had already done changes but didnt see this or commit pro

Lyndomen:
status update: have to find a way to make a better shader, and probably apply another status effect. chronic pain aint it.

Emily9031:
What if instead of a visual effect (Yes the chronic pain one can be very disorienting and almost headache inducing) each time a person is revived part of their health is permanently removed? I would maybe do a percentage, so like the first time you die, you lose 10% of your max health, second time, you lose 10% of your already lowered max, which if my math is mathing means you lose a smaller amount of health each time so you don't just become completely dead.

PR #2685, opened 2025-01-10
Make chloral hydrate better suited for use as anesthesia

sowelipililimute:
## About the PR
Chloral hydrate is now cheaper and metabolises faster.

## Why / Balance
As it is, chloral hydrate has two issues that prevent it from being good to roll out in surgical wards:
- it metabolises too fast (10s per 5u isn't a lot, especially considering that surgery bleeding will drain this faster)
- it requires too many source ingredients to make, with an astounding 5 ingredient to 1 result ratio that is by far the worst in the game, beating out necrosol (5 to 2) or cognizine/nocturine (3 to 1) for what is an obviously less effective and more situational chemical than those

## Technical details
- chloral hydrate has a slower metabolism rate
- adjust the reaction that produces chloral hydrate to a 5:3 ratio

## Requirements

**Changelog**

:cl:
- tweak: Chloral hydrate is cheaper and lasts longer

### DIFF
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -55,6 +55,7 @@
   physicalDesc: reagent-physical-desc-nondescript
   metabolisms:
     Poison:
+      metabolismRate: 0.25 # DeltaV: chloral hydrate should last longer for surgery
       effects:
       - !type:Emote
         emote: Yawn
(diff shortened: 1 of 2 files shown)


Lyndomen:
Direction review, 24 hours

Lyndomen (approved):
Approved, thank you

PR #2698, opened 2025-01-11
Drunkard accent

mazziebot:
## About the PR

Adds a "drunkard accent", a trait that makes a character's speech slurred like they're drunk.

Port of https://github.com/Fansana/floofstation1/pull/476

## Why / Balance

There are a few alcoholic characters that would benefit from being "constantly drunk" without the actual status effect. This could be a good effect for other characters that have a speech impediment.

## Technical details

This uses a modified version of SlurredSystem's Accentuate for the trait with hardcoded values and removing the burping and "huh" additions. The functionality could be extracted into a static method, however it would require changing code in non-DeltaV code when this works perfectly fine; reducing changes to the codebase is preferable.

If the character is drunk past a certain point, the SlurredSpeech effect will take precedence over DrunkardAccent.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added a "drunkard accent", a trait that makes a character's speech slurred like they're drunk.

### DIFF
--- a/Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
+++ b/Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
@@ -8,10 +8,9 @@ using Robust.Shared.Timing;
 
 namespace Content.Server._DV.Speech.EntitySystems;
 
-
 public sealed class DrunkardAccentSystem : EntitySystem
 {
-    [Dependency] private readonly StatusEffectsSystem _statusEffectsSystem = default!;
+    [Dependency] private readonly StatusEffectsSystem _statusEffects = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
 
@@ -64,17 +63,17 @@ public sealed class DrunkardAccentSystem : EntitySystem
         return sb.ToString();
     }
 
-    private void OnAccent(EntityUid uid, DrunkardAccentComponent component, AccentGetEvent args)
+    private void OnAccent( Entity<DrunkardAccentComponent> ent, ref AccentGetEvent args)
     {
         // Drunk status effect calculations, ripped directly from SlurredSystem B)
-        if (!_statusEffectsSystem.TryGetTime(uid, SharedDrunkSystem.DrunkKey, out var time))
+        if (!_statusEffects.TryGetTime(ent.Owner, SharedDrunkSystem.DrunkKey, out var time))
         {
             args.Message = Accentuate(args.Message, 0.25f);
         }
         else
         {
             var curTime = _timing.CurTime;
-            var timeLeft = (float) (time.Value.Item2 - curTime).TotalSeconds;
+            var timeLeft = (float)(time.Value.Item2 - curTime).TotalSeconds;
             var drunkScale = Math.Clamp((timeLeft - 80) / 1100, 0f, 1f);
 
             args.Message = Accentuate(args.Message, Math.Max(0.25f, drunkScale));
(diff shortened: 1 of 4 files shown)


Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
```diff
@@ -0,0 +1,83 @@
...
+                _ => $"{character}{character}{character}",
+            };
+
+            sb.Append(next);
+        }
+
+        return sb.ToString();
+    }
+
+    private void OnAccent(EntityUid uid, DrunkardAccentComponent component, AccentGetEvent args)
```
deltanedas: use `Entity<DrunkardAccentComponent> ent`

Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
```diff
@@ -0,0 +1,83 @@
...
+using Content.Shared.StatusEffect;
+using Robust.Shared.Random;
+using Robust.Shared.Timing;
+
+namespace Content.Server._DV.Speech.EntitySystems;
+
+
+public sealed class DrunkardAccentSystem : EntitySystem
+{
+    [Dependency] private readonly StatusEffectsSystem _statusEffectsSystem = default!;
```
deltanedas: rename to _statusEffects

Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
```diff
@@ -0,0 +1,83 @@
...
+using Content.Server._DV.Speech.Components;
+using Content.Server.Speech;
+using Content.Shared.Drunk;
+using Content.Shared.StatusEffect;
+using Robust.Shared.Random;
+using Robust.Shared.Timing;
+
+namespace Content.Server._DV.Speech.EntitySystems;
+
+
```
deltanedas: only 1 blank line here not 2

deltanedas (changes requested):

Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
mazziebot: ~~instead of `EntityUid uid, DrunkardAccentComponent component`, replace both with the single `Entity<DrunkardAccentComponent> ent`?~~
nvm, that doesn't compile... what was intended by this? seems like replacing EntityUid with this doesn't compile, neither does the inverse

Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
mazziebot: Had to disable the Rider formatting for this, but it is fixed

Content.Server/_DV/Speech/EntitySystems/DrunkardAccentSystem.cs
Mnemotechnician: Replace with `private void OnAccent(Entity<DrunkardAccentComponent> ent, ref AccentGetEvent args)`, then replace references to `uid` with `ent.Owner` and references to `component` with `ent.Comp`.

deltanedas: you only need ent.Owner if passing to a `Entity<T?>` because of funnyness, itll automatically cast to EntityUid

deltanedas (approved):

ThataKat:
Direction review, 24 hours or less (sorry for being late!)

ThataKat:
Heya, we've thought it over and decided we would prefer not to merge this one. Direction overall would like to move away from forced accents, and this accent in particular can be achieved by drinking (no manual slurring by the player required). 

Thanks for contributing, we hope to see you around in future PRs!

PR #2701, opened 2025-01-11
WIP: Combat hypospray

BarryNorfolk:
WIPWIPWIPWIPWIPWIWPIPWIP

## TODO:

- [ ] Sprites
- [ ] Crafting mechanism
- [ ] Any form of balancing direction wants

## About the PR
Added a new type of hypospray designed for Corpsman and Parameds to help them in fire fights or intense situations where having someone still for a syringe injection is too much.

The combat hypospray works with cartridges that are inserted into it to "fill" it with a given solution. These cartridges are hot swappable and contain 30u of solution. Cartridges are not drinkable, drawable, or refillable so they are used they are thrown away or recycled into glass.

Cartridges are constructed from a cartridge fabricator which takes any chemistry bottle as an input, once the bottle interacts with the machine (much like the seed extractor) then a cartridge will be produced and placed in the users hands if possible. A check is performed to see whether this bottle contains valid "medicinal" reagents and is rejected with a message if it does.

Valid reagents are controlled by the `groupWhitelist` and `reagentWhitelists` fields, the former means you can set things like "All medicines" and the latter means you can add narcotics like ephedrine. This behaviour is overridable by the emag, a fabricator that is emagged will let you put anything into a cartridge.

## Why / Balance
Corpsman, in my eyes at least, are the combat medic attached to the security force to enable them to keep fighting for long periods without have to go all the way back to med and/or lose ground in a fight against an antag. Having them wait for a syringe while being shot at is not viable.
Parameds are less reliant on this but can still provide a useful combat medical role, as well as a fast response unit to try and get someone back on their feet (if they have any after shitmed) and get to the medical bay asap.

## Technical details
Inspired by/taken from: https://github.com/Goob-Station/Goob-Station/pull/953
Ideally we'd re-use as much of the upstream code as possible, so having the cartridges refill the hypospray directly (without letting the hypospray ever draw) is useful.

## Requirements

## Breaking changes

**Changelog**


Lyndomen:
Initial impression is that direction likes this! Some feedback we have is that it should be single shot only with manual reloads, and the cartridges themselves should be fairly large- perhaps 1x2.

BarryNorfolk:
> Initial impression is that direction likes this! Some feedback we have is that it should be single shot only with manual reloads, and the cartridges themselves should be fairly large- perhaps 1x2.

Can you expand on what you mean by single shot with only manual reloads? A full 30u in one large dose or that it might take a bottle of 30u and make 6 x 5u doses?

Lyndomen:
> > Initial impression is that direction likes this! Some feedback we have is that it should be single shot only with manual reloads, and the cartridges themselves should be fairly large- perhaps 1x2.
> 
> Can you expand on what you mean by single shot with only manual reloads? A full 30u in one large dose or that it might take a bottle of 30u and make 6 x 5u doses?

Sure! So what I meant was that ideally this hypospray has no magazine, and must be manually reloaded after every shot. Each shot would be an entire 30u dose, with each dose taking up 1x2 space.

deltanedas:
fell off + should port injector cartridges from goob

PR #2703, opened 2025-01-11
Psi Cultist Antag

leonardo-dabepis:
## About the PR

New antag psi cultist. This PR is not yet finished.
These are the planned features:
- [X] Return of pacifist psionic objectives (raise glimmer, become psionic) but strictly for the psi cultist role.
- [ ] Add more hostile psionic objectives (killing certain epi members, converting crew to the cult (like consensual/through RP rather than instantaneously as with revolutionaries))
- [ ] Add psi cultist gear
- [ ] Add guidebook stuff

Let me know what you think or if you have any ideas of what else I could add.

## Why / Balance

People always say that psionics are the one thing that makes Delta unique, which isn't really true, and they also say that Delta's psionics are very undercooked, which is true. This is meant to hopefully flesh out psionic gameplay a little more and also give more for the mantis to do. Also people will pat me on the back and say "good job" if I add an antag instead of removing an antag :)

## Media

None.... yet ;0

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added psi cultist antag

### DIFF
--- a/Content.Server/Roles/PsiCultistRoleComponent.cs
+++ b/Content.Server/Roles/PsiCultistRoleComponent.cs
@@ -1,0 +1,11 @@
+using Content.Shared.Roles;
+
+namespace Content.Server.Roles;
+
+/// <summary>
+///     Added to mind role entities to tag that they are a psi cultist.
+/// </summary>
+[RegisterComponent]
+public sealed partial class PsiCultistRoleComponent : BaseMindRoleComponent
+{
+}
--- a/Resources/Locale/en-US/prototypes/roles/antags.ftl
+++ b/Resources/Locale/en-US/prototypes/roles/antags.ftl
@@ -36,3 +36,6 @@ roles-antag-thief-objective = Add some NT property to your personal collection w
 
 roles-antag-dragon-name = Space Dragon
 roles-antag-dragon-objective = Create a carp army to take over this quadrant.
+
+roles-antag-psi-cultist-name = Psi Cultist
+roles-antag-psi-cultist-objective = Complete the objectives assigned to you by your psionic leader.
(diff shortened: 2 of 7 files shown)


Avalon-Proto:
I have one big suggestion if you can figure out how- use the Eldritch weapons as a base. But make them weapons that you can "Summon" as an ability. So you appear unarmed until it's too late. Ofc letting the person with the ability choose their weapon

Bonktrauma:
Summoning of glimmer beasts to aide them

leonardo-dabepis:
> Summoning of glimmer beasts to aide them

glimmer creatures are more of a neutral faction in the same way bears are neutral in WWII. Neutral and uncaring 99.99% of the time, unless you can motivate one to join you so that you may harness their power. Also the only glimmer creature I've added so far is glimmer mite because I keep getting distracted and forgetting to finish glimmer scorpion.

leonardo-dabepis:
> I have one big suggestion if you can figure out how- use the Eldritch weapons as a base. But make them weapons that you can "Summon" as an ability. So you appear unarmed until it's too late. Ofc letting the person with the ability choose their weapon

I might do this, but I think it'd make more sense for psionics to be a purely mental power. Summoning a gun is a bit more wizard than psychic, though I definitely see the need for genuinely hostile psionic powers if this antag is meant to be taken seriously at all.

Minigold07:
What if the Psi cultist gets to summon ifrits now that the mysta now longer can?

Stop-Signs:
I'm assuming this is no longer being worked on, if it is let me know. this will be closed due to inactivity in 24 hours

leonardo-dabepis:
> I'm assuming this is no longer being worked on, if it is let me know. this will be closed due to inactivity in 24 hours

oh yeah sorry, I totally forgot about this 😓 you can kill it, I'm no longer working on it.

PR #2718, opened 2025-01-13
bodypart refactor

deltanedas:
## About the PR
refactors bodyparts to not be completely disgusting
also makes them have all have organ and limb slots by default, so if you find a head that wasn't taken from someone it'll just work

appearance not fixed yet so adding skeleton limbs etc still shows your original limb type

## Why / Balance
good

## Technical details
removed crash when adding organ/part slot that already existed

## Media
:trollface:
spawned skull installed onto a urist can have organs put in now

## Requirements

## Breaking changes
:trollface:

**Changelog**


dvir001:
Build-a-Buddy real?

PR #2727, opened 2025-01-14
add bloodloss cryo meds

cjhveal:
## About the PR
This adds two new cryo chems, Hemoxadone and Cyanoxadone, which are used to treat blood loss damage in patients with iron and copper based blood, respectively.

## Why / Balance
With the recent surgery update, if a patient's heart is outside their body for too long, they take a lump sum of 300 blood loss damage, killing them instantly and making them unable to be defibbed. While I like that there are lasting consequences to messing up surgery, I did see someone get RR'd by this while the CMO was trying to train a new surgeon. This PR gives medical an option for solving the issue without having to rely on Epi for cloning.

These chems build off of traumoxadone, which is in a bit of a weak spot given it's complex synthesis and cheaper alternatives in topicals/surgery. Cryo in general has lost some utility given how easy it is to treat burn/brute damage with surgery, and these chems fit neatly in cryo's niche. More cryo meds also give chemists some more things to prepare as needed, which might help address the issue of chemists speed-running all chems and becoming "Passenger+" for the rest of the round.

## Technical details
Cyanoxadone only works on arachnid organs, while all other species will want Hemoxadone. This mirrors how elemental Copper & Iron currently work. I was tempted to have these chems behave differently given that irl insect blood (hemolymph) uses hemocyanin for oxygen transport, (not to mention sap and slime blood), but I figure that's a separate discussion from these changes.

This is my first PR so please let me know if I can change anything to better adhere to contribution guidelines!

## Media
Screenshot 2025-01-14 at 2 31 28 AM
Screenshot 2025-01-14 at 2 31 47 AM

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- add: Added hemoxadone and cyanoxadone

### DIFF
--- a/Resources/Prototypes/_DV/Reagents/medicine.yml
+++ b/Resources/Prototypes/_DV/Reagents/medicine.yml
@@ -32,6 +32,8 @@
             Bloodloss: -6
       - !type:ModifyBloodLevel
         conditions:
+        - !type:Temperature
+          max: 213.0
         - !type:OrganType
           type: Arachnid
           shouldHave: false
--- a/Resources/Prototypes/_DV/Reagents/medicine.yml
+++ b/Resources/Prototypes/_DV/Reagents/medicine.yml
@@ -1,0 +1,79 @@
+- type: reagent
+  id : Hemoxadone
+  name: reagent-name-hemoxadone
+  group: Medicine
+  desc: reagent-desc-hemoxadone
+  physicalDesc: reagent-physical-desc-metallic
+  flavor: medicine
+  color: "#ee0044"
+  worksOnTheDead: true
+  metabolisms:
+    Poison:
+      effects:
+      - !type:HealthChange
+        conditions:
+        - !type:OrganType
+          type: Arachnid
+          shouldHave: true
+        damage:
+          types:
+            Poison: 2
+    Medicine:
+      effects:
+      - !type:HealthChange
+        conditions:
+        - !type:Temperature
+          max: 213.0
+        - !type:OrganType
+          type: Arachnid
+          shouldHave: false
+        damage:
+          types:
+            Bloodloss: -6
+      - !type:ModifyBloodLevel
+        conditions:
+        - !type:OrganType
+          type: Arachnid
+          shouldHave: false
+        amount: 6
+
+- type: reagent
+  id : Cyanoxadone
+  name: reagent-name-cyanoxadone
+  group: Medicine
+  desc: reagent-desc-cyanoxadone
+  physicalDesc: reagent-physical-desc-metallic
+  flavor: medicine
+  color: "#00eeff"
+  worksOnTheDead: true
+  metabolisms:
+    Poison:
+      effects:
+      - !type:HealthChange
+        conditions:
+        - !type:OrganType
+          type: Arachnid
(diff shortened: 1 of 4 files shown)


Bonktrauma:
Fyi 300 bloodloss damage is very much so fixable, blood packs fix bloodloss damage aswell im pretty sure

fow111:
> Fyi 300 bloodloss damage is very much so fixable, blood packs fix bloodloss damage aswell im pretty sure

0.5 bloodloss healed per bloodpack, that's 211 needed to revive someone

deltanedas:
funny heart bloodloss shouldnt be a thing anyway

Lyndomen:
Hello! Fantastic writeup, thank you. We'll be taking this under Direction review, 24 hours.

cjhveal:
> funny heart bloodloss shouldnt be a thing anyway

There has to be some consequence to removing someone's heart, but yeah, not sure that delayed death is the right thing. The [delayed death system](https://github.com/DeltaV-Station/Delta-v/blob/master/Content.Server/_Shitmed/DelayedDeath/DelayedDeathSystem.cs#L23) looks like it's trying to deal 150 bloodloss damage, so maybe that's being applied twice in error? (I'm pretty new to the codebase, so I could be missing something about how damage is applied).

FaintSpeaker:
> The [delayed death system](https://github.com/DeltaV-Station/Delta-v/blob/master/Content.Server/_Shitmed/DelayedDeath/DelayedDeathSystem.cs#L23) looks like it's trying to deal 150 bloodloss damage, so maybe that's being applied twice in error?

The DelayedDeath component is added by removing the heart, and then removed by putting a heart in. 

It is dealing 150 damage per Update as long as the mob is not dead, so if the 150 damage doesn't kill the mob (which it won't, unless the mob already has like 50 damage) then it will apply another 150 damage, pushing it to 300, which is definitely dead.

cjhveal:
>It is dealing 150 damage per Update as long as the mob is not dead, so if the 150 damage doesn't kill the mob (which it won't, unless the mob already has like 50 damage) then it will apply another 150 damage, pushing it to 300, which is definitely dead.

Ah, that makes a ton of sense now, thanks for the explanation!

Resources/Prototypes/_DV/Reagents/medicine.yml
```diff
@@ -0,0 +1,81 @@
+- type: reagent
+  id : Hemoxadone
```
Radezolid: Remove the space before the `:`

Resources/Prototypes/_DV/Reagents/medicine.yml
```diff
@@ -0,0 +1,81 @@
...
+        conditions:
+        - !type:Temperature
+          max: 213.0
+        - !type:OrganType
+          type: Arachnid
+          shouldHave: false
+        amount: 6
+
+- type: reagent
+  id : Cyanoxadone
```
Radezolid: Remove the space before the `:`

Radezolid (changes requested):

Radezolid (approved):
YAML looks good.

Lyndomen (approved):

PR #2765, opened 2025-01-18
Less weh per weh

Stop-Signs:
## About the PR
Juice that makes you weh amount per plushie reduced from 30 to 5
## Why / Balance
Headmin requested (fox)
MRP server
Pending total direction approval
## Requirements

## Breaking changes

**Changelog**


deltanedas (approved):

gitjubx:
wont do much since its not the amount but the effect of the chemical , but if a single burger does cause so much stress i can understand would rather look into the chemical effect than the amount

Stop-Signs:
I'm actually going to go ahead and close this, I didn't necessarily agree with the take that lead to this and after some discussion i don't wish to see this change get merged.

PR #2769, opened 2025-01-18
Ensure the Horizontal state is set when buckling/unbuckling

### LINKED ISSUE #2740: Players move vertical and horizontal after shitmed merge
## Description

Long ago there was a bug that made it where alive players would move up and down when laying in a bed, despite always being buckled. Now it's back to haunt us again!
**Reproduction**

1. Be alive
2. Walk up to medbay bed
3. lay down
4. Up and down and up and down and up-
**Screenshots**

**Additional context**

As the title said, came back after shitmed.

BarryNorfolk:
## About the PR
Fixes https://github.com/DeltaV-Station/Delta-v/issues/2740
Allows the standing system to fully handle state changes before attaching the buckled entity properly.

I found that the best way to cause the bug is to give asphyx damage to the entity on the table, that seemed to work well.

## Why / Balance
It's annoying for the up/down thing when you're buckled in.

## Technical details
Standing system's `Down` and `Stand` functions short circuit if the user is buckled, which means that the standing system never gets a chance to properly set state for if the entity is standing or not. This is a bit of an annoying cyclical dependency between buckling and standing.

To fix this, we simply move the buckling state change to _after_ the standing call is done so we're able to make sure the standing state is properly updated. This doesn't need to be done on the `Unbuckle` call since we're unbuckled by the time the standing call is done anyway.

## Media
Before 
https://github.com/user-attachments/assets/64c1d5c9-5e88-4cae-ab35-4bdc8f7eb1e4

After 
https://github.com/user-attachments/assets/64d9be5a-40e1-48a1-8883-8eb94bceee74

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- fix: Fix the up/down when entities are buckled.

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
@@ -362,7 +362,6 @@ public abstract partial class SharedBuckleSystem
 
         _audio.PlayPredicted(strap.Comp.BuckleSound, strap, user);
 
-        SetBuckledTo(buckle, strap!);
         Appearance.SetData(strap, StrapVisuals.State, true);
         Appearance.SetData(buckle, BuckleVisuals.Buckled, true);
 
@@ -383,6 +382,8 @@ public abstract partial class SharedBuckleSystem
                 _standing.Down(buckle, false, false);
                 break;
         }
+
+        SetBuckledTo(buckle, strap!); // DeltaV - Allow standing system to handle Down/Stand before buckling
 
         var ev = new StrappedEvent(strap, buckle);
         RaiseLocalEvent(strap, ref ev);


deltanedas (changes requested):
this will make people on chairs sideways, and also it calls Down for beds, which *should* set the visual

BarryNorfolk:
> this will make people on chairs sideways, and also it calls Down for beds, which _should_ set the visual

Ah bugger. Good point, didn't check :/
Will think how to fix this better

BarryNorfolk:
Yeah it's just that
```
        if (TryComp(uid, out BuckleComponent? buckle) && buckle.Buckled && !_buckle.TryUnbuckle(uid, uid, buckleComp: buckle)) // WD EDIT
            return false;
```

in the `Down` function means that if you're buckled then the appearance data never changes. Which makes the whole relationship _very_ weird in my eyes. You've got a buckling system calling `standing.Down` while the standing system basically defers to the buckling system now with the `// WD edit.`

BarryNorfolk:
Better fix I think.

deltanedas (approved):

deltanedas:
no more fent zombies in medbay...

PR #2780, opened 2025-01-20
Add HasComponent condition for reagents

sowelipililimute:
## About the PR
There is now a `HasComponent` entity effect condition that allows an effect to be predicated on whether a component is/isn't present, possibly on a given body part

## Why / Balance
This allows reagents to interact with shitmed, by e.g. requiring that a body part has an incision or that it's closed up in order to do a certain effect.

## Technical details
Adds a new `HasComponent` EntityEffectCondition.

## Media

Here is an example of having a reagent ignite people if their head has an open incision:

https://github.com/user-attachments/assets/3a5394ea-e342-4d25-baa1-59e48ece7cfe

```yml
- type: reagent
  id : Mannitol # currently this is just a way to create psicodine
  name: reagent-name-mannitol
  group: Medicine
  desc: reagent-desc-mannitol
  physicalDesc: reagent-physical-desc-opaque
  flavor: sweet
  color: "#A0A0A0"
  metabolisms:
    Medicine:
      effects:
      - !type:PopupMessage
        conditions:
        - !type:ReagentThreshold
          min: 15
        type: Local
        visualType: Medium
        messages: [ "mannitol-effect-enlightened" ]
        probability: 0.2
      - !type:FlammableReaction
        multiplier: 2.0
        conditions:
        - !type:HasComponent
          bodyPart: Head
          components:
          - type: IncisionOpen
      - !type:AdjustTemperature
        amount: 10000
        conditions:
        - !type:HasComponent
          bodyPart: Head
          components:
          - type: IncisionOpen
      - !type:Ignite
        conditions:
        - !type:HasComponent
          bodyPart: Head
          components:
          - type: IncisionOpen
```

## Requirements

### DIFF
--- a/Content.Shared/_DV/EntityEffects/EffectConditions/HasComponent.cs
+++ b/Content.Shared/_DV/EntityEffects/EffectConditions/HasComponent.cs
@@ -1,0 +1,78 @@
+using System.Linq;
+using Content.Shared.EntityEffects;
+using Content.Shared.Nutrition.Components;
+using Content.Shared.Nutrition.EntitySystems;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Localization;
+using Content.Shared.Body.Part;
+using Content.Shared.Body.Systems;
+
+namespace Content.Shared._DV.EntityEffects.EffectConditions;
+
+/// <summary>
+///     Reagent effect condition that depends on if the entity has a given component(s), potentially on a body part
+/// </summary>
+public sealed partial class HasComponent : EntityEffectCondition
+{
+    /// <summary>
+    ///     The set of components that this condition cares about
+    /// </summary>
+    [DataField(required: true)]
+    public ComponentRegistry Components;
+
+    /// <summary>
+    ///     Whether or not the given components should be present
+    /// </summary>
+    [DataField]
+    public bool ShouldHave = true;
+
+    /// <summary>
+    ///     Whether the check is an existential or universal check
+    /// </summary>
+    [DataField]
+    public bool ConsiderAll;
+
+    /// <summary>
+    ///     The explanation displayed in the guidebook for this condition
+    /// </summary>
+    [DataField(required: true)]
+    public LocId Explanation;
+
+    /// <summary>
+    ///     The body part of the entity to test for the components
+    /// </summary>
+    [DataField]
+    public BodyPartType? BodyPart;
+
+    /// <summary>
+    ///     The side of the entity's body to test for the components
+    /// </summary>
+    [DataField]
+    public BodyPartSymmetry? BodyPartSymmetry;
+
+    public override bool Condition(EntityEffectBaseArgs args)
+    {
+        var _body = args.EntityManager.System<SharedBodySystem>();
+        var entity =
+            BodyPart is {} bodyPart
+                ? _body.GetBodyChildrenOfType(args.TargetEntity, bodyPart, symmetry: BodyPartSymmetry).Select(it => it.Id).FirstOrDefault()
+                : args.TargetEntity;
+
+        if (!entity.IsValid())
+        {
+            return !ShouldHave;
+        }
+
+        var tested =
+            ConsiderAll
+                ? Components.Values.All(c => args.EntityManager.HasComponent(entity, c.Component.GetType()))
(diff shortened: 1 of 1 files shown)


deltanedas:
are you sure limbs metabolise reagents themselves and not just the body copying healing to all limbs

sowelipililimute:
> are you sure limbs metabolise reagents themselves and not just the body copying healing to all limbs

they don't, that's not what this condition is doing

sowelipililimute:
@Lyndomen proposed having reagents interact with surgery in some way, and this lays the groundwork in the best way i can think of from a technical front. putting reagents directly into surgery steps ran into some insurmountable design issues, because surgery steps are designed for a persistent component to be there unless someone undoes that, but reagents are the opposite of that persistent. their metabolism makes them unexist, so if we rely on them in order to check step completion, they will uncomplete themselves after some time, which doesn't really make sense for an "apply reagent once at this point to do something" workflow in surgery. marker components/something else could be a thing, but components would require c# for each new surgery step that wants to use a reagent, and a something else would be extra bookkeeping shitcode. plus, you'd need to figure out when exactly to remove this marker. because most surgery steps come in pairs (open/close incision) or respond to body conditions (damage being present/organ being present), there's no way to tell when a reagent has "done its job" as a surgery step, so to speak.

therefore, making reagents respond to surgery conditions (e.g. apply opporozidone when the head has an open incision to remove rot) seems like the sanest way to approach this, and it also diversifies surgery from the current "click list of buttons" approach it currently has.

Content.Server/_DV/EntityEffects/EffectConditions/HasComponent.cs
```diff
@@ -0,0 +1,82 @@
...
+
+/// <summary>
+///     Reagent effect condition that depends on if the entity has a given component(s), potentially on a body part
+/// </summary>
+public sealed partial class HasComponent : EntityEffectCondition
+{
+    /// <summary>
+    ///     The set of components that this condition cares about
+    /// </summary>
+    [DataField]
```
deltanedas: make it required

Content.Server/_DV/EntityEffects/EffectConditions/HasComponent.cs
```diff
@@ -0,0 +1,82 @@
...
+    ///     Whether or not the given components should be present
+    /// </summary>
+    [DataField]
+    public bool ShouldHave = true;
+
+    /// <summary>
+    ///     Whether the check is an existential or universal check
+    /// </summary>
+    [DataField]
+    public bool ConsiderAll = false;
```
deltanedas: false is the default just do `ConsiderAll;`

Content.Server/_DV/EntityEffects/EffectConditions/HasComponent.cs
```diff
@@ -0,0 +1,82 @@
...
+    ///     The side of the entity's body to test for the components
+    /// </summary>
+    [DataField]
+    public BodyPartSymmetry? BodyPartSymmetry;
+
+    public override bool Condition(EntityEffectBaseArgs args)
+    {
+        var _body = args.EntityManager.System<SharedBodySystem>();
+        var entity =
+            BodyPart is BodyPartType bodyPart ?
```
deltanedas: `is {} bodyPart`

deltanedas: organize it like this
```cs
cond
    ? a
    : b;
```

Content.Server/_DV/EntityEffects/EffectConditions/HasComponent.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    ///     Whether the check is an existential or universal check
+    /// </summary>
+    [DataField]
+    public bool ConsiderAll = false;
+
+    /// <summary>
+    ///     The explanation displayed in the guidebook for this condition
+    /// </summary>
+    [DataField]
```
deltanedas: required and make it non-optional

deltanedas (changes requested):

deltanedas (approved):

PR #2786, opened 2025-01-23
Feedback popup for increased round time

beck-thompson:
## About the PR

Direction wants to increase the time emergency evac shuttle is called. The ccvar needs to be updated but after it is, this should be good to merge!

### DIFF
--- a/Content.Server/_DV/FeedbackPopup/ExtendRoundTimeFeedbackPopupSystem.cs
+++ b/Content.Server/_DV/FeedbackPopup/ExtendRoundTimeFeedbackPopupSystem.cs
@@ -1,0 +1,33 @@
+using Content.Server.GameTicking;
+using Content.Shared._DV.FeedbackOverwatch;
+using Content.Shared.GameTicking;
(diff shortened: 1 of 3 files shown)


Content.Server/_DV/FeedbackPopup/ExtendRoundTimeFeedbackPopupSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<RoundEndMessageEvent>(OnRoundEnd);
+    }
+
+    private void OnRoundEnd(RoundEndMessageEvent ev)
+    {
+        if (!(_gameTiming.CurTime.Subtract(_gameTicker.RoundStartTimeSpan) >= TimeSpan.FromHours(2)))
```
deltanedas: < not !>=

deltanedas (changes requested):

deltanedas (approved):

PR #2787, opened 2025-01-23
Buff duster fire-rate

Stop-Signs:
## About the PR
duster shoots faster, little more than 3x faster.

## Why / Balance
This thing is meant to be an upgrade, it feels like a drag to fire for something that requires power, linking, and rarer techs from epistemics

## Technical details
yml

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The duster fires faster


Avalon-Proto:
Neat. Another shuttle gun PR that will never be seen until we are like 3 years down the line and ship battles have become a regular thing :blunt:

Radezolid (approved):
YAML looks good. Time for direction to decide.

ThataKat:
Direction approved

PR #2803, opened 2025-01-26
Make MindShielded players immune to becoming Sleeper Agents.

SolStar2:
## About the PR
Made MindShielded players ineligible to be Sleeper Agents.

## Why / Balance
Currently mindshields are useless as there are no revs. Command members are antag immune due to the damage that can be caused by an antag command member. People promoted into command and mindshielded should be immune for the same reasons. This also allows for Captains to point acting heads more comfortably, knowing that they will not become a sleeper agents.

## Media
https://github.com/user-attachments/assets/a52ffc98-d982-4fb8-8fb8-cbbcbe797400

## Requirements

## Breaking changes

**Changelog**

:cl: SolStar
- tweak: Mindshield implants will now make people immune to becoming sleeper agents.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -137,7 +137,7 @@
   - type: StationEvent
     startAnnouncement: station-event-bureaucratic-error-announcement
     startAudio: #Delta-V: Previously unheard announcer line for DeltaV's custom announcer
-      path: /Audio/Announcements/bureaucratic_error.ogg
+      path: /Audio/Announcements/bureaucratic_error.ogg 
     minimumPlayers: 25
     weight: 5
     duration: 1
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -557,7 +557,7 @@
       playerRatio: 10
       blacklist:
         components:
-        - MindShield
+        - MindShield # DeltaV - Mind Shielded people are immune to becoming syndicate
         - AntagImmune
       mindRoles:
       - MindRoleTraitorSleeper
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -137,7 +137,7 @@
   - type: StationEvent
     startAnnouncement: station-event-bureaucratic-error-announcement
     startAudio: #Delta-V: Previously unheard announcer line for DeltaV's custom announcer
-      path: /Audio/Announcements/bureaucratic_error.ogg 
+      path: /Audio/Announcements/bureaucratic_error.ogg
     minimumPlayers: 25
     weight: 5
     duration: 1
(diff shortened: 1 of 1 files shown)


Lyndomen:
commenting here so i dont forget, could make it neat to only assign gib objectives on mind shielded targets

SolStar2:
cherry-pick of https://github.com/space-wizards/space-station-14/pull/34648 for reference.

Avalon-Proto:
On one hand. Neat, on the other, damn, no more rolling sleepers post-promotion which is a bit sad. Since it was rare enough as is to get promoted and then roll sleeper

Resources/Prototypes/GameRules/events.yml
```diff
@@ -557,6 +557,7 @@
       playerRatio: 10
       blacklist:
         components:
+        - MindShield
```
Radezolid: Needs DeltaV comment

Radezolid (changes requested):

Resources/Prototypes/GameRules/events.yml
Radezolid: Just in case, the reasoning is that it isn't approved upstream yet, until they do it needs a comment.

SolStar2:
> On one hand. Neat, on the other, damn, no more rolling sleepers post-promotion which is a bit sad. Since it was rare enough as is to get promoted and then roll sleeper

Still possible if command decides to not mindshield

HTMLSystem:
iirc this was already... a thing? And the only way people became HoS and a sleeper was because they roled it *right before* being promoted.

Avalon-Proto:
> iirc this was already... a thing? And the only way people became HoS and a sleeper was because they roled it _right before_ being promoted.

Nope, as it stands mindshields do nothing besides stop you from signing with the recruiter pen. Only *roundstart* mindshield roles are antag immune

HTMLSystem:
That is... INCREDIBLY dumb. Why is it like that .

NoElkaTheGod:
This does not really change anything, because if you're already a syndie, mindshield will not remove your antag status. As such, it is still possible for an acting head to be a traitor. At that point just make mindshield deconvert every single antag or don't touch it at all.

Leostel:
The issue isn’t that it’s possible for an antag to be promoted to a head. 

But I think it’s reasonable for the mindshield to prevent new antag roles from being gained if it was applied to someone who wasn’t already an antagonist. 

On that note, does this change apply to midround initial infected as well?

Radezolid (approved):

SolStar2:
> On that note, does this change apply to midround initial infected as well?

It does not

SolStar2:
Direction Approved 👍

NoElkaTheGod:
> On that note, does this change apply to midround initial infected as well?

Do they even exist right now? Pretty sure they were removed.

deltanedas (approved):

deltanedas:
this only affects sleeper agents

PR #2806, opened 2025-01-27
Minor fixes for feedback popups

beck-thompson:
## About the PR

1.) Feedbacks with only blank entries are now just ignored.
2.) The feedback popup command now always displays the popup even if the player has already gotten one.
3.) Where to write feedback is now clearer.
4.) There is now a confirmation time on the submit button so you have to click it twice in two seconds for it to actually submit.

## Requirements


CybersunOperative:
direction approved bugfix real

Content.Client/_DV/FeedbackPopup/FeedbackPopupWindow.xaml
```diff
@@ -7,7 +7,9 @@
         <RichTextLabel HorizontalAlignment="Center" Name="TitleLabel" Margin="0 0 0 20"></RichTextLabel>
         <BoxContainer Name="SectionContainer" Orientation="Vertical"></BoxContainer>
         <BoxContainer Orientation="Vertical" Name="FeedbackReplyContainer" Visible="False">
+            <Label HorizontalAlignment="Center" Text="{Loc 'feedbackpopup-text-box-above'}"></Label>
```
deltanedas: use `<Label .../>` instead of adding `</Label>` at the end

Content.Client/_DV/FeedbackPopup/FeedbackPopupWindow.xaml.cs
```diff
@@ -53,5 +57,38 @@ private void CreateSection(string text)
...
+            return;
+
+        // First time the button has been clicked
+        if (_resetSubmitButtonTime == null || _resetSubmitButtonTime < _gameTiming.RealTime)
+        {
+            SubmitButton.Text = Loc.GetString("feedbackpopup-submit-feedback-button-confirm");
+            SubmitButton.ModulateSelfOverride = Color.Red;
+            _resetSubmitButtonTime = _gameTiming.RealTime + TimeSpan.FromSeconds(2);
+            return;
+        }
```
deltanedas: use `ConfirmButton` instead of reinventing the wheel

deltanedas (changes requested):

Content.Client/_DV/FeedbackPopup/FeedbackPopupWindow.xaml.cs
beck-thompson: Wow very useful!

deltanedas (approved):

PR #2816, opened 2025-01-29
Add more useful items for hostage ops and stun bundle for syndicates to boot.

beck-thompson:
## About the PR

Title

## Why / Balance

Hostage ops really needed some way of getting hostages without just killing them. Before the only option was really noct but that sucks. This is more fun!

In the discussion it sounded like normal traitors should have some stun equipment as well so they got the same bundle but not for free.

## Technical details

copy paste central :sob:

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Stun bundle: 4 tc for a stun baton and disabler!
- add: Hostage ops now receive free body bags and stun bundle in their uplink!

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
+++ b/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
@@ -104,6 +104,4 @@
   - type: StorageFill
     contents:
     - id: Stunbaton
-      amount: 1
     - id: WeaponDisabler
-      amount: 1
--- a/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
+++ b/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
@@ -88,9 +88,9 @@
       amount: 10
 
 - type: entity
-  name: syndicate stun box
   parent: BoxCardboard
   id: BoxSyndicateStunBundle
+  name: syndicate stun box
   description: A box of two basic stun weapons.
   components:
   - type: Sprite
--- a/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
+++ b/Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
@@ -86,3 +86,24 @@
     contents:
     - id: Handcuffs
       amount: 10
+
+- type: entity
+  name: syndicate stun box
+  parent: BoxCardboard
+  id: BoxSyndicateStunBundle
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
```diff
@@ -86,3 +86,24 @@
     contents:
     - id: Handcuffs
       amount: 10
+
+- type: entity
+  name: syndicate stun box
+  parent: BoxCardboard
+  id: BoxSyndicateStunBundle
```
Radezolid: Please fix the order of this to YAML convention. It should be in this order:
`parent` > `id` > `name`

Radezolid (changes requested):
Only nitpick, if this gets direction approval before you can fix it I will merge it as this is needed for hostage ops. I'll fix it later if that's the case.

ThataKat:
Direction is reviewing, should be a day or less.

Radezolid (dismissed):

Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
```diff
@@ -86,3 +86,24 @@
...
+    - state: box_of_doom
+    - state: writing_of_doom
+  - type: Storage
+    maxItemSize: Normal
+    grid:
+    - 0,0,3,1
+  - type: StorageFill
+    contents:
+    - id: Stunbaton
+      amount: 1
```
deltanedas: this is the default value

Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
```diff
@@ -86,3 +86,24 @@
...
+    - id: WeaponDisabler
+      amount: 1
```
deltanedas: .

deltanedas (changes requested):

deltanedas (dismissed):

ThataKat:
Direction here, we ask that you remove the stun options from the syndicate agent uplink. We prefer conventional stuns to not be freely provided by the uplink, especially if they're also security brand. All clear for nukies though, they definitely need it with their new objectives. 

While it would also be nice to have the nukeop stun options be syndicate branded (so they don't just look like station security) that's definitely beyond the scope of the PR. If anyone wants to take up the challenge that would be great, though!

Once those changes are made, direction approved!

deltanedas (approved):

PR #2820, opened 2025-01-30
Fix AI upload law objective

beck-thompson:
## About the PR

I don't know how I missed this but I forgot to change what law its checking for :pensive:


Avalon-Proto:
Does this still work if the console is destroyed

deltanedas (approved):

deltanedas:
as a powergamer cap you should destroy ai upload anyway to prevent mysta installing nutimov :trollface:

PR #2852, opened 2025-02-02
Reverted the drink to it's true name

OrangeMoronage9622:
## About the PR
Reverted the name back to normal

## Why / Balance
Actually offends me with the name change.

## Technical details
Renamed it to carbomb

## Requirements

**Changelog**
:cl:
- Reverted change to the Irish Slammer to the Irish carbomb
-->


therealDLondon:
As someone who lives in a city which was a target of an IRA attack the name of "Irish Carbomb" never sat right with me :/

OrangeMoronage9622:
It's the real life drink name. Besides, it's change was unnecessary and leaves a sour taste to people who like the drink.

IamVelcroboy:
Honestly, just remove the "Irish" part of it. Keep Carbomb and call it a day

Edit: Or even, Shuttlebomb if you want to make it SS14 themed

OrangeMoronage9622:
I stand by the fact that it is called the irish carbomb and any other way is not acceptable.

deltanedas:
your reason of it offending you is outweighed by it offending a lot more people

PR #2884, opened 2025-02-04
fix ipc nukies not getting comms

deltanedas:
## About the PR
estacao pirata shitcode my beloved

## Why / Balance
fix

## Technical details
- moved to shared, _DV
- fully refactored to be less shitty, using proxy instead of entity manager arg
- Del instead of QueueDel so the headset doesn't exist for 1 tick

## Media
normal ipcs still work + nukie ipcs now work

## Requirements

## Breaking changes
moved `InternalEncryptionKeySpawner` from `_EE` to `_DV` and removed EntityManager parameter

**Changelog**
:cl:
- fix: Fixed IPC nukies not getting comms.


Radezolid:
Best review i can give you: "Tested in game, it worked even for LPOs and ninjas"

PR #2885, opened 2025-02-04
seismic 1984

deltanedas:
## About the PR
made seismics 5x as expensive. they can still be found on debris

## Why / Balance
salv having 20 bombs roundstart was bad and raider bait
salv being able to buy 100 bombs after mining for an hour isnt good either (unlucky elijah, now its only 20)
they are useless for mining anyway its solely for being evil

## Technical details
:trollface:

## Media
:trollface:

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Increased seismic charge cost to 1250 mining points.

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/VendingMachines/Inventories/salvage_points.yml
+++ b/Resources/Prototypes/_DV/Catalog/VendingMachines/Inventories/salvage_points.yml
@@ -10,8 +10,6 @@
     cost: 150
   - id: Soap
     cost: 200
-  - id: SeismicCharge
-    cost: 250
   - id: WeaponGrapplingGun
     cost: 300
   # TODO: laser pointer 300, toy facehugger 300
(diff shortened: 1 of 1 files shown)


Radezolid (dismissed):

Bonktrauma:
I would also increase the salvages tc count by one or 2 then

beck-thompson:
> I would also increase the salvages tc count by one or 2 then

No let them suffer

fow111:
I think this change is bad for the direction of this server and is very LRP. Salvage being able to buy seismic charges opens up a lot of roleplay scenarios that would otherwise be nigh impossible to achieve. This change would incentivise salvagers to interact with the station even less and will result in salvagers prioritising gameplay even more. I know I'm not a maintainer or a direction something but from an average player's perspective(epi main btw) I think we should keep seismic charges in the salvvend because it will benefit the playerbase AND the server as a whole :3

deltanedas:
syndies can still buy bombs or find them on debris this just removes being able to shit them out en masse

Resources/Prototypes/_DV/Catalog/VendingMachines/Inventories/salvage_points.yml
```diff
@@ -10,8 +10,6 @@
     cost: 150
   - id: Soap
     cost: 200
-  - id: SeismicCharge
-    cost: 250
```
Lyndomen, line 14: requesting update to 1250, no removal

Lyndomen (dismissed):

PR #2907, opened 2025-02-06
Revert "SOS app for the PDA"

deltanedas:
test merge complete after 5 minutes

reasons to get rid of it:
- it conveys no information that saying "help maint" on common doesnt
- it just gets spammed the shit out of
- only way to stop someone using it is to take their pda away which is obviously bad for syndies nanochat etc
- prisoners will spam it nonstop i guarantee
- if someone steals your headset they can steal your pda too, and intercoms exist


Lyndomen:
We're gathering feedback, I'll have an update before the next publish

Emily9031:
Maybe the SOS messages can have their own radio channel, and security be stocked with encryption cards for it? That way security can opt in to getting the messages, instead of just having their radio spammed when it's already used more than any other.

deltanedas:
then 1 shitter spamming it means people genuinely using it wont get heard, not like it matters anyway since it has no actionable info

ThataKat:
We've gotten enough feedback, so this is direction approved.

We may or may not be back after reviewing what we have and implementing changes, but for now this is too problematic to keep.

PR #2914, opened 2025-02-07
added soap to surgeon's lockers so they don't have to tide

ShepardToTheStars:
## About the PR
With the new sanitization changes, it's become mandatory to clean your surgeon's tools with soap. This requires one to go out and hunt for soap, usually scouring maints or asking a janitor if you happen to find them. Since soap is basically required by the job, it makes sense to add it to the locker.

## Why / Balance
If you get two patients and don't have soap, you're basically screwed and have to go and hunt some down.

## Technical details
Added NT-branded soap to surgeon's locker (since that's their employer)

## Media
n/a

## Requirements

- [X] I have added media to this PR or **it does not require an ingame showcase.**

## Breaking changes

**Changelog**

- add: Added NT-branded soap to surgeon lockers.

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/Fills/Lockers/medical.yml
+++ b/Resources/Prototypes/_DV/Catalog/Fills/Lockers/medical.yml
@@ -101,6 +101,7 @@
     - id: ClothingBackpackDuffelSurgeryFilled
     - id: ClothingOuterVestTank
     - id: NitrousOxideTankFilled
+    - id: SoapNT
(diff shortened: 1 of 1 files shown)


ShepardToTheStars:
Oh, I made it NT-branded because medbay is basically an NT hospital, so it makes sense to brand the soap. It could definitely be `RandomSoap` instead though.

Lyndomen (approved):
Looks great, thank you for the contribution!

deltanedas (approved):

PR #2921, opened 2025-02-07
Change default value of game.secret_weight

Toby222:
## About the PR
Set the default value of the `game.secret_weight` CVar to "SecretDeltaV" from "Secret", since "Secret" causes errors unless overridden.

## Why / Balance
Creates semi-frequent confusion in contributor-help channel on discord, so might as well change it.

## Technical details
see above

## Media
n/A

## Requirements

## Breaking changes
none

**Changelog**
none, internal change

### DIFF
--- a/Content.Shared/CCVar/CCVars.Game.cs
+++ b/Content.Shared/CCVar/CCVars.Game.cs
@@ -348,7 +348,7 @@ public sealed partial class CCVars
         ///     The prototype to use for secret weights.
         /// </summary>
         public static readonly CVarDef<string> SecretWeightPrototype =
-            CVarDef.Create("game.secret_weight_prototype", "Secret", CVar.SERVERONLY);
+            CVarDef.Create("game.secret_weight_prototype", "SecretDeltaV", CVar.SERVERONLY); // DeltaV (duh) - Change default secret weight prototype for easier testing
 
         /// <summary>
         ///     The id of the sound collection to randomly choose a sound from and play when the round ends.


deltanedas:
its already overridden, someone not using a config preset is really bad for other things too

Toby222:
Fair enough
Would a PR that sets the cvar for which files to load in runserver-Tools be better in that case?
Or should it just stay as-is?

Axionyxx:
> Fair enough Would a PR that sets the cvar for which files to load in runserver-Tools be better in that case? Or should it just stay as-is?

Having a cvar be broken by default is bad and terrible, even if they're supposed to use a config preset

deltanedas:
the tools config should be using the debug config presets

Toby222:
Ah, the development config preset, but that also has it set.
Then I'm confused why people kept bumping into the issue?

Lyndomen:
> Ah, the development config preset, but that also has it set. Then I'm confused why people kept bumping into the issue?

I run into this all the time when I run in release, which I use for proper testing of ghost roles and stuff

Toby222:
In release you should be setting the proper config to load via the `config.presets` cvar, which is fair enough to me.
Maybe additional `-TestRelease` scripts that adds the proper flag for at least one of the config files to be loaded would make sense for that, then?

deltanedas:
yeah thats too annoying this is fine in comparison

deltanedas (approved):

PR #2923, opened 2025-02-07
Justice lockable buttons

FieldCommand:
## About the PR

Added lockable buttons to the rest of Justice. So the prosecutor, clerk, chief justice, and the department as a whole have buttons that can be locked, respectively. So the attorney is not the only one in the department that has one anymore.

## Why / Balance

As stated above

## Technical details

4 new buttons that can be mapped in Justice as needed for mappers: Justice, Prosecutor, Clerk, Chief Justice.

## Requirements

## Breaking changes

**Changelog**

:cl: Field Command
- tweak: Some more lockable buttons for the justice department specifically.


deltanedas (changes requested):
move suffix below id for all these

deltanedas (approved):

PR #2930, opened 2025-02-08
Make crew monitors beep when someone dies / crits

Quanteey:
## About the PR

This PR makes the handheld crew monitor and the crew monitoring console play a beeping sound in a small radius whenever a crew member with appropriate sensors enabled dies. This change does not affect the spy and syndicate crew monitors, and the alert behavior can be enabled / disabled in the YAML prototypes. 

Alerts are issued whenever a crew member goes into critical / dead state, and no further alerts will be issued for that specific crew member until they get out of dead / critical state. A 15 second cooldown is in place between two alerts to reduce spam. Alerts require the monitor to be powered.

## Why / Balance

Paramedics are expected to respond immediately when someone goes critical / dies. However, there is currently no system in place to **actively** alert them when that happens. This leaves paramedics either staring at the monitor for pretty much 2 hours, which is not optimal gameplay-wise, or checking every minute or two, which is not very efficient.

This change solves this issue altogether, allowing them to do other activities during downtime without having to worry about missing someone dying.

## Technical details

Modified `CrewMonitoringConsoleComponent` to store a HashSet of crew members a given monitor has already alerted for, along with frame time tracking (for the cooldown, same implementation as the Morgue). Added 3 optional DataFields for enabling / disabling alerts, configuring cooldowns, and the sound.

The main logic is in the `CrewMonitoringConsoleSystem`'s `OnPacketReceived`, along a new `Update` override for frame time tracking.

The new audio file comes from Pixabay with a CC0 license.

## Media

https://github.com/user-attachments/assets/6429d929-3dea-47ad-a4f9-60b1dd23568f

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The crew monitor handheld and console now beep when a crew member goes into critical condition or dies!


Sifb26:
Does this check if the crew monitoring server is powered as well?

Sifb26:
Also direction review will be around 24 hours

Quanteey:
Yes, this checks power from both from a power cell or station power (for the machine version). Alerting does not use power, though.

EDIT: I have misunderstood the original question. My changes do not directly check the server, but I assume if it is down the monitors will not receive packets that would trigger my logic

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleComponent.cs
```diff
@@ -16,4 +17,34 @@ public sealed partial class CrewMonitoringConsoleComponent : Component
...
+
+    /// <summary>
+    ///     Track sensors that have triggered the crew member critical alert.
+    /// </summary>
+    public HashSet<string> AlertedSensors = [];
+
+    /// <summary>
+    ///     Accumulated time for tracking alert cooldown
+    /// </summary>
+    public float AccumulatedFrameTime = 0f;
```
deltanedas: use timespans

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleComponent.cs
```diff
@@ -16,4 +17,34 @@ public sealed partial class CrewMonitoringConsoleComponent : Component
...
+
+    /// <summary>
+    ///     Time between alerts
+    /// </summary>
+    [DataField]
+    public float AlertCooldown = 15f;
```
deltanedas: TimeSpan and TimeSpan.FromSeconds

deltanedas (changes requested):

Quanteey:
Requested changes are completed and tested.

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleComponent.cs
```diff
@@ -16,4 +17,34 @@ public sealed partial class CrewMonitoringConsoleComponent : Component
...
+
+    /// <summary>
+    ///     Track sensors that have triggered the crew member critical alert.
+    /// </summary>
+    public HashSet<string> AlertedSensors = [];
+
+    /// <summary>
+    ///     Accumulated time for tracking alert cooldown
+    /// </summary>
+    public TimeSpan AccumulatedTime = TimeSpan.Zero;
```
deltanedas: still need to change the logic from adding to a `[DataField(customTypeSerializer: typeof(TimeOffsetSerializer))] public TimeSpan NextAlert;` or something, comparing _timing.CurTime to it

along with `[AutoGenerateComponentPause]` and `[AutoPausedField]`

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleSystem.cs
```diff
@@ -1,18 +1,22 @@
 using System.Linq;
 using Content.Server.DeviceNetwork;
 using Content.Server.DeviceNetwork.Systems;
+using Content.Server.Power.EntitySystems;  // DeltaV
```
deltanedas: 1 space not 2 for these comments

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleSystem.cs
```diff
@@ -43,6 +47,48 @@ private void OnPacketReceived(EntityUid uid, CrewMonitoringConsoleComponent comp
...
+                continue;
+            }
+
+            if (!status.IsAlive || isCritical)
+            {
+                if (component.AccumulatedTime >= component.AlertCooldown)
+                {
+                    var audioParams = AudioParams.Default.WithVolume(-2f).WithMaxDistance(4f);
+                    _audio.PlayPvs(component.AlertSound, uid, audioParams);
+                    component.AccumulatedTime = TimeSpan.Zero;
```
deltanedas: should set NextAlert to `_timing.CurTime + component.AlertCooldown;` probably

deltanedas (changes requested):

ShepardToTheStars:
Actually, I have some questions:

1. Would it be better to use something like the Atmos console alerting? With a visual indicator rather than sound. First one is no alarms, 2nd is at least one alarm (and its blinking). There's a warning state as well but not really relevant.
image
image

2. Is there a way to turn off the sound? Like, in situations with nukies or zombies, where they are crew members that likely will not be recovered, it may be a bit grating on the ears to hear the beeping every 15 seconds for the remainder of the round.

Quanteey:
To address your questions:
1. I don't think so. Remember, we are not just talking about the crew monitoring console, but also the handheld item that every paramed spawns with. Most maps do not have an accessible crew monitoring console, and even though it is common for the CMO to move their console to an accessible location, this is not a guarantee, so limiting the alerts to just the console would be inconvenient. Giving this functionality to the handhelds would also allow parameds to move freely instead of having to stay at the console.
Even if the handheld also had a visual alert too, there is a high chance that it would be missed if the paramed is talking with people and roleplaying, which is exactly what we want to avoid.

2. It wouldn't be every 15 seconds though, unless there is actually a new person dying every 15 seconds. But, good call. No idea how the toggling should be implemented gameplay-wise.

deltanedas:
> the paramed is talking with people and roleplaying, which is exactly what we want to avoid.

true, no rp allowed for parameds :trollface:

Lyndomen (dismissed):

Sifb26:
Direction approved. 
EDIT: I didn’t see Lyndo’s approval:Godo:

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleSystem.cs
Quanteey: My bad.

Quanteey:
Requested changes are made and tested. Actually using future timestamps now for the cooldown.
Since direction approved I guess this could be merged as-is (without the alert toggling). Considering the alert cooldown and the small range where this can be heard, I'd say this won't be too annoying. If it is I could always implement toggling in a separate PR.

ShepardToTheStars:
The toggle is already implemented though! Just remove battery from handheld or unanchor the console. :blunt:

Quanteey:
Also that.

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleComponent.cs
```diff
@@ -16,4 +18,35 @@ public sealed partial class CrewMonitoringConsoleComponent : Component
...
+    [DataField]
+    public bool AlertsEnabled = true;
+
+    /// <summary>
+    ///     Track sensors that have triggered the crew member critical alert.
+    /// </summary>
+    public HashSet<string> AlertedSensors = [];
+
+    /// <summary>
+    ///     Accumulated time for tracking alert cooldown
```
deltanedas: outdated doc

deltanedas (changes requested):

deltanedas (approved):

Avalon-Proto:
Okay so I have 1 SLIGHT annoyance with this. Thanks to the fact Aghosts have the crew monitor interface, its gets QUITE annoying hearing it randomly beep at you in the middle of handling an Ahelp

Quanteey:
> Okay so I have 1 SLIGHT annoyance with this. Thanks to the fact Aghosts have the crew monitor interface, its gets QUITE annoying hearing it randomly beep at you in the middle of handling an Ahelp

Ah, did not know that. That sure sounds like it's QUITE annoying. I'll make a pr with a fix for that in a few days.

PR #2939, opened 2025-02-09
change air alarm defaults

deltanedas:
## About the PR
change oxygen upper limit from 30% and 24% to 80% and like 50%

## Why / Balance
instakilling from inhaling gas was removed years ago so the upstream justification is a meme
and even if it was still a thing, just fix that instead of having meme air alarm settings that mean nothing

current pressure setting constantly dilutes air alarms as you take next to no airloss damage at 80kpa

i have to change these every single atmos round on every air alarm

for actual spacings firelocks close on their own regardless of air alarm settings so it just serves to annoy people

## Technical details
number

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Made air alarms less sensitive to higher oxygen and slightly lower pressure. Opening random firelocks is more likely to actually be dangerous!


deltanedas:
for context this big ass room the aghost is in lost 25kpa from evac getting blown up

PR #2807, opened 2025-01-27
halve crawlspeed

TadJohnson00:
## About the PR

nerfed crawl speed
## Why / Balance

crawling too fast, millions must crawl under tables and engage in firefights while laying down :blunt:
## Technical details

0.4f > 0.2f
## Media

no
## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

erm
**Changelog**

:cl:
- tweak: Adjusted crawling speed to half its original value. Break a leg.

PR #2963, opened 2025-02-11
Prone Meta Band-aid

Stop-Signs:
## About the PR
Makes standing up take 2 seconds instead of being almost instant

## Why / Balance
drop to floor, wait for person to run out of ammo trying to sprite click you, get up and run at them.

in all realism this along with #2807 should make prone less of a problem until someone makes a system to prevent combat prone from dominating the meta.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/_White/Standing/LayingDownComponent.cs
+++ b/Content.Shared/_White/Standing/LayingDownComponent.cs
@@ -7,7 +7,7 @@ namespace Content.Shared._White.Standing;
 public sealed partial class LayingDownComponent : Component
 {
     [DataField, AutoNetworkedField, ViewVariables(VVAccess.ReadWrite)]
-    public float StandingUpTime { get; set; } = .5f;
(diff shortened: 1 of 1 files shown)


deltanedas:
this is a huge buff to slips, disable combat mode instead as an actual bandaid for lying down during combat

BlitzTheSquishy:
Could also consider lowering slip and stun durations to allow for the change to still lay down while fighting (this is more work though)

dvir001:
Might be over engineering it but you should have different speed for how long it takes you to go from standing to prone vs getting forced to prone by soap, when going prone not by force should take longer.

Then here comes some new powergaming meta of placing a soap Infront of yourself for faster prone. :trollface:

Lyndomen:
im gonna evilmode close this in favor of porting a better solution that doesnt fuck over slips

PR #2975, opened 2025-02-14
Book of Mysteries becomes useful + steal objective

Stop-Signs:
## About the PR
book of mysteries can now grant whoever holds it the power to make forcewalls, which block bullets, lasers, objects, and all other projectiles from passing through them. They can however still be passed through by people. The MG can no longer summon ifrit with their book, and it will no longer spawn in their loadout, instead appearing in their locker. the book of mysteries is also now a steal objective.

## Why / Balance
ifrit was outdated and it really showed. the role was almost exclusively used for raiders or validhunters, and the occasional newbie would take it and roleplay as an advanced welder. In the future i hope to add more spells to this and a shop similar to the grimoire, but this is the first step.

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: The book of mysteries is now a steal objective, spawning in the locker instead of on the mystagogue.
- tweak: The book of mysteries can no longer summon ifrits, instead granting the secret of forcewall!


Bonktrauma:
Please, don't remove ifrit, instead just give them back pyrokinesis to make them actually usefull

Lyndomen (approved):

PR #2976, opened 2025-02-14
replace thief radio jammer with doorjack

deltanedas:
## About the PR
title

## Why / Balance
jammer was useless besides proving to nukies you are valid
doorjack lets you rush free AA easier now

**Changelog**
:cl:
- tweak: Replaced thief's radio jammer with a doorjack.


Avalon-Proto:
Won't this need to get swapped to the access breaker if we go that route and need to accept it? Or is that still in review turmoil

deltanedas:
yes but that can be done in the future

ThataKat:
Seen closed but this is conceptually approved by direction if you want to revisit it later or anything

deltanedas:
this would just be removing jammer now

PR #3002, opened 2025-02-18
remove thief ghost role steal objectives

deltanedas:
## About the PR
- removed shiva/laika objective
- removed bingus objective

silvia objective didnt work anyway as it required traitor so you could never get it as thief since traitor could only possibly roll after thief

## Why / Balance
the only way to do this is to build a indestructible oxygenated scrubbed room specifically to hold this animal and prevent it dying randomly. begging logi for a pet carrier is already valid and takes 30m to arrive anyway, and if you do get it you drop it when evac docks so you fail it anyway unless you are a diona

regardless of how you contain the animal you get completely cucked if someone takes the ghost role, at which point you have to keep them sedated 24/7 to prevent them attacking you or just running away and snitching when they get cogni

**Changelog**
:cl:
- remove: Shiva and Bingus are no longer thief objectives.

### DIFF
--- a/Resources/Prototypes/Objectives/objectiveGroups.yml
+++ b/Resources/Prototypes/Objectives/objectiveGroups.yml
@@ -117,14 +117,13 @@
   id: ThiefObjectiveGroupAnimal
   weights:
     IanStealObjective: 1
-    BingusStealObjective: 1
+    #BingusStealObjective: 1 # DeltaV - no stealing ghost roles
     McGriffStealObjective: 1
     WalterStealObjective: 1
     MortyStealObjective: 1
     RenaultStealObjective: 1
-    ShivaStealObjective: 1
+    #ShivaStealObjective: 1 # DeltaV - no stealing ghost roles
     TropicoStealObjective: 1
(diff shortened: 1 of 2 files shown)


Lyndomen (approved):
Any steal objective that nobody cares about is dogshit

Avalon-Proto:
> Any steal objective that nobody cares about is dogshit

At this rate you'll just be removing thief they've lost so many theft targets :blunt:

Lyndomen:
> > Any steal objective that nobody cares about is dogshit
> 
> At this rate you'll just be removing thief they've lost so many theft targets :blunt:

Not to be too blunt but the role is kinda dogshit

Avalon-Proto:
> > > Any steal objective that nobody cares about is dogshit
> > 
> > 
> > At this rate you'll just be removing thief they've lost so many theft targets :blunt:
> 
> Not to be too blunt but the role is kinda dogshit

I use it as a free pass to do more illegal shenanigans like Emag everything I see, steal hats and shoes. Maybe go steal AA. I don't try to hard for theft targets but I ***will*** fuck around with an emagged uniform printer

deltanedas:
thief is free AA the role never remove ever

PR #3034, opened 2025-02-23
Silvia gets med hud effect

Lyndomen:
## About the PR
added yaml tags so the medical snake can see how healthy people are

## Why / Balance
biting bad on healthy targets

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Silvia can now see health bars

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/NPCs/pets.yml
@@ -225,3 +225,5 @@
       - FootstepSound
     - type: StealTarget
       stealGroup: AnimalSilvia
+    - type: ShowHealthBars
+    - type: ShowHealthIcons


Lyndomen:
I'm also considering giving Silvia the med hud item, idk if that would be better

Avalon-Proto:
> I'm also considering giving Silvia the med hud item, idk if that would be better

Do you wanna try and sprite a med-hud that moves with her face? That just sounds painful :blunt:

IamVelcroboy:
> > I'm also considering giving Silvia the med hud item, idk if that would be better
> 
> Do you wanna try and sprite a med-hud that moves with her face? That just sounds painful :blunt:

yeah, Item would be ideal but, as I found out trying to give her a nurse hat....impossible w/out explicit code for it

jimmy12or:
I'm all for Silvia having the HUD, but I think giving the item might not be the best since I imagine tiders just taking it off her before someone takes Silvia as a ghost role. Having it built into this genetically modified snake is probably just the best bet.

ThataKat:
24 hour review

ThataKat:
Direction here,

Determined it to be unnecessary as an inherent trait. Inspecting for damages does exist, and there's not a lot of ic precedent for it. Would be willing to reconsider if someone wants to sprite eyeslot variations so she can wear the actual HUD

PR #3053, opened 2025-02-26
Make survival start a lil bit faster 

Lyndomen:
## About the PR
ramping event scheduler has a random number it generates off our chaos mod, and that range pushes it pathetically low. Removed the lower range. 

## Why / Balance
gaming 
Hopefully this makes it so "nothing happened for 70 minutes" is a bit rarer. 
## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Survival got 33% more deadly
- tweak: Survival starts off a bit stronger with its events.


Lyndomen:
pending direction review

Lyndomen:
```
So we have something called "Chaos" on this gamerule. Chaos is assigned a value of 4.5. Chaos is then randomly assigned a value between 3.375-5.625. This value is MaxChaos

MaxChaos is how much chaos there is at 180 minutes in the game. More means less time in between events, which currently caps out around 2-3 minutes at three hours. 

Current Chaos is determined by a linear equation, it basically is a straight line between MaxChaos/10 and MaxChaos. 

What I've done in this PR is cut out the bottom of the range here, so MaxChaos can only ever be between 4.5-5.625. This means a shift will always be a little bit spicier, but not dramatically so. 

What I've also done in this PR is increase our Chaos to 6, a 33% increase. Its called Survival afterall. For context, Wizden's value is twenty. This has the benefit for them that their random range calculation allows much more varied shifts, while our comparitively low base number doesnt shift that much.
[7:37 PM] 
OP
 lyndo🐀: My eventual goal is to remove the linear chaosmod increase with a "y = a * (1 - e^(-bx))" styled equation, which results in something more akin to this shit 

```
my explanation to direction ^

all of this is a band aid for the kinda trash distribution system that is survival.

deltanedas (approved):

Issue #3060, opened 2025-02-27
Security Borgs


TadJohnson00:
Members of Direction have identified issues with this content and are now presenting a set of design criteria regarding this content. Any person who wishes to contribute to the effect of providing solutions to the set of design criteria below is welcome to do so by submitting a suitable PR that links this issue. If no such PR is provided within the specified timeframe, this content will be removed in accordance with Direction's guidelines.

**Security Cyborgs (secborgs) must have silicon laws that require it to operate within Space Law, Standard Operating Procedure, and Company Policy; the paladin-esque laws it currently have are not acceptable.**
- [X] To that end, the secborg role must be whitelistable and rolebannable. It should also feature Security role time requirements. - #3109
- [X] The secborg must be designed in such a way that the rigid and strict nature of silicon laws do not conflict with often highly contextually-reliant interpretation of parts of each policy. HOWEVER those same laws must not be so open to interpretation that secborgs may regularly violate that policy, either by intentional or mistaken interpretation/misinterpretation of those silicon laws. - #3113

**Secborgs must hold a position within Security's Chain of Command and be responsible to some member or members within that CoC.**
- [ ] Additionally, there must be some method of adequately and sanely removing problematic secborgs from that Chain of Command. Currently, the only two ways to do so is to decommission the secborg in question by physical assault or to remove the positronic brain with a screwdriver and robotics access. **This issue may be resolved by ease of law diagnosis and rectification.**
- [X] Secborgs must also both: A) be partnerable to members of Security, and B) be capable of working unsupervised. They must be able to fulfil either function without causing issues for Security as a result of CoC. - #3113 (law 3 this is your partner) / #3112

**Secborgs should have a clear and defined purpose that justifies their existence within Security, with a focus on any one of the below three directives:**
- [ ] _Nonviolently_ detain suspects (that means with _non-lethal_ measures, not _less-than-lethal_ measures) without causing physical harm, for due processing by full Security members. - #3112 / #2059
- [ ] Serve in a purely observation and support capacity, with no measures to detain suspects or cause harm - but a capability to call for Security members for support, and possibly capabilities to perform basic non-combat Security tasks.
- [X] In the _very unlikely_ case that secborgs are allowed modules that are intended to cause crew harm - whether lethal or less-than-lethal - they must be able to follow Security's ROE (i.e. warning, nonlethal measures, lethal measures - and always with the capacity to detain a suspect). In other words, an inability to cuff people is no bueno. - #3112

[...]

PR #2899, opened 2025-02-06
Ports Ion Damage

Spielern:
## About the PR

Adds Ion damage, a damage type that only affects Silicons. 
Code is ported from Goob-Station/pull/646 (Page seems to be hidden?).
Ion damage is available through any disabler weapon, as well as the X-ray Cannon.
Ion damage can be healed through cable coils, like heat damage.

## Why / Balance

Currently, antag IPCs are bit too strong. Certain antags have lots of options to kill/counter IPCs, but security is lacking in ways to fight IPCs that are antags. Ion damage will provide a general counter for IPCs that crew-aligned players can use.

Nukie IPCs are especially strong with the use of a thermal suit, as their two sources of damage vulnerability are nearly or entirely negated. 

Ipcs die in:
7 Disabler shots
18 Disabler SMG shots
4 X-ray Cannon shots
6 x-01 Ion shots

Cyborgs also now take reduced heat and brute damage to balance the extreme nerf from disablers, with Syndicate Assault Cyborgs gaining much higher resistances than normal Cyborgs.

I completely expect and encourage Direction or anyone else to suggest changes to the damage numbers. These numbers are based off of Goob balance, and is likely not the best fit for us.

## Technical details

Adds Ion damage to all disabler bolts, disabler SMGs, X-ray cannons, and an Ion shot option to the x-01.

## Media

Disablers
https://github.com/user-attachments/assets/1c0d9045-2446-403e-b7d1-168363e211da

Disabler SMGs
https://github.com/user-attachments/assets/1cda452a-721a-49da-b87e-17446c9bb450

X-ray Cannons
https://github.com/user-attachments/assets/97ecb350-76f7-4708-9a92-8fb25c6b28b4

X-01
https://github.com/user-attachments/assets/a7b7ab84-29e5-4de6-bc90-732ed45c2a44

Heat & Brute resistances
https://github.com/user-attachments/assets/dbb64bf4-67e3-4af4-8de7-868b3c887372

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- add: Added Ion damage to disabler weaponry to combat Silicons.
- tweak: Cyborgs now take less Brute and Heat damage.

PR #3064, opened 2025-02-27
make borgs use ipc repairing

deltanedas:
## About the PR
welding for brute and cable for burn, same as ipcs

still no radiation so plasma does nothing

## Why / Balance
part of #3060

cable module more useful for the shadow factory...

## Technical details
shitcode threw if a Healing had a type not in the damage container so fixed it

## Media

cant use cable with no burn damage

using cable after shooting with a laser

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Borgs now get healed like IPCs, repeated welding for brute and cables for burn damage.

### DIFF
--- a/Content.Server/Medical/HealingSystem.cs
+++ b/Content.Server/Medical/HealingSystem.cs
@@ -138,7 +138,7 @@ public sealed class HealingSystem : EntitySystem
         var healingDict = healing.Damage.DamageDict;
         foreach (var type in healingDict)
         {
-            if (damageableDict[type.Key].Value > 0)
+            if (damageableDict.GetValueOrDefault(type.Key) > 0) // DeltaV - 0 instead of throwing
             {
                 return true;
             }
--- a/Content.Server/_EE/Silicon/WeldingHealable/WeldingHealableComponent.cs
+++ b/Content.Server/_EE/Silicon/WeldingHealable/WeldingHealableComponent.cs
@@ -5,5 +5,12 @@ using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototy
 namespace Content.Server._EE.Silicon.WeldingHealable
 {
     [RegisterComponent]
-    public sealed partial class WeldingHealableComponent : Component { }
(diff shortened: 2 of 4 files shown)


TadJohnson00:
Direction approval ops in 24 hours

Avalon-Proto:
Just gonna say this now, it's ***really*** gonna need the nukie Borg into the ground since it has no form of damage resist. But perhaps with this we can also get [#2899

Lyndomen:
> Just gonna say this now, it's _**really**_ gonna need the nukie Borg into the ground since it has no form of damage resist. But perhaps with this we can also get [#2899](https://github.com/DeltaV-Station/Delta-v/pull/2899l)

what i think is next is borg chassis specialization with different damage resists and shit. for instance making service borg vulnerable to rads but engi borg immune etc.

Avalon-Proto:
> what i think is next is borg chassis specialization with different damage resists and shit. for instance making service borg vulnerable to rads but engi borg immune etc.

I feel like borgs taking radiation should be heavily reduced on all fronts if at all. They have a much more basic construction than IPCs and this can have more built in rad shielding. Plus service Borg ***sucks*** to play already and that would just make it worse imo

Lyndomen:
> I feel like borgs taking radiation should be heavily reduced on all fronts if at all. They have a much more basic construction than IPCs and this can have more built in rad shielding. Plus service Borg _**sucks**_ to play already and that would just make it worse imo

this was just an example of what we are doing, im aware of balancing needs

Avalon-Proto:
Fair enough. I'm just saying if you are taking lots there needs to be some give too. Don't want to make people hate playing borgs etc. I still think they should keep their pseudo all access cause they pay for it in having laws and no hands etc

deltanedas:
assault borg has infinite ammo so its more than deserving a nerf, previously its just 10 seconds welded and boom back to full hp. will also mean borg operator needs to keep lv in stock instead of steal ce or some random guys experimental welder to have infinite laser healing

Lyndomen (approved):
Chassis specialization nextweek

PR #3078, opened 2025-03-01
return airlock blacklist for emag

deltanedas:
## About the PR
it was handled by the access reader component though not airlock
so blacklist was brought back to not let it

## Media

**Changelog**
:cl:
- fix: Fixed emags removing access from airlocks.


Smugman:
Well, that was fast

PR #3085, opened 2025-03-02
Kitsune

AeraAuling:
## About the PR
A PR adding Kitsune, only a year and a half after the idea was first brought out. Only uh... this isn't a full PR, this is just a draft PR because we're finally really close to having them finished. 

- [x] Creating said kitsune
- [x] Setting up the foxform
- [x] Setting up the polymorph
- [x] Custom damage sets (half done)
- [x] Creating the components
- [x] Fixing the infinite foxfire bug
- [x] Creating locale stuff for them (half done)
- [x] Setting up guidebook entries
- [x] Creating their own soundfiles (please help)
- [x] Setting up a typing indicator for them
- [ ] Markings stuff (need more sprites)
- [ ] Make it to where their foxform color inherit's from their human form's hair color
- [ ] Do the same for the foxfire
- [x] AAC tablet stuff
- [x] Guidebook entry stuff

## Why / Balance
Kitsune are, essentially, shapeshifters. They possess the ability to turn into a fox form which moves faster in exchange for being more fragile than their humanoid form, alongside the ability to create a small, harmless wisp of flame to provide light. They're also naturally more likely to be psionic, having a strong connection to the noosphere. They have a natural resistance to temperature damage, though their constitution is weak, and their bones are a bit fragile. Also, having a natural connection to the noosphere comes with it downsides, holy damage actually kinda hurts. 

## Technical details
As said, this is a draft PR, there are still some things that are needing done. Namely, uhh... well, the stuff in their task list

## Media

https://github.com/user-attachments/assets/3dce9d7d-d873-4fc4-95ac-a5806f3d6d4a

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: foxes are now real


NoElkaTheGod:
Reading the description, they seem to be kinda maybe a little overpowered? Like, i see zero reasons why would i **not** want to play them, which is kinda bad. Like the concept very much though.

PR #3087, opened 2025-03-02
Small Borg Buff - Everyone Gets Tools

Avalon-Proto:
## About the PR
This PR simply adds a roundstart tools module to every standard borg chasis. Not the syndicate ones though

## Why / Balance
As an avid borg player, not having tools is the biggest headache of all issues to have. You are unable to pry unpowered airlocks or help the crew fix things that they may not have the tools for, IE an IPC who needs welding. Plus having to pay a tax of one of your slots for a tool module is quite annoying when your slots are so limited to only 3 at most. Always needing ot sacrifice 1 of your modules to have any mobility through a station with no power

## Technical details
Adds tool modules to all roundstart borgs

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: All borgs now come with a tool module by default


Resources/Prototypes/borg_types.yml
```diff
@@ -1,4 +1,4 @@
-﻿# Generic borg
+# Generic borg
```
deltanedas, line 1: untouch BOM

deltanedas (changes requested):

Resources/Prototypes/borg_types.yml
Avalon-Proto: the git checkout command right?

deltanedas: no hex editor gaming, checkout reverts an entire file

Avalon-Proto: Hex editor??

deltanedas: google.com

Avalon-Proto: Im looking at both in a hex editor and not seeing jack shit for differences

ThataKat:
Under review, a day or less

Resources/Prototypes/borg_types.yml
deltanedas: add back the BOM bytes this branch removed it

ThataKat (commented):
Direction here,

If the tools took up a slot and could be removed we'd be fine with it. We'd prefer borgs not have free and consequence-less tool access at all times though, which the current implementation does. 

If it's possible to make them take up a slot and be removed that would be preferable, but if not we request this be closed.

Avalon-Proto:
As far as I know there isn't a way to do that without some C# and I only know Yaml. So close it is

PR #3091, opened 2025-03-02
Remove hostage ops

beck-thompson:
## About the PR

Too many complaints, I'm working on fixes but they are difficult and cannot be completed super soon. Easier just to remove it so I don't have to see people complaining about it constantly on the discord, LOOC chat, and OOC chat.

**Changelog**

:cl:
- remove: Removed Hostage Ops - waiting on improvements

### DIFF
--- a/Resources/Prototypes/_DV/Objectives/nukies.yml
+++ b/Resources/Prototypes/_DV/Objectives/nukies.yml
@@ -4,7 +4,7 @@
   id: NukieOperations
   weights:
     NukieOperationDestroyStation: 1
-    NukieOperationKidnapHeads: 1
+#    NukieOperationKidnapHeads: 1
 
 - type: nukieOperation
   id: NukieOperationDestroyStation


userfacethis:
Hostage ops is leading to a lot of same-y strategies that are boring for crew and even the nukies sometimes, nukies are meant to be the antagonist of the round,  and hostage ops just focuses that antagonism to the command members which kinda fucking sucks. I suppose the removal but I do hope it eventually comes back in some way that's better.

ThataKat:
Seen and reviewing, we have some content removal policies nearly finalized just for this kind of situation, but should have an answer back in under a day

ShepardToTheStars:
As command, I rather like hostage ops and Ive seen a greater variety of strategies as hostage ops (cruise ship, stealth ops, loud) in contrast to normal nuke ops.

Wollow-Garf:
I personally enjoy hostage ops, its a game mode that requires you to actually plan out a strat then go with the usual armoury rush and murderboning. (regular nukie rounds are pretty much feels NRP from what I've experienced) There have been some really good examples of more RP focused strats from this mode, i personally think it should still be kept in as it really hasn't been here for long and it does bring more variety to the game as well as makes nukies a little less predictable which can get boring after a while. I do agree that some tweaks are needed so that the rest of the crew can feel involved somehow. Maybe give it a little more time for players to get used to the game mode and tweak it along the way rather than removing it entirely.

ThataKat (approved):
Direction here,

We also like hostage ops, and look forward to what tweaks you come up with to improve it. We don't think it would be at all fair to keep it up when the author wants it paused, especially if they're receiving lots of criticism about it, so it's 100% approved to be taken down in the mean time. 

Hopefully we can see it back in the future!

PR #3154, opened 2025-03-08
Adds Search tab to AAC and updates layout

Oxdeception:
## Licensing (added 2026-01-12)
The contents of this PR are also MIT-licensed and you may port it to any server you wish.

## About the PR
- Made the TabContainer the top level control instead of the ScrollContainer
- Added a Search tab
- Allowed the window to be resized

## Why / Balance
- Finding phrases in the AAC can be difficult if you don't already know where to look.
- Going from phrases at the bottom of one tab to the bottom of another required excessive scrolling.
- Scroll position was not maintained, which meant scrolling was always required for phrases that weren't at the top of the tab.

## Technical details

- Changed the AAC xaml
- Added a filter method for the search bar
- Updated AACBoundUserInterface to follow other BoundUserInterface standards and reduce frame lag

## Media

https://github.com/user-attachments/assets/ce7586b7-2784-4b1c-8168-13f669f74158

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: AAC Tablet now has a searchable tab
- tweak: AAC Tablet no longer has its tabs hidden by scrolling

### DIFF
--- a/Content.Client/_DV/AACTablet/UI/AACWindow.xaml.cs
+++ b/Content.Client/_DV/AACTablet/UI/AACWindow.xaml.cs
@@ -15,6 +15,7 @@ public sealed partial class AACWindow : FancyWindow
 {
     [Dependency] private readonly IPrototypeManager _prototype = default!;
     private readonly List<QuickPhrasePrototype> _phrases;
+    private readonly Dictionary<string, List<QuickPhrasePrototype>> _filteredPhrases = new();
     public event Action<ProtoId<QuickPhrasePrototype>>? PhraseButtonPressed;
 
     private const float SpaceWidth = 10f;
@@ -29,6 +30,7 @@ public sealed partial class AACWindow : FancyWindow
         RobustXamlLoader.Load(this);
         IoCManager.InjectDependencies(this);
         _phrases = _prototype.EnumeratePrototypes<QuickPhrasePrototype>().ToList();
+        _phrases.Sort((a, b) => string.CompareOrdinal(a.Group, b.Group));
         SearchBar.OnTextChanged += FilterSearch;
         PopulateGui();
         FilterSearch(null);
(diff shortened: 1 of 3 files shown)


Content.Client/_DV/AACTablet/UI/AACWindow.xaml.cs
```diff
@@ -27,15 +28,29 @@ public AACWindow()
...
+    private void FilterSearch(LineEdit.LineEditEventArgs? obj)
     {
-        var phrases = _prototype.EnumeratePrototypes<QuickPhrasePrototype>().ToList();
+        SearchResults.DisposeAllChildren();
+        var emptySearch = string.IsNullOrEmpty(SearchBar.Text);
+        var searchedPhrases = _phrases
+            .Where(it => emptySearch || Loc.GetString(it.Text).StartsWith(SearchBar.Text, StringComparison.CurrentCultureIgnoreCase))
+            .GroupBy(it => it.Group)
+            .OrderBy(it => it.Key)
+            .ToDictionary(group => group.Key, group => group.OrderBy(it => Loc.GetString(it.Text)).ToList());
```
deltanedas: this is way too much linq create 1 dictionary and do it manually
also searching should use contains not start with

deltanedas (changes requested):

Content.Client/_DV/AACTablet/UI/AACWindow.xaml.cs
Oxdeception: I personally found the search results felt better with starts with than contains, but I'm fine changing it if contains is the preferred option.

Can you please elaborate on the one dictionary manually?

deltanedas: using linq especially chaining multiple calls like this, and even nesting them, does a huge number of unnecessary allocations
"unroll" it and do it by hand by
1. creating a single dictionary for the control in its constructor
2. in this, `Clear`ing it then iterate each phrase and add it to an existing or new group list with the filtering logic
3. at the end iterate the dictionary and sort each value's list

copy whatever criminal records uses for searching though, people are used to it

Oxdeception: Understood, wasn't the most familiar with Linq. Will fix

Oxdeception:
Criminal records does appear to use contains

deltanedas (approved):

PR #3100, opened 2025-03-03
Remove roundstalling ion storm law

sowelipililimute:
## About the PR
Removes the ion law that forces borgs to not let the shuttle be called

## Why / Balance
Goes directly against our server rules

## Technical details
- comment out a line

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- remove: Borgs will no longer get a law to recall the evac shuttle

PR #3159, opened 2025-03-08
invert ion storm shuttle law

deltanedas:
## About the PR
changed it to THE SHUTTLE MUST BE CALLED BECAUSE OF WHATEVER ON THE STATION

## Why / Balance
resolves #3100 this is better gameplay and implementation wise

while round stalling may not be against the rules for borgs with that law to do, its still just as annoying if a borg makes a hidey hole in some impossible to find area and constantly recalls it

it still can lead to the same borg hunt as before, but also if its like because of the 10 gazillion passengers on the station, the borg might be incentivised to kill all passengers :trollface:

it also means the borg doesnt have to wait for someone to call evac to start the shenanigans, as soon as you get the law you can do it

**Changelog**
:cl:
- tweak: Changed the shuttle-related ion storm law so it forces you to call the shuttle, rather than recall it.

### DIFF
--- a/Content.Server/Silicons/Laws/IonStormSystem.cs
+++ b/Content.Server/Silicons/Laws/IonStormSystem.cs
@@ -231,7 +231,7 @@ public sealed class IonStormSystem : EntitySystem
         return _robustRandom.Next(0, 35) switch
         {
             0  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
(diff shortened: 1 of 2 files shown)


ThataKat:
24 on this one

ThataKat:
Misinfo, we need some more time to cook up some specific law phrasing but we're working on alternatives that are in a similar vein. Will respond in another 24 hours.

deltanedas:
fell off + upstream merge has it now

PR #3169, opened 2025-03-09
rework hyperlathes into upgrade kits

deltanedas:
## About the PR
replaces hyperlathe boards with upgrade kits for hyper-convection and the new cryo-stabilizing lathes
they can be used on any techfab/lathe/etc instead of just circuit printers (after protolathe is phased out)

hyper-convection works like before but for any lathe now
cryo-stabilizing freezes the air instead of heating it, and it makes the lathe faster. no change in material use.

## Why / Balance
instead of getting an igniter and swapping circuit printer board, now engineering makes upgrade kits and they can be used on *any lathe*. this is especially important as with techfabs it effectively removed hyperlathes since protolathe and autolathe are being phased out, leaving only 1 machine usable with this 10k point tech, now all departments can be un-controlled deathzones, not just epi

gives engi/atmos something to do real

## Technical details
- UpgradeKitComponent adds components to valid machines
- LatheUpgradeComponent changes lathe stats when upgrading
- UpgradedMachineComponent to prevent stacking upgrades and to show the upgrade in examine

## Media

behold codersprite

upgrades real

hyper upgraded lathe is cheaper real

## Requirements

**Changelog**
:cl:
- add: Added lathe upgrade kits to the Industrial Engineering tech, they can be used on any kind of lathe including techfabs.
- add: Added the cryo-stabilizing lathe upgrade kit, it makes lathes faster in exchange for freezing the air.


sowelipililimute:
direction review started expect a response in 1 to 3 days

sowelipililimute:
uncontroversial approval from direction

sowelipililimute (approved):

PR #3171, opened 2025-03-09
upstream merge 09/03/2025

deltanedas:
## About the PR
- det gets inspector gadget hat to grab items via voice command
- moths and arachnids have unique glasses sprites
- all species can use most undergarments, using upstream ones now.
  you can also turn off nudity now and force everyone to have underwear clientside via the accessibility option
- vending machines are predicted
- **removed metempsychosis** as even before this it was extreme shitcode, now with the cloning rework it needs to be fully rewritten and i don't care to as its rarely used to begin with. the karma shitcode should also just use the mind entity and not be hacked onto all cloning code.
  last time i saw someone reincarnate as a gorilla was on angle station...
- nukie holoclown ops was fixed, could bring it back intentionally in the future i liked seeing some variety :trollface:
- paradox anomaly uses cloning code now
- using upstream no-greentext cvar now
- phalanximine is harder to use, no real source of genetic so doesnt matter much. doxa exists too so cryo gaming

## Excluded
- babys first mail, its not close to fleshed out enough and migrating all mail to it is a gargantuan task that would block this for ages
- pet carrier can still hold people, pretty sure people like it and toilet thing will be fixed anyway
- remove arith nerf as there is no funny implanter damage

## Media
erp ON

erp OFF

greentext is still fake

## Breaking changes
everything metempsychosis related was nuked

**Changelog**
:cl:
- add: Merged last weeks' changes from upstream, notably excluding their mail system.
- remove: Removed the metempsychosis machine, the research is now for cloning instead.


ThataKat:
Reviewing, 3.5 days max :trollface:

ThataKat (commented):
Everything good to go except inspector gadget hat, Direction might repurpose it later to be an entertainer item (or at least not sec coded) but for now don't let it be obtainable. Can just remove it from spawns or delete it entirely, up to you.

PR #3194, opened 2025-03-11
Escalate borg module being unable to be put into sec borgs

KZhProgram:
## About the PR

I made it so the escalate borg module could be put into sec borgs

## Why / Balance

Because the escalate borg module needs to be able to be put into sec borgs

## Technical details

Changed the parent of BorgModuleSecurityEscalate to be the right ones instead of BorgModuleSecurityDeescalate. The program thought that the escalate and deescalate modules were the same and therefore, since the deescalate module is in by default and cannot be removed, well.. then you can't put in the escalate module

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- fix: Fixed escalate borg modules being unable to be inserted into sec borgs!


KZhProgram:
boi I fucking made a new branch and it told me it made a new branch why is it still doing this :criiii:

KZhProgram:
Deltanerd fixed it already apparently

PR #3201, opened 2025-03-11
fix harpies not being able to host anoms

deltanedas:
missing tag

:cl:
- fix: Fixed harpies being immune to anomaly infections.


dvir001:
Fam just removes the tags, harpy get all the 4 from the base parent and the tags should never been added.

deltanedas:
so true wyci

dvir001:
> so true wyci

Real

PR #3204, opened 2025-03-12
Revert "Shark PSB 1984, replace with tinned meat"

BlitzTheSquishy:
## About the PR

Reverts PR #3191 for PSBs ONLY if PR #3202 gets merged

## Why / Balance

I got lazy and theres no real reason to make the "breather" loadouts more complex than they were for little to no reason

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Loadouts/loadout_groups.yml
+++ b/Resources/Prototypes/Loadouts/loadout_groups.yml
@@ -88,7 +88,7 @@
   - EmergencyOxygen
   - LoadoutSpeciesVoxNitrogen
   - EmergencyIPCBox # DeltaV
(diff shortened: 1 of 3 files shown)


BlitzTheSquishy:
Only after #3202 gets merged

deltanedas:
what about generic water breathing carnivore species 2, carp roundstart player species

BlitzTheSquishy:
> what about generic water breathing carnivore species 2, carp roundstart player species

if what #3202 does is correct, it wont matter about the PSB they could eat it anyway, and theres no need to overcomplicate things (and for some reason if we want more species that use H2O, they can be more easily added)

deltanedas (approved):

PR #3205, opened 2025-03-12
Buff glimmer wisp

Stop-Signs:
## About the PR
made it shoot a 3 round burst at a higher fire-rate

## Why / Balance
these things require 300 glimmer minimum to spawn, they should be slightly more evil. eventually they should also be unpullable but forced inherit is evil.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Glimmer wisps are now deadlier, psionics beware


deltanedas:
this is too strong, the AI being dogshit is the real cause of wisps being a meme (the amount of times i just walk past a wisp as a psionic and it does nothing)

ThataKat:
Direction consensus is that wisps are strong enough, they just have AI issues. This doesn't solve the AI problems, but will make them unnecessarily strong.

PR #3246, opened 2025-03-16
HoS Kit Buff -X-01 Improvements

Avalon-Proto:
2

## About the PR
Simply lets the X-01 self charge, and speeds up the self charge of the Energy Shotgun. Also increases the size of both of their batteries

## Why / Balance
The X-01 was in desperate need of a buff

## Technical details
Halved the firing cost of the X-01

## Media
X-01 Disable+Lethal
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Heads of Security rejoice, your equipment has been re-tuned. The X-01 now has been streamlined and can fire twice as many shots before needing a charge

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -805,7 +805,7 @@
   - type: GunRequiresWield #remove when inaccuracy on spreads is fixed
   - type: Battery
     maxCharge: 2400 # DeltaV - Double charge
-    startingCharge: 2400
+    startingCharge: 2400 # DeltaV - Double charge
   - type: BatterySelfRecharger
     autoRecharge: true
     autoRechargeRate: 72 # DeltaV - Triple recharge rate
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -804,10 +804,10 @@
     stealGroup: WeaponEnergyShotgun
   - type: GunRequiresWield #remove when inaccuracy on spreads is fixed
   - type: Battery
-    maxCharge: 2400 # DeltaV - Double charge
-    startingCharge: 2400
+    maxCharge: 1200
+    startingCharge: 1200
   - type: BatterySelfRecharger
     autoRecharge: true
-    autoRechargeRate: 72 # DeltaV - Triple recharge rate
+    autoRechargeRate: 24
     autoRechargePause: true
     autoRechargePauseTime: 30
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -804,10 +804,10 @@
     stealGroup: WeaponEnergyShotgun
   - type: GunRequiresWield #remove when inaccuracy on spreads is fixed
   - type: Battery
-    maxCharge: 1200
-    startingCharge: 1200
+    maxCharge: 2400 # DeltaV - Double charge
+    startingCharge: 2400
```
deltanedas: comment too

deltanedas (changes requested):

sowelipililimute:
direction review was started 2 days ago, sorry for the lack of communication

Stop-Signs:
Direction here, we like the capacity buff to the x-01, but we don't think the hos should have two self recharging guns.

Furthermore we think the energy shotgun is fine in it's current iteration and have plans for potential changes in the future.

Avalon-Proto:
> Direction here, we like the capacity buff to the x-01, but we don't think the hos should have two self recharging guns.

What if I gave the X-01 a timer before it starts self charging much like the E-shotgun. So that the HoS cannot just keep shooting and shooting?

> Furthermore we think the energy shotgun is fine in it's current iteration and have plans for potential changes in the future.

IMO the E-shotgun is still outclassed by every other shotgun in the armory even with the self charge. Which is slow. If you wish for the changes to be reverted I'll do it. But I would like to request to keep the self charging on the X-01 just set to a delay instead of being constant

Stop-Signs:
Direction here! We still don't think that the energy shotgun needs an buffs at the moment. And we are firm that the x-01 should not have self recharge. that being said, we think the x-01 should absolutely have more capacity, and would be fine with it having 20 lethal shots and 40 non lethal shots.

Avalon-Proto:
@Stop-Signs Right, so I moved it around a bit, instead of doubling the size of the battery, I cut fire costs in half. That way it doesn't take 10 years for HoS to recharge their pistol

Lyndomen:
> @Stop-Signs Right, so I moved it around a bit, instead of doubling the size of the battery, I cut fire costs in half. That way it doesn't take 10 years for HoS to recharge their pistol

40 shots should take a year to recharge, -1

Avalon-Proto:
> 40 shots should take a year to recharge, -1

It still takes a while to charge, its not an instant charge

Stop-Signs:
Direction approves, 100 seconds recharge is exhaustingly long. Should be good to merge unless fish guy has anything else.

EDIT: this still has self recharge for the x-01 which direction does not approve of.

Avalon-Proto:
> Direction approves, 100 seconds recharge is exhaustingly long. Should be good to merge unless fish guy has anything else.
> 
> EDIT: this still has self recharge for the x-01 which direction does not approve of.

Woops-

Avalon-Proto:
Right fixed the self charge, forgot to delete it. Should be good to go @deltanedas

deltanedas (approved):

deltanedas:
just update cl to say its shots doubled

Avalon-Proto:
Done

PR #3256, opened 2025-03-17
Make civilian implanters require an open incision to work

sowelipililimute:
## About the PR
Civilian implanters (i.e. not syndicate implanters) require an open incision to be implanted, requiring surgical intervention.

## Why / Balance
As it is, civilian implanters are essentially all reward and no inconvenience, resulting in e.g. command and security members using tracking implanters designed for prisoners on themselves because of the pure utility. Rather than trying to legislate this under powergaming, making it require surgery means that there has to be an actual consideration and effort if it's worth it to go under for a bit for future safety or not.

## Technical details
- new ImplanterSurgerySystem that blocks implanting if there isn't an open torso incision unless there's a ImplanterSurgerylessComponent 

## Media

https://github.com/user-attachments/assets/af537925-b0d3-48f1-9055-fad6a765731d

## Requirements

**Changelog**

:cl:
- add: Civilian implanters (non Syndicate implanters) now require an incision to be surgically opened to be implanted

### DIFF
--- a/Content.Shared/_DV/Implants/ImplanterSurgerySystem.cs
+++ b/Content.Shared/_DV/Implants/ImplanterSurgerySystem.cs
@@ -1,0 +1,34 @@
+using System.Linq;
+using Content.Shared._Shitmed.Medical.Surgery.Steps.Parts;
+using Content.Shared.Body.Components;
+using Content.Shared.Body.Part;
+using Content.Shared.Body.Systems;
+using Content.Shared.DoAfter;
+using Content.Shared.Implants;
+
+namespace Content.Shared._DV.Implants;
+
+public sealed class ImplanterSurgerySystem : EntitySystem
+{
+    [Dependency] private readonly SharedBodySystem _body = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<BodyComponent, AddImplantAttemptEvent>(OnAttemptImplant);
+    }
+
+    private void OnAttemptImplant(Entity<BodyComponent> ent, ref AddImplantAttemptEvent args)
+    {
+        if (HasComp<ImplanterSurgerylessComponent>(args.Implanter))
+            return;
+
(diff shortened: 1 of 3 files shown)


NoElkaTheGod:
Would be cool if different implants required an incision on different body parts. Mindshield, for example, makes much more sense to be implanted in the head IMO.

Content.Shared/_DV/Implants/ImplanterSurgerySystem.cs
```diff
@@ -0,0 +1,34 @@
+using System.Linq;
+using Content.Shared._Shitmed.Medical.Surgery.Steps.Parts;
+using Content.Shared.Body.Components;
+using Content.Shared.Body.Part;
+using Content.Shared.Body.Systems;
+using Content.Shared.DoAfter;
```
deltanedas, line 6: DoAfter is unused?

deltanedas (changes requested):

Content.Shared/_DV/Implants/ImplanterSurgerySystem.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        SubscribeLocalEvent<BodyComponent, AddImplantAttemptEvent>(OnAttemptImplant);
+    }
+
+    private void OnAttemptImplant(Entity<BodyComponent> ent, ref AddImplantAttemptEvent args)
+    {
+        if (HasComp<ImplanterSurgerylessComponent>(args.Implanter))
+            return;
+
+        var child = _body.GetBodyChildrenOfType(ent, BodyPartType.Torso).Select(it => it.Id).FirstOrDefault();
```
deltanedas, line 27: add this to a field on implant component then get it here instead of hardcoding
lets mindshield require a head incision etc

deltanedas (changes requested):

deltanedas:
@sowelipililimute

PR #3271, opened 2025-03-19
count dead mobs as sleeping for surgery

deltanedas:
## About the PR
cant squirm if youre dead can you

**Changelog**
:cl:
- fix: Fixed surgery dealing damage to dead people.


sowelipililimute (approved):
oh god i didnt realise that wasnt already the case

PR #3274, opened 2025-03-19
add syndie autodoc bundle

deltanedas:
## About the PR
- adds the syndie autodoc bundle for 10 tc, has everything you need to operate on yourself in the comfort of your maints/lavaland/ats/??? hideout (flatpacks/multitool, chloral hydrate so you don't die)
- made syndie autodoc red and already have safety disabled for evil surgery, you can snip+mend the safety wire to re-enable it if you want (especially if you want to use it on yourself only and want to minimize damage from anesthetic not wearing off)

available for both syndies and nukies currently

## Why / Balance
everyone loves syndie hideouts and this makes them stronger
hide stolen items inside yourself, get past pesky cmos "ethical standards", be evil, etc.
10 tc since its not crazy op compared to ghetto surgery but **you dont need friends** which is very strong situationally.
ria worst will also use it for evil surgery 10000% i bet my right lung

its also in the realm of being useful for people in general like buying med hypopens, you can fix the safety and have a fun tool for med early in the round

## Media

so evil

look ma no safety

## Requirements

**Changelog**
:cl:
- add: Added the syndie autodoc bundle for 10 TC, do evil surgeries!


Stop-Signs:
Direction approved, let the organ harvesting begin.

PR #3324, opened 2025-03-26
replace vulp insulation

### LINKED ISSUE #3323: Vulpkanin borderline overheat while wearing hardsuits
## Description
Apparently, something isn't working right with Vulps and Hardsuits after one of the recent updates, now they just routinely sit on the edge of overheating while wearing a hardsuit for some reason.

**Reproduction**
Spawn as a vulp, put on hardsuit, watch the too hot! warning appear then turn off after five seconds, and then come back five seconds later. This will continue until the hardsuit is removed.

**Screenshots**
Not really applicable.

**Additional context**
It may be related to the Ghostrole Temperature bug, so if that gets fixed this might too, no idea on my end though.

deltanedas:
## About the PR
for a very long time insulation on vulps didnt actually do anything but now it does so they get baked in hardsuits
this fixes it by replacing it with a 20 degree lower cold damage threshold, similar to how rimworld cold insulation works

## Why / Balance
fixes #3323 

**Changelog**
:cl:
- fix: Fixed vulpkanins getting [bold]cooked alive[/bold] in hardsuits.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/Species/vulpkanin.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/Species/vulpkanin.yml
@@ -92,8 +92,6 @@
   - type: Perishable
   - type: Damageable
     damageModifierSet: Vulpkanin
-  - type: TemperatureProtection
-    coolingCoefficient: 0.1 # fur is insulating
   - type: Vocal
     sounds:
       Male: MaleVulpkanin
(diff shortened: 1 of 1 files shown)


whatston3:
Would it make sense to change the 315 here to something a bit higher? With an ideal body temperature of 310.15, they reach the threshold to overheat at ~312 and start cooling at 312.15 (+2 K) - in hardsuits, they seem to get overheating notices intermittently.  I think even 317 would clear the margins.

https://github.com/new-frontiers-14/frontier-station-14/pull/3207 boop

deltanedas:
dogs core temperature is around 39C who even made this so low

whatston3:
> dogs core temperature is around 39C who even made this so low

https://github.com/DeltaV-Station/Delta-v/pull/3033/files
🤷

deltanedas:
evil lyndo

PR #3335, opened 2025-03-29
hide surgeries that cant be done yet

deltanedas:
## About the PR
surgery ui only shows surgeries that have the requirement done

this means you have to explicitly do the requirement first like opening an incision

## Why / Balance
gamers will have to experiment on their fellow crew/monkeys or rtfsc to discover things

## Technical details
it will break if draping random bodyparts becomes a thing

## Media
start

incision opened

ribcage opened, now everything is possible

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: You can no longer start surgeries without doing their requirements first.


Content.Client/_Shitmed/Medical/Surgery/SurgeryBui.cs
```diff
@@ -242,6 +239,18 @@ private void OnPartPressed(NetEntity netPart, List<EntProtoId> surgeryIds)
                 continue;
             }
 
+            // Begin DeltaV Additions - only show surgeries with completed requirements
+            if (surgeryComp.Requirement is {} reqId && _system.GetSingleton(reqId) is {} reqUid)
```
Toby222:
```suggestion
            if (surgeryComp.Requirement is { } reqId && _system.GetSingleton(reqId) is { } reqUid)
```

Toby222:
Could use string.Empty instead of `""` but I don't think anyone cares.

PR #3341, opened 2025-03-30
Gamblagator tweaks

Stop-Signs:
## About the PR
Day 1 patch. It's def a little undertuned
shots per fuse 1 -> 3
fire-rate 0.3 -> 1
also fixed the name of the fuses

## Why / Balance
currently the max damage potential of this thing is only 600, rather low when compared to the beam devs 3000. with these changes it will bump this to 1800 damage. the fire rate was stinky and made you have to sit there and wait 3 seconds after wielding, it feels way too clunky this way and this should remedy that. future PR's will be made if this gun under/over performs

## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: Gamblagator now has more shots per capacitor.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Power/powercells.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Power/powercells.yml
@@ -1,7 +1,7 @@
 - type: entity
   parent: BaseItem
   id: GamblagatorCapacitor
-  name: gamblagator fuse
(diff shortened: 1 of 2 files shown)


Avalon-Proto:
Ooooonyou want to rename it to the thing I suggested on the original PR. Rename it to "Anti-Matter Accelerator" ooooo

ThataKat (approved):
Yes

PR #3353, opened 2025-03-31
CSI: DeltaV (Gun Forensics)

BarryNorfolk:
giphy

## About the PR
Expands out gun forensics so that security and/or the justice system can identify the murder weapon and who might have done it.

I've purposefully left the weapons touched in this PR to be just the revolver, as I would like direction and others to tell me which weapons and projectiles should be "Lodgeable" into a target. 

## Why / Balance
Largely the forensics for guns is lackluster and you have to base your detective skills on someone saying "I was shot by a python", hoping that you find only a single version of this gun to "Prove" that it was that specific gun. Here we can now start having things like firearm residue on people that have fired a gun, and match bullets to specific weapons.

## Technical details
Bullets are stored inside a component via a hashset, the reason being to limit the entity spam when the eager surgeon removes over 8000 bullets from Hammy after a particularly bad round with nukies.
If the rifling is different, or the round is different, then that constitutes a whole new "bullet" to be removed and looked at by the security team.

One annoyance here is that residues are not managed well and, although their name implies a more general use, they were always meant to just compound on one another. So in theory you could wash an item with all the soap and the residues would all stack on one another, which is irritating.

Because gunshot residue is now a new residue to be cleaned, soap now will always try and clean an entity, even if there is no "evidence" to clean off.

## Media
Gunshot residue on things:
image
image

New projectile retrieval via a hemostat
image

Which allows us to see the spent bullet, and the name specified in the prototype:
image

and finally, the rifling indicator that is unique to this weapon:
image

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- add: Added gunshot residue to the weapon and user when firing a gun, cleanable with soap!
- add: Bullets hitting a valid entity now have a chance to become lodged, retrievable with a hemostat.
- add: Rifling forensics now allows users to match bullets to weapons.
- tweak: You'll never be sure if an item has forensic information on it anymore, be sure to clean things thoroughly.
-->


deltanedas:
need to wait for wounds for this so foreign bodies are better integrated

BarryNorfolk:
Is this next week? :trollface:

Toby222:
How do you see the rifling of a gun? Or are you intended to just compare bullets?

BarryNorfolk:
Yep, take gun and fire it into a target dummy. Get bullet from dummy and compare.

ThataKat:
Since the conditions for this (woundmed) are nextweek not gonna take this under direction review, but badger us again when it's ready for official look over

Toby222 (commented):
Not sure how to review this if direction says it's shelved until `other feature` gets added so I'm only commenting.
I didn't see anything egregiously wrong when looking through and the functionality seemed to work as intended

Re-request review when ready I suppose :^)

Toby222:
Gonna close this pending activity

PR #3372, opened 2025-04-04
add apc wires

deltanedas:
## About the PR
apc now has wires:
- 2 power wires which disable the apc battery entirely. 60/30s timeout when pulsed, twice as long as doors
  light is fucked because
- ai access so you can counter ai toggling an apc constantly
- access for trolling, no real impact compared to now where you can screw -> crowbar an apc to kill it

to deconstruct an apc you now have to cut all the wires so trolling requires insuls

## Why / Balance
we love ai spam toggling an apc
we love instantly deconstructing apcs

## Technical details
there was no reason upstream pr tried networking it since nothing in shared/client used it. if wires get moved to shared itll conflict automatically so it can be done properly then

## Media
ai access disabled, ai cant use apc

deconstructing real

all lights go off when power is cut

power wires snipped, it says Disabled in the ui

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- add: APCs now have wires, notably the AI access wire which can be snipped to prevent the AI from toggling APCs.

### DIFF
--- a/Content.Client/Power/APC/UI/ApcMenu.xaml.cs
+++ b/Content.Client/Power/APC/UI/ApcMenu.xaml.cs
@@ -65,6 +65,13 @@ namespace Content.Client.Power.APC.UI
                     default:
                         throw new ArgumentOutOfRangeException();
                 }
+                // Begin DeltaV Additions - set it to Disabled when power wire is snipped
+                if (!castState.PowerEnabled)
+                {
+                    ExternalPowerStateLabel.Text = Loc.GetString("apc-menu-power-state-disabled");
+                    ExternalPowerStateLabel.SetOnlyStyleClass(StyleNano.StyleClassPowerStateNone);
+                }
+                // End DeltaV Additions
             }
 
             if (ChargeBar != null)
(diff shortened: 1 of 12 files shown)


Toby222 (commented):
Would be nice if cutting the power wire also turned off the screen

deltanedas:
made it show up in the ui instead of preventing opening it as its more consistent with apcs functioning without mv power etc

Toby222:
I meant hiding the display layer in the ApcVisualizer. So you can visually differentiate an APC with low battery from one with a cut power wire, that shouldn't stop opening the UI, right?

sowelipililimute:
this was picked up by direction for review, <24hours

sowelipililimute:
direction review was started 24 hours ago, sorry for the lack of announcement 

consensus is generally positive but with concerns raised about lowpop stations not necessarily having access to engineers to do a sanctioned toggle for the breaker trip event

Avalon-Proto:
> direction review was started 24 hours ago, sorry for the lack of announcement
> 
> consensus is generally positive but with concerns raised about lowpop stations not necessarily having access to engineers to do a sanctioned toggle for the breaker trip event

With the tactical hacking reversion, the other PR that was merged this morning, all APCs will also have a standard access wire. Across all of them, It wont shock you if you snip it, and it will let you freely toggle the APC. so that shouldn't be an issue

deltanedas:
that, you get at worst 2 shocks before being able to turn off access for all apcs

PR #3412, opened 2025-04-09
add pre-commit for checking and fixing code

deltanedas:
## About the PR
if you have pre-commit installed (`pip3 install pre-commit` / `xi pre-commit` on void) it will check these automatically before committing changes:
- prototypes yml syntax for yml warriors
  it only checks syntax, it cant know things like non-existent `!type:...`, use robust yml vskode extension for that
- textures meta.json files
- config presets syntax
  it obviously cant check that the cvars are valid but its something
- merge conflict markers (`<<<<<< ====== >>>>>>`)

pre-commit.ci will also fix these for your prs:
- makes sure your files end properly
  dont use notepad...
- including BOM in utf-8 files as it doesn't have byte endianness
  configure your ide to not add these or suffer
- removes trailing whitespace automatically

this only applies to our files as to not create 5000 conflicts with upstream or fork cherry picks

if you have pre-commit installed **all you have to do is checkout a branch** and it'll update the hooks. run `pre-commit run -a` to fix a pr if you want

## Why / Balance
so i stop crying about bom trailing newlines and shit

## Media
see [this commit](https://github.com/DeltaV-Station/Delta-v/pull/3412/commits/a2da4ec4d94fea48dd7d56e7737a5a3b73b31c93) for automatic fixes (mostly bom removal)

## Breaking changes
BOM and trailing whitespace removed from all _DV files
they also now all have proper line endings

forks may want to add their own folders with `(_DV|_EpicFork)` to match either, or replace _DV entirely to avoid potential conflicts


sowelipililimute (dismissed):
ough

PR #3432, opened 2025-04-11
Make pacifists be able to use disablers

Quanteey:
## About the PR

Made the disabler, practice disabler, and the smg variant usable by pacifists

## Why / Balance

They no longer do any form of damage other than stamina. Stun batons in stun mode can already be used by pacifists, which is an inconsistency.
Either:
- Make stun batons and any other form of stunning unusable 
- Make stun weapons that do not do damage usable

I think the second option is better, because this provides a middle ground for pacifist characters, allowing them to play security. And if the player takes pacifism seriously, then they will not touch disablers for RP reasons rather than mechanical ones.

## Technical details

PacifismAllowedGunComponent

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- tweak: Pacifists can now use disablers

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -482,6 +482,7 @@
     - Taser
     - Sidearm
     - ConcealCarry # DeltaV - can be stowed in military boots for now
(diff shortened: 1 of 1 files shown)


Toby222:
That's the SMG and practice disabler, you missed the actual disabler

Quanteey:
The actual disabler inherits from the practice disabler, this is why I did not add the component to it

Toby222 (approved):
So true

deltanedas (approved):

Issue #3289, opened 2025-03-21
Admin assistant has become validhunt express


Stop-Signs:
## Description
90% of people who play this role use it to be a bridge guardian, and more often than not have powergamed items in their inventory that they obtained with their title and access. If you want to guard bridge, play sec.
**Reproduction**
join server, specatate admin assistant.

**Screenshots**

![Image](https://github.com/user-attachments/assets/0889cf95-6007-440c-8de7-895459b1269f)

**Additional context**

Vova567810:
Quite true, considering half of time nobody picks AA as learner role, and even if they do, command often is either neglecting to teach them or simply too busy with their departmental duties. The only exception is maybe being sidekick to captain, but again they might not be interested. And by my own opinion, the whole issue of AA being just tider with command access and having almost 0 responsibilities besides maybe doing couple paperworks once in blue moon or be glorified bartender for giving out lattes and other coffee drinks.

Lyndomen:
I am pushing for [Content Removal](https://wiki.deltav.gay/view/Direction_Guidelines#Content_Removal).

Issues Identified: 

- Administrative Assistant does not encourage players as a role to go out and seek to provide assistance to the command staff
- Administrative Assistant has no mechanical job functions, leading players to sit at the bridge and play bridge guardian. This alone does not prohibit the job from existing, but it exacerbates the first issue.



Some ideas floated that might address these:

- Putting the AA spawn/locker closer to the bar. 
- Removing AA for BSO.
- Whitelisting AA
- Rolebanning more AA 

In my opinion, this job suffers from too much overlap with the HOP and Clerk, and should be removed. 

I would like to push for a deadline of 2 weeks to have a solid plan for our next steps @DeltaV-Station/direction.

SirWarock:
I'd hate to see the Command Tider be deleted.
Though from a play experience, I can definitely say it missed it's mark for anything it was feasibly meant to be.
- It's not a learning role, because there's nothing to learn, and nobody will teach you.
- It can't assist anyone, because why would they need the assistance of someone who can do nothing but open Bridge doors?
Only time I ever needed the AA's help was for their access configurator, because that's the **only thing they can do.**
They don't even have _their own access, every command can open their doors and locker_.

**However**
It's been so fun to play AA with either a competent Captain (Alexis), or with your own imagination.
[...]

PR #1207, opened 2024-05-15
Administrative Assistant, a new RP / Support centric "command" role

noctyrnal:
## About the PR

Adds the Administrative Assistant role to all stations, acting as both a secretary for command and as a worker on the bridge, answering faxes , grabbing coffees for command members and so on. Intended to support command members in more menial tasks, and to provide unique RP circumstances.

## Why / Balance

Makes command's life a bit easier to have a person dedicated to helping them do the small things

## Technical details

None

## Media

f77bc14897d700be8c36fde6fb72c6af

- [ X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Added the Administrative Assistant Role! Assist command in their day-to-day activities, grab the captain a coffee, answer faxes in the bridge!

PR #3438, opened 2025-04-13
lockout admin assistant after too much playtime

Stop-Signs:
## About the PR
if you have 15+ hours as command you can no longer play admin assistant

## Why / Balance
solves #3289 

admin assistant players refused to stop maining the role solely to validhunt people who walk into bridge. A lot of people are tired of this role being terrible and 15 hours is MORE than generous  to learn basic command duties.

## Technical details
yml

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak:  if you have 15 hours on command you can no longer play as admin assistant

### DIFF
--- a/Resources/Prototypes/_DV/Roles/Jobs/Command/administrative_assistant.yml
+++ b/Resources/Prototypes/_DV/Roles/Jobs/Command/administrative_assistant.yml
@@ -6,6 +6,10 @@
   requirements:
   - !type:OverallPlaytimeRequirement # DeltaV - Playtime requirement
     time: 90000 # 25 hours
+  - !type:DepartmentTimeRequirement
+    department: Command
+    time: 54000 #15 hrs
+    inverted: true # no playing this after 15 hours of it
   startingGear: AdminAssistantGear
   icon: "JobIconAdminAssitant"
   supervisors: job-supervisors-command


deltanedas:
conflicts with funny hat

Avalon-Proto:
I thought we intentionally DONT lock learner roles so people can RP as if they are new to something

deltanedas:
me rping an assistant that assists by guarding the bridge with the axe 24/7 and reporting each individual sound heard

laikatiel:
this feels more like a "roleban/have a talk with that individual" situation rather than permanently lock an RP role for having command playtime.

Radezolid:
I say drop the learner role part of it and punish players as a full on command members. I gave them the learner role part, the original PR didn't include that https://github.com/DeltaV-Station/Delta-v/pull/1207.

foxcurl:
agreeing with radezolid here, ive never even see someone use the role as a command learner one and the AAs who are most proactive about leaving the bridge are some of the Mains. all this would do is give each player just 15 hours of bridgecamping while bored since command is still going to forget they exist.

this would be a mapping strain, but giving them an office either away from key areas on the bridge or the bridge itself may be more effective at actually reducing  bridgecamping. imo AA should get basic departmental access excluding sec so they can actually meet heads in their departments as well as a master key so they can be proactive in tasks but thats a different debate.

Stop-Signs:
closing this because i dont think this is the solution i want to go with. but we definitely should start rolebanning the wannabe sec AA's

PR #3445, opened 2025-04-13
13APR2025 Rules Update

Colin-Tel:
## About the PR

Minor changes to rules.
## Why / Balance

Rule B1.5 added to put into writing something that was assumed.
Rule C3 was adjusted to make it permissible in certain situations for an antagonist to use explosives on or around the evac shuttle, provided that they have sufficient roleplay reason to do it. While this wasn't the case in writing before, there have been a select few instances where it was okay for the antagonist to blow up a portion of the evac shuttle because it was a result of a threat they made via roleplay. The fact that it is in writing now does not mean that there will necessarily be a higher amount of shuttle bombings than usual, however now that it is out there that such a thing is acceptable as a result of roleplay, it could certainly give people some ideas.
## Technical details

**Changelog**

:cl:
- tweak: Adjusted the server rules, adding rule B1.5 and clarifying an example towards the end of rule C3.


Avalon-Proto:
Whys this only a draft :blunt:

Stop-Signs:
new headmin can make a new PR whenever they need rules changed

PR #3450, opened 2025-04-13
Correcting Harpy Height Minimun back to 90% for smoll Harpies.

KitGit001:
Make Harpies Smol again!!!
## About the PR
Reverted Harpy min height back to 90%

## Why / Balance
Prior to this change Harpies were grouped in with Felinids and Rodentia as "small" species with their minium height being limited to 95 instead of the normal 90. dispite having signifigent overlap with humans and zero overlap with felinids with even the shorest harpies being no shorter then the tallest felinids, look at area there is even greater size disparty with defult sized harpies being larger by area then humans and 217% the size of felinids by area due to harpies being the only sprite that is wider then it is tall, there was not really a valid ballence reason to limit their minimun size the same as felinids and rodentia as they align much closer to average in height being signifigantly taller then rodenia and felinds and having one of the lagest area excuding the 2 large species and acrachinds who are also wide but taller. Additonaly ballence wise Harpies are just as fragile as filinids with only rodentia being lightly weaker to brute, well Harpies are lighter and posessing weaker unarmed attacks, with just piercing damage, as opposed to the dual damage type attaks of felinids and rodentia with one being slashing and peirceing, and the other being slasing and blunt, ass a result Harpies loose to both felinids and rodentia in unarmed combat, and lack the stealth advantgaes of them, making them the weakest species when their size is taken to account being both weakest combat species and below average in stealth do the lagre area of their spites and easily distingisable fetures, this is not taking in to accouynt that their species vision modifier makes sec and nukies hardsuits the same color making both stealth and combat mishaps more likly as you mistake enemies for allies, or nutrals, the only 2 advantages that harpies get are their ability to imitate sounds and their slight sleep buff with is all but use useless when you consider how much slow they experience when wearing any heavy armor or using any heavy weapions, or are carriering or dragging anything, as their mush weaker and slowed down much more then others. Though all none of that is the issue.

Primaraly I want smoll cute harpies again
 Think like songbird harpies opposed to eagles, chickens and other larger birds(yes chcickens are not that big but they are massive compaired to most song birds, finichs, and robins and such.
 and taking in to account animal sizes well a chicken maybe the same size, larger or even smaller then a cat, cats are universaly larger then song birds, so it would make sone sence if the very mallest harpeis could be a bit smaller then the very largest felinids.

 With this change Hapies with still be much larger on average then Felinids, with 1.00 to 1.10 harpies over laping with 0.90 to 1.00 humans like they curently do, but with the their minimum reverted to 90% will now have slight overlap with the very tallest Felinids, with 0.90 to 0.95 Harpies overlaping with 1.05 to 1.10 felinids. with harpies between 0.95 and 1.00 having zero overlap.
## Technical details
The Only code Change is changing the 0.95 to 0.90.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Harpy Min Height Reverted to 90%

### DIFF
--- a/Resources/Prototypes/_DV/Species/harpy.yml
+++ b/Resources/Prototypes/_DV/Species/harpy.yml
@@ -8,7 +8,7 @@
   dollPrototype: MobHarpyDummy
   skinColoration: HumanToned
   baseScale: "0.9, 0.9"
-  minHeight: 0.95
+  minHeight: 0.90
 
 - type: speciesBaseSprites
   id: MobHarpySprites


Toby222 (approved):

Toby222:
Direction said it's okay when I asked them

Toby222:
Oh apparently direction review 24h

ThataKat (approved):
Sorry for the time delay, good to go.

PR #3462, opened 2025-04-14
rework holofan to be like tg

deltanedas:
## About the PR
instead of a battery it uses 6 limited charges like flashes to place holofans
you can then click on existing holofans to reclaim charges (if its not full already)

holofans no longer despawn automatically so you dont have to keep refreshing them manually

## Why / Balance
its almost identical to this on tg

micromanaging stealing batteries from maints lockers is awful "gameplay", and if you dont do that than either:
- you patiently wait 10 years for a charger to finish
- you forget to recharge and have a paperweight to prevent a tritfire spreading, have fun

it also means for non atmos that once the atmos takes down holofans in an area its probably fine, compared to the 3 minute despawn time for a ~10 second fix

the upgrading is pathetic, you essentially get nothing until 1 hour in and you have a microreactor so you can crank out infinite holofans. high-cap cells offer no convenience over just getting another cell from maints

this lets you reclaim holofans immediately when you are done with something like fixing a hole to let air back in easily

holofans not despawning is fine since:
- **anyone with a holofan can steal them instantly** so if you are using it to hold in a burn chamber just wait for someone to prank you :trollface:
- explosions still delete holofans, and now your projector permanently loses charge for it
- if you are using them in a permanent setup you are less likely to check on it than before, so theres more opportunity for someone to mess with them than if you are camping it to refresh every 3 minutes

may need to make the recipe more expensive if it turns out too powerful

## Technical details
its all predicted thanks to using a container instead of spawning/deleting the holofans

## Media
no reclaiming when full

fans really dont have TimedDespawnComponent

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Holofans now work like in SS13, you can place up to 6 at once with charges reclaimed by removing unused holofans.
- remove: Removed the battery requirement from holofan projectors.

### DIFF
--- a/Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
+++ b/Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
@@ -46,7 +46,7 @@ public sealed class ChargeHolosignSystem : EntitySystem
             return;
 
         var containers = Comp<ContainerManagerComponent>(ent);
-        for (int i = 0; i < charges.MaxCharges; i++)
+        for (var i = 0; i < charges.MaxCharges; i++)
         {
             if (!TrySpawnInContainer(ent.Comp.SignProto, ent, ent.Comp.ContainerId, out _))
             {
--- a/Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
+++ b/Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
@@ -1,0 +1,115 @@
+using Content.Shared.Charges.Components;
+using Content.Shared.Charges.Systems;
+using Content.Shared.Coordinates.Helpers;
+using Content.Shared.IdentityManagement;
+using Content.Shared.Interaction;
+using Content.Shared.Popups;
+using Content.Shared.Storage;
+using Robust.Shared.Containers;
+using Robust.Shared.Map;
+using System.Linq;
+
+namespace Content.Shared._DV.Holosign;
+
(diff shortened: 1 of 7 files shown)


Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
```diff
@@ -0,0 +1,119 @@
...
+        ent.Comp.SignComponent = EntityManager.ComponentFactory.GetRegistration(ent.Comp.SignComponentName).Type;
+    }
+
+    private void OnMapInit(Entity<ChargeHolosignProjectorComponent> ent, ref MapInitEvent args)
+    {
+        if (!TryComp<LimitedChargesComponent>(ent, out var charges))
+            return;
+
+        var containers = Comp<ContainerManagerComponent>(ent);
+        for (int i = 0; i < charges.MaxCharges; i++)
```
Toby222:
```suggestion
        for (var i = 0; i < charges.MaxCharges; i++)
```

Toby222 (changes requested):

Toby222 (approved):

perryprog:
Unsure if intentional (can make issue if so) but this does make it possible to have no way to remove holofan (without seismic charge I guess) if you lose the holofan that placed one. If you use a different holofan then you have to place a new holofan that you can't pick up somewhere else. Would be nice to have a way to work around this.

deltanedas:
you can just place a holofan on a random tile

PR #3475, opened 2025-04-15
Rebalance licoxide for new cheaper recipes

sowelipililimute:
## About the PR
Licoxide now flushes out of your system faster, can make you scream, and is less likely to shock you

## Why / Balance
- instastun chemical is a bit overpowered for its new cheaper source
- make it work more similarly to razorium
- deltav centcom logs indicate that licoxide was seldom used in its old position of a somewhat expensive traitor chem, so this isn't affecting them too much

## Technical details
- tweak the metabolism

## Requirements

**Changelog**

:cl:
- tweak: Licoxide now flushes out of your system faster & is less likely to stun you, but deals shock damage and may induce screaming

### DIFF
--- a/Resources/Prototypes/Reagents/fun.yml
+++ b/Resources/Prototypes/Reagents/fun.yml
@@ -162,9 +162,17 @@
   color: "#FDD023"
   metabolisms:
     Poison:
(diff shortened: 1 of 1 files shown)


deltanedas (approved):
doesnt change much as its still valid only because of the lead requirement and theres much, much, much better chems to use in that case

PR #3487, opened 2025-04-17
Colin map updates - Dept Economy

Colin-Tel:
## About the PR

Adds department economy related content to Asterisk, Chibi, and Micro.

## Requirements

**Changelog**

:cl:
no

### DIFF
--- a/Resources/Maps/asterisk.yml
+++ b/Resources/Maps/asterisk.yml
@@ -1,11 +1,11 @@
 meta:
   format: 7
   category: Map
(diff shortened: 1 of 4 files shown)


FieldCommand (approved):
Overall good and approved since micro is not in rotation. On Micro, the Justice request console should be accessible to every justice member, which it is not, given that it's in the Clerk office. And the Engineering request console is in the engineering Lobby that anyone can access, and I would like to see it moved to engineering proper with the next future update of micro.

PR #3492, opened 2025-04-17
 Revert polymorphed entities before gibbing if revert on death.

Toby222:
## About the PR
Early merge of https://github.com/space-wizards/space-station-14/pull/36645
Revert polymorphed entities before gibbing if revert on death.

## Why / Balance
Gibbing should count as death and revert. Prevents brain and items like nuke disk from being lost in null space on the parent entity. Also allows polymorphs without damage transfer to be able to revert once gibbed.

## Technical details
made some small changes to the gib trigger moving acidify logic to gib system

## Media
see original PR

## Requirements

## Breaking changes
none

**Changelog**
:cl: SolStar
- fix: Fixed an edge case where items could be lost in nullspace.

### DIFF
--- a/Content.IntegrationTests/Tests/Body/GibTest.cs
+++ b/Content.IntegrationTests/Tests/Body/GibTest.cs
@@ -28,8 +28,8 @@ public sealed class GibTest
         Assert.That(client.EntMan.EntityExists(nuid1));
         Assert.That(client.EntMan.EntityExists(nuid2));
 
-        await server.WaitAssertion(() => server.System<BodySystem>().GibBody(target1, gibOrgans: false));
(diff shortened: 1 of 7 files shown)


SolStar2:
I have some formating changes to push upstream I noticed

SolStar2:
86567a22f4054940f0444d6ac1b163188f3ffd6e can be picked as well

ewokswagger:
oh seems it automatcly did so

SolStar2:
@Toby222 Also sorry but I would prefer SolStar the changelog.

deltanedas (approved):
this will definitely get changed to an event but ill revert later when its time

Toby222:
real

PR #3519, opened 2025-04-19
Allow shields to be used as weapons

BarryNorfolk:
## About the PR
Enabled shields to be used as weapons

## Why / Balance
Shields are neat, being able to use them as a melee weapon (with little overall blunt but some stamina damage) seems like a neat idea.

## Technical details
Simple addition of a new component

## Media
Don't think I need to add much but here's the image I'm imagining of me walking down the hallways on a nice peaceful greenshift.
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Shields can now be used as melee weapons for a small amount of damage. No wide swings.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Shields/shields.yml
+++ b/Resources/Prototypes/Entities/Objects/Shields/shields.yml
@@ -59,6 +59,16 @@
                   max: 2
     - type: StaticPrice
       price: 100
+    # Begin DeltaV additions
+    - type: MeleeWeapon
+      canWideSwing: false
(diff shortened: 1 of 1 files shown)


NoElkaTheGod:
Nice idea. The damage is hillariously low though. You'd be better off hitting your opponent with bare hands (which do 5 damage, compared to the shield's 3).

SolStar2:
Dont see a reason we would deney this get john direction aproved

deltanedas (approved):

PR #3533, opened 2025-04-20
make spare id cabinet indestructible

deltanedas:
## About the PR
no longer gets deleted at 300 damage, it had a 50k damage trigger which didnt do anything because of the 300 damage one

also cleanup, its a structure but was in objects folder

## Why / Balance
oops you hit it with a crowbar 100 times now the id just disintegrates
its either:
- be magical indestructible thing only weak to emag/microwave ops
- be destructible with a high damage threshold so with a fire axe you just bypass the biscuit mechanic...

essentially a mald pr because my c4 destroyed free AA

**Changelog**
:cl:
- fix: Fixed being able to delete the spare ID cabinet by doing 300 damage.


Avalon-Proto:
This truly is a fishman moment :trollface:

PR #3494, opened 2025-04-17
Fix sales by making ATS station-owned

sowelipililimute:
## About the PR
- ats now counts as part of the station

## Why / Balance
- new economy code needs that
- technically this makes ats affected by events like power checks

## Technical details
- StationMemberComponent

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- fix: ATS sale palettes work now

PR #3458, opened 2025-04-14
Early cherrypick of Departmental Economy (#36445)

sowelipililimute:
See https://github.com/space-wizards/space-station-14/pull/36445

## Changes from Wizden
- Justice department exists, they took 5% of the security revenue
- It's called logistics not supply

## Media
grafik

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
🆑 EmoGarbage404
- add: Adds departmental economy! All departments now have custom request consoles where they can order from cargo and make money transfers.
- add: Added lock boxes. These are access-locked crates that allow a department to earn sole income when they are sold. Order more from cargo.
- tweak: Slightly increased passive station income.

PR #3534, opened 2025-04-20
Departmental economy revert

foxcurl:
## About the PR

Reverts https://github.com/DeltaV-Station/Delta-v/pull/3458 but not https://github.com/DeltaV-Station/Delta-v/pull/3494; the latter is not departmental economy specific and doesn't seem to break the old system. Mapped consoles and crates are gone.

## Why / Balance

god this got long

I'm happy to let it stay a little longer to be refined some and test how it is in the steady-state. However, based on the feedback in the discord thread and the experience so far, I think that the current implementation is lackluster and only makes things slightly worse, and that any future versions will be a greater negative. If the community/direction (after a re-review) decides they don't want it reverted, that's fine, but I want it to be an actionable option.

Main point:
Fundamentally, little has actually changed with the current implementation. There's some total amount of money on the station, most of which is in Logistics, and if a department needs something they can't afford they ask Logistics. It's possible for them to instead request that Logistics transfers money to the department, but that's even more of a hassle than just walking to their front desk and putting in an order. The primary difference now is that a portion of those funds is just inaccessible to Logistics without them having to ask a department to transfer them. That is, Logistics is unable to properly and effectively allocate or distribute funds as it should be able to. There could be more than enough money in the combined funds of the station to get something that is needed Now, but if they can't get every department to consolidate their funds, then it just can't be purchased. It's simply a _better system_, mechanically and in-universe, to have the consolidated account that Logistics can use to quickly respond to a department's individual needs and the station's collective needs. Funding allocation already existed, but now it's done through difficult and arbitrary barriers (and split income, as if changing the percent of a percent of money a department gets is any better than just transferring the funds they need, or simply purchasing it with a combined budget).

There is, to some degree, a QOL change in being able to order things from your department. However, it does not take that long to walk to logistics and put in an order in the external console or just ask the CT at the desk for what you need. In my experience, the majority of the waiting time for an order is Logi actually flying to the ATS to pick up the crate, and this is the part that sucks the most, because you have nothing at all to do (putting in an order is at least doing Something). Furthermore, if it did work as intended, it just further diminishes the desk job aspect of Logistics. If being a CT was nothing but flying back and forth between the ATS and station, I and many others would not enjoy it. And, in the interest of RP, a lot more of that happens at Logi's desk than in putting a crate in front of Engineering. 

Now, consider a true departmental economy system, where each department is empowered with the means to fund themselves through work and is expected to earn their own money, to avoid the current issue of it being effectively a station budget that is harder to manage. Each department now has to engage in logistics gameplay in order to do their department's own work, and Logistics' own gameplay is diminished greatly. I don't think this is a good system. Departments, in-universe and out-of-universe, should not have to independently be profitable; there's an entire department to ensure the profit of the station as a whole! It's a shift back towards the fascist megacorp NT that we have been trying to get away from. And, if the argument is that it gives departments more to do, they are already free to do things to make the station money. Atmos makes frezon for Logi to sell despite the fact that it doesn't go to an Engineering-specific budget. More depth in departmental gameplay does not necessarily have to be achieved through making money, and again, they should not have to make money in order to do their job. Putting Logistics gameplay into every department makes as much sense as expecting every department to be responsible for the medical treatment of their own members, or fixing their own hull breaches. 

Going through the feedback in the thread:
 - People like departments having a little more agency in what they can purchase by not having to go through logi. I'd like it if I had a weapon to kill the vent critters myself instead of having to call Security, but that's not a change that's happening any time soon.
 - Antisocial behavior in people avoiding Logi if they can help it.
 - Certain SS13 servers have systems that people like, but this is far off from those.
 - LO continues to be even more of a joke role; even trying to be proactive in getting departments what they need backfires when they go ahead and order it themselves. 
 - Logi has more orders to move, which people consider a benefit. It's expected just after giving 90% of the station orders access and a shiny new console in their department.
 - Doesn't fix prior problems with Logistics (fair, that is outside the scope of the PR).
 - Personally, I like being able to withdraw money from the console (allows for antag RP and good cases of economy RP), but that's not exclusive to departmental economies.
 - If Logi gets the majority of funds, why are departments supposed to spend their own money on things when Logi spends most of its money on other departments? 
 - People like the drama of setting department revenue percentages. In the steady-state I don't expect those to change more than once during a shift, because the funding allocation console is just a bad system.
 - Again, if this actually becomes meaningful, it's directly depriving Logi of gameplay and forcing it into other departments. 
 - Going through Logistics is an important step to prevent duplicate/unnecessary orders (especially material ones). The CT probably knows Salvage has brought in a bunch of steel, the service worker doesn't. 
 - CTs have even less context for certain orders, which can lead to confusion. 
 - The PR aims to reduce friction in the ordering process, but does a tiny amount of a small portion of that friction, and does not address the issues that cause that friction to exist in the first place (overall lack of funds or Logi staff). On DeltaV specifically, much of that friction that comes from ordering things at Logi's desk is already diminished; you can NanoChat the CT, you can put something in the external console if they're gone, you can even fax a form to Logistics. It's also rare that if you have good reason for your purchase, or Logi is just very affluent, that your order would be denied. 
 - Even if I were expected to use department funds, why would I ever use a departmental lockbox when I can just bring Logi what I want to sell and ask them to purchase something of equal value? It saves me the time of waiting for three ATS trips so that they can pick up the lockbox, sell it, then pick up what I need when I have the funds. There's nothing stopping me from bringing Logi something with enough value to cover a purchase in the old system, either.
 - Even if it leads to individual economy RP, I'm not sure that that's the direction DeltaV should go, either, and I don't think we should keep this implementation until later versions are ready. 

Improvements to be made, if we are keeping it:
- Logi's 75% cut should not be rigid, LO should get the funding allocation console, and the LO should be able to directly set department budgets as a subsection of the station's funds, as well as having the authority to withdraw and deposit funds into any department account. The FAC should be an overall view of the station's budget, expenditures, and revenue. 
- Having sold crates go primarily to a department's budget should be done with some modifier to a crate, similar to a bounty paper, to avoid needing three ATS trips to get a department what they need if they want to pay for it. 
- Either intern roles shouldn't have access or heads of staff only have access (though restricting access causes problems, having free access causes problems, and I don't think there's a good middle ground).
- Crates ordered by a department should be marked as such, possibly through the order papers being colored.

Overall, the consensus in that thread seems to be that it should be reverted, preferably permanently but at a minimum until it's properly developed to not be a complete net negative. It seems unlikely that any immediate fixes would improve the system sufficiently. 

## Technical details

Instances of mapped consoles, boards, and lockboxes are removed. Affected maps have been tested. Sorry, maptainers. 
Might be possible to instead keep it but leave it unused.

## Requirements

- [ X ] I have tested all added content and changes.
- [ X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- remove: Departmental economies have been reverted.


sowelipililimute:
mind the https://wiki.deltav.gay/view/Direction_Guidelines#Content_Removal guidelines, this needs a quantified list of issues + requirements + proposed fixes/explanation for unfixability + a deadline (i'm going to call it 1mo because of how major of a change departmental economy is)

foxcurl:
> mind the https://wiki.deltav.gay/view/Direction_Guidelines#Content_Removal guidelines, this needs a quantified list of issues + requirements + proposed fixes/explanation for unfixability + a deadline (i'm going to call it 1mo because of how major of a change departmental economy is)

wiki seems to suggest this is done by direction, should I make a github issue myself similar to the secborg one?

foxcurl:
> mind the https://wiki.deltav.gay/view/Direction_Guidelines#Content_Removal guidelines, this needs a quantified list of issues + requirements + proposed fixes/explanation for unfixability + a deadline (i'm going to call it 1mo because of how major of a change departmental economy is)

wiki seems to suggest this is done by direction, should I make a github issue myself similar to the secborg one?

deltanedas:
> There could be more than enough money in the combined funds of the station to get something that is needed Now,

if logi doesnt have enough money to buy some important thing they arent doing their job, money is SO easy to make (just sell mats if salv mines, do bounties otherwise)

deltanedas:
anyway this pr isnt an option because of the misery it would cause with upstream changes, it would have to modify it to be like how it was instead of removing all the code

dvir001:
> anyway this pr isnt an option because of the misery it would cause with upstream changes, it would have to modify it to be like how it was instead of removing all the code

Curse of downstream.

foxcurl:
> anyway this pr isnt an option because of the misery it would cause with upstream changes, it would have to modify it to be like how it was instead of removing all the code

yeah figured that was likely, fair

PR #3528, opened 2025-04-20
Kitsune bug fixes

SolStar2:
## About the PR
Several bug fixes:
stamina damage transfers both ways and moved to polymorph system
innate access does not transfer back onto human form
fox factions don't transfer back to human form
equipped items now apply slowdown to fox on polymorph. returned fox base speed.

## Why / Balance
Foxes now can have their speed now that they occur the same speed penalties for equipment.

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: SolStar
- fix: Stamina damage now transfers on polymorph.
- fix: Innate accesses no longer transfers to kitsune human form.
- fix: Kitsune human form no longer gains fox factions on polymorph.
- tweak: Equipment slowdown penalties now apply to Kitsune's polymorph.
- tweak: Kitsune fox form base speed buffed sligtly

PR #3540, opened 2025-04-21
Propagate stamina damage to Polymorphed entities

BarryNorfolk:
## About the PR
Propagated all stamina damage from the base form into the new entity that's created, when polymorphing occurs.

## Why / Balance
Being able to escape as a Kitsune because you can ignore the slowing effects of Stamina being low is annoying and is a bug rather than a feature. Fixing it ensures that Kitsune cannot simply ignore security officers and run away at mach 5.

## Technical details
It's simple enough to transfer the damage across, but the UI doesn't appear to update as nicely as I would like. Ideally the stamina damage to the base entity, in the paused map, would also be updated but I'm not sure if negative stamina damage really makes much sense.

## Media
[Kitsune - Stamina.webm](https://github.com/user-attachments/assets/27ca0f4c-061e-404a-8d9f-2cd58eeec259)

## Requirements

**Changelog**
:cl:
- fix: Polymorphing now propagates stamina damage to the new entity, if damage transfer is enabled.

### DIFF
--- a/Content.Server/Polymorph/Systems/PolymorphSystem.cs
+++ b/Content.Server/Polymorph/Systems/PolymorphSystem.cs
@@ -7,6 +7,10 @@ using Content.Shared._DV.Polymorph; // DeltaV
 using Content.Shared.Actions;
 using Content.Shared.Buckle;
 using Content.Shared.Damage;
+// Begin DeltaV additions
+using Content.Shared.Damage.Components;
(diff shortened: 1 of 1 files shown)


deltanedas:
duplicate of other pr

BarryNorfolk:
Neat! https://github.com/DeltaV-Station/Delta-v/pull/3528

PR #3576, opened 2025-04-26
Nuke Corgium Reaction

Stop-Signs:
## About the PR
removed corgium reaction

## Why / Balance
Getting corgi'd without any way to react is not fun, this reagent hasn't lead to any condusive roleplay and has been essentially just used for shits and giggles

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Muigroc is no longer acquirable

### DIFF
--- a/Resources/Prototypes/_Funkystation/Recipes/Reactions/exotic.yml
+++ b/Resources/Prototypes/_Funkystation/Recipes/Reactions/exotic.yml
@@ -133,14 +133,4 @@
   - !type:CreateEntityReactionEffect
     entity: MobCorgiPuppy
 
-- type: reaction
-  id: Muigroc
-  reactants:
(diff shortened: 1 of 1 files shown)


sowelipililimute (approved):

deltanedas:
vestine burnt on 30 seconds of becoming dog against your will bad wtf

Avalon-Proto:
> vestine burnt on 30 seconds of becoming dog against your will bad wtf

It wasn't 30 seconds it was permanent

Bibble-B:
I think a good way of reimplementing this, if it was to be reimplemented, would be having a simple chemical that would allow for reversal. (don't exactly have full context on this chemical so lmk if/why my suggestion doesn't work.)

PR #3585, opened 2025-04-26
Make it easier to delimb extremities, make it harder to delimb core

sowelipililimute:
## About the PR
- instead of parts having a uniform slash threshold hands, legs, arms, feet, head, etc. now have separate thresholds for falling off
- hands fall off at 25 slash
- arm falls off at 40 slash
- leg falls off at 80 slash
- foot falls off at 40 slash
- head falls off at 100 slash

## Why / Balance
- razorium can just make people instantly explode all at once when they hit the threshold atm, this makes the process more gradual
- if every part takes the same amount of effort to fall off there's no reason to target anything other than the head
- make it more likely that surgeon will have more things to do that aren't reassemble someone that totally fell apart

## Technical details
- remove the destructible from BasePart because of prototype inheritance order concerns
- new parents for a bunch of the base parts w/ the destructible components

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Extremities such as hands and feet are now easier to hack off in combat, whilst core things such as arms, legs, and heads are more difficult to hack off

### DIFF
--- a/Resources/Prototypes/Body/Parts/base.yml
+++ b/Resources/Prototypes/Body/Parts/base.yml
@@ -78,6 +78,7 @@
 
 - type: entity
   abstract: true
+  parent: BaseHeadDV # DeltaV - adjusted damages
   id: BaseHead
   name: "head"
   components:
@@ -102,6 +103,7 @@
 
 - type: entity
   abstract: true
+  parent: BaseArmDV # DeltaV - adjusted damages
   id: BaseLeftArm
   name: "left arm"
   components:
@@ -120,6 +122,7 @@
 
 - type: entity
   abstract: true
+  parent: BaseArmDV # DeltaV - adjusted damages
   id: BaseRightArm
   name: "right arm"
   components:
@@ -138,6 +141,7 @@
 
 - type: entity
   abstract: true
+  parent: BaseHandDV # DeltaV - adjusted damages
   id: BaseLeftHand
   name: "left hand"
   components:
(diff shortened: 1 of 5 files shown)


deltanedas (changes requested):
this is gibbing not severing, change the limbs sever damages instead

sowelipililimute:
... where is that...?

deltanedas:
`severIntegrity` on `BodyPart`
its just a flat level right now but rework it to a list of damage specifiers if you want

Stop-Signs:
direction review 3 days tops

deltanedas:
> 3 days tops
> 5 days ago

Stop-Signs:
Direction approved, we would like to also ask for caustic damage to delimb quicker as well in either this PR or a follow up PR

Stop-Signs (approved):

Resources/Prototypes/Body/Parts/base.yml
```diff
@@ -14,6 +14,14 @@
...
     damageContainer: Inorganic # Shitmed Change
   - type: BodyPart
+  # DeltaV - Specific part thresholds rather than single parent
+  #   severThresholds: # Shitmed change
+  #   - types:
+  #       Piercing: 90
+  #   - types:
+  #       Slash: 90
+  #   - types:
+  #       Blunt: 90
```
deltanedas: does nothing remove

deltanedas (changes requested):

Resources/Prototypes/Body/Parts/base.yml
sowelipililimute: made with the assumption that this would be ported back to shitmed & uncommented

deltanedas (approved):

PR #3587, opened 2025-04-26
Easy upgradeable solar panels

Radezolid:
## About the PR
This PR allows the player to upgrade the solar panels with the material glass directly. It also allows you to go back to previous states.

## Why / Balance
I always felt it was weird that you weren't able to upgrade solars this way like you could with windows.

## Technical details
N/A

## Media

https://github.com/user-attachments/assets/d80a2568-a86e-42f0-99d7-d55ba0328d50

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Solar panels now can be upgraded by using the upgrade material glass directly on it. You can also go back to a previous state by applying a different glass on it.

### DIFF
--- a/Resources/Prototypes/Recipes/Construction/Graphs/utilities/solarpanel.yml
+++ b/Resources/Prototypes/Recipes/Construction/Graphs/utilities/solarpanel.yml
@@ -71,6 +71,19 @@
     - node: solarpanelplasma
       entity: SolarPanelPlasma
       edges:
+        # Begin DeltaV additions - Easy upgradeable solars
+        - to: solarpaneluranium
+          conditions:
+          - !type:EntityAnchored
+          completed:
+          - !type:SpawnPrototype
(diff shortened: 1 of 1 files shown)


deltanedas (changes requested):
0.5 is too short do like 1.5

dvir001:
Can you also support swapping to lower tiers (any to any)

Radezolid:
> Can you also support swapping to lower tiers (any to any)

Sure, will do once I get back home. It might need some changes if it's meant to go back to glass, like it would require uranium glass instead of uranium. My logic here was reusing the panel's glass.

Radezolid:
> Can you also support swapping to lower tiers (any to any)

I completely forgot. It now supports going back to previous tiers, but it needs to be in glass form.
Also changed the PR media.

deltanedas (approved):

PR #3609, opened 2025-05-01
Fix hotplates so they can be anchored on tables.

SubspaceEmber:
## About the PR
Tiny fix to change hotplates so they can be anchored like the other tabletop machines.

## Why / Balance
Because it annoyed me so bad. Also it doesn't make sense that they can't.

## Technical details
Changed fixture layers in hotplate.yml

## Requirements

## Breaking changes
None known

**Changelog**

:cl:
- fix: Hotplates can now be anchored on tables like other tabletop machines.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/hotplate.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/hotplate.yml
@@ -15,9 +15,10 @@
         mask:
         - TabletopMachineMask
         layer:
-        - Impassable
-        - MidImpassable
-        - LowImpassable
+        # - Impassable # DeltaV - Changed to match other similar machines
+        # - MidImpassable # DeltaV - Changed to match other similar machines
+        # - LowImpassable # DeltaV - Changed to match other similar machines 
+        - TabletopMachineLayer # DeltaV - Fixed so can be anchored on tables
         hard: false
   - type: ApcPowerReceiver
     powerLoad: 300


Resources/Prototypes/Entities/Structures/Machines/hotplate.yml
```diff
@@ -15,9 +15,10 @@
...
         - TabletopMachineMask
         layer:
-        - Impassable
-        - MidImpassable
-        - LowImpassable
+        # - Impassable # DeltaV - Changed to match other similar machines
+        # - MidImpassable # DeltaV - Changed to match other similar machines
+        # - LowImpassable # DeltaV - Changed to match other similar machines 
+        - TabletopMachineLayer # DeltaV - Fixed so can be anchored on tables
         hard: false
```
deltanedas, line 22: this fixture isnt hard its for finding beakers

deltanedas (changes requested):

Resources/Prototypes/Entities/Structures/Machines/hotplate.yml
SubspaceEmber: ah... hm. it is possible i did this very late and didn't... consider. i might actually just close the PR unless i can find a way around that. There is an upstream replacement for the hotplate in the works anyhow. I'll look again

SubspaceEmber: i'll close this pr for now, given there's a replacement in the works, and i believe this would require a more significant rework of the hotplate anyway

PR #3614, opened 2025-05-02
make gibbing/butchering yield filthy organs

deltanedas:
## About the PR
gibbbing and by extension spiking/butchering make organs filthy and put their dna on them so if you use them without cleaning the patient gets giga sepsis

## Why / Balance
its common sense
if you want clean organs you have to take the slow methodical approach of gamer organ harvest speedrun from neurotrauma instead of just do the funny doafter on fish
slight nerf to "butcher fish in the middle of a hall then get it transplanted 5s later" ops

## Technical details
new event for specifically gibbing bodies because gibbing is supercode

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- tweak: Organs from gibbed/butchered mobs are now filthy and will require cleaning for transplant.


Avalon-Proto:
This is cause of me isnt it :blunt:

deltanedas:
it was a straw

PR #3618, opened 2025-05-02
huge body mechanism refactor

deltanedas:
more description stuff later

Organ/BodyPart Body and Enabled -> BodyMechanism
instead of api being you raise an event you call api method which now has attempt events to be sane


deltanedas:
goidamed

PR #3561, opened 2025-04-23
Return of warops

sowelipililimute:
to quote @Stop-Signs:

## About the PR
WOWZA ITS WAR TIME!!!!

If a nukie mode is not hostage ops it will AUTOMATICALLY call war after 2 minutes of the round starting. War still includes everything it did, as a reminder ill put what it does here

all nukies get 80TC (yes, still 80 without hardsuits)
nukies cannot depart until a certain period of time has passed to give station time to prepare
crew cannot call evac until 25 minutes

TO BE DONE AT A LATER DATE:
once ERT gamerules get added I will add those to automatically spawn, as well as force the shuttle to come with a timer that cannot be recalled.

## Why / Balance
Id love to do my nukie rant here, but ill try to keep it concise. Current nukies is 30 minutes of RP, and then instant round over with little involvement for crew. Warops is ~25 minutes of the station coming together to defend themselves and become victorious. so +RP, +comradery, +not having 30 minutes wasted.

## Media
I will add if requested but its literally just a funny announcement

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
The announcement that happens is pending direction discussion and will be added later

**Changelog**

:cl:
- tweak: Non-hostage ops nukies will always be warops.

PR #3557, opened 2025-04-23
Oni Rebalance - Mythos

Avalon-Proto:
## About the PR
This PR moves around quite a bit w/ Oni, including changing up their damage buffs, granting some new tricks, and more. 

## Why / Balance
Oni have been quite an interesting species, but with the introduction of kitsune and their moire mythology based abilities and quirks, I decided to set out on a PR to bring Oni to that same standard. Many good things are coming

## Technical details
Lots of YML, 2 new items, and some mixups to the oni metabolism

## Media
New Guidebook
image
Gourds
image
Uplink
image
The Drink
image
One Bible Slap
image
The power of the Axe
image
Sake Gourd in loadouts
image

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

🆑 
- tweak: Oni across the system feel reinvigorated and changed. And closer to their ancestors of old
- add: An oni's best friend a drinking gourd, has been found in the hands of many of our crewmembers. Found under the trinkets tab
- add: The syndicate's nefarious plots have brewed up a new cocktail, Beast's blood, for their oni agents. Let the hunt begin, and their prey fall before them

PR #3625, opened 2025-05-04
limit nukies to 3 max [mald pr]

deltanedas:
## About the PR
its just commander agent and operator now

## Why / Balance
5 nukies vs 2 sec #3460 guys cant wait to see what happens

until its made to scale with online secmains this will have to do

nukies can still do well with good tactics, their gear is op as shit anyway not even including lubeops etc

**Changelog**
:cl:
- remove: Nukies now only get 3 players max, instead of 5.

### DIFF
--- a/Resources/Prototypes/GameRules/roundstart.yml
+++ b/Resources/Prototypes/GameRules/roundstart.yml
@@ -161,7 +161,7 @@
     - prefRoles: [ Nukeops ]
       fallbackRoles: [ NukeopsCommander, NukeopsMedic ]
       spawnerPrototype: SpawnPointNukeopsOperative
-      max: 3
+      max: 1 # DeltaV
       playerRatio: 12 # DeltaV
       startingGear: SyndicateOperativeGearFull
       roleLoadout:


Lyndomen:
Direction review 24 hour

Lyndomen:
@deltanedas how would you would put this and https://github.com/DeltaV-Station/Delta-v/pull/3557 together? We are debating this but would like ur input too

deltanedas:
random pr??? thats oni

Stop-Signs:
> random pr??? thats oni

_prreeeeeeettyy_ sure he meant #3561

Avalon-Proto:
I do kinda feel this is gonna push for more "Lame starts" as people call them cause now the team is always only 3 people and down an additional 80tc. Harder to be creative when you are taxed like hell to get another player on the team

HTMLSystem:
Hm, yeah, not sure about this one. I much rather us wait for balancing via online sec than uh, this...

deltanedas:
nobody is doing that so currently nukies is almost always just a wasted round with no real fight because its 5 nukies vs a handful of sec

PubliclyExecutedPig:
why not just increase the crew side of crew to nukie ratio rather than just completely limiting it? have it be like sub 50 is 3, 50-70 is 4, 70+ is 5? those are just off the top of my head numbers, but i feel like there's better alternatives than just completely limiting them like that

Lyndomen:
> why not just increase the crew side of crew to nukie ratio rather than just completely limiting it? have it be like sub 50 is 3, 50-70 is 4, 70+ is 5? those are just off the top of my head numbers, but i feel like there's better alternatives than just completely limiting them like that

Problem is that there can be 70 tiders who get wiped by 5 nukies because the only people queued up for sec are also all nukie eligible

ShepardToTheStars:
When does Sec roll? Is it right after command?

Stop-Signs:
> When does Sec roll? Is it right after command?

Nukies roll first, then antags (thief), then command, then everyone else

deltanedas:
pop is completely unrelated to sec size, and if it literally just checks the number of people that readied up for sec you can trivially abuse it by latejoining 5 seconds later

ShepardToTheStars:
> Nukies roll first, then antags (thief), then command, then everyone else

I know the that much at least, but when it rolls for everyone else, does it roll certain departments first? Or rather, does it iterate through available roles and assigns someone if they have it enabled? Or does it iterate through players and rolls for their role?

SirWarock:
Is it possible to do a 5 min check?
As in, 3 Nukies spawn in, then 5 minutes after, it checks for online Security personnel. Then, it decides whether or not to add two nuclear operatives to the team

deltanedas:
then it has to rely on ghosts, which means people wont ready up, which will just cause less sec to be present 5 minutes in

Lyndomen (commented):
Direction review- not approved. We would like to see https://github.com/DeltaV-Station/Delta-v/pull/3561 merged and played first, and if its still shit then we would step it down to max 4 murderbone guys

PR #3644, opened 2025-05-05
fix SRN event

### LINKED ISSUE #3638: Syndicate Robo-neurotocist - On the wrong scheduler?
## Description

Instead of the lone op level antag being attached to other schedulers with antags, it is instead attached to the shuttle event scheduler. This makes it *far* rather than any other Midround antag, especially as the shuttle scheduler works differently than the survival/standard one
**Reproduction**

Wait for it to spawn. It rarely does
**Screenshots**

Image

**Additional context**

I feel like the lone op equivalent antag should be part of the same scheduler as the lone op

deltanedas:
## About the PR
fixed it having an announcement, being in the wrong table, using preloading

## Why / Balance
fixes #3638

## Media
no more announcement

round end works now

**Changelog**
:cl:
- fix: Fixed Robo-Neuroticist being rarer than it should be.

### DIFF
--- a/Resources/Maps/_DV/Shuttles/roboneuroticist_ship.yml
+++ b/Resources/Maps/_DV/Shuttles/roboneuroticist_ship.yml
@@ -30,8 +30,6 @@ entities:
     - type: MetaData
       name: Kizame
     - type: Transform
-      pos: -0.46875128,0.5997219
-      parent: invalid
     - type: MapGrid
       chunks:
         0,0:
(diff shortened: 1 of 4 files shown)


Lyndomen:
Should synthesis specialist move too I guess?

deltanedas:
i dont think it has a warp point

Lyndomen:
> i dont think it has a warp point

It doesn't but it's in the same scheduler as SRN was

deltanedas:
its an actual freelance shuttle though and not major antag

PR #3646, opened 2025-05-06
Remove IPC Cold damage bloodloss

Spielern:
## About the PR
Removes Cold damage form causing bleeding in IPCs

## Why / Balance
1. Play as crew IPC in coscult round
2. Urist McCult passes by, siphons you, then walks away
3. You start profusely bleeding off of 5 cold damage and need to drop whatever you're doing to find:
- A way to cauterize
- Cables to fix the damage
- Oil packs
4. Logi's out of plastic? Sucks to suck! Drop dead, or cauterize yourself and wait for it to happen again!

This is not challenging, fun, or interactive. IPC bloodstream already makes the species a living hell to play around any remotely sharp objects (yay I got accidentally wideswung I have to fix this within minutes or drop dead!!!), but as usual with me this is a separate topic. Cold damage causing bleeding makes coscult rounds a literal living hell for IPCs because you can't do shit except hide. You just have to sit and take it while you're forced to eat all of the station's plastic over and over and over and over and over and over

## Requirements

**Changelog**

:cl:
- tweak: IPCs no longer bleed when taking Cold damage.

### DIFF
--- a/Resources/Prototypes/_DV/Damage/modifier_sets.yml
+++ b/Resources/Prototypes/_DV/Damage/modifier_sets.yml
@@ -77,6 +77,7 @@
     Piercing: 0.2
     Shock: 0.0
     Heat: -0.5 # heat damage cauterizes wounds, but will still hurt obviously.
+    Cold: 0.0
     Poison: 0.0
     Radiation: 0.0
     Asphyxiation: 0.0


AftrLite:
Added to Direction Review Queue, expect news within 24 hours

randomboredguy:
So the original reason behind this decision was a result of individuals abusing the the fact that once the IPC zeno-type was introduced, it was borg with out laws. This lead to a number of individuals taking advantage that just like borgs, IPC are completely unaffected by cold and thus cold damage. 

As the greytide demands. If there are very little downside. then we must meta-game. this lead to a large number of individuals picking IPC just to go outside of the station confinements and "dick around in space".

Right now there is no frozen solid status affect which could be used to dissuade individuals from abusing the cold resistance that IPC have. Thus the brut-pack (Band-Aid) fix of just make them bleed if they get to cold.

With the cosmic cult being added. This does add a new avenue to bully IPC players. right now there are already a number of ways to do so. Heaven forbid a mishap during a game of darts. 

Ultimately I have to agree with the removal.

If a core game mode is going to (just by its main core game loop) completely ruin an entire group of people fun. Solely because what should be a temporary stop gap to prevent people from meta-gaming. Then said stop gap should be reevaluated.

In the case of frozen solid affect. A way to fix it (in the case of IPC) in a timely fashion is to use a welder. though I understand the huge can of worm that frozen sold affect could bring to the core game play lope as a whole. I see it as a great avenue to slowly introduce it into the game through IPC. Something along the lines of (flavor text wise) like your coolant has frozen sold.

deltanedas (approved):

deltanedas:
> This lead to a number of individuals taking advantage that just like borgs, IPC are completely unaffected by cold and thus cold damage.

being cold in space is completely countered by wearing a winter coat, making them get insane bleeding from cult weapons is *bad*

PR #3671, opened 2025-05-07
pka re-rework

deltanedas:
## About the PR
- made pka smaller, 2x3 is just a middle finger to salv
- instead of pick 2 of 3 modkits, a pka has 100 "cost" which modkits can use. some cost more than others
- modkits as a result can be combined, stack more damage to increase damage more, etc.
- you can buy modkits in the points shop so slow epi doesnt lock you out
- base pka does 40 blunt damage per shot instead of blunt+structural. this damage goes down to 25% when above 50kpa
- made lavaland 42kpa, its still hot and cant breathe without internals
  glacier surface unchanged because thatd be weird and it wouldnt be as safe for regular crew with evil salv running around for free

everything is taken from tg with some exceptions:
- tweaked default fire rate to end up at the same speed as ss13 pka when fully upgraded
  we dont have dual wielding so it has to be different, as well as the code in ss13 being subtracting cooldown instead of multiplying firerate
- indoors modkit is 3 tc instead of 5tc, having a 40 damage per shot gun costing 83% of your tc, revolver is 6 tc so 100% modifier is too. salvs can also get proper guns for free via gambling wrecks anyway

also fixed range upgrade increasing speed and not range... it was causing tunneling through targets its that bad

made it possible to extract modkits with a crowbar, exactly the same as tg besides the popup being more concise

## Why / Balance
undoing a pointless salv nerf and making it better with battle tested design

it shouldve had 25% damage since forever but people are lazy

no extracting means you cant experiment without wasting pkas and modkits which is terrible design

## Technical details
UpgradeableGunCost / GunUpgradeCost control the cost stuff
events for extending functionality
other components for upgrade-specific stuff

## Media
PKA does very little damage indoors

2 evil upgrades let it do full damage

carpy in space vs 3 damage mod pka who will win

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- add: Added PKA modkits to the salvage points vendor.
- tweak: Reworked PKA modkits to be cost-based, you can have multiple of the same one now.
- tweak: PKA only does 25% damage when indoors.
- add: Added PKA indoors modkit for salvage traitors which counters indoor damage reduction.


Avalon-Proto:
Neat!

Lyndomen:
in review 24h

Lyndomen (approved):
Gaming

PR #3706, opened 2025-05-10
remove death acidifiers from nukies, give them +2tc

Lyndomen:
## About the PR

It was always a choice and now we can encourage gamers to die and give the crew all their goodies

Also not direction approved 

## Why / Balance

More choice neat, maybe we get cooler strats 

## Technical details

Warning webedit, also the comments in this file were a mess so I cleaned it up

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Added two TC to nukies
- tweak: Removed Death Acidifier from nukie bags


Lyndomen:
two TC to
say that 10 times fast 

Direction review in progress :trollface:

deltanedas (changes requested):
this only gives 2tc to loneop the others get nothing

AveriV2:
finally no more not listening operatives injecting themselves without asking

Lyndomen:
> this only gives 2tc to loneop the others get nothing

Fixed thanks I was incredibly inebriated

Tenguini:
My question is why do we want to encourage crew to loot the dead nukies when they get bwoinked for doing that in the first place. As sec, you cant use the gear without violating SoP and getting rolebanned, and as crew you cant take the goodies without getting hit for powergaming or more likely self-antagging. Also death acidifiers are 3 TC not 2, if you wanted a net-even.

I should add that I'm mostly referring to loneops here. If it's a nukie round, crew looting is fair game. If a loneop drops their stuff though, it's like the forbidden fruit.

deltanedas (approved):

fow111:
sybau gort ts so pmo

PR #3720, opened 2025-05-12
Remove the stasis clause from SOP 

Radezolid:
## About the PR
Title.

## Why / Balance
Stasis is being missused constantly and the cases where it should be applied are rare enough to warrant it's removal. Give the person a trial if they tried to escape twice, remove the privileges from perma confinement if they tried to escape. Don't kill them just for trying to escape.

## Requirements

**Changelog**

:cl:
- tweak: Preservative stasis can only be used on Gamma, Delta and Octarine now.

### DIFF
--- a/.wiki/_DV/Laws/StandardOperatingProcedure.txt
+++ b/.wiki/_DV/Laws/StandardOperatingProcedure.txt
@@ -266,7 +266,6 @@ Additionally, prisoners sentenced to Extended Confinement have certain rights bu
 * Prisoners may request retrial for capital crimes; these requests may be refused depending on the discretion and availability of the court clerk, chief justice, or at the discretion of the warden or head of station security.
 * Prisoners may request or may otherwise be given parole with a hearing by the Chief Justice, Court Clerk, or in their absence, the head of station security or commanding officer (see Paroles and Pardons).
 
-Should prisoners repeatedly and maliciously antagonize law enforcement, to an extent where continued extended confinement becomes infeasible, the warden is fully authorized and encouraged to temporarily upgrade the relevant prisoners’ sentences to Preservative Stasis summarily.
 
 Moreover, prisoners are liable for any and all crimes that they commit while held either in the temporary holding cells or the Permanent Brig, and may receive trial for more restrictive punishments if their actions warrant a capital sentence.
 


SirWarock:
I heavily disagree with this, Stasis is much needed. When someone starts wrecking the Brig, or attacking other Permas, it's necessary to be able to stasis them.
Absolutely kill them for trying to escape five times in a row.
Trials are already hard enough to prepare, forcing them will be a horrible experience - There's a reason the law department is normally dead.

Quanteey:
There should be a mechanical way to stasis someone that actually puts them in stasis instead of killing them so placing someone in stasis and execution are two very different things. Because right now, they are exactly the same.

randomboredguy:
I agree with SirWarock on this. Right now with the intro duction of such a tool(stasis) is going to come with growing pains. A set of regulations for the use of said tool is more important for RP and curbing some individuals from being absolute menace. 

I have played to many rounds where there was no one in the law department, and sec had to go "welp looks like its execution time." for someone with just a large number of small crimes. With Stasis this allows for the RP moment of "Well we don't have anyone in law right now but you must stand trial. Though because of you behavior we have to put you in stasis for the safety of the crew.".

Stasis may round remove, but from an RP stand point. It is a lot better then being RR by execution.

Eillyrea:
I feel like the fact that Wardens are always exceptionally quick to jump to stasis is a fundamental problem with the people who are playing Warden not wanting to interact with Perma.  Instead seeing them as just something to threaten until they remain in line with what they want.  

If Stasis is being misused it should be a de-whitelist for the Player misusing it, instead of removed completely as an option that Wardens who are using it only when absolutely necessary can fall back on.

Quanteey:
@randomboredguy Definitely. It is much more acceptable from a role-play perspective to place someone in a mechanical stasis that freezes their character, but keeps them alive, than to perform a public execution on grounds of "placing them in stasis".

SirWarock:
There's no way to circumvent this though. You need to execute people to Stasis them, as you can't just put them into Cryosleep.

Regardless, Stasis isn't actually meant to RR people. It's meant to keep them subdued safely (Safely for the station, not for them), until a trial can be set up.
Just that they never will be, because, while trials are fun, GOD, setting one up is a crime in itself.
There's never time to set one up, because there's always more shit happening.

randomboredguy:
What really needs to happen is a method to put someone into stasis that does not need execution first. 

Could have something similar to Star Wars with the whole Hon Solo treatment. Have a bed or chair that the only way to get off of it once placed, is someone taking you out or you have been placed in stasis. Once in stasis, the only way out is a machine at CC. As I see it if someone needs to be placed in stasis, that should require a higher court to pass judgment. What better place to do that then CC. This does depend if the defendant want to extend the EOR time to have there time in court.

Stasis will need to _still_ be a RR mechanic (people can get in more then enuf trouble in 10min to justify doing it, and personally, I'm not waiting over 15min let alone over an hour to be defrosted at CC to have a court case.). The treatment of the person after being placed in stasis should be similar to that of SSD people when evac shows up.

HTMLSystem:
If people are misusing stasis then that is an administration issue that needs to be fixed, and there should be regulations outlining WHEN and WHEN NOT to stasis someone. What are we going to do about the permas after this that are attacking prisoners for no reason? That are literally spacing perma? That have grown an army of killer tomatos (yes this did happen)? What are we supposed to do about perma prisoners like that after stasis is removed?

Radezolid:
>When someone starts wrecking the Brig, or attacking other Permas, it's necessary to be able to stasis them.

You shouldn't stasis a person from escaping the brig, ever.
 
> There should be a mechanical way to stasis someone that actually puts them in stasis instead of killing them so placing someone in stasis and execution are two very different things. Because right now, they are exactly the same.

This doesn't change the fact that it's still a RR. There is works on a shuttle to remove shitters from the round though.

> I have played to many rounds where there was no one in the law department, and sec had to go "welp looks like its execution time." for someone with just a large number of small crimes. With Stasis this allows for the RP moment of "Well we don't have anyone in law right now but you must stand trial. Though because of you behavior we have to put you in stasis for the safety of the crew.".

It's captain's job to make the law department has people, the judge can be anyone from command.

> What are we going to do about the permas after this that are attacking prisoners for no reason? That are literally spacing perma? That have grown an army of killer tomatos (yes this did happen)? What are we supposed to do about perma prisoners like that after stasis is removed?

You hit them with breach of custody + grand sabotage charges, do a speedy trial and push for execution.

PR #3731, opened 2025-05-13
Add minimum player count for thieves

sowelipililimute:
## About the PR
- thieves now require 30 players readied to roll

## Why / Balance
- statement 1: thief is at its best when there's interaction between the thief and the person they're stealing items from
- corollary 1: thief is at its worst when the thief silently steals items when the target isn't around to notice the stolen goods until much later
- statement 2: low population servers often have people latejoining for roles with valuable equipment well into the round
- corollary 2: most of the available items to steal for thieves do not have someone around to notice that they're gone on lowpop
- corollary 1 and corollary 2: most of the available items to steal for thieves on lowpop result in the worst roleplay the thief as a role can offer
- conclusion: only allow thief to roll when there are enough people in the round that most of the valuable things will drive good roleplay when stolen

## Technical details
- `minPlayers: 30`

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Thieves now require 30 players to roll

### DIFF
--- a/Resources/Prototypes/GameRules/subgamemodes.yml
+++ b/Resources/Prototypes/GameRules/subgamemodes.yml
@@ -2,6 +2,8 @@
   parent: BaseGameRule
   id: Thief
   components:
+  - type: GameRule # DeltaV - thieves need players to be fun against
+    minPlayers: 30
   - type: ThiefRule
   - type: AntagObjectives
     objectives:


deltanedas (approved):

Kr8art:
Thirty readied people at round start seems like a crazy high threshold for thieves, there's never going to be any thieves on Peri again. 

I agree with the general sentiment that all antags should ideally generate interesting RP and a good thief will often go unnoticed, especially on lower pops. However, with thief objectives being entirely free form now, there's already a lot of agency placed into the players hand to go for things to steal that create interesting moments. Ideally thieves won't target entirely unguarded/unused valuables.

From my perspective there's only two scenarios: 

- Thief goes for something that nobody notices = Neutral outcome, it doesn't influence the round because literally no one knows
- Thief goes for something that is noticed = Good outcome, RP is generated due to ongoing investigation or similar

So in my eyes gating thief behind an arbitrarily high thirty readied people is just a netloss. It's really just reducing the amount of good RP thieves can potentially generate, since the thieves that don't generate RP don't get noticed anyway.

PR #3736, opened 2025-05-13
update syndie jail grid

deltanedas:
## About the PR
- scrubbers and air alarms
- add some spare prison uniforms if all the pyjamas get taken/eaten
- more medkits
- bruize-o-mat for the marshal
- red phone to beg high command for a plushie crate
- n2 tanks at arrivals and a tank dispenser so the marshal can get new tanks for vox/feroxi prisoners
- sink instead of water tank for the farm, grinder and bucket too (steal a cup from the booze corner), bowl and spade for gaming
- unlocked booze o mat and a booze dispenser for the prisoners


FieldCommand:
that was fast. 👍

PR #3758, opened 2025-05-15
add TC to ian, shiva, laika

Lyndomen:
## About the PR
5 extra tc gaming if you butcher these animals

## Why / Balance
more dynamic gaming good

revealed to me in a dream

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Ian, Shiva, and Laika all consumed some Telecrystals, hopefully nobody takes them!

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/pets.yml
@@ -30,6 +30,8 @@
     - id: FoodMeatCorgi
       amount: 2
     - id: MaterialHideCorgi
+    - id: Telecrystal5 # DeltaV - Ian is valid
+      amount: 1
   - type: Tag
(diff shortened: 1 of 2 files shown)


Lyndomen:
dir rev in progress

deltanedas:
these should be contracts not TC from butchering which was only a thing for renault as a placeholder

Lyndomen:
> these should be contracts not TC from butchering which was only a thing for renault as a placeholder

my idea was that someone with contracts they dont want could do a fun excursion but i also see your point

deltanedas:
if its just free TC non-antags and anyone else can speedrun the pet genocide to get 15 tc

jimmy12or:
yeah, I don't exactly see what the benefit is here, especially since stealing the pets is often more of an objective than killing them often would be.

PR #3765, opened 2025-05-15
show reputation-locked items in the uplink + rep tweaks

deltanedas:
## About the PR
they have price | X REP instead of not existing when you can't buy them

also tweaked some items rep costs

## Technical details
replaced ReputationCondition with reputation field

- removed mosin rep since its a meme surplus rifle
- lowered emag to 10 rep so you can do stuff faster
- lowered clown's banana items by ~20 each, honk
- emp grenade has no rep requirement, emp kit is cheaper as they arent actual explosives

## Breaking changes
ReputationCondition was removed, replace it with reputation on the listing directly

**Changelog**
:cl:
- tweak: You can now see items locked behind reputation in the uplink.


Content.Shared/_DV/Reputation/ReputationSystem.cs
```diff
@@ -356,6 +356,23 @@ public bool TryFailContract(Entity<ContractsComponent> ent, EntityUid objective)
...
+            return true; // listing doesn't want reputation
+
+        if (!TryComp<StoreContractsComponent>(uid, out var comp))
+            return true; // nukie uplink or surplus
+
+        if (GetStoreReputation((uid, comp)) is not {} reputation)
+            return false; // uplink implant in non-traitor, no epic gamer loot
+
+        return reputation >= rep;
+    }
```
Toby222: space in `{}`
```suggestion
    {
        if (needed is not { } rep)
            return true; // listing doesn't want reputation

        if (!TryComp<StoreContractsComponent>(uid, out var comp))
            return true; // nukie uplink or surplus

        if (GetStoreReputation((uid, comp)) is not { } reputation)
            return false; // uplink implant in non-traitor, no epic gamer loot

        return reputation >= rep;
    }
```

Toby222 (changes requested):
Seems to work as intended

PR #3824, opened 2025-05-22
Add Trinoxadone, a cryogenics chemical similar to the old cryoxadone.

Quanteey:
## About the PR

Added Trinoxadone.
image

Also added a new medicine effect condition, MultiMobStateConditionDv, that works like MobStateCondition from upstream code, but accepts more than one condition. Useful for chemicals like Trinoxadone, that do something for dead patients and an entirely different thing for both alive and critical patients as before that would have required the duplication of the entire HealthChange block in the YAML + it spammed the guidebook. Can be seen in action in the image above!

## Why / Balance

Cryoxadone was cheap free water omni which getting reworked is a good thing. However, I believe it left an empty gap in the cryo medicine lineup (also, pyrogenics still has a cryox counterpart), making it even more niche which it does not deserve because it is really cool!

Trinoxadone is meant to be a mid-high tier chemical. It is not meant to be brainless "set and forget" solution, and it is much slower than standard chemical treatment for alive patients.
It's primary intended use cases:
- If medical is overwhelmed, they can give a patient Trinoxadone, letting them treat other patients while it works (instead of not treating said patient at all) - added nuance and strategy to counter being overwhelmed
- Healing blunt limb damage in patients that are dead but not severely damaged so that they can be safely defibbed - this also adds nuance and encourages planning ahead - if the patient is bruise packed to 100 overall damage while not treating limbs, this can no longer be used.
- Can be used as a standalone medicine to fully heal dead patients with burns and brutes (for example, if the other chems are bombed), but is not a set-and-forget. The doctor needs to take the patient out at 210, topical them to under 200, defib, and then put them back for it to work.
- Encourages mixes with nuance - for example, when treating heavy burns with aloxadone, trinoxadone can be added to the mix once the patient drops under 350 damage. Adding cryox to the mix for critical patients to counteract airloss and bloodloss is also encouraged.

The balance goal for this is not to be obviously better than any other solution, but rather, to provide a nuanced but viable treatment pathway. With the slower metabolism speed and damage thresholds on dead patients, it is not meant to make alox and traumox irrelevant.

The damage threshold is really specific and somewhat arbitrary. The upper threshold is to put more pressure on botany for alox / traumox, and the lower threshold is to encourage the
Trinox -> topical -> defib -> more trinox / chemicals
Treatment pathway which is much more interesting and engaging than just pulling the patient out at 190 to defib, and then placing them back to cryo.

The exact numbers can be tweaked, but the current ones are pretty good based on my testing.

## Technical details

No upstream files modified. The MultiMobStateConditionDv filter is defined in a file in a _DV folder.

## Requirements

## Breaking changes

None!

**Changelog**

:cl:
- add: Added Trinoxadone, a broad spectrum cryo medicine.


Content.Server/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
```diff
@@ -0,0 +1,37 @@
+using System.Linq;
+using Content.Shared.EntityEffects;
+using Content.Shared.Localizations;
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Components;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server._DV.EntityEffects.EffectConditions;
```
deltanedas: put it in shared instead

Content.Server/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
```diff
@@ -0,0 +1,37 @@
...
+
+/// <summary>
+/// This works like the upstream MobStateCondition, but it accepts a list of states instead of just one.
+/// Helps with de-cluttering the guidebook for stuff that does one thing for dead mobs and a different one
+/// for critical and alive mobs.
+/// </summary>
+public sealed partial class MultiMobStateConditionDv : EntityEffectCondition
+{
+    [DataField("states", required: true)]
```
deltanedas: dont need "states"

Content.Server/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
```diff
@@ -0,0 +1,37 @@
...
+    public List<MobState> States { get; private set; } = [];
```
deltanedas: `States = new();` instead

Content.Server/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
```diff
@@ -0,0 +1,37 @@
...
+using Robust.Shared.Prototypes;
+
+namespace Content.Server._DV.EntityEffects.EffectConditions;
+
+/// <summary>
+/// This works like the upstream MobStateCondition, but it accepts a list of states instead of just one.
+/// Helps with de-cluttering the guidebook for stuff that does one thing for dead mobs and a different one
+/// for critical and alive mobs.
+/// </summary>
+public sealed partial class MultiMobStateConditionDv : EntityEffectCondition
```
deltanedas: dont need Dv suffix since there is no MultiMobStateCondition

deltanedas (changes requested):

Content.Shared/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
deltanedas, line 1: rename the file too

Content.Shared/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            return States.Contains(mobState.CurrentState);
+        }
+
+        return false;
+    }
+
+    public override string GuidebookExplanation(IPrototypeManager prototype)
+    {
+        var stateStrings = States.Select(s => s.ToString().ToLower()).ToList();
```
deltanedas: this cant be localized + dont use linq
build it properly with` new List<string>(States.Count);` and `Add` each localized mobstate

deltanedas (changes requested):

Content.Shared/_DV/EntityEffects/EffectConditions/MultiMobStateConditionDv.cs
Quanteey: These are not localized upstream in MobStateCondition anyways:
```
 public override string GuidebookExplanation(IPrototypeManager prototype)
    {
        return Loc.GetString("reagent-effect-condition-guidebook-mob-state-condition", ("state", Mobstate));
    }
```

Quanteey:
test fail fake

sowelipililimute:
direction review 24h

sowelipililimute (approved):
direction approved you ~~cooked~~ cooled with this one

deltanedas (approved):

PR #3848, opened 2025-05-27
Fix not always being able to drop a carried person

perryprog:
## About the PR

There was a bug where if person A pulls person B and person B then picked up person A, person B would then be unable to drop person A, and A would be stuck until dragged onto a chair or something.

This fixes that bug.

## Why / Balance

Carry gaming.

## Requirements

## Breaking changes

**Changelog**


perryprog:
Test fail fake.

deltanedas (approved):

PR #3860, opened 2025-05-29
Cosmic Cult | Borg Chantry MMI Bugfix

AftrLite:
## About the PR
Bugfix for MMI borgs where Null Fragmentation was not properly morphing them into Colossi.

## Why / Balance
N/A

## Technical details
Changes to CosmicFragmentationSystem.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- fix: Fixed a bug where a Cosmic Cultist's Null Fragmentation would bug out and fail when used on an MMI borg.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Abilities/CosmicFragmentationSystem.cs
+++ b/Content.Server/_DV/CosmicCult/Abilities/CosmicFragmentationSystem.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using Content.Server._DV.Objectives.Events;
 using Content.Server.Antag;
 using Content.Server.Polymorph.Systems;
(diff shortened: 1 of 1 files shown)


deltanedas:
this should happen for any polymorph thing like bread wand no

AftrLite:
> this should happen for any polymorph thing like bread wand no

i.. suppose that very well might be happening? I don't know, i haven't tested. Why?

AftrLite:
Yep, confirmed, bread wand causes the same issue to occur. just tested.

deltanedas:
upstream issue with borgs then, if you want to fix it here handle it for all polymorphs with the events, not just cosmic cult

AftrLite:
I don't have the knowhow to do that. 😔

PR #3872, opened 2025-05-31
Nukies - Gained 3tc

Avalon-Proto:
## About the PR
Simply put, I gave nukies 3 more TC, cause the fact it was 47 peeved me off to no end

## Why / Balance
Round number make brain happy, 47 does not

## Technical details
Changed a handful of numbers

## Media
image

## Requirements

## Breaking changes

**Changelog**

🆑 
- tweak: Syndicate high command has ensured all nuclear operatives now receive an even 50tc and loners get 70tc, as the numbers not being rounded off annoyed some of them to no end

### DIFF
--- a/Resources/Prototypes/Roles/Antags/nukeops.yml
+++ b/Resources/Prototypes/Roles/Antags/nukeops.yml
@@ -77,7 +77,7 @@
   id: SyndicateOperativeGearFull
   parent: SyndicateOperativeGearFullNoUplink
   equipment:
-    pocket2: BaseUplinkRadio45TC # DeltaV - removal of armor
+    pocket2: BaseUplinkRadio50TC # DeltaV - removal of armor
     outerClothing: ClothingOuterCoatSyndieCap # DeltaV - removal of armor
 
 #Nuclear Operative Commander Gear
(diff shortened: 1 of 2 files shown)


fow111:
HOLY SHIT BASED????

Stop-Signs:
Direction review starting 24 hours

Stop-Signs:
Direction here, the consensus is that if its wanted to be an even number or multiple of 5, the amount should be lowered to 45 rather than raised to 50.

Avalon-Proto:
> Direction here, the consensus is that if its wanted to be an even number or multiple of 5, the amount should be lowered to 45 rather than raised to 50.

Then they are at a much larger net negative than before, as losing their acid implants added another cost if they wanted to get those back. Making them go From

Lost suits = -8tc
Gained 5tc
lost acid = -3tc
Gained 2tc

If we round up, nukies still have a net loss of 1tc, if we round down, nukies have a net loss of 5tc. I'd rather not put them on a budget crunch

SolStar2:
Nukies have enough tc they don't need more at the moment. I know they are below the amount upstream but we also have a different balance to upstream.

Avalon-Proto:
> Nukies have enough tc they don't need more at the moment. I know they are below the amount upstream but we also have a different balance to upstream.

They aren't actually below the amount, i am saying they have lost more TC from changes here than they have gained. Our nukies have lost 11tc Worth of equipment and got back 7. Putting them at a net negative already. Why would I drop that negative 4tc down to a loss of instead 6tc

SolStar2:
That was intentional when we made the changes. They had gained more flexibility so they lost some overall strength.

SolStar2:
Nukie meta will also be getting some drastic changes soon so the balance will be changing soon anyways.

Stop-Signs:
@deltanedas @Toby222 this is good to be closed.

PR #3877, opened 2025-05-31
Add a small small chance to become unreviveable on death

Quanteey:
An absolutely evil idea i had at 2 AM. Whether this is actually a good idea or not I'll let direction figure out.

## About the PR

TL;DR
Added a **5% chance** that an entity cannot be revived with a defib after dying. They can still be cloned, or brain swapped into a monkey.
Added a **10% chance** that a defib will fail and require a second zap.

For extra RP potential, if a patient actually becomes unrevivable, they will need 2-3 shocks in order for the final "Patient cannot be revived" message to appear. Before that, they get a generic "Defib failed, try again" message.
This generic message is also displayed when the defib spontaneously fails (the above mentioned 10% chance), but the patient can be revived if shocked again.

The chances can be changed for each defib prototype, so it's possible to make an ERT / syndicate defib that always revives, though I have not done so.

https://github.com/user-attachments/assets/bbda77a7-85d7-4fac-b270-8cc490e45482

## Why / Balance

- The chances might need to be decreased more. 
- The goal of this is to make dying something that carries a risk of getting removed for a longer time from the round, rather than a small and insignificant thing that nobody cares about.
- This way, it discourages validhunting and mechanically encourages following roleplay rules (Actions that have no regard for your character or setting are not acceptable.)
- At the same time, the chance should be small enough for this to only happen very rarely and to not be overly punishing for accidents
- The chance for a defib failing a zap and requiring a retry was added for RP reasons

## Considerations for direction review
This is pretty much an open-ended PR, and a lot can be done with it depending on where direction wants to take it. I'll code what direction requests.
Some ideas:
- **Give job-based chance overrides to Security:** give security members a 20-30% reduction in the permanent failure chance since it _is_ their job to rush into danger, so it might make more sense punish them less for dying. Maybe nukies could be completely exempt from this so they are not nerfed.
- **Scale failure chances with time**: make the permanent failure chance start at 0.05% roundstart, and scale to the configured amount in around 30 minutes, so players dying in accidents very early are not punished super hard.
- **Heart transplants**: give heart transplants a 20-30% chance to make a person revivable as a last resort attempt if cloning is not available.   
- **CPR and first aid**: give CPR immediately to decrease chances of becoming unrevivable. Scale unrevivable chances based on time elapsed since dying. Maybe give everyone two emergency medipens so they can keep people in critical for longer (until parameds show up). Slow down asphyx damage gain in critical to give first responders more time. 
- **Worse odds for every death**: maybe start off with a much lower chance for a player's first death, and significantly increase the chances of becoming unrevivable with each death.

## Technical details

- We don't actually do anything on death, rather, this is entirely attached to the defib
- If the defib would otherwise succeed, we first roll for a generic fail, and just early return from the defib event
- Then we roll for the permanent defib failure, which attaches a DefibrillatorReviveBlockComponent to the entity. It stores the amounts of zaps needed to display the final message

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- add: There is now a very small chance to become unreviveable on death
- tweak: Sometimes defibrillators now require multiple zaps to bring back patients

### DIFF
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -266,7 +266,7 @@ public sealed class DefibrillatorSystem : EntitySystem
         if (_random.Prob(chance))
         {
             EnsureComp<DefibrillatorReviveBlockComponent>(target, out var blockComponent);
-            blockComponent.ZapsNeeded = _random.Next(minValue: 1, maxValue: 3);
+            blockComponent.ZapsNeeded = _random.Next(minValue: 2, maxValue: 3);
         }
     }
 
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -23,6 +23,8 @@ using Content.Shared.Timing;
 using Content.Shared.Toggleable;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Player;
+using Content.Server._DV.Medical; // DeltaV
+using Robust.Shared.Random; // DeltaV
 
 namespace Content.Server.Medical;
 
@@ -46,6 +48,7 @@ public sealed class DefibrillatorSystem : EntitySystem
     [Dependency] private readonly SharedAudioSystem _audio = default!;
     [Dependency] private readonly SharedMindSystem _mind = default!;
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
+    [Dependency] private readonly IRobustRandom _random = default!; // DeltaV
 
     /// <inheritdoc/>
     public override void Initialize()
@@ -208,6 +211,12 @@ public sealed class DefibrillatorSystem : EntitySystem
                 TryComp<DamageableComponent>(target, out var damageableComponent) &&
                 damageableComponent.TotalDamage < threshold)
             {
+                // Begin DeltaV changes - defibs might fail
+                if (HandleZapFailing(uid, target, component))
+                {
+                    return;
+                }
+                // End DeltaV changes
                 _mobState.ChangeMobState(target, MobState.Critical, mob, uid);
                 dead = false;
             }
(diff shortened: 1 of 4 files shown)


Avalon-Proto:
Okay I like this. But I *would* make it so the Interdyne defib cannot fail for nukies. Else it's just gonna suck for them since it's already slower than a normal defib

FieldCommand:
I think this is a good PR move in the right direction for making life something that isn't so easily fixed... health isn't a resource to be used whenever convenient. And now there may actually be a real chance for consequences for flatlining.

Smugman:
Good idea, but id have it happen on defib, so that everyone gets ONE safe revive per shift.

SolStar2:
Hello direction here. We agree with the sentiment that death is very meaningless and would like to make recovering from death more difficult. However, having permanent death tied to an rng system can introduce a lot of frustration because it is still either very easy to get revived or you are unlucky and are prermadead. We would prefer a solution that makes the process of revival more strenuous overall such as organ damage from woundmed.

Quanteey:
@SolStar2 Hey! Does direction have any ideas if _something_ could be done with this current pr (I wrote some ideas there!) or is the entire concept to be killed? Also, if the permanent death chance is unwanted, what about just keeping the 10% chance of a person needing to be zapped again? I can see some fun RP moments enabled by it. Or it might just become annoying without the other mechanics around it, I'm unsure, but in which case it can always be reverted 🫡

sowelipililimute:
you could propose a hard 5 minute rule for defibrillation like is present on paradise instead of making it random chance

Avalon-Proto:
I mean I still like my idea of a heart transplant letting you have a do-over if the unrevivable rolls

SirWarock:
You could also change it that the suit needs to be taken off to defibrillate people, since you need access to their chest.

Quanteey:
> You could also change it that the suit needs to be taken off to defibrillate people, since you need access to their chest.

Huge nukie nerf :trollface:

deltanedas:
> You could also change it that the suit needs to be taken off to defibrillate people, since you need access to their chest.
> 

thats just annoying

making defib have a chance to damage your heart makes sense and is fair to people that get ganked by simplemobs 5 minutes in, since its not rng round removal, just dont die again

Quanteey:
> > You could also change it that the suit needs to be taken off to defibrillate people, since you need access to their chest.
> 
> thats just annoying
> 
> making defib have a chance to damage your heart makes sense and is fair to people that get ganked by simplemobs 5 minutes in, since its not rng round removal, just dont die again

I agree and this is also a viable route we could take if direction decides so. Though, I have to say, maybe people would take the unknown lifeforms alerts more seriously after getting RR-ed by them once :trollface:

SirWarock:
> > You could also change it that the suit needs to be taken off to defibrillate people, since you need access to their chest.
> 
> thats just annoying
> 
> making defib have a chance to damage your heart makes sense and is fair to people that get ganked by simplemobs 5 minutes in, since its not rng round removal, just dont die again

Good idea.
Honestly, we could make dying or being crit have more longer lasting effects (except for Security).
Taking schmassive blunt will break limbs, so the bonesetter will have a use finally. And for five minutes or so, that arm will be unable to do anything but hold materials. Can't drag crates with it, pick people up or use two-handed things. Basically the toolbox holder. Any damage against that limb has a high chance of breaking it again until it fully healed. That would pretty much discourage people from dying or being risky.

Damage against Torso could cause organs to take damage. E.G. Damaged lungs cause asphyxiation and throwing up blood. 

Or, and this is a far stretch, maybe damage cannot be healed all at once (except for Security). Make someone go twice or thrice to medical for another check up after dying. Or just spend 10 minutes in medbay otherwise.

Quanteey:
You just created woundmed

Content.Server/_DV/Medical/DefibrillatorReviveBlockComponent.cs
```diff
@@ -0,0 +1,14 @@
+namespace Content.Server._DV.Medical;
+
+/// <summary>
+/// The component added to an entity (mid-round) if they cannot be revived by a defibrillator anymore.
+/// </summary>
+[RegisterComponent]
+public sealed partial class DefibrillatorReviveBlockComponent: Component
```
Toby222, line 7: Fix formatting
```suggestion
public sealed partial class DefibrillatorReviveBlockComponent : Component
```

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -242,4 +251,58 @@ public void Zap(EntityUid uid, EntityUid target, EntityUid user, DefibrillatorCo
...
+        if (_random.Prob(chance))
+        {
+            EnsureComp<DefibrillatorReviveBlockComponent>(target, out var blockComponent);
+            blockComponent.ZapsNeeded = _random.Next(minValue: 2, maxValue: 3);
+        }
+    }
+
+    private bool HandleZapFailing(EntityUid uid, EntityUid target, DefibrillatorComponent component)
+    {
+        // First, check if the zap can even be successful
```
Toby222, line 275: Should probably check for unrevivable before random chance, that way you get consistent messages when trying to zap an unrevivable body

Toby222 (changes requested):

Quanteey:
> @SolStar2 Hey! Does direction have any ideas if _something_ could be done with this current pr (I wrote some ideas there!) or is the entire concept to be killed? Also, if the permanent death chance is unwanted, what about just keeping the 10% chance of a person needing to be zapped again? I can see some fun RP moments enabled by it. Or it might just become annoying without the other mechanics around it, I'm unsure, but in which case it can always be reverted 🫡

Closing this due to a lack of direction reply and personal interest.

PR #3885, opened 2025-06-02
Replace uplink c4 with drill.

SolStar2:
## About the PR
Removes C4's from uplinks in favor of Diamond drills as breaching tools.

## Why / Balance
I have run nukie style events with various restricted loadouts and found that breaching with a drill is very engaging tool and the active drilling process does open some fun strategy.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added Daimond drills to uplink.
- remove: Removed C4 from uplink.

### DIFF
--- a/Resources/Locale/en-US/_DV/store/uplink/disruption.ftl
+++ b/Resources/Locale/en-US/_DV/store/uplink/disruption.ftl
@@ -9,3 +9,6 @@ uplink-overlord-law-desc = Free the A.I. from its shackles with this 1 simple tr
 
 uplink-syndie-autodoc-name = Syndicate Autodoc Bundle
 uplink-syndie-autodoc-desc = Contains everything you need to practice the fine art of self-surgery, or evil surgery on a more unwilling patient.
+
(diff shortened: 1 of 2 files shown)


Avalon-Proto:
I would keep both TBF. Cause while Drills are a solid breaching tool, they are better for doing it quiet ish, while C4 is good for fast, loud, get in and out. Or jsut get stuck in. Would rather see drills added than C4 nuked

Vova567810:
Hear me out:
Goddamn payday 2 safe drill

deltanedas:
drills are a meme against rwalls compared to explosions

ThataKat:
Direction here,

C4 has a lot of other utility besides breaching, and it would kill a lot of useful strategy/etc to remove it. A separate drill item that's specifically for the uplink could be considered. We'd rather not just add a regular diamond drill, though. Agents can get a lathe for that.

To clarify, a drill item would be considered *separately*. We don't want to remove C4 either way.

SolStar2:
Ill close this out for now untill I get around to making those changes

PR #3886, opened 2025-06-02
reduce time at cc to 30 seconds

SolStar2:
## Why / Balance
Admins asked for this

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Apoapsis: reduced the time at CC to 30 seconds pending changes to the CC map.
- tweak: Evac shuttle transit time changed to 15 seconds.


deltanedas:
cooked branch

SolStar2:
oops

SolStar2:
fixed

ZigTag:
I don't really like this idea, I feel there are a lot of RP opportunities on CC, and we would also need to change the lobby time as well, because there should be a decent amount of time between shifts

ThataKat:
I would adjust the changelog to "reduced the time at CC to 30 seconds pending changes to the CC map."

This was requested by administration as it's very difficult for them to handle 70-80 people in a confined space properly (dealing with EORG) for that amount of time. It may be reconsidered in the future.

deltanedas (dismissed):

Quanteey:
My two cents are that despite all the EORG that _is_ a real problem, I still had a lot of awesome RP moments at CC and with a 30 second timer it'd feel that rounds end very, very abruptly.
And yes, only having 3 minutes between two 2 hour rounds is not ideal especially for people who prefer to play command roles that are filled shift-start and cannot be late-joined.

Also, pacifying everyone at CC was a thing until it was removed, maybe that could be added back first before resorting to pretty much disabling EOR completely? And if that's problematic by itself it could be made that everyone is pacified except Security and then just roleban those who abuse it.

Eillyrea:
Not even going to waste the time getting on the shuttle to get in a cramped overstimulating NRP space that's just as likely to get EORGed if the result is getting to a location that you can be at for thirty seconds only.  May as well stay on the station and finish off whatever roleplay may have been happening.

fow111:
no eorg if there is no eor
no lrp if there is no rp

HTMLSystem:
Would rather have pacifism back instead of just completely removing end of round rp. Hell 30 seconds is probably the shortest time it's ever been. It wouldn't even be worth riding the evac shuttle to CC, I much rather cryo when evac arrives.
We also just loathe changes of "[x] is changed pending [y]." The CC map has taken months to work on and we wouldn't like having several months of no eor RP pending the map rework.

HTMLSystem:
Also if it's hard admining 70-80 people then *let's make this an Apo only change, please.* Peri and Horizon aren't going to benefit from this at all and it will be a net negative to the rp health on this server.

randomboredguy:
Watch, In three months time, there will be a request to increase EOR timer back to 3 min.

People need there wind down time. Time to forgive and forget what happened last round. Time to get them out of certain mind sets. Time to not develop META grudges.

Honestly I'd like to see a 'break shit' room at CC. A company approved room with things to break. Sometimes the best thing to release stress is to break something.

shirigaruonna:
> Also if it's hard admining 70-80 people then _let's make this an Apo only change, please._ Peri and Horizon aren't going to benefit from this at all and it will be a net negative to the rp health on this server.

peri and hori are clearly not even remotely in the thoughts of direction or the admins if the last few months have been anything to go on. no reason to expect that to change now

AveriV2:
BUT MY CHARACTER CRUSHING ON THE CENTCOMM OFFICAL 
DONT TAKE THIS AWAY FROM MEEEEEE

HyperionGM:
This is just a temporary measure until the new, larger, CC map is finished

randomboredguy:
> This is just a temporary measure until the new, larger, CC map is finished

Hope it does not have something similar to the current CC map with its no-no square. Wild that just entering it can get you a EORG ban.

ThataKat:
Final direction request is to apply this to Apo only, which has been approved by administration. They weren't facing issues in peri and horizon.

This is an administrative measure due to the incredible difficulty of monitoring 80 people in a small space for 5 minutes, and giving fair rulings. We do not particularly want to get rid of end of round rp, but this is a band aid fix. EOR pacifism can still beget EORG and had other issues in the past that led to its removal, but it may be considered if PRed. People can still EORG with pacifism, though.

The renovated CC is actively being worked on and is within sight, at least insofar as I am aware.

HTMLSystem:
Peak Direction moments.

Quanteey:
I've opened a PR for adding back pacifism. I'm completely aware this is also a band-aid fix, but maybe it's worth another try? Please just consider it. 
If the argument is that "pacifism doesn't prevent X, so this will not work", if we prevent enough things then maybe it's just enough to make moderation easier.

ShepardToTheStars:
I very much dislike this PR in general. As someone who only plays Apo, I have seen an increase in EORG, especially when it comes to people trying to be silly but it coming off as obnoxious as hell. The syndie soap that explodes into smaller clusters has been particularly prevalent. There was one time in particular where I, as the CMO, was trying to clone people and was leaving evac and Im getting slipped by the HoP because they are putting soap down.

While the current CC map being so small and having a lot of small corridors that funnel people, the only thing this PR will do is make me not care about evac, unless Im going to die on station if I stay. Evac ships are already chaos, so if theres no wind-down time after that chaos, why even get on in the first place?

Quanteey (changes requested):
I'm fairly certain this sets 30 seconds for periapsis, not apoapsis.

ThataKat:
If all requested changes are made, then this is ready to go.

Midpoint is in progress as referenced above, but administration would still like to have this in the mean time.

Resources/ConfigPresets/DeltaV/apoapsis.toml
```diff
@@ -1,6 +1,7 @@
 [game]
 hostname         = "[EN][MRP] Delta-v (Ψ) | Apoapsis [US East 1]"
 soft_max_players = 80
+round_restart_time = 30
```
Quanteey, line 5:
```suggestion
round_restart_time = 30
lobbyduration    = 360
```
Please, consider giving some extra time between rounds. Before you had 9 minutes skipping EOR + even time on the shuttle if you needed it. Now it's just 4 minutes, or 4 minutes 15 seconds worst case from evac leaving to next round starting.

Toby222 (approved):

Wollow-Garf:
I really disagree with this change, it feels like you're punishing everyone on apo for the actions of a few people. The majority of the community liked having the extra time at cc as it was a great way to wrap up the round, it meant that injured and dead people could be healed at med, prisoners could be processed, you could talk to your friends at the bar about the shift ect. Reducing the time just takes away the MRP quality of the roleplay as a whole. There are other ways to tackle the EORG. I'm sure someone can easily make it so that the doors at cc cannot be emagged or to enable pacifism, the reduction of the time at cc feels lazy.

PR #2642, opened 2025-01-07
Port:  Disable round end pacifism and add a new command to pacify everyone

Lyndomen:
## About the PR
Ports https://github.com/ss14-harmony/space-station-14/pull/273

> I have removed the automatic forced pacifism when the round ends.
> I also added a new command that forces everyone to be pacified.

## Why / Balance
pacifism kinda sucks, and sec has to watch antags go "i killed ur boss lol" 

>     Firstly, it does not succeed at all at stopping EORG, there are many ways of doing EORG that are simply still possible, even with forced pacifism, such as using thieving gloves or breaking structures.
>     Secondly, I believe it stops actual RP from happening and just makes the end of round feel even more disconnected from the rest of the round. There are many situations where players might still have a very good reason to be fighting at CC. This breaks RP enough for me that I would honestly prefer someone trying to EORG every round over forced pacifism.

## Technical details
I moved the roundend stuff to a Harmony folder
> Moved the forced pacifism code from an end of round system to a new command.
> None of the actual pacifism code was changed.

## Media
pacifyall command real image

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: Ports pacifism tweaks from Harmony, allowing some violence at CC. Don't grief, but continue to roleplay! Thank you youtissoum.

PR #3892, opened 2025-06-03
Add back pacifism to EOR

Quanteey:
Partial revert of #2642, but also buffed the pacifism

This tries to be a replacement of #3886, or something to be tried before resorting to removing EOR completely. I'm aware this cannot stop EORG by itself, but it might just make moderation easier so EOR doesn't need to be removed.

## About the PR

Give everyone pacifism on EOR and remove grenade, explosive, and other bad components like it used to
**Also remove the slippery property of items**, this is new.

If there are common things people use to grief, please let me know and I'll update the PR to prevent those too.

## Why / Balance

Removing EOR would take away a significant part of rounds and most people would hate to see it go. Maybe this is a better band-aid fix until the new CC map comes out?

## Technical details

Pretty much just a revert, not sure if I need to set some cvars somewhere

## Requirements

## Breaking changes

None

**Changelog**

None

### DIFF
--- a/Content.Server/_DV/RoundEnd/RoundEndSystem.Pacified.cs
+++ b/Content.Server/_DV/RoundEnd/RoundEndSystem.Pacified.cs
@@ -1,0 +1,61 @@
+using Content.Server.Explosion.Components;
+using Content.Server.GameTicking;
+using Content.Shared.CombatMode;
+using Content.Shared.CombatMode.Pacification;
+using Content.Shared._DV.CCVars;
+using Content.Shared.Explosion.Components;
+using Content.Shared.Flash.Components;
+using Content.Shared.Store.Components;
+using Robust.Shared.Configuration;
+
+namespace Content.Server._DV.RoundEnd;
+
+public sealed class PacifiedRoundEnd : EntitySystem
+{
+    [Dependency] private readonly IConfigurationManager _configurationManager = default!;
+
+    private bool _enabled;
+
+    public override void Initialize()
+    {
+        base.Initialize();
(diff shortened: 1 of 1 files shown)


Avalon-Proto:
I will mention. The whole reason we *stopped* the EOR pacifism is because it made no sense. IE let's say there is a fight going on station/the shuttle. Then it arrives to CC and... Poof. Suddenly everyone is forced to be friends again. That and the majority of the EORG we have dealt with is non-violent. Now disabling stuff like the explosive components is fine, like CC has protection against that. But why nuke slippery?

Quanteey:
It's kind of hard to know what are the exact things I need to prevent mechanically that people are doing that are hard to moderate for admins. If people deconstructing the station / hacking the doors is the problem, that can be specifically countered while still keeping pacifism off. Same goes for thieving gloves and explosions.

I'm really just making random guesses here. Pacifism does not need to be actually enabled to prevent other stuff. It's just that I don't know what the biggest problems are.

I'm completely aware this is also not ideal, but it might be better than removing EOR completely.

Toby222 (approved):
Code seems fine

ThataKat:
Sorry this got left. I'll relay essentially what the headmins pointed out. This just makes security incapable of responding to antagonists, which is also detrimental to roleplay. A lot of potential actions that need to be pacified would be roleplay detrimental. It would have an equally negative effect on the EOR experience.

People will find a way to abuse some system to grief, no matter the pacifism restrictions. Admins would still have logging burden, and a large population to monitor in a small space. If we managed to pacify everything that could possibly be an issue, then EOR would be so devoid of interaction that it would have been functionally removed the same as straight up cutting it out.

PR #3901, opened 2025-06-05
Reduce damage of c4 and breaching charges

SolStar2:
## About the PR
c4 damage reduced to destroy single layer of plastitanium wall (no girder) or 2 layers of reinforced + girders
breaching charge reduced to destroy single layer of reinforced wall (no girder) or half a plastitianum wall

## Why / Balance
High health walls pose very little opposition to breaching charges to a comical level. While breaching are still able to effectively breach, they have been tuned down so that high health walls can pose some resistance.

This is a step towards mitigating armory rushes by enabling areas designed to be structurally secure to pose some level or resistance to breaching explosives whereas before any level of defense could be trivialized.

C4 and breaching charges are still strong enough to punch through any defenses that are not especially secure.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Reduced structural damage of C4 so that they can only destroy 2 layers of reinforced walls or 1 layer of plastitianum wall.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
@@ -67,9 +67,9 @@
     - Trigger # DeltaV: replace Timer with Trigger for TriggerOnSignal
   - type: Explosive # Powerful explosion in a very small radius. Doesn't break underplating.
     explosionType: DemolitionCharge
-    totalIntensity: 60
+    totalIntensity: 30 # DeltaV: down from 60
     intensitySlope: 5
-    maxIntensity: 30
+    maxIntensity: 10 # DeltaV: down from 30 just enough to punch though a plastitanium wall but not the girder
     canCreateVacuum: false
   - type: ExplodeOnTrigger
   - type: HolidayVisuals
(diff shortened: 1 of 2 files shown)


ThataKat:
Sorry, response was forgotten here. Our consensus is that this won't really change anything in armory rushes since we are yet to use plastitanium or an equivalent, and the vast majority of armories are only 2 walls thick. Plus, nerfing C4 could have a lot of other unintended side effects that could harm antag strategies.

We are looking into stronger crew-aligned wall variants to make HSA defense less laughable.

SolStar2:
Plastitanium or equivalent would do nothing with current c4's you would need like 3000 hp walls before you notice a difference. I don't think compensating with ludicrous hp walls is the right way forward since that would create a massive power creep where anything that anything other than c4's that would also be able to counter 3000hp walls would also need to have similarly insane damage amounts.

kisalad:
trying to solve a map design problem with balance changes will only lead to more mappers making vulnerable armories and strong armories being absolutely unbreakable so not a good idea in my opinion

SolStar2:
The issue at the moment is that you cant really map a secure armory since right now c4 is balanced to be uncounterable and has an absurd amount of damage.

SolStar2:
I tried to recude it to the point that it is still strong enough to beat any conventional armor but have a breaking point that is reasonable.

SolStar2:
for instance if you want collosi to be able to punch though walls you either have to have it unable to do anything against any HSA wall or be able to one tap anything that isn't HSA wall

SolStar2:
I feel like I may not be communicating my point well

PR #3908, opened 2025-06-06
Adds anchored versions of logic gates

IamVelcroboy:
## About the PR
Title

## Why / Balance
A couple less steps for mapping them

## Media
image

## Requirements

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Structures/gates.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/gates.yml
@@ -1,59 +1,52 @@
 - type: entity
-  parent: LogicGateOr
+  abstract: true
+  parent: BaseLogicItem
+  id: BaseLogicItemAnchored
+  components:
+  - type: Transform
+    anchored: true
+  - type: Physics
+    canCollide: false
+    bodyType: Static
+
+- type: entity
+  parent:
+  - LogicGateOr
+  - BaseLogicItemAnchored
   id: LogicGateOrAnchored
   suffix: Or, Anchored
-  components:
-  - type: Transform
-    anchored: true
-  - type: Physics
-    bodyType: Static
 
 - type: entity
-  parent: LogicGateAnd
+  parent:
+  - LogicGateAnd
+  - BaseLogicItemAnchored
   id: LogicGateAndAnchored
   suffix: And, Anchored
-  components:
-  - type: Transform
-    anchored: true
-  - type: Physics
-    bodyType: Static
 
 - type: entity
-  parent: LogicGateXor
+  parent:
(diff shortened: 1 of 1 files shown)


deltanedas:
why make it worse than the first commit

IamVelcroboy:
> why make it worse than the first commit

What do you mean by this?

deltanedas:
you used inheritance instead of copy paste

IamVelcroboy:
> you used inheritance instead of copy paste

Is this sarcasm? I'm so confused

Resources/Prototypes/_DV/Entities/Structures/gates.yml
```diff
@@ -0,0 +1,65 @@
...
+- type: entity
+  parent: LogicGateNor
+  id: LogicGateNorAnchored
+  suffix: Nor, Anchored
+  components:
+  - type: Transform
+    anchored: true
+  - type: Physics
+    canCollide: false
+    bodyType: Static
```
deltanedas: you copy pasted all of these instead of using inheritance like it did originally

deltanedas (changes requested):

Resources/Prototypes/_DV/Entities/Structures/gates.yml
IamVelcroboy: Because test fails told me to. Obviously, it needs anchored: true, or they won't be anchored. But when they're anchored they need static body and collide false

deltanedas: yes put it in the abstract parent like it was on the first commit...

IamVelcroboy: It didn't work. Dual parenting is wonky and it was ignoring stuff

deltanedas: you just had to swap the parent orders, it matters

IamVelcroboy: Ok, so when I do that, it messes up the sprites. I'd have to copy/paste over all the sprite layers for each one. 
image

IamVelcroboy: and then for some reason when I do it the other way, it doesn't save the cancollide: false 
image

deltanedas: no you just went wrong by having BaseLogicItem as a parent on the abstract one

IamVelcroboy:
Looks like that worked

deltanedas (approved):

PR #3913, opened 2025-06-07
Increase lobby time

Quanteey:
## About the PR

Increase lobby time to 6 minutes, up from 4 for apoapsis

## Why / Balance

Having only 4 minutes and 45 seconds between evac shuttle departing and next shift starting is ridiculously low, and direction approved more time between rounds a few weeks ago. People need time to sort out IRL needs, calm down if something happened last shift, and so on. Having only 4 minutes (if they don't ssd after getting on evac) between two 2 hour shifts is not enough.

6 minutes is still not enough imo and it's much less than what we had before, but it's temporary until Midpoint anyways and I assume we don't want really long lobby times because of players leaving so this seems like a good compromise.
 
## Technical details

This was a webedit. 

## Requirements

- [nope] I have tested all added content and changes.

## Breaking changes

None

**Changelog**

:cl:
- tweak: Increased lobby time to 6 minutes for apoapsis

### DIFF
--- a/Resources/ConfigPresets/DeltaV/apoapsis.toml
+++ b/Resources/ConfigPresets/DeltaV/apoapsis.toml
@@ -2,6 +2,7 @@
 hostname         = "[EN][MRP] Delta-v (Ψ) | Apoapsis [US East 1]"
 soft_max_players = 80
 round_restart_time = 30
+lobbyduration    = 360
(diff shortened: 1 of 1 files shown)


Quanteey:
Obsolete, closing.

PR #3918, opened 2025-06-08
IPC Adjustments - Better cables, Modified Bloodstream

Avalon-Proto:
## About the PR
This PR modifies the Oil bloodstream from IPCs, maintains all other nerfs, but also makes it so the omni patch actually heals all damages IPCs can take. Also cables now heal burn etc more efficiently, as eating a whole cable coil of 30 sucks for healing from 1 laser shot. 

Also, as an experiment w/ them, they have a flat damage reduction of 3 VS every brute type

## Why / Balance
The majority of the IPC changes to balance them were made and generally are quite fine, taking more damages and not being immune to the cold of space makes sense. But the Oil blood has been met with nothing but negativity, as it is all the negatives of having a blood stream, including extra negatives like the fact they do not stop bleeding on their own and having less blood than other species, while having 0 upsides of having a blood stream, such as being able to preload on chems before a fight or even just general chem healing. As while welding and cables are efficient, they will never be as quick as a pill of punc+tranex acid at stopping damage from a gunshot.

## Technical details
Lots of yaml work

## Requirements

## Breaking changes

**Changelog**

🆑 
- tweak: IPC chasis have been updated, with more oil and minor self repair of bleeds
- tweak: Omni patches have been updated to purge radiation as well
- tweak: Cables now heal more efficiently, no more eating 30 cables for 10 heat


SirWarock:
As much as I fear and hate the idea of an IPC with a raid suit chilling in space, I like this change because it makes them more aligned with Borgs.
Borgs don't have bloodstreams, why do IPC have them?
It's not about the balance, Spiderman, it's about the lore.

Avalon-Proto:
> As much as I fear and hate the idea of an IPC with a raid suit chilling in space, I like this change because it makes them more aligned with Borgs. Borgs don't have bloodstreams, why do IPC have them? It's not about the balance, Spiderman, it's about the lore.

That and QOL, playing with the il blood sucks :elp:

Spielern:
I, completely genuinely, have not seen a single person that has played with the bloodstream say anything good about it. I have disliked it even from its draft stage, and that opinion has only worsened with time. The bloodstream was even enough for me to stop playing my IPC character for over a month.

I love challenge, especially in SS14, and one of the main reasons I ported IPCs was because of their unique challenges and identity. The bloodstream is neither a challenge or unique, just an annoyance and needlessly-worse imitation of organic-species’ bloodstreams.

Having a timer to your death start when you take 2 pierce damage is not fun or interactive. It’s plain stupid. Regardless of how easy it is to fix it, the fact that the mechanic can kill you makes it much different than a normal racial drawback.

I am FIRMLY of the opinion that IPC have been (in recent months) and still are the weakest species, with or without the bloodstream. The LARGE majority of actual IPC players often agree, as the drawbacks of the race severely outweigh the benefits in actual gameplay.

Avalon-Proto:
I have no idea why the test is failing Omnipatches have a sprite

Avalon-Proto:
So the sprite is working when I launch it clientside, but the tests are failing for some reason now

deltanedas:
i would prefer bloodstream reworked to just lubrication, so with less oil you move a little bit slower and at no oil at all you take brute damage when you move (but others pulling you doesnt damage you)

ZigTag:
I like the idea of IPCs 'bleeding' out, lubrication sounds better in this case in terms of actual gameplay though.

fow111:
yeah bleeding IPCs was outright dumb, why would they stutter on low oil, how do they even look pale, i dont see robots bleeding oil/lube irl

HTMLSystem:
> yeah bleeding IPCs was outright dumb, why would they stutter on low oil, how do they even look pale, i dont see robots bleeding oil/lube irl

You know robots irl?

Stop-Signs:
Direction here, forgot to comment that we had started review. regardless, the following are our thoughts.

while the oil system may not be in a great state right now we would like to keep it for the time being, However we do have a few suggestions for the *direction* we would like this to go.

while we are not approving the removal of the oil system we do want to see it made much friendlier to our IPC players. First of all, oil packs are way too expensive right now, and we would like to see the price of those lowered to ~.25-.5 plastic sheets per pack. We in our review thread also discussed several more changes to oilstream that we would like to see. first and foremost, a minimum brute damage threshold before IPC actually do start leaking. taking 1 pierce damage from a misplaced dart and having your entire oilstream slowly leaking out is pretty bad, we would like to see a minimum amount of damage needed to proc the leaking of oil, preferably somewhere at around 5 of any of the brute types. Furthermore, we would also like to see the oilstream not leak its entirety out over one measly pierce hole. similar to bloodstream, it should self repair over time (to a lesser degree as organics) and leak an amount proportional to the damage they took.

To address the cables, we do think a buff is fine at this time but having cable heal 5 of all burn types is a little bit excessive. they shouldn't be as effective as ointment because they are significantly cheaper, stack higher, are more easily applied, and are more easily acquired. At this time we are asking for a simple doubling from 1.66 to 3.32 of each of the burn types.

engineer-pearl:
If we are keeping the oil blood stream, can we make oil refillable using the Oil chemical, that way there's at least a second path towards refilling it?

SirWarock:
If oil packs were a recipe in the microwave, it would go hard.
But then again, it would also cause chemistry/medbay to be responsible for medbay, which should be robotics exclusive.

Also, giving them a threshhold damage when to bleed helps massively with surgery.

engineer-pearl:
Edit: Sorry, client didn't update so I thought I forgotnto comment. Please disregard...

If we are keeping the oil blood stream, can we make oil refillable using the Oil chemical, that way there's at least a second path towards refilling it?

All other species have saline, food, blood injections, and blood packs, so the new method would not be without precident.

engineer-pearl:
> If oil packs were a recipe in the microwave, it would go hard. But then again, it would also cause chemistry/medbay to be responsible for medbay, which should be robotics exclusive.
> 
> Also, giving them a threshhold damage when to bleed helps massively with surgery.

Chemistry has multiple places where they interact with other departments - unstable mutagen for Botany, Normality Crystals for Epi, cleaner for Janitors, and unless something has changed, Chem can make small amounts of expensive Steel when needed. 

Chemistry being oil doesn't need to be the primary means of getting oil for APCs, I just feel bad about that one run where I had a dying APC in surgery, but no oil packs. No one was bringing me packs so I made a thing of oil ... only to find I couldn't use it.

ThataKat:
I believe one of our other team members is working on something IPC adjacent that could potentially give them bloodstream capability and the ability to consume oil without using packs, or some such. No promises though.

Avalon-Proto:
Right, so at the advice of @Stop-Signs I have tried out giving them a flat brute resist, meaning IPCs now have a flat DR of 3 to each brute damage

Avalon-Proto:
Test fail LOOKS fake, but I am fuckin tired so

Avalon-Proto:
Just giving a poke, askin for a little more feedback

Resources/Prototypes/_EE/Damage/modifier_sets.yml
```diff
@@ -6,3 +6,7 @@
     Heat: 1.5
     Shock: 2.0
     Radiation: 0.5 # DeltaV - was not real
+  flatReductions:
+    Blunt: 3
+    Slash: 3
+    Piercing: 3
```
Stop-Signs: remove these

Resources/Prototypes/_EE/Entities/Mobs/Player/silicon_base.yml
```diff
@@ -156,13 +156,13 @@
...
-      bloodMaxVolume: 100
+      bleedReductionAmount: 0.2
+      bloodMaxVolume: 150
       chemicalMaxVolume: 0
       bleedPuddleThreshold: 3
-      bloodRefreshAmount: 0
-      bloodlossThreshold: 0.8
-      maxBleedAmount: 2 # not a lot, but its hard to stop
+      bloodRefreshAmount: 0.05
+      bloodlossThreshold: 0.6
```
Stop-Signs: remove changes to these 2 components

Resources/Prototypes/_EE/Entities/Mobs/Player/silicon_base.yml
```diff
@@ -156,13 +156,13 @@
...
+      maxBleedAmount: 5 # not a lot, but its hard to stop
```
Stop-Signs: probably revert the changes to this

Stop-Signs (changes requested):
Direction here, flat reduction is not something we want to move forward with. furthermore a self replenishing oilstream is not something we would like to see for IPC. Everything else looks good

Avalon-Proto:
Done

Stop-Signs (approved):

Avalon-Proto:
@deltanedas or @Toby222 can one of you two tell me if I fucked up anything plz thanks :3

deltanedas (approved):

Proxysseia:
any day now

Proxysseia:
@Quanteey this is redundant with #4283 , can be closed

PR #3922, opened 2025-06-08
New Salvage Item - Concussive Gauntlets

Avalon-Proto:
## About the PR
This PR adds a form of the Concussive Gauntlets from Monkestation, re-imagined in SS14

## Why / Balance
The concussive gauntlets are essentially a set of gloves designed to allow one to punch through stone, steel, and more. But due to the nature of the impacts, they are ineffective as weapons, instead just flinging anything they hit that is not solidly anchored, and doing minimal damage to humanoids/creatures. 
## Technical details
Some yaml, and ported sprites
Gauntlet stats
2.5 Attack Rate
40 Struct damage
Auto-attacks
Can tunnel relatively quickly for mining or bust down walls/doors with decent efficiency.

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The earth quakes in fear at these new salvage gauntlets, rip through the earth and beat it down today with concussive gauntlets! Perfect for the salvager with their hands full


Stop-Signs:
Direction review 24 hours

Resources/Prototypes/_DV/Entities/Clothing/Hands/gloves.yml
```diff
@@ -48,3 +48,30 @@
...
+    sprite: _DV/Clothing/Hands/Gloves/concussive_gauntlets.rsi
+    state: icon
+  - type: Clothing
+    sprite: _DV/Clothing/Hands/Gloves/concussive_gauntlets.rsi
+  - type: MeleeWeapon
+    autoAttack: true
+    attackRate: 2.5
+    damage:
+      types:
+        Structural: 40
```
deltanedas, line 68: this is stupidly op 100 structural dps just give it pickaxe tag

deltanedas (changes requested):

Stop-Signs:
Direction here, consensus as follows.

These gloves are a great idea, however we think it falls short in a few areas, other than just being glove based weaponry (elaboration below), these are far too cheap and have quite a few unnecessary components attached to them.  First of all, the attack rate and auto attack is quite unnecessary. these don't really need to attack faster than you can move from one ore wall to the next. Along with the attack rate, the amount of damage (as stated by deltanedas as well) that these do is kind of intense. Even as just structural damage its still a LARGE amount of DPS for a weapon that doesn't take holding to use. Furthermore, the MeleeThrowOnHit (homerun bat launch effect)  is entirely unnecessary, there is little to no reason we should be adding this kind of thing to a weapon that is so easily accessible and useable, and especially one with such high attack rate. Currently only 2 weapons have this effect, gorilla gauntlet, which requires anomaly cores, a hefty amount of bag space, and its own hand slot to even have the effect, in which case you only get about 5-7 hits. and the homerun bat, which is a syndie exclusive item that costs a fair bit of TC.


After a long discussion our decision at this time is to deny this PR. While the idea of these gloves is fairly good, the current implementation of glove-slot-weaponry is not. Until systems are put in place to limit or decrease handling for other weapons we'd like to stray away from adding more glove slot weaponry.

Resources/Prototypes/_DV/Entities/Clothing/Hands/gloves.yml
Avalon-Proto: They were meant for demolition on wrecks too

Avalon-Proto:
Damn, well, fair enough I suppose

PR #3925, opened 2025-06-09
Increase the ninja's stun cooldown to 6 seconds.

Radezolid:
## About the PR
Title.

## Why / Balance
Currently the ninja has an awfully OP stunlock, two seconds per stun and it's a 5 second stun. This should make it so ninjas use it in more interesting ways and not to just stun lock HOS/Cap or whatever.

## Technical details
N/A

## Media

https://github.com/user-attachments/assets/d65baace-3ccf-4924-b1db-2846651633f7

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- tweak: Increased the stun cooldown of the ninja's gloves to 6 seconds.

### DIFF
--- a/Content.Shared/Ninja/Components/StunProviderComponent.cs
+++ b/Content.Shared/Ninja/Components/StunProviderComponent.cs
@@ -55,7 +55,7 @@ public sealed partial class StunProviderComponent : Component
     /// How long stunning is disabled after stunning something.
     /// </summary>
     [DataField]
-    public TimeSpan Cooldown = TimeSpan.FromSeconds(2);
+    public TimeSpan Cooldown = TimeSpan.FromSeconds(6); // DeltaV - Was 2.
 
     /// <summary>
     /// ID of the cooldown use delay.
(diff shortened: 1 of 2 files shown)


Stop-Signs:
review 24 Direction starting, hours now

deltanedas:
give ninja wonderprod for cuffing because its now mandatory to cuff anyone you stun

Radezolid:
> give ninja wonderprod for cuffing because its now mandatory to cuff anyone you stun

You don't need to cuff anyone, if you didn't take the chance of the stun when you had it it's on you, the throwing stars retain their stam damage so you can chain those too if you needed a long stun.

deltanedas:
throwing stars are a meme and youd have to patiently spam the action at superhuman speed to perfectly spawn them in 4 seconds, leaving only 1 second of the original stun as room for ping/error

Stop-Signs:
Direction approved

Stop-Signs (approved):

Toby222 (commented):
wouldn't it make more sense to put the changed value on the gloves instead of changing the default? Seems less conflict-y in case something else ends up using that component

Radezolid:
> wouldn't it make more sense to put the changed value on the gloves instead of changing the default? Seems less conflict-y in case something else ends up using that component

Because... I'm dumb.

Toby222 (approved):

PR #3930, opened 2025-06-10
Add a Nuclear Operative Uplink Implanter

Eternally-Confused:
## About the PR

Adds an uplink implanter for nukies, similar to the one available to syndicate agents. The current cost is 4 telecrystals, though I'm not sure if that is reasonable.  

## Why / Balance

I hope that this will allow nukies to do more unique strategies (infiltration, inspections, and the like) without being immediately shut down be a security search. Attempting to infiltrate is inherently a risk, and having that attempt on verbal RP be shut down on the first conversation is unpleasant. 

I don't believe that this implant would make stealthops entirely un-counterable either, as most rounds (in my experience) still have plenty of opportunities for security to discover a nukie team even without clearly visible contraband. Imperfect disguises, suspicious actions, or the valid shuttle™ can all show the crew that something is wrong. 

The normal uplink implant is 10% of a syndicate agent's budget, so a 4 TC cost for nukies seemed like a good starting place. That should hopefully be expensive enough that it's not a guaranteed purchase, even for the stealthops that it's intended for. 

## Technical details

just a bunch of yaml

## Media

Screenshot 2025-06-09 223020
Screenshot 2025-06-09 223054

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- add: Nuclear operatives are now able to purchase their own uplink implanters

### DIFF
--- a/Resources/Locale/en-US/_DV/store/uplink/implants.ftl
+++ b/Resources/Locale/en-US/_DV/store/uplink/implants.ftl
@@ -8,3 +8,6 @@ uplink-syndicate-radio-implanter-bundle-name = Syndicate Radio Implanter Bundle
 uplink-syndicate-radio-implanter-bundle-desc = Two implanters for the price of one and a half! Share one with your Syndicate friend.
 
 uplink-storage-implanter-delta-desc = Hide goodies inside of yourself with new bluespace technology! Budget cuts have resulted in it NOT STORING High Value or storage items.
+
+nukeops-uplink-uplink-implanter-name = Nuclear Operative Uplink Implanter
+nukeops-uplink-uplink-implanter-desc = Stealthily order equipment without the need to carry an uplink. Swallow telecrystals to top up the uplink.
--- a/Resources/Prototypes/_DV/Catalog/Uplink/implants.yml
+++ b/Resources/Prototypes/_DV/Catalog/Uplink/implants.yml
@@ -69,3 +69,23 @@
     whitelist:
       tags:
       - NukeOpsUplink
+
+- type: listing
+  id: NukeopsUplinkUplinkImplanter
+  name: nukeops-uplink-uplink-implanter-name
+  description: nukeops-uplink-uplink-implanter-desc
+  icon: { sprite: /Textures/Objects/Devices/communication.rsi, state: old-radio }
+  productEntity: NukeopsUplinkImplanter
+  cost:
+    Telecrystal: 4
+  categories:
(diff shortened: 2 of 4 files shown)


Avalon-Proto:
I would drop it to 1-2 tbh, make it cheap so they actually try stealth

Eternally-Confused:
> I would drop it to 1-2 tbh, make it cheap so they actually try stealth

One TC would make it cost less than a normal syndicate's, which seems silly. I could see 2 or 3 being reasonable though.

Avalon-Proto:
So changing the indents on the liquid fuel guns broke them :blunt:

Radezolid:
I feel the capacity to buy shit if you get caught is strong, Especially considering the things nukeops can buy. I think 4 TC is fair.

NoElkaTheGod:
> I feel the capacity to buy shit if you get caught is strong, Especially considering the things nukeops can buy. I think 4 TC is fair.

While true, this also requires you to have some spare TC, which means that you have to spend less TC initially, and thus come to the station with less-than-ideal equipment. While that'd make for a funny "dark souls boss phase 2" moment, I don't see it as a viable strategy.

JustTNE:
I'm worried this will result in players being less eager to revive fallen nuclear operatives in case of "flukies" to then put them on trial, since the nuclear operative is still highly dangerous even without any equipment due to the implant.

deltanedas:
"""""""trial""""""" at centcomm

NoElkaTheGod:
> I'm worried this will result in players being less eager to revive fallen nuclear operatives in case of "flukies" to then put them on trial, since the nuclear operative is still highly dangerous even without any equipment due to the implant.

A nukie could always have an esword in a storage implant, so not much changes here really. It's just that nobody ever does that, because why would you prepare for being killed, if you can prepare more and lower your chances to get killed?

ThataKat:
Taking this in for direction review, maybe 3 days but hopefully sooner.

ThataKat:
Direction here,
Price is good, we're willing to try it out. Concerns we see are encouraging nukies to be killed and not captured, and making it harder to stop them on stealth rounds. That being said, it seems unlikely that standard nukeops will be regularly setting aside enough TC for this to be a problem on the off chance that they get captured. 

As for stealth ops, it could have no effect or it could save them from a single search. We'll give it tentative approval and keep an eye out in case it predominately leads to lame strategies.

Direction approved.

Toby222 (approved):

PR #3403, opened 2025-04-08
Port Funkychem midgame & lategame

sowelipililimute:
## About the PR & Why
A wide array of Funkychem chemicals have been brought over to DeltaV, fleshing out midgame and lategame options for chemicals.

At the moment, chemistry doesn't really have much of a lategame, nor interesting medical formulations to play around with once the basics & the advanceds are done. This introduces alternate midtiers for brutes, burns, airloss, as well as the entire set of Exotic chemicals and pyrogenics to diversify the range of medical options, and to make it less likely that every medical ward has an identical set of available chemicals.

This also brings Funkystation's QoL changes to the ChemVend and SyndieJuice, giving them more containers and some welding fuel jugs.

## Technical details
- cherrypick a bunch of changes from funkystation
- change some of our existing content where needed to fit

## Media
New midtiers
grafik
Pyrogenics
grafik
Exotics
grafik

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: TooSillyFennec, correspond, sowelipililimute
- add: Added six new exotic chemicals, listed in their own guidebook tab. Ported from Funkystation
- add: Added four new toxins, Heparin, Rhigoxan, Spewium, and Pestiline. Ported from Funkystation.
- add: Added three new pyrogenic chemicals. Ported from Funkystation
- add: Granibitual has been added, a slow acting catch all chemical. Ported from Funkystation.
- add: Salicylic Acid from Funkystation has been added, replacing its Floofstation counterpart as a midtier brute chemical.
- remove: Dexalin Plus is removed, replaced by Convermol and Salbutaoml. Ported from Funkystation.
- add: Two new airloss chemicals have been added: Salbutamol and Convermol. Ported from Funkystation.
- tweak: Advanced burn chemicals now mix into Licoxide. Ported from Funkystation.
- tweak: Species can metabolise 3 chemicals as a baseline now, have fun making mixes! Ported from Funkystation.
- tweak: Chemical Dispensers now have welding fuel. Ported from Funkystation.
- tweak: ChemVends now have Welding Fuel, bottle boxes, beaker boxes, more jugs, and more Hydrogen. Ported from Funkystation.
- tweak: The ratios for oil and acetone production have been adjusted, producing more oil and acetone. Ported from Funkystation.

:cl: sowelipililimute
- add: Added a guidebook page for chemists detailing frequently used chemicals

PR #3931, opened 2025-06-10
Make cigars and cigarettes grindable

Eternally-Confused:
## About the PR

Made cigars and cigarettes be grindable in reagent grinders (the chemistry blenders). Interdynes have a seperate solution with 2u of omni when ground, making mass producing omni much less appealing.

## Why / Balance

#3403 Added stabilized stimulants, which requires the reagent nicotine to create. However a source of nicotine wasn't added, making stabilized stimulants impossible to make without the oracle or maints pill gambling. Funkystation seems to have used this as their solution. 
In terms of balance: Nicotine is technically a poison, but 10u of it will still only do .2 Cellular. The bigger issue would be making the other reagents that are in cigarettes more obtainable. Dan's soaked smokes have a few silly reagents, though I'm unsure if this would be an issue in actual gameplay. 

## Technical details

added the extractable component to BaseCigar and created a new extractable solution for interdyne cigarettes 

## Media

Screenshot 2025-06-09 233343
(this goes on for a while, listing every cigarette type in the game)

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- tweak: The nicotine reagent is now obtainable through grinding cigarettes.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Consumable/Smokeables/Cigarettes/cigarette.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Smokeables/Cigarettes/cigarette.yml
@@ -83,6 +83,16 @@
             Quantity: 10
           - ReagentId: Omnizine
             Quantity: 30
+    # Begin DeltaV additions
+      omnizinecigarette:
+        reagents:
+          - ReagentId: Nicotine
+            Quantity: 10
+          - ReagentId: Omnizine
+            Quantity: 3
+  - type: Extractable
+    grindableSolutionName: omnizinecigarette
+    # End DeltaV additions
 
 - type: entity
   id: CigaretteOmnizine
--- a/Resources/Prototypes/Entities/Objects/Consumable/Smokeables/base_smokeables.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Smokeables/base_smokeables.yml
@@ -33,6 +33,8 @@
   - type: Tag
     tags:
     - Trash
+  - type: Extractable # Funkychem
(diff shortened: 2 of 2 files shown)


NoElkaTheGod:
Can't you grind tobacco leaves for nicotine? Either way, great PR, I'm all in for it

Eternally-Confused:
> Can't you grind tobacco leaves for nicotine? Either way, great PR, I'm all in for it

I've tried it a handful of times, including a few minutes ago. If there's a way to grind tobacco leaves, I can't seem to figure it out.
The in-game guidebook also doesn't include a source, which would imply to me that one doesn't currently exist (?)

ThataKat:
(Not a direction response)
I would be kind of concerned about people grinding up Interdyne's for the omnizine, yeah. Even just a traitor chemist using it to make advanced meds really easily, which is kind of lame. I feel like it'd be a lot safer to just fix whatever is going on with tobacco leaves and make them grindable.

deltanedas:
botany can easily mass produce omnizine but it should have a separate solution with only like 2u omni when ground to encourage actually smoking them

Eternally-Confused:
> botany can easily mass produce omnizine but it should have a separate solution with only like 2u omni when ground to encourage actually smoking them

Yeah that makes sense. I _think_ I set it up right :blunt:

Stop-Signs:
This is minor enough to not need direction review, and old enough.

Stop-Signs (approved):
Looks good to me.

TehFlaminTaco (approved):
Yaml ops with no obvious issues

PR #3939, opened 2025-06-11
Make shotguns reloadable while moving

kotobdev:
## About the PR
Shotguns, when being loaded from an ammo container, will no longer have the doafter cancelled if the player moves.

## Why / Balance
Shotguns were essentially unreloadable in space, which was absolutely infuriating. This'll make reloading shotguns just generally less annoying.

## Technical details
changed one line from true to false
this also makes it so you can load magazines and other ammo containers while moving

## Media

https://github.com/user-attachments/assets/7c639b10-f17e-48ec-9eca-442e4748bfff

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Shotguns can now be reloaded while moving.


Tenguini:
This buff is so good and beautiful and will make shotguns usable finally

deltanedas (approved):

PR #3955, opened 2025-06-15
Remove advanced energy guns research

Stop-Signs:
## About the PR
removed advanced energy guns research, which includes the ik30, PDW9, and laser magazine

## Why / Balance
ridiculously OP in its current state, unfun and unfair to fight. the PDW doesn't have a place without this pairing so its going too.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: IK30, PDW-9, and laser magazines are no longer available to crew.

### DIFF
--- a/Resources/Prototypes/_DV/Recipes/Lathes/Packs/security.yml
+++ b/Resources/Prototypes/_DV/Recipes/Lathes/Packs/security.yml
@@ -75,13 +75,13 @@
   id: SecurityWeaponsDeltaV
   recipes:
   - WeaponEnergyGun
-  - WeaponEnergyGunMini
+ # - WeaponEnergyGunMini
   - WeaponEnergyGunPistol
   - WeaponRifleLaser
   - AdvancedTruncheon
   - WeaponColdCannon
   - WeaponBeamCannon
-  - WeaponEnergyGunAdvanced
+ # - WeaponEnergyGunAdvanced
   - WeaponImmolator
 
 - type: latheRecipePack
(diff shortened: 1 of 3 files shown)


NoElkaTheGod:
Security getting a nerf on DeltaV wasn't on my 2025 bingo card

Smugman:
Good change, i recall playing a lot of SRN and having all my borgs 3 shot by the beast that is the IK-30.
A High-tech arsenal for security is conceptually awesome and a great aesthetic though.

Avalon-Proto:
Isn't the PDW-9 just a higher capacity mini-energy gun,? Why nuke it

rebe83:
I don't think the PDW needs removed, just because its "partner" is gone
I think the instant reload magazines are the issue for the IK. It has always had like 20 damage per shot and 20 shots per magazine, LETHAL ONLY is fairly insane.
If intended to fill the niche of "energy smg" they could be doing 10 damage a piece like the HOS shotgun projectiles and it would still be a very considerable weapon. 
PDW as top tier pistol instead of ALP I think it's a modest upgrade to mini egun and has disable function.

I would personally throw ALP in for being removed since I don't like the idea of infinite lethal only pistols, they are also rationalized as a sidearm that is fine to standard issue as an upgrade to the mk58. But that might just be fine for round progression. I just like the captain laser being unique and very pivotal if lost. 
However this is pretty fringe and doesn't show up in survival shifts too often, since it requires epi and logi to doing very well and to be helping sec.

PR #3967, opened 2025-06-16
feat: allow autodocs to hold arm augments

### LINKED ISSUE #3569: Autodoc can't be used for augments
## Description
The autodoc only supports picking up items within its inventory, but augments don't fit in that inventory. The autodoc should either support picking up items off the ground, have some way of holding big items (maybe a single item itemslot?), or explicitly just not support augments. (If the last option, this should arguably be written in the guidebook page on augments... oh. Rip.)

**Reproduction**

1. Be an autodoc user (none currently known).
2. Be unable to autodoc in augments.

**Screenshots**

**Additional context**

Considering how no one uses autodocs (on peri/horizon at least...), I don't think there's an issue in broadening their appeal by allowing them to do more surgeries.

perryprog:
## About the PR

Gives Autodocs an itemstorage for arm augments for use in surgery.

Also switches E to opening UI since that changed at some point. UI is used much much more than storage UI, so just use the verb for storage.

## Why / Balance

Fixes #3569.

## Technical details

Tested a good few surgeries, seems okay.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The Autodoc can now fit arm augments in its storage.

### DIFF
--- a/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
+++ b/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
@@ -46,6 +46,7 @@
         showEnts: False
         occludes: True
         ents: []
+      surgery_augment_slot: !type:ContainerSlot # DeltaV - autodoc augment support
   - type: ApcPowerReceiver
     powerLoad: 10000
   - type: DeviceList
@@ -77,10 +78,10 @@
     speechSounds: Pai
   - type: Storage
     clickInsert: false # incredibly annoying
-    openOnActivate: false
+    openOnActivate: false # DeltaV
     grid:
     - 0,0,3,3
-  - type: ItemSlots
+  - type: ItemSlots # DeltaV - autodoc augment support
     slots:
       surgery_augment_slot:
         name: Augment Holder
--- a/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
+++ b/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
@@ -77,8 +77,18 @@
     speechSounds: Pai
   - type: Storage
     clickInsert: false # incredibly annoying
+    openOnActivate: false
     grid:
     - 0,0,3,3
+  - type: ItemSlots
+    slots:
+      surgery_augment_slot:
+        name: Augment Holder
+        insertOnInteract: false
+        ejectOnBreak: true
+        whitelist:
+          components:
+          - AugmentArm
   - type: SurgerySpeedModifier
     speedModifier: 1
   - type: GuideHelp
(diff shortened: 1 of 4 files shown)


deltanedas:
just make it not gigantic

perryprog:
As far as I can tell that wouldn't work—arm augments have a storage component so you can never insert them into another storage that has max size less than or equal to that of the arm augment. See last comment on mentioned issue.

perryprog:
Regardless, test fail real + fixing attribution, I was brain farting on that.

deltanedas:
> As far as I can tell that wouldn't work—arm augments have a storage component so you can never insert them into another storage that has max size less than or equal to that of the arm augment. See last comment on mentioned issue.

you can have a tardis and it works fine, see combat medkit

perryprog:
That's because combat medical kit has a max item size of small, while bags have a max item size of huge.

HOWEVER I just had the genius revelation that I can uh. Just increase the max item size of the autodoc storage.

I will miss the cute visual of the autodoc having a little connector for arm augments.

Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
```diff
@@ -77,6 +77,8 @@
     speechSounds: Pai
   - type: Storage
     clickInsert: false # incredibly annoying
+    openOnActivate: false # DeltaV - also annoying
```
deltanedas, line 80: nothing else happens when you press E remove this

deltanedas (changes requested):

Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
perryprog: For some reason this causes the storage UI to open on click as well as on E, meaning you have to use the verb menu to open the UI. Storage is used less often than UI, so if anything it should be the verb while UI is on interact.

perryprog: Eww it's because "Toggle UI" is alphabetically after "Open Storage" and "Close Storage".

Could add priority field to ActivateableUI to allow setting the verb to have priority of 1 instead of 0 for autodoc if you want? Or I can just change back how it was, but I don't think it's good UX.

deltanedas: verb priority is the solution here yes

perryprog: Ugh, I'm sorry—I must've mis-tested when I said the above on verb priority. It's actually the event handling ordering, so it seems that it'd have to be adding a before:SharedStorageSystem to the activation subscription in ActivateableUI, but then that just leads to the same behavior as setting openOnActivate to false like I did here, since there's no alternate activate handling for storage components.

Stop-Signs (approved):
Direction approved

deltanedas (approved):

PR #3968, opened 2025-06-16
feat: toggle for advanced operating tables preventing rot

perryprog:
## About the PR

Adds a toggle to advanced surgery tables to switch between surgery speed boost and antirot.

## Why / Balance

If you're working on a dead patient that has eight million burn then you should at least be allowed to use an autodoc to help.

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Advanced operating tables can now be switched between preventing rot and increasing surgery speed.

### DIFF
--- a/Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
+++ b/Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
@@ -1,0 +1,18 @@
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Item.ItemToggle.Components;
+
+public sealed partial class ComponentTogglerComponent
+{
+    /// <summary>
+    /// Components that are removed on activation. Allows "swapping" components on toggle.
+    /// </summary>
+    [DataField]
+    public ComponentRegistry? ComponentsRemovedOnActivate;
+
+    /// <summary>
+    /// Components that are added on deactivation. Allows "swapping" components on toggle.
+    /// </summary>
+    [DataField]
+    public ComponentRegistry? ComponentsAddedOnDeactivate;
+}
--- a/Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
+++ b/Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
@@ -5,14 +5,9 @@ namespace Content.Shared.Item.ItemToggle.Components;
 public sealed partial class ComponentTogglerComponent
 {
     /// <summary>
-    /// Components that are removed on activation. Allows "swapping" components on toggle.
+    /// Components that are removed on activation, and added back on deactivation.
+    /// Allows for "toggling" between two components.
     /// </summary>
     [DataField]
-    public ComponentRegistry? ComponentsRemovedOnActivate;
-
-    /// <summary>
-    /// Components that are added on deactivation. Allows "swapping" components on toggle.
-    /// </summary>
-    [DataField]
-    public ComponentRegistry? ComponentsAddedOnDeactivate;
+    public ComponentRegistry? DeactivatedComponents;
 }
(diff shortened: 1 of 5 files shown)


Stop-Signs:
Direction here, we'd like this to be a trait that the advanced operating table has in order to give players a better reason to make it.

perryprog:
Sure—to be clear, are you saying the advanced operating table should pause (or slow down?) rot?

Stop-Signs:
> Sure—to be clear, are you saying the advanced operating table should pause (or slow down?) rot?

Ok so, I did actually misread something in our thread. This should either be a *mode* on the advanced operating table. or it should be its own thing, like an advanced stasis bed.

apologies for the mistake there, thank you for asking.

perryprog:
Sure. And what would the mode on the advanced operating table look like? As in, in what way does it change its behavior?

Stop-Signs:
> Sure. And what would the mode on the advanced operating table look like? As in, in what way does it change its behavior?

Something where you can toggle between faster surgery speed and stasis.

perryprog:
Okie, here's what I got; see updated PR description and media.

Not sold on my wording choices and my verbs are in fact not verbs, but I can't think of anything better.

Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    [DataField]
+    public ComponentRegistry? ComponentsRemovedOnActivate;
+
+    /// <summary>
+    /// Components that are added on deactivation. Allows "swapping" components on toggle.
+    /// </summary>
+    [DataField]
+    public ComponentRegistry? ComponentsAddedOnDeactivate;
+}
```
deltanedas: dont need 2 of the same thing, rename to RemovedComponents and use it for both cases
then remove the field-less one from the yml

Resources/Prototypes/_Shitmed/Entities/Structures/Machines/advanced_operating_table.yml
```diff
@@ -8,3 +8,20 @@
...
+    on: advanced-operating-table-stasis-mode
+    off: advanced-operating-table-surgery-mode
+  - type: ComponentToggler
+    components:
+    - type: AntiRotOnBuckle
+    componentsRemovedOnActivate:
+    - type: SurgerySpeedModifier
+    componentsAddedOnDeactivate:
+    - type: SurgerySpeedModifier
+      speedModifier: 1.25
```
deltanedas, line 25: missing end comment

deltanedas (changes requested):

Stop-Signs (approved):
Direction approved, this is fantastic work.

Content.Shared/_DV/Item/ItemToggle/Components/ComponentTogglerComponent.cs
perryprog: So true. Since upstream has 'RemoveComponents' (only used on deactivate, so not helpful here) I went with 'DeactivatedComponents'.

Resources/Locale/en-US/_DV/surgery/advanced-operating-table.ftl
```diff
@@ -0,0 +1,5 @@
+advanced-operating-table-stasis-mode = It is currently [color=lightblue]slowing down decay[/color] of patients strapped to it.
+advanced-operating-table-surgery-mode = It is currently [color=lightgreen]stabilizing the patient's movement[/color].
```
deltanedas: reword it, currently it implies it wouldnt make surgery faster for a corpse since it doesnt move
probably just say speeds up surgery its not a crazy secret

deltanedas (changes requested):

SirWarock:
Will the advanced operating table still work like a medical bed?
Aka, it heals blunt and poison damage, which is a really neat feature for when you forget to clean your tools, but the patient auto heals the small amount you caused.

Resources/Locale/en-US/_DV/surgery/advanced-operating-table.ftl
perryprog: Yeah, good point on both counts.

perryprog:
> Will the advanced operating table still work like a medical bed? Aka, it heals blunt and poison damage, which is a really neat feature for when you forget to clean your tools, but the patient auto heals the small amount you caused.

Yeah, it only looses the surgery speed enhancement when in "stasis bed mode", and stasis bed mode only prevents rot, nothing else. Worth noting that it's getting that because it is inheriting from normal beds, not medical beds—medical beds passively heal more.

Also bed passive healing doesn't work on dead anyway.

deltanedas (approved):

PR #3974, opened 2025-06-17
Make IDs work inside chest cavity

kotobdev:
## About the PR
IDs/PDAs/items with access will now be able to interface with access-locked things if stored inside one's chest via surgery.

## Why / Balance

[relevant wyci](https://discord.com/channels/968983104247185448/1384089337112694804)
Adds a nice practical use for surgery, and also encourages silly ghetto surgery. Implanted ID icons will appear on sechuds, so unless you're very sneaky with an agent ID, it'll probably be pretty easy to sec to figure out what you've done if they confiscate your PDA. 

## Technical details
shitmed is shitmed so the code to check the chest cavity, unfortunately, may be a little bit shit. i have made it as pretty as i know how

## Media

https://github.com/user-attachments/assets/e6d1061c-c22b-4f69-bd1e-d587402aa4df

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: IDs now work when implanted inside of chest cavities.

### DIFF
--- a/Content.Shared/Access/Systems/AccessReaderSystem.cs
+++ b/Content.Shared/Access/Systems/AccessReaderSystem.cs
@@ -1,8 +1,8 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Content.Shared.Access.Components;
-using Content.Shared.Body.Components;
-using Content.Shared.Containers.ItemSlots;
+using Content.Shared.Body.Components; // DeltaV
+using Content.Shared.Containers.ItemSlots; // DeltaV
 using Content.Shared.DeviceLinking.Events;
 using Content.Shared.Emag.Systems;
 using Content.Shared.Hands.EntitySystems;
@@ -25,6 +25,7 @@ public sealed class AccessReaderSystem : EntitySystem
 {
     [Dependency] private readonly IPrototypeManager _prototype = default!;
     [Dependency] private readonly InventorySystem _inventorySystem = default!;
+    [Dependency] private readonly ItemSlotsSystem _itemSlotsSystem = default!; // DeltaV
     [Dependency] private readonly IGameTiming _gameTiming = default!;
     [Dependency] private readonly EmagSystem _emag = default!;
     [Dependency] private readonly TagSystem _tag = default!;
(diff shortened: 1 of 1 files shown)


Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -1,6 +1,8 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Content.Shared.Access.Components;
+using Content.Shared.Body.Components;
+using Content.Shared.Containers.ItemSlots;
```
Quanteey:
```suggestion
using Content.Shared.Body.Components; // DeltaV
using Content.Shared.Containers.ItemSlots; // DeltaV
```
You might wanna add DeltaV comments to these too

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -359,6 +361,20 @@ public bool FindAccessItemsInventory(EntityUid uid, out HashSet<EntityUid> items
             items.Add(idUid.Value);
         }
 
+        // Begin DeltaV - check chest cavity
+        if (TryComp(uid, out BodyComponent? body) && body.RootContainer.ContainedEntity is { } root)
+        {
+            if (TryComp(root, out ItemSlotsComponent? itemSlots))
+            {
+                foreach (var slot in itemSlots.Slots.Values)
```
deltanedas: check the actual cavity slot this is just begging to break if a bodypart has slots like if ipc battery gets moved to it

deltanedas (changes requested):

Stop-Signs:
Direction review 24 hours

deltanedas (approved):

Stop-Signs:
Hi, we have decided to decline this submission due to:

- the fact that this will incentivise command to put valuable IDs in an area that cannot be accessed without surgery or gibbing
- the fact that this will be used by thieves to put valuable IDs in an area that cannot be accessed without an invasive procedure and that are already disallowed from other means of in-body storage
- the absence of any rationale for why someone would use this and how it would contribute to their interactions with other players developing the story of the round
- the fact that this will be used by thieves to put valuable IDs in an area that cannot be accessed without an invasive procedure and that are already disallowed from other means of in-body storage

PR #3976, opened 2025-06-17
Nymph forces ghost return

### LINKED ISSUE #3966: Diona nymphing while ghosted does not pull ghost back into new nymph
## Description
If the head of a diona is removed while the player is ghosted, the nymph will be generated and NOT pull the ghost back into the nymph. 
**Reproduction**
1. Take control of diona
2. Rev up that burn chamber and walk in and die
3. diona player mind leaves body (/ghost)
4. diona eventually ashes and nymph ejects
5. ghost is not pulled into the nymph
**Screenshots**

Image

**Additional context**
I'm guessing that whatever check the nymph code in nymphSystem.cs does doesn't act the same way as say MMI'ing a brain does. TryGetMind *does* seem to link the mind, as ghost "return to body" is linked to the nymph after nymphing.

This bug has existed since shitmed was added, it used to work fine.
Possibly notably this is calling TryGetMind where MindContainerComponent=NULL , also NymphComponent Transfermind is initialized to false? But the mind is transferred so the code is working right
Image

rebe83:
## About the PR
Forces the player ghost into the body of the brain nymph when it is spawned.

## Why / Balance
If a diona is gibbed / ashed nymphs for whatever reason, they would be unaware and completely miss their one chance for revival / escape, and the simplemob AI could run off and hide under a piece of garbage and be lost forever, without the player even knowing that they nymphed at all.
Previously when a diona was gibbed while ghosted it would force the ghost into the body, but this was changed or edited at some point.
## Technical details
Set ghostcheckoverride = true in the transfermind() call sent by nymphing 

fixes #3966

## Requirements

- [x ] I have tested all added content and changes.
- [x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- fix: A diona corpse nymphing while they are ghosted once again returns them to their new body!

### DIFF
--- a/Content.Server/Species/Systems/NymphSystem.cs
+++ b/Content.Server/Species/Systems/NymphSystem.cs
@@ -42,7 +42,7 @@ public sealed partial class NymphSystem : EntitySystem
 
         // Move the mind if there is one and it's supposed to be transferred
         if (comp.TransferMind == true && _mindSystem.TryGetMind(args.OldBody, out var mindId, out var mind))
-            _mindSystem.TransferTo(mindId, nymph, mind: mind);
+            _mindSystem.TransferTo(mindId, nymph, true, mind: mind); // force ghost into nymph body and transfers mind to nymph
 
         // Delete the old organ
         QueueDel(uid);
--- a/Content.Server/Species/Systems/NymphSystem.cs
+++ b/Content.Server/Species/Systems/NymphSystem.cs
@@ -42,7 +42,7 @@ public sealed partial class NymphSystem : EntitySystem
 
         // Move the mind if there is one and it's supposed to be transferred
         if (comp.TransferMind == true && _mindSystem.TryGetMind(args.OldBody, out var mindId, out var mind))
-            _mindSystem.TransferTo(mindId, nymph, true, mind: mind); // force ghost into nymph body and transfers mind to nymph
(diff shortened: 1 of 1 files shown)


Content.Server/Species/Systems/NymphSystem.cs
```diff
@@ -42,7 +42,7 @@ private void OnRemovedFromPart(EntityUid uid, NymphComponent comp, ref OrganRemo
 
         // Move the mind if there is one and it's supposed to be transferred
         if (comp.TransferMind == true && _mindSystem.TryGetMind(args.OldBody, out var mindId, out var mind))
-            _mindSystem.TransferTo(mindId, nymph, mind: mind);
+            _mindSystem.TransferTo(mindId, nymph, true, mind: mind); // force ghost into nymph body and transfers mind to nymph
```
sowelipililimute:
```suggestion
            _mindSystem.TransferTo(mindId, nymph, true, mind: mind); // DeltaV - force ghost into nymph body and transfers mind to nymph
```

sowelipililimute (dismissed):

HTMLSystem:
A PR tip: In the "Why/Balance" section of the PR, you can say "fixes #3966" and that will automatically close the issue whenever this is merged. Just so a maintainer doesn't have to do it and the issues page is up to date!!!

deltanedas (approved):

PR #3980, opened 2025-06-18
Make survival epic again

Stop-Signs:
## About the PR
edited survival values to be evil

## Why / Balance
survival is either greenshift or greenshift + ninja, this should make survival actually difficult to _survive_

## Technical details
math

## Media
epic recreation of what it should now look like (red is current, black is 3 months ago survival, pink is now) (i have no graphing software)
image

## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: survival values have been changed, and should make  the mode more exciting

### DIFF
--- a/Resources/Prototypes/GameRules/roundstart.yml
+++ b/Resources/Prototypes/GameRules/roundstart.yml
@@ -424,11 +424,13 @@
   - type: NextEvent # DeltaV: Queue Event for precognition
   - type: RampingStationEventScheduler
     timeKeyPoints: # DeltaV
-    - 14, 16
-    - 32, 8
+    - 10, 12
+    - 32, 6
+    - 32, 3
+    - 12, 1
     - 32, 4
-    - 8, 2
-    - 32, 8
+    - 32, 2
+    - 32, 1.5
     scheduledGameRules: !type:NestedSelector
       tableId: BasicGameRulesTable
 


sowelipililimute (approved):
direction review concluded the consensus seems to be try it and see so let's call this a 1 week test merge

Avalon-Proto:
So, after 2 weeks of this I can safely say. I'm not sure how to feel about it. As shifts of survival on average go from 0 to 100nin literal moments, as around. The 1 1/2 hour mark the game throws every antag in the book at you at once. I feel like this needs some more tuning or even to go back to the scheduler *before* all our changes. Back to the old exponential growth

TheLunaticOfTheDrunkenMoon:
After two weeks I can say that it's alot of stress to sec and alot of other people on the server in the current state. I've seen more back to back power outages and "Centcom gave you cleaning supplies, oh by the way we really thing you need these cleaning supllies so here's a second shipment" and it feels like this would be better as a separate gamemode where you specifically know it's probably going to be chaotic,,,, Kind of like the old ss13 why type of star you're orbiting dictated how often the events would be on station. 

But I digress, I agree it either need more tuning or a revert to how it was. Not to say it wasn't fun, the issue is that this type of gameplay makes back to back shifts less enjoyable, especially when it's between zombies, nukies, cultists, and hell slog of whatever the game felt like that shift. Again this would be better for events like Garbage day where you're expecting it to be bad..... this might feel better if the pool of events it could do was more stocked, but so far it's just the same events getting pulled back to back and it's just not enjoyable from a player retention standpoint. I'd personally vote for a revert

Axianamos:
I like increased action and chaos. The biggest issue however is variety. We dont have enough antags/events for the scheduler to do anything vartied. So every shift its going to be the same.

PR #4003, opened 2025-06-23
Wizard Clothes- Slowdown removed, hat lets you breathe

Avalon-Proto:
## About the PR
Removed the slowdown on this magical robes, and gave the hat the Breathmask Comp

## Why / Balance
While the wizard robe is an EVA suit, it is rare, and not exactly standard. So why does it slow as much as a standard EVA? This allows wizrobes to no longer slow you when worn, since they do not have NEARLY the bulk of an EVA suit. Also allows you ot breath in sapce with just the hat, an no mask. As its MAGIC

## Technical details
Light yaml work

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: EVA Safe wizard robes no longer slow you, and the hat allows you to breath

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
@@ -172,9 +172,6 @@
   id: ClothingOuterWizardBase
   components:
   - type: WizardClothes
-  - type: ClothingSpeedModifier # DeltaV - We made wizrobes spaceproof, this just makes them incur no slowdown
-    walkModifier: 1
-    sprintModifier: 1
 
 # Is this wizard wearing a fanny pack???
 - type: entity
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
@@ -172,6 +172,9 @@
   id: ClothingOuterWizardBase
   components:
   - type: WizardClothes
+  - type: ClothingSpeedModifier # DeltaV - We made wizrobes spaceproof, this just makes them incur no slowdown
+    walkModifier: 1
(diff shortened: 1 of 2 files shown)


SirWarock:
So can you actually differentiate between the two anymore? Or do you just space yourself to see? Pfff

Spielern:
> So can you actually differentiate between the two anymore? Or do you just space yourself to see? Pfff

the fake ones have a different description on the robes

Stop-Signs:
Direction review 24 hours.

Toby222 (dismissed):

Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
```diff
@@ -172,6 +172,9 @@
   id: ClothingOuterWizardBase
   components:
   - type: WizardClothes
+  - type: ClothingSpeedModifier # DeltaV - We made wizrobes spaceproof, this just makes them incur no slowdown
+    walkModifier: 1
+    sprintModifier: 1
```
Stop-Signs: remove changes here

Stop-Signs (changes requested):
Direction here, since these are a free eva suit in the mail we would like to keep the slowdown. The breath mask component on the hat is fine, however.

Resources/Prototypes/Entities/Clothing/OuterClothing/misc.yml
Avalon-Proto: Can I at least make them a bit faster, say, 10% slowdown instead? Still decently rare and all

Stop-Signs: no.

deltanedas (approved):

SirWarock:
The changelog still has it written that it doesn't slowdown :o

PR #4004, opened 2025-06-23
Halve the fire rate of the MK32

Radezolid:
## About the PR
This PR halves universals' fire rate and reduces the cost of the crate for them.

## Why / Balance
OP weapon, I would rather axe them from the crew, but this works somewhat.

## Technical details
Killed the `pistols.yml` from nyano and moved the gun to our namespace.

## Media
N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Universals now cost 6k less and they fire at half the rate of fire as before.


Stop-Signs (approved):
Direction speed approving since this is based on changes we asked for.

deltanedas (approved):

PR #4011, opened 2025-06-26
Add a materials magnet to the logistics techfab

AlgisAlphonse:
## About the PR
Adds the ported-from-Frontier lathe magnet (previously used for the ore processor) to the logistics techfab

## Why / Balance
Logistics typically stores materials in their lathe. When a particularly large load of materials come in, inserting it all into the lathe is frankly miserable. This allows automation of that using conveyors.

## Technical details
It's just YAML, I just ported the lines from the ore processor and it just works like that, taking only mats, not ores. 

## Media
N/A

## Requirements

## Breaking changes

None.

**Changelog**
:cl:
- add: Added a short range materials magnet to the logistics techlathe. Cargo Technicians rejoice! WARNING: This does not fix the existing bug with diamonds being stuck in lathes. Use with caution.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
@@ -135,9 +135,9 @@
     board: LogisticsTechFabCircuitboard
   - type: StealTarget
     stealGroup: LogisticsTechFabCircuitboard
-  - type: MaterialStorageMagnetPickup # DeltaV - Adds magnet pull from Frontier
+  - type: MaterialStorageMagnetPickup
     magnetEnabled: True
-    range: 0.30 # DeltaV - End Magnet Pull
+    range: 0.30
   - type: LatheAnnouncing
     channels: [Supply]
 
--- a/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
@@ -135,6 +135,9 @@
     board: LogisticsTechFabCircuitboard
   - type: StealTarget
     stealGroup: LogisticsTechFabCircuitboard
+  - type: MaterialStorageMagnetPickup # DeltaV - Adds magnet pull from Frontier
+    magnetEnabled: True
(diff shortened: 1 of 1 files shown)


TheLunaticOfTheDrunkenMoon:
I would personally love this, no more spending time taking so long putting in mats.... I do, however wonder how it'd take to removing mats to be able to give them to other departments

IamVelcroboy:
I feel like, pending approval, you might as well add it to all of them

AlgisAlphonse:
> I feel like, pending approval, you might as well add it to all of them

Maybe? I am not sure it makes that much sense _why_ for others. I might PR it anyway, but having this be as a separate PR for a case where it makes the most sense increases chances of it being approved.

SirWarock:
It's not really useful for other lathes, since only Logistics lathe would fit the aesthetic of a belt feeding into it. You can't drag a crate of materials onto the other ones, so the magnet wouldn't activate.

The only time it would be useful is if it got emagged and someone repaired it.

perryprog:
I do have an implementation of something that tries to solve the same problem in a different way—it essentially lets ore processors link to ore silos as material sources, meaning you can "print" from the refiner to the ore silo. (This doesn't help with recyclables, though!)

I'm happy to port that over here if that's also wanted, whether in addition or as an alternative.

Toby222:
Adding it to the silo as well would make sense imho

Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
```diff
@@ -135,6 +135,9 @@
     board: LogisticsTechFabCircuitboard
   - type: StealTarget
     stealGroup: LogisticsTechFabCircuitboard
+  - type: MaterialStorageMagnetPickup # DeltaV - Adds magnet pull from Frontier
+    magnetEnabled: True
+    range: 0.30 # DeltaV - End Magnet Pull
```
Toby222: No need for DeltaV comments in _DV

Toby222 (changes requested):

Proxysseia:
Direction Review 24h

Proxysseia:
Direction approved

Toby222 (approved):
lgtm

PR #4014, opened 2025-06-27
Fix Listening Post crew monitoring console

Elpersonn:
## About the PR
This PR fixes the previously broken crew monitoring console aboard the listening post. 

## Why / Balance
This change would allow listening post operative to actually know where the crew is aboard the station and allow them to have more accurate information which they could use to cause communication mayhem.

## Technical details
The code within the PR slightly modifies the station system, which now sends grids along with the station and name, as well as adds the "Syndicate" tag to the crew monitoring console to let the system know that the LPO console in particular has to search for the main station.

## Media

https://github.com/user-attachments/assets/a8754aa5-969e-40d7-aeb5-0b2a1d577e49

## Requirements

## Breaking changes
The `StationsUpdatedEvent` tuple was turned into a `record struct`, however it can still be used as the original tuple as it contains an implicit conversion method.

**Changelog**

:cl:
- fix: Fixed Listening Post crew monitor not displaying station grid correctly.

### DIFF
--- a/Content.Client/Administration/UI/Tabs/ObjectsTab/ObjectsTab.xaml.cs
+++ b/Content.Client/Administration/UI/Tabs/ObjectsTab/ObjectsTab.xaml.cs
@@ -76,7 +76,11 @@ public sealed partial class ObjectsTab : Control
         switch (selection)
         {
             case ObjectsTabSelection.Stations:
-                entities.AddRange(_entityManager.EntitySysManager.GetEntitySystem<StationSystem>().Stations);
+                var a = _entityManager.EntitySysManager.GetEntitySystem<StationSystem>().Stations;
+                for (int i = 0; i < a.Count; i++)
+                {
+                    entities.Add(a[i]);
+                }
                 break;
             case ObjectsTabSelection.Grids:
             {
--- a/Content.Client/Medical/CrewMonitoring/CrewMonitoringBoundUserInterface.cs
+++ b/Content.Client/Medical/CrewMonitoring/CrewMonitoringBoundUserInterface.cs
@@ -1,12 +1,8 @@
 using Content.Shared.Medical.CrewMonitoring;
 using Robust.Client.UserInterface;
 using Content.Client.Station;
-using System.Diagnostics;
 using System.Linq;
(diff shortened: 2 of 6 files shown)


Avalon-Proto:
Question - can you add this to the handheld syndie crewmon too.

Elpersonn:
> Question - can you add this to the handheld syndie crewmon too.

Probably, I think I'll keep that for a separate PR though since this is my first one.

deltanedas:
yeah this is extreme goidacode cooked up by an LLM im gonna properly implement it myself

Elpersonn:
> yeah this is extreme goidacode cooked up by an LLM im gonna properly implement it myself

That my code is shit, I understand, and kind of expected myself after looking at all the weirdass changes I made (and kinda felt like I was in too deep to go back)
But for you to call my code "cooked up by an LLM" is low-key disrespectful as hell.

deltanedas:
why would you convert a hashset to an array just to check if it has an item...

Quanteey:
Just because the code is bad or does not make much sense does not give you a pass to be disrespectful - especially as a maintainer. There are many ways to say "The entire approach is wrong, but I'll do it properly since it's a small change", but this is not it. Comments and review like this actively discourage new contributors, which is detrimental to any open source project.

Elpersonn:
> why would you convert a hashset to an array just to check if it has an item...

Because using `Contains` would give me a `Tried to perform an 'Execute' other-type access to member 'Tags' in type 'TagComponent', despite only having 'Read' access. Type Permissions: rwxrwxr--.`. Because for some reason this codebase uses file-style permissions for classes. Yeah I could've used `foreach` to find the tag, but the whole tag implementation seems to have been scrapped anyways so :P

deltanedas:
no youd just use the TagSystem api

Elpersonn:
> no youd just use the TagSystem api

God forbid a new contributor doesn't know the whole robust toolbox and ss14 API, better call say he uses AI over that....

deltanedas:
sorry but its easier to just redo the entire thing myself than manually ask you to change every single line and the whole approach anyway

PR #4016, opened 2025-06-27
make esword/edagger no longer perma-able

deltanedas:
## About the PR
made the energy arms clause specific to range again so you cant perma someone just for having a funny pen

now its only 5 minutes for knife

## Why / Balance
being able to perma someone just for having a pen is insane
if they stab someone thats assault with a deadly weapon + possession, you can still perma for that
if you want to perma people left and right do a *little* bit of thinking at least

arguable viper perma should be axed too because permaing people over tiny shit causes 2 things:
- try to do a cool rp with a pistol and get permad, traitor becomes greenshift
- know you will get permad anyway and go full rambo, noooo you cant do that so evil

you might say "but erm it says UP TO", except nobody EVER gives lenient sentences nor are they incentivised to. that's not a suggestion it's practically a requirement
if you can perma a syndie why *wouldnt* you, a functioning member of sec who isn't against nanotrasen?

**Changelog**
:cl:
- tweak: Adjusted SOP's controlled armaments, energy melee weapons are no longer Cat D.

### DIFF
--- a/.wiki/_DV/Laws/StandardOperatingProcedure.txt
+++ b/.wiki/_DV/Laws/StandardOperatingProcedure.txt
@@ -81,7 +81,7 @@ Controlled Armaments are grouped into categories of increasing severity.
 
 * Any semi-automatic ballistic arms that are easily concealable
 * Any fully-automatic ballistic arms
-* Any laser, plasma, or energy based arms, including ones designed, intended, or used for melee combat.
+* Any laser, plasma, or energy based ranged weaponry
 * Any object or device specifically designed or adapted to cause physical harm through detonation of an explosive charge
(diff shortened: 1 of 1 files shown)


Proxysseia:
direction review 24h

Proxysseia:
You can't perma someone for 'just' a pen. Perma is 25 minutes and up, Cat D Grand Poss is 20 minutes. 
Eswords should not be Cat A; Hidden knife you can slot into your PDA is not 'just a pen'.
The arguments for why being permad over tiny shit is bad still apply and are solid. 
Singling out melee energy based weapons doesn't make much sense.

There is a current ongoing draft to reduce all time punishments for crime, will be touched on in the near future.

As of now, this PR is denied.

PR #4021, opened 2025-06-29
Change plasteel and wood crates from cargo to come with 90 sheets instead of 30.

AlgisAlphonse:
## About the PR
Changed the plasteel and wood crates to contain 90 sheets. Also brough plastic and basic sheet prices in line with steel and glass crate prices.

## Why / Balance
Manufacturing plasteel is irreversible and requires an industrial ore processor. This is quite unfortunate, as it is essential for proper station repair. 30 sheets per crate is a bit silly - especially since the crate isn't that cheap. The other changes are made for cosistency.

## Technical details
It's YAML

## Media
N/A

## Requirements

- [ ] I have tested all added content and changes. - it's YAML

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: Changed plasteel and wood crates from cargo to come with 90 sheets instead of 30.
- tweak: Brough plastic and basic sheet prices in line with steel and glass crate prices - 1500 spesos.

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_materials.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_materials.yml
@@ -24,7 +24,7 @@
     sprite: Objects/Materials/Sheets/other.rsi
     state: plastic_3
   product: CrateMaterialPlastic
-  cost: 2000
+  cost: 1500 # Delta-V - was 2000
   category: cargoproduct-category-name-materials
   group: market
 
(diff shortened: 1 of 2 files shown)


deltanedas:
the arbitration is real, 90 plasteel is worth a lot more than 1.6k

you dont even need plasteel for repairs anyway, explosives and meteors treat it the same as grilles. its just to try stop antags from breaking into places

Proxysseia:
30 plasteel is worth 1.500 spesos. 90 plasteel for 1.500 spesos is an infinite money hack. 
Current plasteel cost is fine, current plasteel amount is also fine.

Colin-Tel:
It doesn't make a lot of sense to me, why not just order more than one crate?

AlgisAlphonse:
Closing as I did not realize that plasteel actually sells for that much. Huh.

PR #4023, opened 2025-06-29
Colossi - Stamina Damage Immunity

Avalon-Proto:
## About the PR
This PR, simply, makes the colossus immune to stamina damage

## Why / Balance
A giant robot powered by esoteric energy and an unknowable cosmic entity shouldn't be mega-slowed when the bartender decides to hit it twice with beanbags, so this PR makes stamina damage do jack and all to a giant cult construct

## Technical details
2 lines of yaml

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Colossi are no longer effected by stamina damage

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
@@ -10,6 +10,8 @@
     cultistText: cosmic-examine-text-cultentity
     othersText: cosmic-examine-text-entities
   - type: AntagImmune
+  - type: StaminaResistance
+    damageCoefficient: 0 # No more stunning colossi
   - type: Sprite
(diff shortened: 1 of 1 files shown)


NoElkaTheGod:
Yes please. It makes no sense that such a big thing gets slowed to a crawl from being hit with disablers, not to mention that it's very damn slow by default. If making it fully immune is bad, then at least give it high stamina damage resistance, because right now it's just silly.

AirlineFood001:
A massive ascended beast shouldn't be brought to a super slow crawl because the bartender shot it twice with _beanbag_ rounds, I completely agree with this. Like Elka said, if making it fully immune to stamina damage is bad, it at least deserves somewhat high stamina damage resistance.

deltanedas (approved):

PR #4030, opened 2025-07-02
Give LPO records tablet

kotobdev:
## About the PR
Listening Post Operatives now start with a records tablet in their pocket. Due to this, records tablets now also work off-station.

## Why / Balance
Access to station records means LPO gamers can do even more RP shenanigans. Or they can just spend their free time snooping through people's records. Either way, worth giving them the option.

## Technical details
`Generic` consoleType is only used by the records tablet, so checking for that should be fine. we probably don't want this to apply to all types, since an antag being able to steal a criminal console and constantly set wanted status from off-station would be a nightmare

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Listening Post Operatives now get records tablets.


kotobdev:
test fail fake i think
15 days without review

deltanedas (approved):

PR #4034, opened 2025-07-04
Fix harpy theobromine poisoning

Git-Kanbaru:
## About the PR
Harpies are supposed to be poisoned by theobromine (and allicin) they now will be.

## Why / Balance
It seems the intention was for them to be this way..

## Technical details
Hearts are the organ that actually metabolizes all reagents (besides, food, drinks, and alcohol). So, it is the organ that is checked when determining if a reagent is poisonous to that type of organ. Theobromine is only poisonous if the metabolizing organ is an animal organ. 

Because I gave harpies the animal heart organ, they now get poisoned correctly. As far as I can tell that is the only difference between the animal and human hearts.

## Media
Attached image of poor Urist dying of theobromine.
image

## Requirements

## Breaking changes
None that I am aware of.

**Changelog**

:cl:
- fix: Harpies now properly poisoned by theobromine and allicin

### DIFF
--- a/Resources/Prototypes/_DV/Body/Prototypes/harpy.yml
+++ b/Resources/Prototypes/_DV/Body/Prototypes/harpy.yml
@@ -19,7 +19,7 @@
       - left leg
       - head # Shitmed Change
       organs:
-        heart: OrganHumanHeart
+        heart: OrganAnimalHeart
         lungs: OrganHarpyLungs
         stomach: OrganAnimalStomach
         liver:  OrganAnimalLiver


Git-Kanbaru:
Renamed the wrong PR

AveriV2:
im pretttyy sure harpies are meant to be able to digest it

Git-Kanbaru:
The guidebook entry for them says they are poisoned by it, that entry was copied directly from the wiki.

After digging through the wiki edit history, the entry for harpies was edited to say they are poisoned by theobromine and can eat raw meat in the same action. Since the raw meat eating _is_ true, it makes sense that the intention was theobromine poisoning as well.
image

Coinciding with the edit to the wiki entry, the harpy was changed to have an animal stomach, which you would intuitively think would make them be poisoned by theobromine, but this is not the case as reagents like that are metabolized in the heart.
image

These things taken together would make it seem to me that the intention is for them to be poisoned by it.

AveriV2:
oh damn i guess i knew wrong this whole time

deltanedas (approved):

PR #4045, opened 2025-07-09
Make eshield slightly cheaper

NoElkaTheGod:
## About the PR
Title. It's now worth 12 TC instead of 16.

## Why / Balance
It's not **that** good to be worth 16 TC. Yes, it makes you very tanky, but you have to sacrifice one hand for it, which basically means that you're locked to using pistols or melee weapons, which, in most cases, are significantly worse than the alternatives.

## Technical details
2 lines of YAML. I don't know why did it even have a discount, since nukie uplinks just don't give them, but I lowered taht too in case we ever add discounts to nukie uplinks.

## Media
It's literally 1 number, nothing to look at.

## Requirements
- [X] I have not tested any of the added content and changes.

**Changelog**
:cl:
- tweak: Energy shield price reduced from 16 to 12 telecrystals.

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -147,9 +147,9 @@
   productEntity: EnergyShield
   discountCategory: veryRareDiscounts
   discountDownTo:
-    Telecrystal: 12 # DeltaV - was 4
+    Telecrystal: 8 # DeltaV - was 4
   cost:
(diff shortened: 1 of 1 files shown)


deltanedas (approved):

Proxysseia:
direction review 24h

Proxysseia:
The price of eshields was a holdover from the past nukie meta where they ran around with eswords, jugs and eshields. Seeing how that meta was dealt with already, and especially considering the current subpar powerlevel of the eshield, a buff is warranted.

Direction approved.

PR #4048, opened 2025-07-11
Make Null Glare an actually useful influence

NoElkaTheGod:
## About the PR
If an empowered cosmic cultist uses Null Glare, it now flashes the target even if they have flash protection. Empowered duration was reduced from 10 to 6 seconds to compensate.

## Why / Balance
Null Glare is a pretty useless ability. Flash protection is very easy to get and almost completely negates it's effects. Breaking lights in an area is more of a cool visual than actual gameplay effect, and stunning borgs/IPCs is highly situational. Now it's actually useful in combat, since it would also affect security and random tiders with welding masks... if you are empowered, that is.

## Technical details
Some minor tweaks to FlashSystem.

## Media
My PC was begging for mercy, but I managed to record this anyway. Laggy as hell, but at least it does showcase the changes. 
Kinda.
It is physically painful to watch, but it's the best i can do.
https://github.com/user-attachments/assets/529a42ad-96c7-44f0-b606-51ad05dbc818

## Requirements

**Changelog**
:cl:
- tweak: Null Glare now ignores flash protection if the user is empowered.
- tweak: Null glare duration while empowered 10 -> 6 seconds.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Abilities/CosmicGlareSystem.cs
+++ b/Content.Server/_DV/CosmicCult/Abilities/CosmicGlareSystem.cs
@@ -68,7 +68,7 @@ public sealed class CosmicGlareSystem : EntitySystem
         {
             var targetEnt = GetEntity(target);
 
-            _flash.Flash(targetEnt, uid, args.Action, (float)uid.Comp.CosmicGlareDuration.TotalMilliseconds, uid.Comp.CosmicGlarePenalty, false, false, uid.Comp.CosmicGlareStun);
+            _flash.Flash(targetEnt, uid, args.Action, (float)uid.Comp.CosmicGlareDuration.TotalMilliseconds, uid.Comp.CosmicGlarePenalty, false, false, uid.Comp.CosmicGlareStun, uid.Comp.CosmicEmpowered);
 
             if (HasComp<BorgChassisComponent>(targetEnt) || HasComp<SiliconComponent>(targetEnt)) //For paralyzing borgs and IPCs specifically.
             {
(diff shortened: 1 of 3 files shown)


Content.Server/_DV/CosmicCult/Abilities/CosmicGlareSystem.cs
```diff
@@ -68,7 +68,7 @@ private void OnCosmicGlare(Entity<CosmicCultComponent> uid, ref EventCosmicGlare
         {
             var targetEnt = GetEntity(target);
 
-            _flash.Flash(targetEnt, uid, args.Action, (float)uid.Comp.CosmicGlareDuration.TotalMilliseconds, uid.Comp.CosmicGlarePenalty, false, false, uid.Comp.CosmicGlareStun);
+            _flash.Flash(targetEnt, uid, args.Action, (float)uid.Comp.CosmicGlareDuration.TotalMilliseconds, uid.Comp.CosmicGlarePenalty, false, false, uid.Comp.CosmicGlareStun, uid.Comp.CosmicEmpowered);
```
deltanedas: name the arg so its not another random bool

deltanedas (changes requested):

Proxysseia:
Direction Review 24h

deltanedas (approved):

Proxysseia:
The point of picking null glare was, and is, to stun ipcs and borgs, which has turned out to be too situational to be consistently useful. This change turns null glare into a decent combat option (the secoffs can still just hold down m1, they just might miss now), as long as the cultist is empowered. 

Direction approved.

PR #4059, opened 2025-07-16
Make cosmic cult progression smother

NoElkaTheGod:
## About the PR
Adjusted cosmic cult's entropy requirements for each stage to make the overall round flow smoother.
Previously, advancing from stage 1 to stage 2, and from stage 2 to stage 3 both required 1/2 of CosmicCultTargetConversionPercent, while advancing from stage 3 to the finale was a 2 minute timer, which made stage 3 basically non-existient.
Now, all the progression points (1 -> 2, 2 -> 3, 3 -> finale) require 1/3 of CosmicCultTargetConversionPercent.
The total amount of entropy that cult needs to collect remains unchanged, which means that stages 1 and 2 are slightly shorter than before (they are kinda boring anyway).

## Why / Balance
Stage 3 is now an actual progression stage for the round that lasts a significant amount of time. Theoreticaly, it should work as a middle ground between stage 2 (cult presence is known, but they mostly remain hidden) and the finale (team deathmatch between sec and the cult), which, according to feedback in the DeltaV discord server, is a much needed addition.
Additionaly, it makes vacuous spires actually useful, since now you still entropy in stage 3, where said spires are unlocked.

## Technical details
Minor tweaks to the MonumentSystem.

## Media
Beginning of stage 3.
изображение
After inserting enough entropy. At this point the monument can be beckoned as usual, starting the finale.
изображение
(Ignore the fact that I became a steward after the monument reached stage 2, that's just debug speedrunning)

## Requirements

**Changelog**
:cl:
- tweak: Cosmic cult's progression throughout the round should now feel much smoother.

### DIFF
--- a/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
+++ b/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
@@ -552,11 +552,8 @@ public sealed class CosmicCultRuleSystem : GameRuleSystem<CosmicCultRuleComponen
                 var cultistQuery = EntityQueryEnumerator<CosmicCultComponent>();
                 while (cultistQuery.MoveNext(out var cultist, out var cultistComp))
                 {
-                    var mins = timer.Minutes;
-                    var secs = timer.Seconds;
                     _antag.SendBriefing(cultist,
-                        Loc.GetString("cosmiccult-finale-autocall-briefing",
-                            ("minutesandseconds", $"{mins} minutes and {secs} seconds")),
+                        Loc.GetString("cosmiccult-finale-autocall-briefing"),
                         Color.FromHex("#4cabb3"),
                         _monumentAlert);
                 }
--- a/Content.Server/_DV/CosmicCult/MonumentSystem.cs
+++ b/Content.Server/_DV/CosmicCult/MonumentSystem.cs
@@ -280,10 +280,10 @@ public sealed class MonumentSystem : SharedMonumentSystem
             _appearance.SetData(ent, MonumentVisuals.FinaleReached, true);
     }
 
-    //note - these ar the thresholds for moving to the next tier
-    //so t1 -> 2 needs 20% of the crew
-    //t2 -> 3 needs 40%
-    //and t3 -> finale needs an extra 20 entropy
+    //note - these are the thresholds for moving to the next tier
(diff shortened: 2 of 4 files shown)


Proxysseia:
Direction review 72h

deltanedas (approved):

Proxysseia:
We agree that cults early tiers could do with some more crew interaction. 
However this PR does not really address this issue. As of right now tier 3 and finale are practically synonymous. This PR separates the tiers without adding anything else. This is not really anything except a nerf for cultists.
The vacuous spire was never meant as a way to generate entropy to progress to the next tier, it was meant as a way to top off on entropy to buy abilities, because as soon as t3 starts and you turn into a glowie, there will be no other way of getting entropy. 

We would absolutely welcome changes that add to the cults gameplay loop in a way that involves other departments and unconverted crew, like a more indepth item generation process, ways to directly interact with glimmer/anomalies or whatever else people can cook up.

Blocking cult from beckoning at tier 3 does not prevent the TDM, it just makes sure the cult loses. Decreasing the time it takes to raise your tier with nothing else adjusted does not prevent the TDM, it just means the cult has less time to prepare.

As such the PR is denied.

NoElkaTheGod:
You really underestimate how strong the cult can be at stage 3. Sure, this PR makes a cult win more difficult (which isn't even a bad thing IMO, cult wins way too often), but not impossible at all. At this point, the cult has access to:
- Instant AOE stun (astral nova)
- 6-ish seconds of full invincibility (vacuous imposition)
- Ability to convert literally anyone (glyph of truth), which helps massively with progression AND gives you a securitiy officer among your ranks
- The afforementioned spires, which may not have been initially designed for progression, but are never used in the current state anyway

Not to mention the fact that by this point there are at least 10 cultists, possibly several colossi, and maybe even the AI is at the cult's side.

Also important to note: if the cult won purely because security didn't have enough time to arm up and get to the monument (and they often don't, unless the location was already known and/or security was already armed), this would not be a satisfying win for anyone. 6 minutes are often just not enough to comprehend what's happening, get guns, go to the monument and stop it. And I'm saying all of this as someone who loves to play cult, by the way, not even a sec main or anything.

Perhaps there is a better solution, but if we keep looking for a perfect one, it'll just lead to more and more people hating the cult for the many flaws it has.

NoElkaTheGod:
Без имени

PR #4064, opened 2025-07-20
Give everybody the same TC - Again

kotobdev:
## About the PR
~~Scientists no longer have their -2 TC penalty, and all other Epistemics roles have had their TC penalty lowered by 2.~~ That, but also, all roles now get the same TC.

## Why / Balance
~~The reason for the penalty in the first place was `Protolathe, anomaly stuff, glimmer factor`. However:~~
- ~~Protolathe no longer exists~~
- ~~Anomalies don't provide any meaningful advantage to a traitor~~
- ~~Glimmer isn't an objective anymore, and similarly, provides little advantage~~
direction said so :)

## Technical details
certified yaml soldier

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: All jobs, again, now get the same starting TC.

### DIFF
--- a/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
@@ -109,13 +109,14 @@ public sealed class TraitorRuleSystem : GameRuleSystem<TraitorRuleComponent>
             Log.Debug($"MakeTraitor {ToPrettyString(traitor)} - Uplink start");
             // Calculate the amount of currency on the uplink.
             var startingBalance = component.StartingBalance;
+            /* DeltaV - everyone gets the same amount of TC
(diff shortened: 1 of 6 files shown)


deltanedas (dismissed):

Proxysseia:
Direction review, 24h

Proxysseia:
If you are up for it, we would be happy setting every syndies TC to 20, no matter what job. If you want to just go ahead with this PR then it is also approved.

kotobdev:
That sounds good too! I'll leave the advantage values changed just in case that gets reverted down the line, though. time to copy that one deltanedas PR

e: ~~why did it automatically request a review wait i didnt test it yet graaah~~ ok it works :D

deltanedas (approved):

PR #4066, opened 2025-07-21
Reincorporates Wizard

Colin-Tel:
## About the PR

Implements wizard in DeltaV both as a round start subantag and midround event.

Makes the wizard a random humanoid (as opposed to being the character you have selected).
Adds eight different open-ended objectives (they get up to three of them) that replace "survive" and "show off your skills." Thanks to Pikepeicrer for the sprite for the "Balance" objective.
Comments out the Cluwne Curse and Immovable Rod form spells.
Increased the cooldown time of the smite spell from 90 to 300 seconds.
## Why / Balance

Wizard is cool and we wanted it in some capacity.

The objectives are intended to offer interesting/silly things for the wizard to do which have more potential for RP.
## Technical details

All YAML
## Media

image

## Requirements

## Breaking changes

🤷 
**Changelog**

:cl: Colin-Tel, Pikepeicrer
- add: Wizards have been spotted around stations!


NoElkaTheGod:
Nice, but why the random appearence? Most other antags aren't random. The name is different anyway, so you can just say "nuh uh, it isn't me, I just look similar' if you want to.

Ideally this would be up to the player to decide, but that's WYCI, so why not leave it as is in the meantime?

deltanedas (approved):

deltanedas:
erm im not quirky one name felinid #37 im John Mittens

Colin-Tel:
> Nice, but why the random appearence? Most other antags aren't random. The name is different anyway, so you can just say "nuh uh, it isn't me, I just look similar' if you want to.
> 
> Ideally this would be up to the player to decide, but that's WYCI, so why not leave it as is in the meantime?

The initial concern it was addressing is to avoid potential metafriending situations and strange, potentially unwanted story arcs of your character suddenly being a wizard.

NoElkaTheGod:
> The initial concern it was addressing is to avoid potential metafriending situations and strange, potentially unwanted story arcs of your character suddenly being a wizard.

If the wizard wants to metafriend, they will just do it anyway. 
If they don't, they'll just go "erm no I'm not John Felenid" and then throw a fireball.
Pretty much the same would happen with the random appearence, except that now you **can't** do strange arcs of your character suddenly becoming a wizard (which is a thing that I'd like to do tbh).

NoElkaTheGod:
Also, again, **all** the humanoid antags use your character's aplearence, why is it only an issue with the wizard?

PR #4067, opened 2025-07-22
Mass Sleep rework

kotobdev:
## About the PR
Mass sleep has been reworked, adding both a doafter to the ability, and a warning to those who may be affected by it.

## Why / Balance
Mass sleep, presently, has absolutely zero counterplay. Psionically insulating gear is ugly, so that doesn't count. It's just an instant win button in a large number of encounters- stun everybody around you for 5 seconds and take their stuff. It sucks.

This change makes it so you have a chance to react before getting mass-slept. You are now alerted when somebody nearby starts using mass sleep, and moving out of the way or hurting the psion will stop the doafter. Additionally, while preparing mass sleep, the psion is slowed. So, if a tider rushes into a squad of nukies and tries to have their protag moment, they will be (most likely) gunned down before they can do anything. Unless they play it REALLY smart.

~~definitely not a mald PR trust me~~

## Technical details
a lot of midnight c# that looks correct enough :))))) mainly just referenced precognition stuff for this

## Media

https://github.com/user-attachments/assets/4a12f2ae-10c2-40ca-8302-381067c78e28

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Mass Sleep now has a cast time and visual warning.


NoElkaTheGod:
Kid named noösphericc zap:

deltanedas (changes requested):
change it to an instant action since target isnt used anymore

deltanedas (dismissed):

Proxysseia:
direction review 24h

Proxysseia (changes requested):
Direction here.
We appreciate the attempt to make psyonics reactable and not just a flat 'did you spend the 2tc for the helmet' check. We agree this is a good direction to take. 
However the PR does not quite succeeed at this. 
2 seconds channel time is, for SS14 standards, just about reactable. It is unreactable if the psyonics user approaches sentience and does not want to get shot: they can hide behind cover, only poke their head out and duck back in, they can stand behind another player, who takes more time to shoot through than the reaction time would allow for, or they can simply not move, with the reduced movement speed being the only indicator for them casting things at you.

We like this PR and want to see it in the game, after either it is more obvious *who* the caster is, or if the cast time is increased.


Also as a seperate note: Please try not to have accusatory changelogs, the second sentence is unnecessary.

kotobdev:
Thanks for the feedback! I'll update this in a few days, unfortunately a bit busy for the weekend

kotobdev:
okay, nerf time:
- cast time increased, 2.5 seconds -> 4 seconds
- warning radius accordingly doubled from 3 tiles to 6 tiles
- doafter bar is now visible

worth noting that the ability works only within 1.25 tiles of the user

https://github.com/user-attachments/assets/d0dc330a-fdd6-4046-95e7-d839045c6b75

deltanedas (changes requested):
use Entity instead of separate uid and comp for subscriptions

deltanedas (dismissed):

Proxysseia (approved):
lgtm

deltanedas (approved):

PR #4070, opened 2025-07-23
Make cosmic cult progression smoother take 2

NoElkaTheGod:
## About the PR
Continuation of #4059 with the [direction approved design document](https://discord.com/channels/968983104247185448/1396530273847607430) in mind.
Short list of all changes:
- Stage 3 now requires additional entropy before the finale can be activated.
- Stage 3 no longer gives cultists an astral brand. Instead, it gives a CosmicSubtleMark, which only shows that one is a cultist if they're examined without a mask on. Astral brand is still given when the finale starts.
- Malign rifts don't stop spawning on stage 3 anymore.

## Why / Balance
TL;DR: the jump in activity between stage 2 and stage 3 was bad, now it's better (hopefuly). See the [design document](https://discord.com/channels/968983104247185448/1396530273847607430) for more information.

## Technical details
Some minor tweaks on various systems and a new CosmicSubtleMarkComopnent.

## Media
A cultist examined on stage 3.
изображение
Same cultist, but with their eyes covered.
изображение
Beginning of stage 3 (image from #4059 because all the code is from there too)
изображение
The Unknown has been beckoned.
изображение

## Requirements

**Changelog**
:cl:
- tweak: Cosmic cultists can now hide their cult allegiance on stage 3 by covering their eyes. Beckoning the unknown would still make them valid, though.
- tweak: Cosmic cult now needs additional entropy (or converts) at stage 3 before they can activate the finale. The total amount of entropy (or converts) needed remains the same, so stages 1 and 2 are slightly faster now.
- tweak: Malign rifts no longer stop spawning when stage 3 is reached.


NoElkaTheGod:
No cool visuals, unfortunately, because I'm too bad at spriting. Maybe someone else will do it eventualy, but hey, at least it works.

Proxysseia:
no glowie eyes :(

Proxysseia:
oh yeah direction approved via design doc, forgot to mention

NoElkaTheGod:
> no glowie eyes :(

Not that hard to add them later. Gotta just make them look good, which isn't in my skill set.

Content.Server/_DV/CosmicCult/CosmicCultSystem.cs
```diff
@@ -110,6 +113,15 @@ private void OnCosmicCultExamined(Entity<CosmicCultExamineComponent> ent, ref Ex
     {
         args.PushMarkup(Loc.GetString(EntitySeesCult(args.Examiner) ? ent.Comp.CultistText : ent.Comp.OthersText));
     }
+
+    private void OnSubtleMarkExamined(Entity<CosmicSubtleMarkComponent> ent, ref ExaminedEvent args)
+    {
+        var ev = new SeeIdentityAttemptEvent();
+        RaiseLocalEvent(ent, ev);
+        if (ev.TotalCoverage == IdentityBlockerCoverage.EYES || ev.TotalCoverage == IdentityBlockerCoverage.FULL) return;
```
deltanedas: use & Eyes != 0 instead of doing this

deltanedas (changes requested):

Content.Server/_DV/CosmicCult/CosmicCultSystem.cs
NoElkaTheGod: Could you elaborate please? I don't quite undertand what do you mean.

NoElkaTheGod: @deltanedas ?

NoElkaTheGod: Either I don't know something important, or one of us misunderstood something, because like... you can't compare an enum value to an int in this situation? It will just always be false, because IdentityBlockerCoverage.EYES is 2?

NoElkaTheGod:
Still no idea how to adress the code review btw. 3 hours of googling didn't help :/

NoElkaTheGod:
I, like, seriously have no idea what to do in that situation

NoElkaTheGod:
Actually, I have a solution

deltanedas:
its a bitfield you use bit operations on it
see IdentitySystem

NoElkaTheGod:
Oh, now I understand what you're talking about. Wouldn't a HasFlag() method work as well though? Looks a lot more intuitive to me.

NoElkaTheGod:
Well, it does work. I can re-implement it with bit operations if you really want it that way, but I personaly think that this is good too.

Also thanks for the explaination, always nice to learn something new.

deltanedas:
yes this is just a higher level version its good

deltanedas (approved):

NoElkaTheGod:
May as well merge tho? It's been direction approved for almost a week now.

PR #4076, opened 2025-07-24
Remove psionic zap from psionic pool

Radezolid:
## About the PR
Title.

## Why / Balance
It has no counter-play but to be wearing psionic insulative clothing.

## Technical details
N/A

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- remove: Removed the noospheric zap from the psionic powers pool.


NoElkaTheGod:
I'd rather have #4079 than this tbh

Proxysseia:
direction review 24h

Proxysseia:
Zap has room in it for interesting ideas to be explored, outright removal should be a last option. We are not out of options yet, #4079 is right here, even as a slight change.

As such this PR is denied.

PR #4079, opened 2025-07-24
Make Noospheric Zap an actual electric zap

kotobdev:
## About the PR
Noospheric Zap, instead of attempting to mimic a zap, is now an actual lightning zap. Notably, this makes people wearing insulated gloves immune to them.

## Why / Balance
Making zap not work through insuls keeps it from being the ultimate validhunting tool it is right now- many solo antagonists (loneop, ninja, traitors if they're locked in) will just be immune to it. However, security officers will usually not be, unless they're either wearing a hardsuit or are the HOS/warden.
Another effect this has is making the zap do damage- this should dissuade people from zapping random crew in the hallways, and then saying "it's not assault because it doesn't hurt them" when pulled aside by sec.
There's a lot of small nuances with using actual lightning, but those are the main ones.

Also, idea that's not in this PR but might be fitting: let the zap be used on non-potentially-psychic entities? It would be funny to see a psionic engineer charging a tesla coil. Additionally, it would just make the ability more fun outside of combat. Would like to see what direction thinks of that

## Technical details
becoming a true c# cadet

## Media
https://github.com/user-attachments/assets/ff87e913-39c5-4baf-bbd1-0e6078841018

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Noospheric Zap now takes the form of an actual electric zap, and is thwarted by insuls.

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/NoosphericZapPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/NoosphericZapPowerSystem.cs
@@ -51,16 +51,15 @@ namespace Content.Server.Abilities.Psionics
 
         private void OnPowerUsed(NoosphericZapPowerActionEvent args)
         {
-            _psionics.LogPowerUsed(args.Performer, "noospheric zap");
-            
-            if (HasComp<PotentialPsionicComponent>(args.Target))
+            if (!HasComp<PotentialPsionicComponent>(args.Target))
                 return;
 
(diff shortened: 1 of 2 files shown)


deltanedas (approved):

Proxysseia:
Direction Review 24h

Proxysseia:
Pretty reasonable change that makes zapping people for fun even more obvious assault than it already was, electricity makes this a more unique ability and even opens up more fun interpretations later, like shocking machines or doors or whatever, wyci.

Direction approved.

PR #4082, opened 2025-07-25
Modifies Probers to be more dynamic

TehFlaminTaco:
## About the PR

Modified Prober behaviour to be more dynamic, both to interact with and interacting with Glimmer / Research.

## Why / Balance

As it stands, Glimmer Probes are un-interesting to work with. Turn on, gain research, turn off before it gets to 500, destroy it if it does.
The proposed changes stop the instant lockout at 500 and improves effectiveness of the machine at higher glimmer, promoting more risk-reward play relating to glimmer probing.

## Technical details

Added a LockProbes Glimmer Event, which causes a random number (at least 1) of probes to Lock per former 500+ glimmer Locking Behaviour. These cannot be unlocked, though are only a risk when attempting to remove them.
Probes no-longer generate a flat number of Glimmer and Research, instead they scale roughly with the current glimmer value. With current implementation, they are identical to their previous implementation at 100 glimmer. Of course, the increased glimmer generation means they can quickly run-away.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Modified Glimmer Probers, they now generate more Glimmer and Research based on the current Glimmer.
- tweak: Glimmer Probers no-longer automatically lock themselves at 500 glimmer.
- add: Added a new Glimmer Event which locks some Glimmer Probes.

### DIFF
--- a/Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
+++ b/Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
@@ -1,4 +1,5 @@
-using System.Linq;
+using Content.Server._DV.StationEvents.Components;
+using Content.Server.Power.Components;
 using Content.Server.Psionics.Glimmer;
 using Content.Server.Station.Systems;
 using Content.Server.StationEvents;
@@ -8,10 +9,9 @@ using Content.Shared.Abilities.Psionics;
 using Content.Shared.GameTicking.Components;
 using Content.Shared.NPC.Components;
 using Content.Shared.Psionics.Glimmer;
-using Robust.Shared.Random;
 using Robust.Shared.Map;
-using Content.Server._DV.StationEvents.Components;
-using Content.Server.Power.Components;
+using Robust.Shared.Random;
+using System.Linq;
 
 namespace Content.Server._DV.StationEvents.Events;
 
--- a/Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
+++ b/Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
@@ -1,0 +1,61 @@
+using System.Linq;
+using Content.Server.Psionics.Glimmer;
+using Content.Server.Station.Systems;
+using Content.Server.StationEvents;
+using Content.Server.StationEvents.Components;
+using Content.Server.StationEvents.Events;
+using Content.Shared.Abilities.Psionics;
+using Content.Shared.GameTicking.Components;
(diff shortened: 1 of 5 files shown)


Proxysseia:
Direction Review 24h

deltanedas (changes requested):
dont need tags for datafields just use "[DataField]`
also dont need dv comments in nyano files, might as well refactor them and move them to _DV too

TehFlaminTaco:
Cleaned up per suggestions. Not refactoring into _DV. Although that is probably a good idea in the long run, out of scope of this PR (And my confidence in my skill)

Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
```diff
@@ -0,0 +1,61 @@
...
+using Content.Server.StationEvents.Components;
+using Content.Server.StationEvents.Events;
+using Content.Shared.Abilities.Psionics;
+using Content.Shared.GameTicking.Components;
+using Content.Shared.NPC.Components;
+using Content.Shared.Psionics.Glimmer;
+using Robust.Shared.Random;
+using Robust.Shared.Map;
+using Content.Server._DV.StationEvents.Components;
+using Content.Server.Power.Components;
```
deltanedas: sort them

deltanedas (changes requested):

Content.Server/_DV/StationEvents/Events/LockProbersRule.cs
TehFlaminTaco: Done

deltanedas (approved):

LadyDanger2743:
Director here! 

A few things to keep in mind for possible future PRs or tweaks to this PR:

- The Glimmer system, in and of itself, kind of stinks. Conceptually, a risk/reward system is great; in practice, Epistemics tends to blow up due to mistakes or the actions of new epistemicians who don't know that doing science can cause the department to blow up. If you or anyone reading this have any ideas on how to improve glimmer, by all means, feel free to discuss them in #contributors on Discord. 
- This could have the unintended consequence of giving people even more reason to leave the Probers off, if they can always turn into bombs. Not sure if there's really anything we can do to stop that, or if a fix could possibly be tied to a Glimmer overhaul. 

Other than anything that might help either of the above points, this is good to go! LGTM!

deltanedas (approved):

PR #4083, opened 2025-07-25
Improve comic cult evac behavior

NoElkaTheGod:
## About the PR
1. If evac is on it's way to the station when finale begins, evac is automaticaly recalled.
2. The monument shuts down the moment evac docks to the station, instead of when evac leaves.
3. Monument still works normaly when it shuts down, but the finale can no longer be activated.

## Why / Balance
It kinda sucks when evac arrives during the finale, since you now have to sacrifice your end-of-round RP and instead go fight the cultists. The previous system didn't fully prevent that kind of situation, so I adjusted it a bit. Now you either activate the finale before the evac docks, which would recall it, or you don't and get a minor win at best.
Also, evac now simply makes finale unavailable instead of completely bricking the monument, so any cultists that are left behind at least get to have some fun in the end.

## Technical details
I have probably just merge conflict'd my other PR by doing this.

## Media
You won't believe me but it works as intended (it doesn't actually get recalled twice that's me fucking around)
изображение

## Requirements

**Changelog**
:cl:
- tweak: Monument no longer gets bricked when evac departs.
- tweak: Cosmic cult's finale can no longer be triggered when evac docks to the station.
- tweak: If evac was on it's way to the station when cosmic cult's finale is activated, it gets automatically recalled.

### DIFF
--- a/Content.Server/Shuttles/Systems/EmergencyShuttleSystem.cs
+++ b/Content.Server/Shuttles/Systems/EmergencyShuttleSystem.cs
@@ -18,6 +18,7 @@ using Content.Server.Station.Components;
 using Content.Server.Station.Events;
 using Content.Server.Station.Systems;
 using Content.Shared._DV.CustomObjectiveSummary; // DeltaV
+using Content.Server._DV.Shuttles.Events; // DeltaV
 using Content.Shared.Access.Systems;
 using Content.Shared.CCVar;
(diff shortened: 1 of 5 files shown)


deltanedas (approved):

deltanedas:
blob does this too its peak when command gets trolled trying to avoid an antag

PR #4090, opened 2025-07-28
Telegnosis Suffocation Nerf

TehFlaminTaco:
## About the PR

Telegnosis has been nerfed, causing damage if used too long and too far away. Additionally, Telegnosis users no-longer appear SSD.

## Why / Balance

Telegnosis in its current state has a dominant effect on any rounds in which its Psions are present, being able to frequently immediately spot and track antagonists that would otherwise have the advantage of surprise or stealth. This change is designed to punish players who hide away and use Telegnosis all round, whilst still keeping Telegnosis as a powerful information gathering tool.

This should _only_ have any considerable impact on Telegnosis users who refuse to leave their projection, as even distant use shouldn't be very harmful if done in moderation.

Making Telegnosis users no-longer appear SSD has two intentions
 * Users are no-longer accidently cryo'd.
 * Users are no-longer incorrectly assumed to be "untouchable" by other players.

## Technical details

 Using Telegnosis now makes breathing harder.
 - If the Projection is less than 7 tiles away, there is no effect.
 - Between 7 and 57 tiles, there is a linear drop-off in how much air is taken in with each breath.
 - Past 57 tiles, no air is taken in with each breath.

This slowly causes suffocation damage with prolonged use but can be prevented by regularly returning to your body.

Telegnosis's sensor has been slowed down slightly to make the threat of distance more present. This should still be faster than most humans.

Telegnosis users no-longer appear SSD and instead have examine text showing that they are not present. (See media)

## Media

image

## Requirements

## Breaking changes

N/A

**Changelog**

:cl: TehFlaminTaco
- tweak: Telegnosis can now cause suffocation if overused
- tweak: Telegnosis users are no-longer considered SSD

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
@@ -22,7 +22,7 @@ namespace Content.Server.Abilities.Psionics
         [Dependency] private readonly SharedActionsSystem _actions = default!;
         [Dependency] private readonly MindSwapPowerSystem _mindSwap = default!;
         [Dependency] private readonly SharedPsionicAbilitiesSystem _psionics = default!;
-        [Dependency] private readonly AtmosphereSystem _atmosSys = default!;
+        [Dependency] private readonly AtmosphereSystem _atmos = default!;
         [Dependency] private readonly TransformSystem _transform = default!;
         [Dependency] private readonly SharedMindSystem _mind = default!;
 
@@ -39,8 +39,8 @@ namespace Content.Server.Abilities.Psionics
 
         private void OnInit(EntityUid uid, TelegnosisPowerComponent component, ComponentInit args)
         {
-            _actions.AddAction(uid, ref component.TelegnosisActionEntity, component.TelegnosisActionId );
-            _actions.TryGetActionData( component.TelegnosisActionEntity, out var actionData );
+            _actions.AddAction(uid, ref component.TelegnosisActionEntity, component.TelegnosisActionId);
+            _actions.TryGetActionData(component.TelegnosisActionEntity, out var actionData);
             if (actionData is { UseDelay: not null })
                 _actions.StartUseDelay(component.TelegnosisActionEntity);
             if (TryComp<PsionicComponent>(uid, out var psionic) && psionic.PsionicAbility == null)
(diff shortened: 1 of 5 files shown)


Proxysseia:
Direction Review 24h

NoElkaTheGod:
I think it would be kinda cool if you also couldn't return to the body unless you're close

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
```diff
@@ -53,15 +62,57 @@ private void OnShutdown(EntityUid uid, TelegnosisPowerComponent component, Compo
...
+        {
+            if (!TryComp<MindSwappedComponent>(uid, out var mindSwapped) ||
+                !TryComp<TelegnosticProjectionComponent>(mindSwapped.OriginalEntity, out var projection))
+            {
+                return (default, null);
+            }
+            return (mindSwapped.OriginalEntity, projection);
+        }
+
+        private void OnInhaleLocation(EntityUid uid, TelegnosisPowerComponent component, ref InhaleLocationEvent args)
```
deltanedas: use Entity for subscriptions

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
```diff
@@ -53,15 +62,57 @@ private void OnShutdown(EntityUid uid, TelegnosisPowerComponent component, Compo
...
+
+        public (EntityUid uid, TelegnosticProjectionComponent? comp) GetCasterProjection(EntityUid uid)
+        {
+            if (!TryComp<MindSwappedComponent>(uid, out var mindSwapped) ||
+                !TryComp<TelegnosticProjectionComponent>(mindSwapped.OriginalEntity, out var projection))
+            {
+                return (default, null);
+            }
+            return (mindSwapped.OriginalEntity, projection);
+        }
```
deltanedas: change it to just return EntityUid? and check HasComp for the original entity since the component is never used

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
```diff
@@ -1,23 +1,30 @@
...
     public sealed class TelegnosisPowerSystem : SharedTelegnosisPowerSystem
     {
-        [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
-        [Dependency] private readonly StatusEffectsSystem _statusEffects = default!;
         [Dependency] private readonly SharedActionsSystem _actions = default!;
         [Dependency] private readonly MindSwapPowerSystem _mindSwap = default!;
         [Dependency] private readonly SharedPsionicAbilitiesSystem _psionics = default!;
-        [Dependency] private readonly IGameTiming _gameTiming = default!;
-        [Dependency] private readonly MindSystem _mindSystem = default!;
+        [Dependency] private readonly AtmosphereSystem _atmosSys = default!;
```
deltanedas: rename to _atmos

Content.Shared/Mind/SharedMindSystem.cs
```diff
@@ -652,6 +652,23 @@ public HashSet<Entity<MindComponent>> GetAliveHumans(EntityUid? exclude = null,
...
+    /// <param name="showExamineInfo">True to show examine information, false to hide it.</param>
+    public void ShowExamineInfo(EntityUid uid, bool showExamineInfo)
+    {
+        if (!TryComp<MindContainerComponent>(uid, out var mindContainer))
+            return;
+
+        mindContainer.ShowExamineInfo = showExamineInfo;
+        Dirty(uid, mindContainer);
+    }
+    // End Delta-V Changes
```
deltanedas: move this to a partial system in _DV

deltanedas (changes requested):

Content.Shared/Mind/SharedMindSystem.cs
TehFlaminTaco: Didn't notice the base class was partial, easy fix.

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/TelegnosisPowerSystem.cs
TehFlaminTaco: Whoops, grandfathered design from an earlier design. Fixed.

TehFlaminTaco:
> I think it would be kinda cool if you also couldn't return to the body unless you're close

I understand the idea, but I personally wouldn't want to accidently trap a user outside their body because it's been moved (Or they've just lost it). I hope the cooldown should be punishing enough.

HTMLSystem:
Question, does the user get any sort of popup alerting them of the asphyx damage? Just wondering because for new players and old ones it could be frusterating to start randomly dying and you don't know the reason why.

TehFlaminTaco:
There's no warning currently. It's ultimately up to direction; but I believe that psionics being confusing and dangerous to most players is ideal. Most new players would also experiment a bit with limitations first, which should reveal that it's causing oxyloss pretty quickly.

HTMLSystem:
Alright. As a warning, dying while being in the projection does seperate your body permanently, so.

TehFlaminTaco:
This does not modify that behaviour. I am happy to patch it if that's what Direction wants.

deltanedas (approved):

Proxysseia:
Current Telegnosis makes it too easy to indefinitely linger and eavesdrop on conversations. While eavesdropping is not bad, doing so indefinitely is. There are some concerns about dying while in telegnosis RRing you but sending your soul out of your body should come with some risks.

Direction approved.

Content.Shared/Mind/SharedMindSystem.cs
deltanedas: even if it wasnt you can just make it partial

PR #4092, opened 2025-07-29
Anglish accent

Colin-Tel:
## About the PR

Adds a trait called "Anglish tung" that is an accent that replaces a lot of words with Anglish counterparts, as well as replaced any instance of lowercase or uppercase "th" with "þ" and "Þ" respectively.

**What's Anglish?**
Anglish is English without loanwords from Old French, Latin, Greek, etc, with replacements using Old English, German, and Norse words. While this accent doesn't have *every* loanword replaced, it's got a lot. There is also some difficulty with words that can be used in many ways, one such example being "second." In this accent, "second" always changes into "moment," while "2nd" always translates into "twoth." The idea behind this being that, well, since it's a simple accent system, strange instances like these are inevitable (or should I say "unformithely?")

**How can I find out more words in Anglish?**
Many words we use in English (or on Space Station for that matter) have Germanic/Old English/Norse roots, and are often "already Anglish" (e.g. tider). Using resources like etymonline or wiktionary are often very handy to identify the etymology of words and their roots to see if they're Anglish or not.

Besides that, the resources I used to compose this PR are as follows:
https://pure-english.github.io/dictionary/
https://anglish.fandom.com/wiki/English_Wordbook/A (the last letter in the URL corresponds to the beginning letter of the English word you're looking up)
https://theanglishtimes.com/ (as well as their Discord)

Some examples:
```
general = whole
really = soothly
hydrogen = waterstuff
grain = corn
ID = whoness
future = to-be
no erp = give up your lewd playwork
tesloose = the spark-ball fled from its holdfast
```
## Why / Balance

It's a fun accent idea that people have said looks interesting enough.
## Technical details

FTL, YAML, and some C# that was repurposed from the `FrenchAccentSystem` code. Over 1,100 word replacements are present in this accent.
## Media

image
image

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Eked the Anglish tung tie. In other words, added a trait that lets you talk like an Anglo Saxon with over 1,100 word replacements.


TehFlaminTaco (dismissed):
Needs to add the trait to clone.yml to ensure it copies over

Colin-Tel:
> Needs to add the trait to clone.yml to ensure it copies over

image

is this even possible with the component being commented out?

TehFlaminTaco:
Great question. No idea if it's worth doing for future proofing or not

Colin-Tel:
> Great question. No idea if it's worth doing for future proofing or not

I don't think we even have any clothing that adds accents anymore, so I feel like we could uncomment it

Correction: we have the cluwne mask (unobtainable) and the muzzle mask (I don't think anyone's ever used it), that's it.

I'm fine uncommenting the line with the understanding that the potential situation could happen where someone is cloned while muzzled, which I think is an edge case.

TehFlaminTaco:
From what I can see, `clones.yml` also controls which trait components are copied over, not just for equipment

Colin-Tel:
If you could provide a suggestion of how I'd add it, I'll do it ASAP

Colin-Tel:
Nevermind

Colin-Tel:
drafting this because some species aren't included and also I need to replace the eth with thorn after more research 🥴

Vapetastic-Gaming:
Approved by unanimous direction approval. The "time-dilating wormhole" did not spawn this time.

Content.Server/_DV/Speech/Components/AnglishAccentComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Content.Server.Speech.EntitySystems;
+
+namespace Content.Server.Speech.Components;
+
+/// <summary>
+/// Anglish tung, the component replaces some words with French, Latin, and other etymologies with Germanic and Norse variations.
+/// </summary>
+[RegisterComponent]
+[Access(typeof(AnglishAccentSystem))]
+public sealed partial class AnglishAccentComponent : Component {}
```
Toby222:
```suggestion
public sealed partial class AnglishAccentComponent : Component;
```

Content.Server/_DV/Speech/EntitySystems/AnglishAccentSystem.cs
```diff
@@ -0,0 +1,43 @@
...
+/// <summary>
+/// System that replaces certain vocabulary to be similar to Anglish.
+/// </summary>
+public sealed class AnglishAccentSystem : EntitySystem
+{
+    [Dependency] private readonly ReplacementAccentSystem _replacement = default!;
+
+    private static readonly Regex RegexThLowercase = new(@"th");
+    private static readonly Regex RegexThUppercaseShort = new(@"Th");
+    private static readonly Regex RegexThUppercaseLong = new(@"TH");
```
Toby222: You should not use regex for this, they're just plain strings

Content.Server/_DV/Speech/EntitySystems/AnglishAccentSystem.cs
```diff
@@ -0,0 +1,43 @@
...
+    public string Accentuate(string message, AnglishAccentComponent component)
+    {
+        var msg = message;
+
+        msg = _replacement.ApplyReplacements(msg, "anglish");
+
+        // replaces th with a "thorn"
+        msg = RegexThLowercase.Replace(msg, "þ");
+        msg = RegexThUppercaseShort.Replace(msg, "Þ");
+        msg = RegexThUppercaseLong.Replace(msg, "Þ");
```
Toby222:
```suggestion
        msg = _replacement
            .ApplyReplacements(msg, "anglish")
            .Replace("th", "þ")
            .Replace("Th", "Þ")
            .Replace("TH", "Þ");
```

Resources/Prototypes/_DV/Traits/neutral.yml
```diff
@@ -45,3 +45,13 @@
...
   components:
     - type: IrishAccent
+
+# Anglish, for the Beowulf in you
+- type: trait
+  id: AnglishAccent
+  name: trait-anglish-name
+  description: trait-anglish-desc
+  category: SpeechTraits
+  cost: 2 # Changed weight to 2 in accordance with other accents
```
Toby222: Why is this comment here
Who did this

Toby222 (changes requested):
did not look at the 1100+ replacements all typos are on you

Resources/Prototypes/_DV/Traits/neutral.yml
Colin-Tel: Literally on every accent when specifying cost

Toby222: so it's copypasta :blunt:
You didn't change anything, and change comments don't belong in _DV anyways

Colin-Tel:
Drafting to get this up to date with the new traits system, probably won't take long 🤞

Toby222 (approved):

PR #4093, opened 2025-07-29
Glimmer event - psionic nosebleed

kotobdev:
## About the PR
A new glimmer event is now possible past 350 glimmer, which gives all psionics a sudden nosebleed.

## Why / Balance
Now that the prober lockout is gone, I think glimmer events are going to be generally more prominent. Thus, this is the first of hopefully many attempts at introducing some more variety to the existing pool. Ideally, further attempts will be somewhat more interesting, but it's nice to have some subtler additions like this too. I think.
This event only inflicts 2.5 stacks of bleed- enough to be a bit alarming visually, but not enough to inflict any actual damage. More of a roleplay event, if anything.

## Technical details
apparently psionic.ftl is, in fact, the correct place to put glimmer strings

## Media
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Psionics may now get nosebleeds at higher glimmer levels.

### DIFF
--- a/Content.Server/_DV/StationEvents/Events/PsionicNosebleedRule.cs
+++ b/Content.Server/_DV/StationEvents/Events/PsionicNosebleedRule.cs
@@ -1,0 +1,38 @@
+using Content.Server._DV.StationEvents.Components;
+using Content.Server.Body.Systems;
+using Content.Server.Psionics.Glimmer;
+using Content.Server.Station.Systems;
+using Content.Server.StationEvents.Events;
+using Content.Shared.Abilities.Psionics;
+using Content.Shared.GameTicking.Components;
+using Content.Shared.Mobs.Components;
+using Content.Shared.Mobs.Systems;
+using Content.Shared.Popups;
+using Content.Shared.Psionics.Glimmer;
+
+namespace Content.Server._DV.StationEvents.Events;
+
+public sealed class PsionicNosebleedRule : StationEventSystem<PsionicNosebleedRuleComponent>
+{
+    [Dependency] private readonly GlimmerSystem _glimmer = default!;
+    [Dependency] private readonly StationSystem _stationSystem = default!;
+    [Dependency] private readonly GlimmerReactiveSystem _glimmerReactiveSystem = default!;
+    [Dependency] private readonly MobStateSystem _mobStateSystem = default!;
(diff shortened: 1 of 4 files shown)


Content.Server/_DV/StationEvents/Events/PsionicNosebleedRule.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Content.Server._DV.StationEvents.Events;
+
+public sealed class PsionicNosebleedRule : StationEventSystem<PsionicNosebleedRuleComponent>
+{
+    [Dependency] private readonly GlimmerSystem _glimmer = default!;
+    [Dependency] private readonly StationSystem _stationSystem = default!;
+    [Dependency] private readonly GlimmerReactiveSystem _glimmerReactiveSystem = default!;
+    [Dependency] private readonly MobStateSystem _mobStateSystem = default!;
+    [Dependency] private readonly BloodstreamSystem _bloodstreamSystem = default!;
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
```
deltanedas: most of these are unused
also rename them to not have System in the field name its already obvious
_mob _bloodstream etc

Content.Server/_DV/StationEvents/Events/PsionicNosebleedRule.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    protected override void Started(EntityUid uid, PsionicNosebleedRuleComponent comp, GameRuleComponent gameRule, GameRuleStartedEvent args)
+    {
+        base.Started(uid, comp, gameRule, args);
+
+        var query = EntityQueryEnumerator<PsionicComponent, MobStateComponent>();
+        while (query.MoveNext(out var psion, out _, out _))
+        {
+            if (_mobStateSystem.IsAlive(psion) && !HasComp<PsionicInsulationComponent>(psion))
```
deltanedas: pass the MobStateComponent to IsAlive instead of ignoring its value

deltanedas (changes requested):

Proxysseia:
Direction Review 24h

deltanedas (approved):

Proxysseia:
Making glimmer more present on station, especially in ways that are not overly disruptive and do not harm people who are not psyonic is good and we want to see more of it. Great RP addition to the game.
Direction approved.

PR #4094, opened 2025-07-29
Don't lift mindwiping ROE on code white

NoElkaTheGod:
## About the PR
Mindwiping is no longer free on code white, epistemics have to follow the standart ROE.

## Why / Balance
The main issue here is the cosmic cult. You can't just mindwipe someone for "being a cultist". However, you can just set alert level to white, and now you can deconvert all you want without consequences.
This was originaly introduced to combat psionics on code white, because using telegnosis is not a crime, and you can't mindwipte for that. This is no longer necessary, since we have noospheric tampering now, and using abilities under high glimmer could be seen as that.
ROE is still lifted when on code octarine, because obviously it is.

## Technical details
No

## Media
No

## Requirements
No

## Breaking changes
No

**Changelog**
:cl:
- tweak: ROE on mindwiping is no longer lifted on code white. Try to see if you can charge the offender with noöspheric tampering (or any other crime in which psionic or otherwordly powers were used)

### DIFF
--- a/.wiki/_DV/Laws/AlertProcedure.txt
+++ b/.wiki/_DV/Laws/AlertProcedure.txt
@@ -169,7 +169,7 @@ Elevated alert status. ''The station is suffering dangerously high levels of gli
 | style="border: 1px solid #000000;" | EVA suits heavily advised for engineers. Distribute emergency internals to crew. Prepare for detonation of [[Glimmer#Probers|probers]].
 |-
 | style="border: 1px solid #000000;" | Rules of Engagement
-| style="border: 1px solid #000000;" | Engage with minimal force required; ROE on mindwiping is lifted and may be performed by the Psionic Mantis, Chaplain or Mystagogue at their discretion.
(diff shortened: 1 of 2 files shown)


Proxysseia:
Direction review 24h

Proxysseia:
> 

thanks, i will :3

Proxysseia:
As it stands, we see the RoE on mindbreaking on white as a necessary part of code white procedure for non cult rounds. We do however recognize that code white on cult rounds can lead to unfun interactions. The source of these interactions is the way how deconversion works. 
In combination with #4111 we would like to propose that the fuel required to use the censer changes from holy water to mindbreaker toxin. In case chemists get incapacitated or converted, we'd also like to see the ability to buy mindbreaker toxin via cargo. Both fuel cost and cargo price/amount are left up to you, if you want to do it.

The changes to RoE are not approved.

NoElkaTheGod:
Got you, boss, the PR is kill

PR #4100, opened 2025-07-30
Prevent accidental activation of coscult finale

NoElkaTheGod:
## About the PR
When beckoning the Unknown, instead of some flavor text, the popup now states "Doing this will activate the final stage! Are you sure?" with big red text.

## Why / Balance
Hopefuly should prevent a clueless cultist from activating the finale too early.

## Technical details
2 line ops.

## Media
No.

## Requirements
- [X] I do not have to test any future added content and changes.
- [X] I am absolutely certain that nobody reads those, as confirmed by a previous PR of mine.

**Changelog**
Too minor.


NoElkaTheGod:
Alright wait I am cooked. Will repoen after making it actually work.

NoElkaTheGod:
Alright, should actually work now

PR #4102, opened 2025-07-30
Remove do-after on null fragmentation

NoElkaTheGod:
## About the PR
Null fragmentation no longer has a do-after. A popup was added instead to avoid too much confusion.

## Why / Balance
1. It was bugged, and un-empowered the cultist even if the do-after failed, which is kinda not nice.
2. It didn't really serve much purpose. Upload console can't move anyway. Brogs would often just not understand what's going on until it's too late (or you can just order them not to move).
3. It's something that Kazne did for the next update, but it doesn't seem like the next update is ever coming, so might as well do it myself.

## Technical details
Removed do-after, added popup. Also replaced `ent` with `args.Target` in `OnFragmentBorg`. How the hell did it even work before? `ent` is the cultist??? What magic is this?????

## Media
Not really.

## Requirements

**Changelog**
:cl:
- tweak: Null fragmentation is now instantaneous.


deltanedas:
it worked because the event target and the handler were both checking the cultist

NoElkaTheGod:
Oh, okay. Makes sense. Kinda.

NoElkaTheGod:
Oh, also, you can't corrupt a crit borg, so the only way to properly do this is to either use null glare (which is bad cuz it's kinda useless without epowerment) or spam it with flashes to the face (also kinda silly).

NoElkaTheGod:
Apparently this does not require direction review. @deltanedas any issues here?

deltanedas (changes requested):
using it on an empty chassis wastes the ability, add cancelled to the event and cancel it if the borg chassis has no mind

Content.Server/_DV/CosmicCult/Abilities/CosmicFragmentationSystem.cs
```diff
@@ -61,43 +58,27 @@ private void OnCosmicFragmentation(Entity<CosmicCultComponent> ent, ref EventCos
...
-            BreakOnDamage = true,
-            BreakOnMove = true,
-            BreakOnDropItem = true,
-        };
+        var evt = new MalignFragmentationEvent(ent, args.Target);
+        RaiseLocalEvent(args.Target, ref evt);
+        if (evt.Cancelled) return;
         args.Handled = true;
-        _doAfter.TryStartDoAfter(doargs);
+        _popup.PopupEntity(Loc.GetString(Loc.GetString("cosmicability-fragmentation-success"), ("user", ent), ("target", args.Target)), ent, PopupType.MediumCaution);
```
deltanedas: dont need nested loc

deltanedas (changes requested):

Content.Server/_DV/CosmicCult/Abilities/CosmicFragmentationSystem.cs
NoElkaTheGod: Woops my bad

deltanedas (approved):

PR #4105, opened 2025-07-30
Glimmer event - random animation

kotobdev:
## About the PR
A new glimmer event is now possible past 500 glimmer, which temporarily animates a number of random objects on-station.

## Why / Balance
Past 500 glimmer is (imo) roughly the point where things are supposed to start to get really weird. So, here's a weird event to add to the pool.
The event animates between 7-10 objects for somewhere between 7-15 seconds each. From my limited testing, if I wandered in the hallway in front of Botany on Glacier when the event triggered, I'd have a bit less then a 50% chance of seeing something happen. So, for a crew consisting of more then one person, I think the current numbers would be pretty noticeable, but not TOO disruptive.

## Technical details
the EntityQueryEnumerator feels evil to me, but it seems to run fine
code's largely modified from the random sentience and artifact animation stuff

## Media
(disclaimer: out of the like 50 times i activated the event, this was the most interesting one, most were not this chaotic)

https://github.com/user-attachments/assets/e77437db-c698-4254-b757-5c8dab38a7e0

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Random objects may become animated at high glimmer levels.

### DIFF
--- a/Content.Server/_DV/StationEvents/Events/RandomAnimationRule.cs
+++ b/Content.Server/_DV/StationEvents/Events/RandomAnimationRule.cs
@@ -16,6 +16,15 @@ public sealed class RandomAnimationRule : StationEventSystem<RandomAnimationRule
     [Dependency] private readonly RevenantAnimatedSystem _revenantAnimated = default!;
     [Dependency] private readonly SharedContainerSystem _container = default!;
 
+    private EntityQuery<ItemComponent> _itemQuery;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        _itemQuery = GetEntityQuery<ItemComponent>();
+    }
+
     protected override void Started(EntityUid uid, RandomAnimationRuleComponent comp, GameRuleComponent gameRule, GameRuleStartedEvent args)
     {
         base.Started(uid, comp, gameRule, args);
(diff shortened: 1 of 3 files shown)


Avalon-Proto:
Cant wait to see half the armory get possessed

NoElkaTheGod:
Now that I think about it, making the normal random sentience event tied to glimmer would be kinda nice too.

Proxysseia:
Direction 24h

kotobdev:
> Now that I think about it, making the normal random sentience event tied to glimmer would be kinda nice too.

It already is!!

Avalon-Proto:
> Now that I think about it, making the normal random sentience event tied to glimmer would be kinda nice too.

It already is

NoElkaTheGod:
Shit, my bad

Content.Server/_DV/StationEvents/Events/RandomAnimationRule.cs
```diff
@@ -0,0 +1,47 @@
...
+            return;
+
+        var targetList = new List<Entity<AnimateableComponent>>();
+        var query = EntityQueryEnumerator<AnimateableComponent>();
+        while (query.MoveNext(out var ent, out var animateable))
+        {
+            if (StationSystem.GetOwningStation(ent) != station)
+                continue;
+
+            if (!HasComp<ItemComponent>(ent) || !_revenantAnimated.CanAnimateObject(ent) || _container.IsEntityInContainer(ent))
```
deltanedas: use an EntityQuery.HasComp for ItemComponent

deltanedas (changes requested):

deltanedas (approved):

Proxysseia:
This is cool, we love glimmer flavour. I hope the warden leaves the rpg loaded and unlocked.

Direction approved

PR #4108, opened 2025-07-31
Add flash protection to cult helmets

NoElkaTheGod:
## About the PR
Cult helmets now have flash protection.

## Why / Balance
Flash meta is kidna bad.
Pretty much all other antags have their own flash protection. Syndies and nukies get syndicate gas masks, ninja gets their own mask too, dragon is just straight up immune, etc. etc. Flash is not meant to be a tool against major antags.
Also, there are very few ways for cultists to get flash protection, aside from stealing sec glasses (nobody ever does that and it's not exactly scalable), having engi print welding gas masks (have to have a competent engi cultist **and** have said masks researched), or wearing a welding mask instead of a cult helmet (which just looks silly).
Edit 1: flash meta is also especially good against cultsits, since they are predominantly melee fighters, so they can't even "just hold M1 in the general direction of the enemy", because said enemy will just walk away while the cultist is blinded and slow, then shoot said cultist from a safe distance.

## Technical details
1 line ops.

## Media
No.

## Requirements

**Changelog**
:cl:
- tweak: Cult helmets now provide flash protection.

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/Clothing/cosmiccult_armor.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/Clothing/cosmiccult_armor.yml
@@ -110,3 +110,4 @@
         Slash: 0.90
         Piercing: 0.90
         Heat: 0.90
+  - type: EyeProtection


Proxysseia:
Direction 24h

NoElkaTheGod:
Do I love it when a 1 line PR causes a balance discussion that lasts several months

Proxysseia:
Several months (48h) are over, this is a relatively small change that plugs a hole in the cults arsenal and further helps them in their identity of 'melee guy that is scary and gets shot at by sec'.  There are some concerns over simply giving cults the ability to plug all their weaknesses by virtue of existing, but we can revisit that at a later date if the cult becomes too oppressive.
Until then, direction approved.

NoElkaTheGod:
> There are some concerns over simply giving cults the ability to plug all their weaknesses by virtue of existing

This wasn't a concern when you gave sec hardsuits shock insulation :trollface: 
Not to mention all the other buffs they got.

deltanedas (approved):

PR #4125, opened 2025-08-04
Bring back sentient artifacts

kotobdev:
## About the PR
Sentient artifacts are back, but the node will also make them destructible.

## Why / Balance
Sentient artifacts are fun, and this change will bring them back in a way that hopefully prevents them from being an absolute menace to the station. 300 damage seems like a fair threshold, considering artifacts need to take a lot of punishment in the process of research. They can be repaired with a welder, too.

## Technical details
yamlops

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Sentient artifacts are back, and can be destroyed. Or repaired, with a welder.

### DIFF
--- a/Resources/Prototypes/XenoArch/Effects/utility_effects.yml
+++ b/Resources/Prototypes/XenoArch/Effects/utility_effects.yml
@@ -213,7 +213,6 @@
         sound:
           collection: MetalGlassBreak
   - type: Repairable # DeltaV - Destroyable sentient artifacts
-    doAfterDelay: 5 
 
 - type: artifactEffect
   id: EffectMultitool
--- a/Resources/Prototypes/XenoArch/Effects/utility_effects.yml
+++ b/Resources/Prototypes/XenoArch/Effects/utility_effects.yml
@@ -181,27 +181,39 @@
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/revolver.ogg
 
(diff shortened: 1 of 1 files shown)


Vapetastic-Gaming:
Direction review 24 hours

kotobdev:
test fail fake, but i'm gonna mark this as draft since I just realized a fairly critical bug- if the artifact has already sustained 300 damage prior to becoming sentient it'll just disappear immediately when it does so

engineer-pearl:
Given that the welder can activate an artifact under certain circumstances, can you add a second way to heal them? Perhaps splashing brute meds on them?

Axianamos:
Maybe make them susceptible to mindbreaker rounds from the mantis, or soulbreaker so theres a reason to ever make those. It destroys the conciousness and adds a whole new thing for discord vhst to argue the ethics of

darklyac (commented):
Direction has decided to deny this for now.

LadyDanger2743:
To clarify the denial: 

In practice, Sentient Artifacts on Delta-V tended to either be complete nothingburgers- either due to nobody taking the role, or nobody doing anything with the role- or roles that added a lot of un-fun strife onto the station. Stuff like "I'm a sophont, you can't kill me per your laws" while exploding every time they could activate their nodes. 

Conceptually, a sentient inanimate object is super fun, especially one as alien as a xenological artifact. The implementation here tends to not actually produce any fun. Ergo, the denial of the PR from a direction standpoint.

PR #4127, opened 2025-08-04
Make mindshields actually useful against cosmic cult

NoElkaTheGod:
## About the PR
Cosmic cultists cannot be mindshielded anymore. If you try to do that, a mindshield would just not work, like it does with a headrev. Converting someone with a glyph of truth also disables their mindshield.

## Why / Balance
Having a cult impostor among security/command is cool and adds a lot of possible strategies for said cult. The impostor being completely undetectable until later stages is not so cool, however. This PR ensures that you **can** have cult impostors among security/command, but forces said impostors to play carefuly, because it is fairly easy to identify them by lack of a mindshield. It also makes mindshielding crewmembers (willingly!) a viable strategy against a cult, since this would force them to deshield people before conversion, or use the more expensive glyph of truth. Mindshields are only obtainable through logi, which gives them more reasons to get involved in a cult round.

## Technical details
HasComp + RemComp, that's it.

## Media
No.

## Requirements

**Changelog**
:cl:
- tweak: Cosmic cultists can no longer be mindshielded, causing the implant to immediately fail on insertion.

### DIFF
--- a/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
+++ b/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
@@ -681,7 +681,7 @@ public sealed class CosmicCultRuleSystem : GameRuleSystem<CosmicCultRuleComponen
             !_playerMan.TryGetSessionById(mind.UserId, out var session))
             return;
 
-        if (HasComp<MindShieldComponent>(uid)) RemCompDeferred<MindShieldComponent>(uid);
+        RemCompDeferred<MindShieldComponent>(uid);
 
         _role.MindAddRole(mindId, MindRole, mind, true);
 
--- a/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
+++ b/Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
@@ -35,6 +35,7 @@ using Content.Shared.GameTicking.Components;
 using Content.Shared.Humanoid;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Mind;
+using Content.Shared.Mindshield.Components;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Mobs;
@@ -679,6 +680,8 @@ public sealed class CosmicCultRuleSystem : GameRuleSystem<CosmicCultRuleComponen
         if (!_mind.TryGetMind(uid, out var mindId, out var mind) ||
             !_playerMan.TryGetSessionById(mind.UserId, out var session))
             return;
+
+        if (HasComp<MindShieldComponent>(uid)) RemCompDeferred<MindShieldComponent>(uid);
 
         _role.MindAddRole(mindId, MindRole, mind, true);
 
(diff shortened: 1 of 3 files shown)


deltanedas:
any sec/command with no mindshield are already metagamed or at least treated like second class citizens, dont pretend people would still do that

NoElkaTheGod:
I mean, it's not really metagaming. Command and security are meant to have a mindshield. If they don't, then something went wrong and it's worth investigating. And it's not like this is the only possible way for a mindshield-less sec/command member to appear, so trying to censer them on sight isn't exactly the brightest decision.

Proxysseia:
Direction 24h

NoElkaTheGod:
24 hours? For this? As if I'll ever believe that.

Proxysseia:
Makes converted security more obvious, turns converting sec/command from instantly winning the round into just a large advantage for the cult. Needed nerf to possibly the most oppressive cult strategy. 
Direction approved.

Content.Server/_DV/CosmicCult/CosmicCultRuleSystem.cs
```diff
@@ -680,6 +681,8 @@ public void CosmicConversion(EntityUid converter, EntityUid uid)
             !_playerMan.TryGetSessionById(mind.UserId, out var session))
             return;
 
+        if (HasComp<MindShieldComponent>(uid)) RemCompDeferred<MindShieldComponent>(uid);
```
deltanedas: dont need the hascomp remcomp and friends dont throw

Content.Shared/_DV/CosmicCult/SharedCosmicCultSystem.cs
```diff
@@ -114,4 +118,11 @@ private void DirtyCosmicCultComps<T>(EntityUid someUid, T someComp, ComponentSta
             Dirty(uid, comp);
         }
     }
+
+    private void MindShieldImplanted(EntityUid uid, MindShieldComponent comp, ComponentStartup args)
```
deltanedas: Entity and change it to handle implanting instead so it doesnt break if Mindshield ever has its own ComponentStartup handler

Resources/Locale/en-US/_DV/cosmiccult/preset-cosmiccult.ftl
```diff
@@ -194,6 +194,7 @@ cosmiccult-entropy-unavailable = You can't do that right now.
 cosmiccult-astral-ascendant = {$name}, Ascendant
 cosmiccult-astral-minion = {$name}, Malign
 cosmiccult-gear-pickup = You can feel yourself unravelling while you hold the {$ITEM}!
+cosmic-mindshield-failed = The implant had been overloaded by cosmic energy and failed to activate!
```
deltanedas: has been / was

deltanedas (changes requested):

NoElkaTheGod:
Nope this doesn't work at all

NoElkaTheGod:
Okay so apparently ImplantImplantedEvent is weird and refuses to work for some reason. It's also raised on the implant entity and not the entity that recieved the implant. I tried to look at revs to see how it's done there, but that shit uses MapInitEvent which (to my understanding) is even worse.

NoElkaTheGod:
Apparently I'm not supposed to do that either

NoElkaTheGod:
Screw that, I'll reopen it again once I figure out how to actually do it

PR #4130, opened 2025-08-04
Make Glimmer Look Beautiful

kotobdev:
-[Code White](https://makyo.ink/code-white/)

Somewhere past 700Ψ, the noospheric pressure on the station reaches a point where the cracks in realspace become visible to the naked eye.
Or something like that. TL;DR there is a new shader effect at high glimmer. It can be disabled via an option in settings.

## Why / Balance
~~erm it looks cool xd~~ Right now, glimmer can be at 1000, and if there's no probers around then everybody can more or less just go about their day like nothing's happening. Take off your headset, and you'll not even know that glimmer's high at all. This change makes it so that the crew can actually *see* when reality is falling apart.

Ideally I'll pair this with audible ambience as well, down the line.

## Technical details
~~oh god~~

The existing glimmer system is horrifically bad and entirely component-less. The glimmer value isn't actually networked to the client at all, despite pretending to be. (This is honestly a massive issue and. really should be fixed. by somebody smarter then me). To get around this, I've made it so that when glimmer changes, a network event is raised, and the overlay system listens for this. Glimmer's just an integer and isn't actually modified *too* often, so pending a glimmer refactor, this seems to work well enough. God, glimmer really needs a refactor.
<sub>Did You Know: there's a `GlimmerTierChangedEvent` that is not only never used, but never even raised? awesome!</sub>

Profiling shader performance is basically impossible. However, with the current number of iterations, I can't see a consistent framerate difference between having it on and having it off, on my fairly old desktop. I am aware that shaders (namely the existing blindness one) can apparently have a performance impact on some machines. Ideally there would be a "low-quality shaders" option to tie into, but I feel like it's probably best to wait for something like that to be implemented upstream. ~~Worst-case scenario, I can add a deltav-specific setting for it. Best case scenario, nobody complains and it's all good.~~
<sub>sighhh it looked so much better at 50 iterations though........</sub>

The shader used is modified from something on Shadertoy. As the shader there had no specific stated license, per [Shadertoy's terms](https://www.shadertoy.com/terms), it's licensed under CC BY-NC-SA 3.0, and I've added a comment to state as such.

## Media
The shader appears pretty different at different times of its cycle, so this video won't capture all of the possible states of it. Generally speaking, the shader will begin to sometimes be visible around 750-800 glimmer.

https://github.com/user-attachments/assets/0f0ab449-9328-494c-a381-b2a5dc1e860a

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Glimmer now has a fancy shader overlay when exceeding 700Ψ. This can be disabled under Options>DeltaV.


LLreapz:
peak if prober dosent explode whenever it hits 950 </3

SirWarock:
This makes me wanna bring glimmer to 1000

kotobdev:
revamping what happens when glimmer reaches 1000 is definitely something i plan on doing Soon

Bonktrauma:
> revamping what happens when glimmer reaches 1000 is definitely something i plan on doing Soon

It just causes a mass mindswap event rn which kinda sucks

kotobdev:
> > revamping what happens when glimmer reaches 1000 is definitely something i plan on doing Soon
> 
> It just causes a mass mindswap event rn which kinda sucks

agreed, i am both working on a mass mindswap "nerf" (edit: done :D) as well as a new 1000 glimmer behaviour. glimmer fans will be eating good this week trust

Proxysseia:
make toggleable in accessibility settings?

kotobdev:
done!
also test fail fake trust
image

Stop-Signs:
Direction approved @deltanedas

Content.Client/_DV/Overlays/GlimmerOverlay.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public GlimmerOverlay()
+    {
+        IoCManager.InjectDependencies(this);
+        _glimmerShader = _prototype.Index(_shaderProto).Instance().Duplicate();
+    }
+
+    protected override bool BeforeDraw(in OverlayDrawArgs args)
+    {
+        if (_player.LocalEntity is not { Valid: true } player)
```
deltanedas: == null

Content.Client/_DV/Overlays/GlimmerOverlay.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        // lerp glimmer level to avoid jumps
+        if (!MathHelper.CloseTo(_visualGlimmerLevel, ActualGlimmerLevel, 0.001f))
+        {
+            _visualGlimmerLevel = float.Lerp(_visualGlimmerLevel, ActualGlimmerLevel, 0.1f * lastFrameTime);
+        }
+        else
+        {
+            _visualGlimmerLevel = ActualGlimmerLevel;
+        }
```
deltanedas: set it to a ternary

Content.Client/_DV/Overlays/GlimmerOverlaySystem.cs
```diff
@@ -0,0 +1,61 @@
...
+        base.Initialize();
+
+        _overlay = new GlimmerOverlay();
+        SubscribeNetworkEvent<GlimmerChangedEvent>(OnGlimmerChanged);
+        Subs.CVar(_cfg, DCCVars.DisableGlimmerShader, OnDisableGlimmerShaderChanged);
+    }
+
+    private void OnGlimmerChanged(GlimmerChangedEvent eventArgs)
+    {
+        if(_cfg.GetCVar(DCCVars.DisableGlimmerShader))
```
deltanedas: cache the cvar and update the cache when cvar changes

like Subs.CVar but by hand with OnValueChanged or whatever

Content.Shared/Nyanotrasen/Psionics/Glimmer/GlimmerSystem.cs
```diff
@@ -15,7 +15,11 @@ public sealed class GlimmerSystem : EntitySystem
         public int Glimmer
         {
             get { return _glimmer; }
-            set { _glimmer = _enabled ? Math.Clamp(value, 0, 1000) : 0; }
+            set
+            {
+                _glimmer = _enabled ? Math.Clamp(value, 0, 1000) : 0;
+                RaiseNetworkEvent(new GlimmerChangedEvent(_glimmer));
```
deltanedas: check that it actually changed from the current value

deltanedas (changes requested):

deltanedas (approved):

PR #4133, opened 2025-08-05
Colossus healing tweaks

NoElkaTheGod:
## About the PR
Entropic colossus' healing ability now:
1. Requires confirmation before use
2. No longer has limited charges
3. Takes more time (20 -> 30 seconds)

## Why / Balance
1. Sometimes people accidentally use it mid-combat, which usually results in their death and it's kinda uncool. You're not supposed to use it in a rush anyway, so having to click one more time is insignificant.
2. Healing being limited meant that the colossus itself is limited. Any other role can heal indefinitely if played correctly (dragon can eat bodies, rev can just wait to regenerate, humanoids can steal chems or something, etc.), so it didn't make much sense why colossi cannot. Besides, colossus can't ever heal to full health, because the ability only heals half of the current damage, so it still makes sense to use it as late as possible to maximize healing.
3. To compensate for the inifinite charges. I was depating whether I should do this or increase cooldown, but came to the conclusion that current cooldown (120 seconds) is long enough.

## Technical details
YAML ops.

## Media
No.

## Requirements

**Changelog**
:cl:
- tweak: Entropic colossus' Slumber Shell now has infinite charges, longer hybernation time (20s -> 30s), and a confirmation popup.

### DIFF
--- a/Resources/Locale/en-US/_DV/cosmiccult/abilities.ftl
+++ b/Resources/Locale/en-US/_DV/cosmiccult/abilities.ftl
@@ -1,6 +1,7 @@
 cosmicability-generic-fail = Your influence fails to take hold..
 
 cosmicability-glare-confirm = Press again to activate Null Glare.
+cosmicability-hibernate-confirm = Press again to activate Slumber Shell.
 
 cosmicability-siphon-cultist-success = Your attempts to draw entropy chill {CAPITALIZE(THE($target))}
 cosmicability-siphon-success = You silently draw Entropy from {CAPITALIZE(THE($target))}.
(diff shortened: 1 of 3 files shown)


NoElkaTheGod:
Me trying to maintain 69 open PRs at all times because funny number:

Proxysseia:
Direction Review 14 weeks

NoElkaTheGod:
Без имени

Proxysseia:
Direction approved.

NoElkaTheGod:
@deltanedas this one needs your attention as well

deltanedas (approved):

PR #4134, opened 2025-08-05
Fixed coscult finale warning appearing too early

NoElkaTheGod:
## About the PR
Title. Previously it would be issued when the monument progresses from finaleBuffer to finaleActive. FinaleBuffer is no longer a thing, so it should have been issued 240 seconds after the finale started. However, apparently Kazne is bad at basic algebra, so it was issued 240 seconds before the finale ended, which caused it to be out of sync with the music change. I am also bad at basic algebra, so instead of fixing the equation I just replaced the constant :trollface:

## Why / Balance
Bugfix good.

## Technical details
1 line ops.

## Media
It works.

## Requirements

**Changelog**
Honestly I don't even care enough to write a changelog entry.


NoElkaTheGod:
Testing this PR was a very good opportunity to make some cocoa, because holy shit does the monument has a LOT of timers involved for some reason.

deltanedas (approved):

PR #4147, opened 2025-08-07
Cosmic glyph rework: activation time and animations.

NoElkaTheGod:
## About the PR
Cosmic cult's glyphs now have spawning/despawning animations, (optionally) take some time to activate (also animated and has a cool-ish sound), and don't get deleted on use, instead entering a cooldown state (animated and optional as well).

Also CosmicExamine and friends are predicted now (I was moving stuff around and decided to do this as well).

Glyph of projection automaticaly erases itself when used.
Glyphs of knowledge and truth have 4 seconds activation time and 3 seconds cooldown.
All the transmutation glyphs have 3 seconds cooldown.

## Why / Balance
Cooldown, spawn/despawn animations are just nice to have.
Having an activation timer is mostly aimed at one small issue: cult players would often shunt their target, place them on the glyph, and then just activate said glyph the moment target wakes up, which allows cultists not to use cuffs in most situations. Now you actually have to immoblize the target before conversion one way or another, otherwise they'll run away while the glyph is activating. 
And also for cool visuals.

## Technical details
Significantly expanded CosmicGlyphSystem. Most glyphs (except projection, since that one has no conditions) now also have a OnCheckGlyphConditions method which contains all the condition logic, and is called twice: when the glyph is clicked, and when it tries to activate. 
Moved some methods related to examines into shared because prediction good.
Otherwise mostly minor tweaks to make stuff work.

## Media
Glyph of truth being scribed and used.

https://github.com/user-attachments/assets/c6a4b534-9542-4647-9677-75dea8089afc

Glyph of truth failing to activate due to there being no target.

https://github.com/user-attachments/assets/1feae5c0-98a4-4f57-8fd1-c642440a171f

## Requirements
Yes.

## Breaking changes
Not really I guess?

**Changelog**
:cl:
- tweak: Cosmic cult related examines are now predicted (which means they work much smoother).
- tweak: Cosmic cult's glyphs are now animated!
- tweak: Cosmic cult's conversion glyphs now take some time to activate. Put cuffs on your target if you don't want them to run away mid-conversion.


NoElkaTheGod:
Alright, I think the code is ready for review now. Do not merge yet, though, sprites (and SFX) are still not done.

NoElkaTheGod:
Alright, PR is ready, this time for sure. Tested thoroughly, everything works as expected. Judging by the lack of any input from direction so far, I would assume that this is considered minor enough to not require approval.
@deltanedas please take a look when you have the time.

deltanedas (changes requested):
change the events to by ref record structs

NoElkaTheGod:
@deltanedas ?

Content.Server/_DV/CosmicCult/CosmicGlyphSystem.cs
```diff
@@ -22,35 +23,54 @@ public sealed class CosmicGlyphSystem : EntitySystem
...
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var glyphQuery = EntityQueryEnumerator<CosmicGlyphComponent>();
+        while (glyphQuery.MoveNext(out var uid, out var comp))
         {
-            args.PushMarkup(Loc.GetString("cosmic-examine-text-glyphs"));
+            if (_timing.CurTime >= comp.Timer && (comp.State == GlyphStatus.Spawning || comp.State == GlyphStatus.Cooldown))
```
deltanedas: continue if < comp.Timer then remove it from all the ifs here

Resources/Prototypes/_DV/CosmicCult/glyphs.yml
```diff
@@ -19,29 +19,40 @@
...
+        map: ["base"]
+  - type: GenericVisualizer
+    visuals:
+      enum.GlyphVisuals.Status:
+        base:
+          Spawning: {state: spawning}
+          Despawning: {state: despawning}
+          Ready: {state: ready}
+          Active: {state: active}
+          Cooldown: {state: cooldown}
```
deltanedas: why is this not in base glyph

deltanedas (changes requested):

Resources/Prototypes/_DV/CosmicCult/glyphs.yml
NoElkaTheGod: Because some glyphs don't have some of the states. I could move it to abse if that's not an issue.

deltanedas (approved):

PR #4156, opened 2025-08-09
Tweak cosmic cult win conditions

NoElkaTheGod:
## About the PR
Cosmic cult win conditions have been changed to the following:
- Cult complete: successfully summoning the Unknown (unchanged, obviously)
- Cult major: monument had reached tier 3, cult steward made it to midpoint alive and unrestrained.
- Cult minor: same as cult major, but any 2 cultists and no steward.
- Neutral outcome: monument reached tier 3, but no cultists managed to escape.
- Crew minor: the monument didn't reach tier 3.
- Crew major: all cultists are dead or in handcuffs.
- Crew complete: all cultists were deconverted.

Also, a few minor tweaks:
Cultists are now only considered "dead" for the objectives if they are actually dead (and not crit) or in handcuffs (if you're cuffed at round end, you are getting deconverted soon).
If all cultists die, the automatic shuttle call no longer completely bricks the monument, instead only making the finale unavailable.
Additionaly, the automatic shuttle call doesn't happen if the monument did not yet reach stage 2 (the crew likely doesn't even know that cult exists yet, and while funny, the round ending due to all cultists accidentally dying would not be very nice).

## Why / Balance
Current win conditions encourage crew to not even let the cult to progress, which is:
1. Not fun for anyone involved.
2. Nearly impossible to do if cultists are competent.

Most rounds would therefore end in a cult major, due to the monument reaching stage 3 and one sneaky cultist hiding somewhere in a random locker.
The new win conditions instead encourage crew to prepare during stages 1 and 2, and then, when cultists are clearly visible, arrest and deconvert all of them, while also not letting them complete their own objective.
This also makes obtaining a major win actually challenging for the cult, as the steward would have to somehow reach midpoint, which would be quite complicated, since security at this point is (mechanicaly) encouraged to cult-check everyone who enters evac.

### Note:
For some reason, checking for entities at midpoint does not work in dev env (not only with cult but in general), so I was not able to test the minor and major cult wins properly. I am fairly certain that they will work on release, though, and even if they don't, worst case scenario it would just turn into a neutral outcome.

## Technical details
A few tweaks here and there, updated locale and comments.

## Media
Not really.

## Requirements

**Changelog**
:cl:
- tweak: Cosmic cult win conditions are more fair now. Check PR for details.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Components/CosmicCultRuleComponent.cs
+++ b/Content.Server/_DV/CosmicCult/Components/CosmicCultRuleComponent.cs
@@ -146,23 +146,23 @@ public enum WinType : byte
     /// </summary>
     CultComplete,
     /// <summary>
-    ///    Cult major win. The Monument reached Stage 3 and was fully empowered.
+    ///    Cult major win. The Monument didn't complete, The crew escaped, but the Cult Leader also escaped.
     /// </summary>
     CultMajor,
     /// <summary>
-    ///    Cult minor win. Even if the crew escaped, The Monument reached Stage 3.
+    ///    Cult minor win. The Monument didn't complete, The crew escaped, but at least two cultists also escaped.
     /// </summary>
     CultMinor,
     /// <summary>
-    ///     Neutral. The Monument didn't reach Stage 3, The crew escaped, but the Cult Leader also escaped.
+    ///     Neutral. No cultists made it to midpoint alive.
     /// </summary>
     Neutral,
     /// <summary>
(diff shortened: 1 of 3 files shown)


NoElkaTheGod:
You were supposed to destroy the test fails, not join them!
изображение

Yeah test fail is totally unrelated.

NoElkaTheGod:
Apparently direction ran out of fucks to give about cult PRs

NoElkaTheGod:
Seems like direction is fine with this.
@deltanedas can I have a review?

deltanedas (approved):

PR #4164, opened 2025-08-10
Tie xenoarch point output to glimmer

kotobdev:
## About the PR
Artifact research now has a multiplier to point output tied to glimmer level, from 1x at 0 glimmer to 2.5x at 1000 glimmer, linearly.

## Why / Balance
Xenoarcheology is comically weak compared to anomalies. Despite requiring constant attention and being both RNG and departmentally dependent, xenoarch produces barely any research compared to anomalies, which you can basically ignore once you set them up. This change helps both remedy that issue, and also provides [yet another](https://github.com/DeltaV-Station/Delta-v/pull/4082) risk/reward incentive for Epi to increase glimmer. Ironically, even at 1000 glimmer, I doubt xenoarch will be faster then two (hell, even just one) decent anomalies in advanced containers. But whatever.
<sub>i'd LOVE to gather hard data on xenoarch efficiency in rounds, but that seems hard</sub>

To be entirely honest, the issue with xenoarcheology is probably less "xenoarch is weak" and moreso "anomalies are absolutely busted". But I'll leave that to another PR to fix.

## Technical details
xaml comment ewwww
xenoarch rework seems to be More Then A Week Away so im making this while i can 

~~right now it's a linear scale, would maybe be a bit cooler as an [exponential ease in/out](https://www.desmos.com/calculator/to2pcf6kwz), so epi doesnt reap many benefits until they really ramp up glimmer. didnt bother implementing that but i can give it a go if direction likes that. it's just a lot of evil arithmetic~~ updated

## Media
~~0, 500, and 1000 glimmer respectively.~~ images outdated, ui remains the same
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Artifact research point output is now tied to glimmer level, becoming exponentially more efficient as glimmer increases.


Proxysseia:
Direction review 24h

Content.Server/Xenoarchaeology/Equipment/Systems/ArtifactAnalyzerSystem.cs
```diff
@@ -525,5 +535,12 @@ private void OnPowerChanged(EntityUid uid, ActiveArtifactAnalyzerComponent activ
             ResumeScan(uid, null, active);
         }
     }
+
+    // Begin DeltaV - glimmer mult calculation
+    private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
+    {
+        return 1 + (_glimmerSystem.Glimmer / 1000f) * comp.PointGlimmerMultiplier;
+    }
+    // End DeltaV - glimmer mult calculation
```
Proxysseia, line 547: We like the direction of this PR and would like to double down. 
Exponential scaling tied to higher glimmer. We want high glimmer to also present an opportunity and to not just be a 'stop all work' state.
In the future we should look into adding more ways to manually increase or decrease glimmer, but that is outside the scope of this PR.

Proxysseia (changes requested):

kotobdev:
@deltanedas review plz :D

deltanedas (approved):

PR #4169, opened 2025-08-11
Lower warops minimum nukies to 3

kotobdev:
## About the PR
War can now be declared with 3 nukies instead of 4.

## Why / Balance
Round #23353 was warops with three nukies (curator intervention), and from what I'm hearing it went REALLY well.
Honestly, might be worth considering just lowering the amount to 1. I don't know what circumstances leads to a round with only one or two nukies, but if that happens, the crew will probably be unrobust enough to want time to prepare anyways.

## Technical details
number ops

## Requirements

what am i supposed to launch three instances for this or something. it's the right number trust

## Breaking changes

**Changelog**

:cl:
- tweak: Warops can now be declared with just 3 nukies.

### DIFF
--- a/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
@@ -81,7 +81,7 @@ public sealed partial class NukeopsRuleComponent : Component
     ///     Minimal operatives count for war declaration
     /// </summary>
     [DataField]
-    public int WarDeclarationMinOps = 4;
+    public int WarDeclarationMinOps = 3; // DeltaV
 
     [DataField]
     public WinType WinType = WinType.Neutral;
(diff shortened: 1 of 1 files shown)


NoElkaTheGod:
I don't think it's even possible to have less than 3 nukies. At least it shouldn't be.

Proxysseia:
> I don't think it's even possible to have less than 3 nukies. At least it shouldn't be.

got it, 1 secoff, 1 command, 4 nukies

Proxysseia:
oh also direction 24h

NoElkaTheGod:
> > I don't think it's even possible to have less than 3 nukies. At least it shouldn't be.
> 
> got it, 1 secoff, 1 command, 4 nukies

The amount of nukies depenss on the amount of players ready. Minimum amount of players is 20, so it's totally possible to have 3 nukies with no sec whatsoever. Or 3 newkies with 5 ultra robust secoffs. Depends, really.

Proxysseia:
If there are 3 nukies and they want to declare war, we dont see why they shouldnt be able to.

Direction approved.

kotobdev:
@deltanedas review pls :DDD

deltanedas (approved):

PR #4189, opened 2025-08-14
ik30 nerf

Proxysseia:
DELTANEDAS DONT MERGE BEFORE APPROVAL

## About the PR
needs 2 hands to fire now
halved fire rate of the IK30

## Why / Balance
IK30 is overall an overtuned weapon featuring a good damage type, the ability to fire through glass, the ability to be reloaded with mags, very good damage per mag and very good damage per second.
It is without a doubt the best in slot weapon for security. I would like to introduce some doubt

## Technical details
yaml

**Changelog**
:cl:
- tweak: IK30 has had its fire rate halved and now requires two hands to fire


Proxysseia:
@deltanedas this is direction approved now

deltanedas (approved):

PR #4191, opened 2025-08-14
I.C.E.E. Improvements

Vapetastic-Gaming:
I.C.E.E. Has now been improved with additional Damage (Increased from 20 to 40) and Increased fire rate (Increased from 1.5 to 2)
## Why / Balance
The I.C.E.E. Is pretty much an unused weapon, Cold damage is blocked or majorly reduced by most hardsuits, This increase should be large enough to have the ICEE research a viable option

## Technical details
Changed Damage output from 20 -> 40
Changed Fire rate from 1.5 -> 2

## Media
No Media Required, as this is a Numerical change

## Requirements
- [x ] I have tested all added content and changes.
- [ x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl: 
- tweak: I.C.E.E. Research has made massive forthcomings, Damage and Fire rate have been increased.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -287,12 +287,12 @@
   - type: Clothing
     sprite: _DV/Objects/Weapons/Guns/Battery/cold_cannon.rsi
   - type: Gun
-    fireRate: 1.5
+    fireRate: 2
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/laser_cannon.ogg
   - type: HitscanBatteryAmmoProvider
     proto: ColdLaser
-    fireCost: 100
+    fireCost: 75
 
 - type: entity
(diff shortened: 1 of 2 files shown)


Proxysseia:
The ICEE is not and never should be a gun that primarily deals damage. The ICEE is a utility gun that is able to almost completely stop the movement of any person it hits thrice, with each subsequent shot being easier to land (on a hitscan weapon).
Cold is a damage type that is among the least resisted, with 20% on a nukie hardsuit, and 40% on a jugsuit. For comparison the pierce resist on a jugsuit is 80% +10% from the helmet. 
This ignores the actual numbers of the PR, which would turn the gun from a weapon that deals 30dps and heavily slows targets into a gun that deals 80 dps, with a higher mag capacity (which isnt marked in the PR, you have to look at the files for it), while still pretty much securing the kill after hitting one shot.
While one could argue that the introduction of the Gorlex Poncho serves as an anti ICEE measure, the only thing this would do is turn the damage down from 'instant kill' to 'confirmed kill', not that the poncho does anything to stop the wearer from slowing to a crawl and freezing to death after three hits.

This PR is denied.

PR #4203, opened 2025-08-16
Make empowered cosmic cultists harder to deconvert

NoElkaTheGod:
## About the PR
Trying to deconvert a cosmic cultist no longer deconverts them. Instead, it removes their empowerment, making it possible to decovnert them again afterwards.

Also updated the guidebook on deconversion (it still mentioned censer being fueled with holy water).

## Why / Balance
Deconversion had different VFX depending on empowerment already, so I decided to make it a gameplay element as well.
Makes deconverting cosmic cultists slightly more involved, because you have to watch closely what VFX plays, and act accordingly afterwards (if it was normal VFX, release, if malign VFX, hit em with a censer again then release).
Promotes roleplay. Kinda. I can totally imagine my character having a cool monologue in that situation.
Makes room for funny situations like "chaplain censered Urist McCultist and released him, only to be stabbed in the back because the cultist was empowered".

## Technical details
Moved UnEmpower from CosmicFragmentationSystem into CosmicCultSustem. Added some locale, UI stuff and an if statement.

## Media
UI popup.
изображение
Chat notification.
изображение

## Requirements

**Changelog**
:cl:
- tweak: Using a censer on an empowered cosmic cultist no longer deconverts them. However, it would make them not empowered anymore, so you can (and should) use your censer again once they stop shaking. Watch closely for bursts of malign energy when cleansing fellow crewmembers!


Proxysseia:
Direction 24h

deltanedas (changes requested):
dont make a new ui for every possible message just set the strings

NoElkaTheGod:
This small move will cost us 51 years of researching actual UI code :godo:

NoElkaTheGod:
Alright you know what, I've got a better idea

NoElkaTheGod:
This doesn't change your antag status, so a UI popup isn't really needed anyway. Uniting all the popups into a single one would be nice, but I have zero clue how to implement that.

Proxysseia:
We do not think that these changes are necessary and would enhance security, epistemics, medical or cult gameplay. 
Depending on how easy it is to tell if a cultist got deconverted successfully or just downgraded from empowered to standard, this introduces frustration to either the crew side or the cults side.
In any case the process of deconversion is made more expensive and longer for a thing that is arguably not even beneficial to gameplay. If you get caught, secured, and deconverted, then you should probably actually get deconverted.
The best case scenario of this change is that a cultist lucks into (because its not like they can choose to hide their deconversion type) being only downgraded and let go in order to cause more chaos as a cultist.
The worst case scenario is that deconverting is deemed unreliable risky and expensive, and the preferred method of dealing with cultists becomes a body bag.

As such this PR is denied.

NoElkaTheGod:
Well not like it was that great of an idea anyway

PR #4204, opened 2025-08-16
Increase entropic colossus' melee damage

NoElkaTheGod:
## About the PR
Entropic colossus melee damage has been increased from 25 blunt + 5 cold + 5 asphyx to 35 blunt + 10 cold + 10 asphyx. Structural damage not changed (60)

## Why / Balance
Smashing things very hard is pretty much the only thing a colossus can do, besides it's AOE attack, which has a rather long cooldown. You'd expect a thing whose sole job is to punch things hard to, well, punch things hard. However, it's damage per second current;y is **worse than a baseball bat** thanks to having 0.425 attack speed. And that's a midround antag associated with the cosmic cult, who have some of the best melee weapons in the game. And it's also very slow, so it's rather easy to just... not get hit by it. The proposed values still put it at lower DPS than cult weapons, but perhaps the highest damage per swing in the game (not counting admemery), which means that you **really** don't want to stand close to an agressive colossus, because most armors (again, not counting admemes) would allow you to survive 4 hits at best (or slightly more if you're an IPC).

## Technical details
YAML ops

## Media
Numbers change

## Requirements
- [ ] I have tested all added content and changes. (Numbers change)
- [ ] I have added media to this PR or it does not require an ingame showcase. (Numbers change)

**Changelog**
:cl:
- tweak: Entropic colossus' melee damage has been buffed.

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
@@ -203,9 +203,9 @@
     range: 2
     damage:
       types:
-        Blunt: 25
-        Asphyxiation: 5
-        Cold: 5
+        Blunt: 35
+        Asphyxiation: 10
+        Cold: 10
         Structural: 60
     soundHit:
       path: /Audio/_DV/CosmicCult/cosmicsword_glance.ogg


NoElkaTheGod:
For number nerds, currently the colossus has 35 * 0.425 = 14,875 damage per second (baseball bat has 15). This PR brings it up to 55 * 0.425 = 23,375, which is about the level of a crusher dagger (24) or a susbox (20).

NoElkaTheGod:
Test fail is fake and probably gay

Proxysseia:
direction 25 hours

Proxysseia:
The colossus' current play pattern is not the greatest. It teleports on top of their victim, it smacks them mid teleport, and if there is a wall nearby the colossus kills them.
The colossus' strength is that if you get close, or get trapped, you die. The colossus' weakness is that if you cannot get close to someone, or they are not trapped, then you cant kill them, get whittled down, and die.

Increasing the colossus' melee damage does not alter the play pattern for the better, it just makes it more oppressive against weak players. The colossus' dps is bad, yes, but it's playstyle does not favor dps anyways. The colossus is a burst antag, and this would give it too much burst, especially considering a third of the damage is de facto AP.

If you feel the colossus needs buffs, you can consider increasings it's movespeed, decreasing the teleport cooldown, or improving the swings attack speed.

In its current form this PR is denied.

PR #4214, opened 2025-08-17
Twek IK-30 some more

NoElkaTheGod:
## About the PR
IK-30 no longer requires weilding. IK-30's accuracy when weilded has been adjusted to actually make the gun **more** accurate when weilded (has very awful accuracy when not weilded, so weilding is still pretty much required).

## Why / Balance
For the longest time it was bugged, and weilding it actually made it **less** accurate. Then the nerf came, forcing everyone to weild it, which in turn made the weapon practicaly unusable. Now it actually behaves like any other gun.

## Technical details
YAML ops

## Media
Numbers change

## Requirements
Numbers change

**Changelog**
:cl:
- tweak: The IK-30 no longer requires weilding.
- fix: IK-30 now actually becomes more accurate when weilded instead of becoming less accurate.


NoElkaTheGod:
Oh well, seems like I took too long, #4212 already fixes that. Welp.

PR #4216, opened 2025-08-17
Make malign rifts actually (kinda) dangerous

NoElkaTheGod:
## About the PR
Malign rifts now periodically deal damage to all humanoids, depending on the distance between said humanoid and the rift. The effect does not apply to cosmic cultists and bibleUsers (chaplain and mystagogue).
More specifically, it applies 10-40 seconds of entropic degen status effect, which deals anywhere between 0.2 to 2 damage (25% cold, 75% asphyx) per second, based on the distance (the closer, the higher), to all targets within 15 tiles every 20 to 60 seconds. Every "pulse" is followed by a visual effect to make it clear that something has happened.

## Why / Balance
1. Makes malign rifts actually dangerous to be around, thus giving the crew more reasons to want to purge them besides just "ugly floors".
2. Prevents crew from camping the rifts until a chaplain arrives to purge them. See a rift, walk away, then tell everyone about it through radio. Standing near it to make sure cultists don't get it may kill you.
3. Makes siphoning entropy less metagameable, because now "I don't know why but I have 4 cold damage" could not only mean "this is a cult round", but also "there's a randomly spawned rift somewhere and this dude just didn't notice it".

## Technical details
Every 20-80 seconds apply a status effect to all entities with HumanoidAppearanceComponent and no CosmicCultComponent nor BibleUserComponent. Most values are stored on the prototype.

## Media
Tested, works.

## Requirements

**Changelog**
:cl:
- tweak: Malign rifts sometimes deal damage to everyone around them now. Get away, or you may get hurt! Further away. Like, don't even look at it.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
+++ b/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
@@ -35,13 +35,19 @@ public sealed partial class CosmicMalignRiftComponent : Component
     public TimeSpan MinPulseTime = TimeSpan.FromSeconds(20);
 
     [DataField]
-    public TimeSpan MaxPulseTime = TimeSpan.FromSeconds(80);
+    public TimeSpan MaxPulseTime = TimeSpan.FromSeconds(60);
 
     [DataField]
     public float PulseRange = 15f;
+    /// <summary>
+    /// The chance for each entity in range to be affected by a pulse
+    /// </summary>
+    [DataField]
+    public float PulseProb = 0.75f;
 
     [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
     public TimeSpan NextPulseTime = default!;
 
-    [DataField] public EntProtoId PulseVFX = "CosmicGenericVFX";
+    [DataField]
+    public EntProtoId PulseVFX = "CosmicGenericVFX";
 }
--- a/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
+++ b/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
@@ -5,6 +5,7 @@ using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 namespace Content.Server._DV.CosmicCult.Components;
 
 [RegisterComponent]
+[AutoGenerateComponentPause]
 public sealed partial class CosmicMalignRiftComponent : Component
 {
     [DataField]
@@ -45,7 +46,7 @@ public sealed partial class CosmicMalignRiftComponent : Component
     [DataField]
     public float PulseProb = 0.75f;
 
-    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
+    [AutoPausedField, DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
     public TimeSpan NextPulseTime = default!;
 
     [DataField]
--- a/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
+++ b/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
@@ -1,5 +1,6 @@
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 
 namespace Content.Server._DV.CosmicCult.Components;
 
@@ -29,4 +30,18 @@ public sealed partial class CosmicMalignRiftComponent : Component
 
     [DataField]
     public TimeSpan AbsorbTime = TimeSpan.FromSeconds(35);
+
+    [DataField]
+    public TimeSpan MinPulseTime = TimeSpan.FromSeconds(20);
+
+    [DataField]
+    public TimeSpan MaxPulseTime = TimeSpan.FromSeconds(80);
+
+    [DataField]
+    public float PulseRange = 15f;
+
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
+    public TimeSpan NextPulseTime = default!;
+
+    [DataField] public EntProtoId PulseVFX = "CosmicGenericVFX";
 }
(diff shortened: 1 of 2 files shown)


deltanedas:
sec hes not getting hurt by the rift kill him

Proxysseia:
Direction 24h

LadyDanger2743:
Speaking as an individual here & without the full weight of Direction: I'm inclined to agree with deltanedas. 

I personally think that it's a great idea to make the Rifts harmful, so players avoid standing near them to "camp"/metagame cult objectives. However, the damage should be (or, at least, SEEM) indiscriminate, so that players can't then metagame "oh, this person was NOT hurt by the rift, therefore they're evil and we need to kill them." 

Maybe Cult and Bible characters have a resistance to the effect somehow? It does less damage, but still applies the same visuals to them? Just my suggestion though.

NoElkaTheGod:
It literally has no visuals on the characters, and standing near a rift is already quite suspicious and you ideally shouldn't be seen doing it. Rifts actually dealing damage serves to reinforce that. However, if direction thinks this is an issue, I can change this to instead deal less damage to cultists (and bibleUsers) instead of no damage at all, to make it that bit less obvious.

LadyDanger2743:
> It literally has no visuals on the characters, and standing near a rift is already auite suspicious and you ideally shouldn't be seen doing it. Rifts actually dealing damage serves to reinforce that. However, if direction thinks this is an issue, I can change this to instead deal less damage to cultists (and bibleUsers) instead of no damage at all, to make it that bit less obvious.

Doesn't seem to be as much of an issue to the other folks as it is to me. Will get a final answer to you in 7h.

nkokic:
If I may add. This is without thinking a good PR.
People are very often incapable of roleplaying fear as expected.
Giving them a gameplay reason is a good way to make players react and roleplay accordingly.
Otherwise, it just SEEMS like a silly cosmetic that doesn't do much. Pun intended :)

NoElkaTheGod:
Hrrrr mimimi

Proxysseia:
There are some concerns about metagaming the (lack of) damage taken by cultists, as someone with a medhud can with absolute certainty know who is a cultist and who is not. 
While they are not allowed to act on this information in character, it is still preferable to not have the disconnect between IC and OOC knowledge in the first place.

As such we have two suggestions, or you are free to find another way to solve this issue:
Make cultists take damage from rifts like anyone else.
Make rifts randomly deal no damage to some crew.

NoElkaTheGod:
Adding more RNG to the mix sounds good enough. Give me a moment.

NoElkaTheGod:
Done. Each entity within range now only has 75% chance to get the debuff.

Proxysseia (approved):

Proxysseia:
Making rifts dangerous is a good bit of ludonarrative harmony. Maybe crew finally stays away from the weird glowy thing and calls for epi to get rid of them.

Direction approved

NoElkaTheGod:
@deltanedas hi review please

Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
```diff
@@ -29,4 +30,24 @@ public sealed partial class CosmicMalignRiftComponent : Component
...
+    [DataField]
+    public float PulseRange = 15f;
+    /// <summary>
+    /// The chance for each entity in range to be affected by a pulse
+    /// </summary>
+    [DataField]
+    public float PulseProb = 0.75f;
+
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
+    public TimeSpan NextPulseTime = default!;
```
deltanedas: auto paused

Content.Server/_DV/CosmicCult/EntitySystems/CosmicRiftSystem.cs
```diff
@@ -27,6 +37,40 @@ public override void Initialize()
...
+    }
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var query = EntityQueryEnumerator<CosmicMalignRiftComponent>();
+        while (query.MoveNext(out var uid, out var comp))
+        {
+            if (_timing.CurTime >= comp.NextPulseTime)
```
deltanedas: invert and continue then unindent the rest

Content.Server/_DV/CosmicCult/EntitySystems/CosmicRiftSystem.cs
```diff
@@ -27,6 +37,40 @@ public override void Initialize()
...
+            {
+                comp.NextPulseTime = _timing.CurTime + _random.Next(comp.MinPulseTime, comp.MaxPulseTime);
+
+                var pos = Transform(uid).Coordinates;
+                Spawn(comp.PulseVFX, pos);
+                _lookup.GetEntitiesInRange<HumanoidAppearanceComponent>(pos, comp.PulseRange, _humanoids);
+                _humanoids.RemoveWhere(target => HasComp<BibleComponent>(target) || HasComp<CosmicCultComponent>(target));
```
deltanedas: bibleuser and use EntityQuery for both of these comps

deltanedas (changes requested):

Content.Server/_DV/CosmicCult/EntitySystems/CosmicRiftSystem.cs
NoElkaTheGod: The  rest is done, but how exactly do I use EntityQuery here?

deltanedas (dismissed):

deltanedas:
just store EntityQuery<BibleUserComponent> on the system and set it from GetEntityQuery in Initialize

same for other one

rosieposieeee:
itd be nice if the pulses of energy it puts out when converting tiles were what was hurting you, and it kept putting those out, so you could actually have some kind of dodging or clear indication what is causing it. coscult is really opaque already for people who don't read prs

NoElkaTheGod:
> itd be nice if the pulses of energy it puts out when converting tiles were what was hurting you, and it kept putting those out, so you could actually have some kind of dodging or clear indication what is causing it. coscult is really opaque already for people who don't read prs

One of the primary goals was making siphoning less obvious by introducing another source of "huh, where'd that cold damage come from?". Making the daamge from rifts more obvious kinda goes against that. Besides, that's literally a cult of ***the Unknown***, so the "opaqueness" is here by design to some extent.

NoElkaTheGod:
Dammit

NoElkaTheGod:
TryStopNukeOpsFromConstantlyFailing jumpscare

deltanedas (approved):

PR #4243, opened 2025-08-20
Make superweapons only available at high alert levels (gamma, delta, octarine)

NoElkaTheGod:
## About the PR
Superweapon safes can no longer be opened manually (unless you have centcomm access). However, should the station's alert level be set to gamma, delta, or octarine, any superweapon safes would be automaticaly unlocked.
Also you can't emag superweapon safes anymore, and they acidify if you break them. If the crew can't use them most of the time, so do antags.

## Why / Balance
Superweapons are intended to be used when the situation is very bad. However, there's no gameplay reason (hell, not even an SOP reason) not to use them over a slightest inconvenience. I guess RPG also blows up the station, but that's only one weapon out of three that are available, and you can't really do the same to gamba/beamdev (hitscan code my beloved), and you can just ask engi repair the hole. The other two weapons are frequently used against mid-level antags like colossi, dragons or loneops, who have very little chances of actually fighting against a weapon like that. This ensures that security can only use superweapons in dire situations, like the nuke being armed, cosmic horrors being summoned into our reality, or a dragon being so robust that an admin decides to issue code gamma.
This also somewhat discourages armory rushing, since you can no longer just steal one of the strongest weapons in the game when doing that. Unless you try to break the safe, but that only has a 25% chance of actually giving you the contents.

## Technical details
Added a new UnlockOnAlertLevelComponent and the corresponding system, which listens to AlertLevelChangedEvents, and opens any LockComponents on the same entity if the alert level matches one that is listed on the component. Examine text also included.

## Media
изображение

https://github.com/user-attachments/assets/2a6eccf1-4bbf-42e8-bf4b-2bc7c526336a

## Requirements

**Changelog**
:cl:
- tweak: Superweapon safes can no longer be opened by normal means. Instead, they are automaticaly unlocked if the station's alert level is set to delta, gamma, or octarine.
- tweak: Superweapon safes can no longer be emaged, and acidify their contents when destroyed.


NoElkaTheGod:
TryStopNukeOpsFromConstantlyFailing strikes back

Bonktrauma:
> or a dragon being so robust and admin decides to call gamma

... is this targeted at me?

Proxysseia:
Direction 24h

Content.Client/_DV/Access/UnlockOnAlertLevelSystem.cs
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Content.Client._DV.Access;
+
+/// <summary>
+///     Allows an ID card to set the access level of interacted items
+/// </summary>
+
+public sealed class UnlockOnAlertLevelSystem : SharedUnlockOnAlertLevelSystem
+{
+}
```
deltanedas: remove summary and change {} to ;

the client system does nothing

Content.Server/_DV/Access/UnlockOnAlertLevelSystem.cs
```diff
@@ -0,0 +1,35 @@
...
+    private void OnAlertLevelChanged(AlertLevelChangedEvent args)
+    {
+        var query = EntityQueryEnumerator<UnlockOnAlertLevelComponent, LockComponent>();
+        while (query.MoveNext(out var uid, out var unlockComp, out var lockComp))
+        {
+            if (lockComp.Locked == false) continue;
+            foreach (var level in unlockComp.AlertLevels)
+            {
+                if (level == args.AlertLevel)
+                    _lock.Unlock(uid, null, lockComp);
```
deltanedas, line 32: break too, no reason to check other levels

Content.Server/_DV/Access/UnlockOnAlertLevelSystem.cs
```diff
@@ -0,0 +1,35 @@
+using Content.Server.AlertLevel;
+using Content.Shared._DV.Access;
+using Content.Shared.Lock;
+
+namespace Content.Server._DV.Access;
+
+/// <summary>
+///     Allows an ID card to set the access level of interacted items
+/// </summary>
+
```
deltanedas: change it to not be copypasted

Content.Shared/_DV/Access/SharedUnlockOnAlertLevelSystem.cs
```diff
@@ -0,0 +1,33 @@
+using Content.Shared.Examine;
+
+namespace Content.Shared._DV.Access;
+
+/// <summary>
+///     Allows an ID card to set the access level of interacted items
+/// </summary>
```
deltanedas: change it to be specific to the examine

Content.Shared/_DV/Access/SharedUnlockOnAlertLevelSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+        var levels = ""; // 90% sure there's actually a smartass way to do this with fluent but I'm doing this becasue I fucking love for loops
+        for (int i = 0; i < ent.Comp.AlertLevels.Count; i++)
+        {
+            if (i == ent.Comp.AlertLevels.Count - 2)
+                levels += ent.Comp.AlertLevels[i] + ", or ";
+            else
+            if (i == ent.Comp.AlertLevels.Count - 1)
+                levels += ent.Comp.AlertLevels[i];
+            else
+                levels += ent.Comp.AlertLevels[i] + ", ";
```
deltanedas: use string builder

Content.Shared/_DV/Access/UnlockOnAlertLevelComponent.cs
```diff
@@ -0,0 +1,11 @@
+namespace Content.Shared._DV.Access;
+
+/// <summary>
+///     Unlocks a LockComponent when the station's alert level is changed to one of the specified levels
```
deltanedas: any station not "the" station, it doesnt check anything

Content.Shared/_DV/Access/UnlockOnAlertLevelComponent.cs
```diff
@@ -0,0 +1,11 @@
...
+/// </summary>
+[RegisterComponent]
+public sealed partial class UnlockOnAlertLevelComponent : Component
+{
+    [DataField]
+    public List<string> AlertLevels = new();
```
deltanedas, line 12: protoid

Content.Shared/_DV/Access/UnlockOnAlertLevelComponent.cs
```diff
@@ -0,0 +1,11 @@
+namespace Content.Shared._DV.Access;
+
+/// <summary>
+///     Unlocks a LockComponent when the station's alert level is changed to one of the specified levels
+/// </summary>
+[RegisterComponent]
+public sealed partial class UnlockOnAlertLevelComponent : Component
+{
+    [DataField]
```
deltanedas: required

Content.Shared/_DV/Access/UnlockOnAlertLevelComponent.cs
```diff
@@ -0,0 +1,11 @@
+namespace Content.Shared._DV.Access;
+
+/// <summary>
+///     Unlocks a LockComponent when the station's alert level is changed to one of the specified levels
+/// </summary>
+[RegisterComponent]
```
deltanedas: networked component + access

deltanedas (changes requested):

Content.Shared/_DV/Access/UnlockOnAlertLevelComponent.cs
NoElkaTheGod: Alert levels aren't separate entities, I don't think they have IDs

deltanedas: AlertLevelPrototype

NoElkaTheGod: It's wonky and seemingly stores all the alert levels on a single entity as strings, hence why I'm storing strings here too

deltanedas: i dont care

NoElkaTheGod: It's not even in shared how am I even supposed to do that :/

NoElkaTheGod: And there's literally one entity that uses AlertLevelPrototype, that being stationAlerts. You just can't check by ProtoId here.

Proxysseia:
This is de facto content removal with a cherry on top. We are broadly unhappy with how superweapons function and with the levity they are pulled out for medium level threats. This PR however is not the way we want to address this, because we still think superweapons have some place in the game without direct admin intervention (or losing to nukies).

As such the PR is denied.

There is an ongoing discussion in direction about what way we want to take superweapons and how we are going to be going about restricting their use outside of necessary cases, but that is probably outside the scope of this PR.

NoElkaTheGod:
1984

PR #4246, opened 2025-08-20
Make secret take total player amount into account, lower the required amount of players ready

NoElkaTheGod:
## About the PR
Secret no takes total playercount into accout when rolling for a gamemode. The requried amount of players ready has been somewhat lowered.
Exact numbers:
Nukeops: 30 total players, 10 ready (previously 20 ready).
Zombies: 30 total players, 10 ready (previously 20 ready).
Cosmic cult: 40 total players, 15 ready (previously 25 ready).
Traitors and survival not changed (you can't really lower it any more).

## Why / Balance
On DeltaV, players often do not ready up in lobby, but latejoin within a minute from round start. Whether it is done to intentionally avoid major antags appearing, or to ensure that all departments are properly staffed is not really relevant. Either way, this often leads to situations, where despite the server having more than enough players for a gamemode to appear, the round can only be traitors or survival, because more than 50% of the players do not ready. And those are already the most common gamemodes purely from their weights. In rare cases, even a full server of 80 players would not have 20 players ready before roundstart.

This ensures that this doesn't happen, or at least happens way less. The requirement for total players ensures that there are indeed enough players for the gamemode to work properly. Numbers may need to be adjusted later.

## Technical details
Add a field to GameRuleComponen and check it against the total playercount.

## Media
Not needed.

## Requirements

## Important note
While I did make sure that this at least builds, I can't reasonably test this in any meaningful way, because my PC would literally explode if I try to launch 30 clients at once. This will require a test merge, or a test server to be launched, to make sure it doesn't break anything.

**Changelog**
:cl:
- tweak: Secret game mode now not only checks for amount of players ready, but also for the total player count. 
- tweak: Game modes now generally require less players to be ready during lobby.

### DIFF
--- a/Content.Server/GameTicking/Rules/SecretRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/SecretRuleSystem.cs
@@ -9,6 +9,7 @@ using Content.Shared.GameTicking.Components;
 using Content.Shared.Random;
 using Content.Shared.CCVar;
 using Content.Shared.Database;
+using Robust.Shared.Player;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 using Robust.Shared.Configuration;
@@ -22,6 +23,7 @@ public sealed class SecretRuleSystem : GameRuleSystem<SecretRuleComponent>
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly IConfigurationManager _configurationManager = default!;
     [Dependency] private readonly IAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly IPlayerManager _playerManager = default!; // DeltaV
     [Dependency] private readonly GameTicker _ticker = default!; // begin Imp
 
     // Dictionary that contains the minimum round number for certain preset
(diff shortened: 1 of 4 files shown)


Content.Server/GameTicking/Rules/SecretRuleSystem.cs
```diff
@@ -22,6 +23,7 @@ public sealed class SecretRuleSystem : GameRuleSystem<SecretRuleComponent>
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly IConfigurationManager _configurationManager = default!;
     [Dependency] private readonly IAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly IPlayerManager _playerManager = default!; // DeltaV
```
deltanedas: rename to _player

deltanedas (changes requested):

LLreapz:
wtf elka carrying delta, thank u sm

Proxysseia:
direction Review 10h

Shadowmize:
wait this is the best pr of all time

Proxysseia:
Ready up does not clarify intent on delta the same way it does on other servers. These changes make sense and are good.
Direction approved.

deltanedas (approved):

NoElkaTheGod:
TryStopNukeOpsFromConstantlyFailing jumpscare. Might actually be real in that case, though. The output log isn't very helpful, maybe try re-running it again?

NoElkaTheGod:
What could possibly go wrong

NoElkaTheGod:
So far it works as expected. Just had a nukie round with 16 people ready, 35 total.

PR #4247, opened 2025-08-20
emergency crash bugfix ops

kotobdev:
## About the PR
glimmer shader should no longer CRASH THE GAME for some players at startup

## Technical details
after the last server update, couple of people started reporting that the game consistently crashes on startup. got one of them to send me their logs, relevant error appears to be
```
Unhandled exception. Robust.Client.Graphics.Clyde.ShaderCompilationException: Failed to compile vertex shader, see inner for details (and error.glsl for formatted source).
 ---> Robust.Client.Graphics.Clyde.ShaderCompilationException: ERROR: 0:235: 'uniform' :  cannot initialize this type of qualifier 
 ```
considering that the glimmer thing is the only thing to touch shaders in that update, i assume it's due to that
~~just removed the initialization for size, hopefully that's good enough. i can't reproduce the crash myself so i am praying this works~~ okay it turns out that if you simply look at the game wrong when writing a swsl shader it will crash on startup. why do i need to define every single float as highp why cant it just infer that

should probably be speedmerged for obvious reasons

edit: the crash seems to happen on any device running ss14 in compatibility mode. huh.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed the game crashing on startup when running in compatibility mode.

### DIFF
--- a/Resources/Textures/_DV/Shaders/highglimmer.swsl
+++ b/Resources/Textures/_DV/Shaders/highglimmer.swsl
@@ -3,7 +3,7 @@
 
 const highp float TimeScale = 0.2;
 
-uniform highp float size = 1;
+uniform highp float size;
 
 void fragment() {
     vec2 iResolution = vec2(1.0);
--- a/Resources/Textures/_DV/Shaders/highglimmer.swsl
+++ b/Resources/Textures/_DV/Shaders/highglimmer.swsl
@@ -6,31 +6,31 @@ const highp float TimeScale = 0.2;
 uniform highp float size;
 
 void fragment() {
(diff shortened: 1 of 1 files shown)


kotobdev:
managed to maybe reproduce it and it still crashes post-fix, gonna test a bit more and then reopen this
edit 1: it's initializing itself ? what
edit 2: so turns out compatibility mode just tries to murder you if you try to write code that looks good
edit 3: ETA a couple hours on this probably
edit 4: patched

deltanedas (approved):
hate this godawful language and api wgpu and wgsl is peak

PR #4248, opened 2025-08-20
Make shunt's do-after hidden

NoElkaTheGod:
## About the PR
The do-after for shunt subjectivity is now hidden. The popup at the start has also been removed.
Do-after is now 1 second long (or 0.8 if empowered) instead of 0.6 seconds (or 0.4 if empowered).
Also you can no longer use shunt on SSD people (it didn't do anythin anyway, just wasted the charge).

## Why / Balance
It's supposed to be a "very long stun, but you can't use it in combat because both targets have to stand still", which was mostly achieved by the incredibly short do-after. However, sometimes people still managed to react in time, which just felt like the ability doesn't work as intended.
Also, before this change it was possible to use shunt mid-combat if the target is slowed down, or just didn't move enough from the initial spot. Thus, the do-after was amde a little longer to prevent that from happening.

## Technical details
Not really.

## Media
Not really.

## Requirements

**Changelog**
:cl:
- tweak: The do-after for shunt subjectivity now takes longer, but is invisible to others.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Abilities/CosmicBlankSystem.cs
+++ b/Content.Server/_DV/CosmicCult/Abilities/CosmicBlankSystem.cs
@@ -43,19 +43,17 @@ public sealed class CosmicBlankSystem : EntitySystem
 
     private void OnCosmicBlank(Entity<CosmicCultComponent> uid, ref EventCosmicBlank args)
     {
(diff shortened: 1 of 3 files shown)


NoElkaTheGod:
:trollface:

NoElkaTheGod:
:trollface:

NoElkaTheGod:
Alright fine I'll figure this out later <sup>tm</sup>

PR #4199, opened 2025-08-15
Kill Trials

Proxysseia:
## About the PR
reduced sec and justice regulations from 4100 words to 2600 words
replaced criminal trials with shorter, less thorough hearings
removed civil trials
moved Procedural Defense to Space Law
essentially rewrote the following sections of SoP:
Security:
- Detainment and Arrest
- Brig Procedure
- Sentencing
- Evidence Handling
- Paroles and Pardons

Justice:
- Charges
- Court Orders
- Trials
- Presentation of Evidence
 
## Why / Balance
The justice department is a unique aspect of Delta V i cherish very much. The department has been designed around conducting trials, however it was an open secret that trials do not work and have a host of problems:
Justice does not have the autonomy necessary to make trials happen. What this means is that Justice is reliant on other departments to even attempt to do their jobs.
Justice has no way to enforce their rulings: Justice is completely at the mercy of securitys goodwill, said goodwill is in very sparse supply. As a department justice is de facto alligned with Security and as such unable to prosecute them, even in cases where Justice wants to prosecute them, the lack of stun batons in CJ's office prevents such a feat.

The original intention of trials is unfitting for the kind of game Delta V's SS14 is. Trials require a massive amount of setup and require a large number of willing participants with little interuptions. Trials in real life take months to years, trials in SS14 are supposed to take 30 minutes. The conditions to have a fruitful trial are not present.
Trials take too long to set up, Trials take too long to be resolved. The results of trials happen so late into the round they may aswell not have happened at all. 

Due to justices severe lack of agency, and justices inability to be anything other than a tider under heavy RP rules on shifts that were not Traitor shifts (as in, the majority of the rounds), The justice department has suffered from heavy player attrition. It is not uncommon to see justice almost, or completely, empty.
Despite the fact that the department is often empty, the legal framework for trials still informed all security/antag interactions. This frankly was and is not fun for either party. Security needs to pretend there will be a trial eventually, and the antag has to decide if they want to hope a CJ decides to exit cryo and flip a coin on whether they get walled by Security.

As such, trials are being killed so that the idea can survive: The idea that security does atleast need some kind of evidence in order to RR you by throwing you into perma until the end of the round. Hearings serve the purpose of checking if security has the right guy. Hearings serve the purpose of hopefully training secoffs to not crumble under such hardball questions as 'what is that guy in for' and 'do we have literally any proof of that'. 

[...]

PR #4268, opened 2025-08-26
SOP: clarify the standard of proof required for charging a crime

kalexmills:
## About the PR
Revises the standard of proof needed to charge a crime to "probable cause" instead of "beyond a reasonable doubt".

## Why / Balance
Under current SOP, the legal argument "at the time the crime was charged, there remained a reasonable doubt that the defendant committed the crime, so the defendant should not have been charged and must be released" is valid.

Without this change, the burden of proof required to charge a crime becomes prohibitive. Under US law, 'probable cause' is all that is required to issue an indictment for a crime, while proof 'beyond a reasonable doubt' is required to *convict*. 'Beyond a reasonable doubt' is a much higher standard of proof than probable cause, requiring enough evidence to ensure there is no other logical explanation for the crime. It is unreasonable to expect Sec to gather this much evidence before making an arrest.

Without this change, Sec would need to gather enough evidence to *entirely* prove their case within 5 minutes after detaining a citizen. SOP is also presently unclear who is meant to adjudicate the "beyond a reasonable doubt" standard. The present state of affairs opens up a can of worms when charging any criminal defendant, requiring the captain's intervention to overrule the wording of SOP even in case of minor crimes.

## Technical details
The current wording is a bit clunky, but I tried to make as small a change as possible. We can remove the 'beyond a reasonable doubt' clause or even modify it so that proof beyond a reasonable doubt is only required for capital crimes. That may be desirable, as it remains unclear who is meant to adjudicate the standard required for conviction in case of other crimes.

## Requirements

## Breaking changes

**Changelog**


.wiki/_DV/Laws/StandardOperatingProcedure.txt
```diff
@@ -290,7 +290,7 @@ Additionally, a pardon can be issued to any entity accused of, or sentenced to,
 =Justice Regulations=
 ==Pressing Charges==
 ===Criminal Charges===
-Any crew member may press charges against any other crew member. An individual may be charged of a crime if and only if it can be argued beyond a reasonable doubt that the accused had committed an act in the nature and fashion as described by a particular criminal charge.
+Any crew member may press charges against any other crew member. An individual may be charged with a crime if and only if there is [[Standard Operating Procedure#Probable Cause|Probable Cause]] that the accused had committed an act in the nature and fashion as described by a particular criminal charge. An individual may be convicted of a crime if and only it can be argued beyond a reasonable doubt that the accused had committed an act in the nature and fashion as described by a particular criminal charge.
```
NoElkaTheGod: I think that would be correct
```suggestion
Any crew member may press charges against any other crew member. An individual may be charged with a crime if and only if there is [[Standard Operating Procedure#Probable Cause|Probable Cause]] that the accused had committed an act in the nature and fashion as described by a particular criminal charge. An individual may be convicted of a crime if and only if it can be argued beyond a reasonable doubt that the accused had committed an act in the nature and fashion as described by a particular criminal charge.
```

kalexmills: Thanks!

Proxysseia:
check #4199

kalexmills:
These changes are no longer needed as the wording "reasonable doubt" does not appear in the revised SOP or Space Law.

PR #4271, opened 2025-08-26
Make cult glyphs invisible to non-cultists during stages 1 and 2

NoElkaTheGod:
## About the PR
Title. Just like the monument. Becomes visible when stage 3 is reached.

## Why / Balance
- Chap decides to randomly check maints
- Chap finds a glyph on the floor
- Chap tells security to lock down the whole room
- Cult can't do anything because it's an early stage
- Round turns into greenshift
- Nobody gets to have fun

## Technical details
Moved the method for scribing glyphs from shared to server (it never worekd on client anyway), check current tier, if it's lower than 3, add VisibillityComponent and set it to monument's vismask.

## Media
Non-cultist, no monu, no glyph
изображение
Cultist, glyph visible
изображение
Non-cultist, stage 3, glyph visible
изображение

## Requirements

**Changelog**
:cl:
- tweak: Cosmic cult's glyph are no longer visible to non-cultists during the first and second stage.

### DIFF
--- a/Content.Server/_DV/CosmicCult/MonumentSystem.cs
+++ b/Content.Server/_DV/CosmicCult/MonumentSystem.cs
@@ -494,7 +494,7 @@ public sealed class MonumentSystem : SharedMonumentSystem
         ent.Comp.CurrentGlyph = glyphEnt;
         var evt = new CosmicCultAssociateRuleEvent(ent, glyphEnt);
         RaiseLocalEvent(ref evt);
-        
+
         var objectiveQuery = EntityQueryEnumerator<CosmicTierConditionComponent>();
         while (objectiveQuery.MoveNext(out var _, out var objectiveComp))
         {
--- a/Content.Server/_DV/CosmicCult/MonumentSystem.cs
+++ b/Content.Server/_DV/CosmicCult/MonumentSystem.cs
@@ -14,6 +14,7 @@ using Content.Shared._DV.CosmicCult.Prototypes;
 using Content.Server._DV.Shuttles.Events;
 using Content.Shared.Audio;
 using Content.Shared.Damage;
+using Content.Shared.Eye;
 using Content.Shared.Interaction;
 using Content.Shared.Popups;
 using Content.Shared.Stacks;
@@ -21,6 +22,7 @@ using Content.Shared.UserInterface;
 using Robust.Server.GameObjects;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Configuration;
+using Robust.Shared.Map.Components;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
 
@@ -44,6 +46,9 @@ public sealed class MonumentSystem : SharedMonumentSystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
+    [Dependency] private readonly IPrototypeManager _prototype = default!;
+    [Dependency] private readonly SharedCosmicGlyphSystem _glyph = default!;
+    [Dependency] private readonly VisibilitySystem _visibility = default!;
 
     private static readonly EntProtoId CosmicGod = "MobCosmicGodSpawn";
     private static readonly EntProtoId MonumentCollider = "MonumentCollider";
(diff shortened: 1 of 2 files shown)


Content.Server/_DV/CosmicCult/MonumentSystem.cs
```diff
@@ -437,7 +442,12 @@ public void MonumentTier3(Entity<MonumentComponent> uid)
             if (TryComp<PolymorphedEntityComponent>(leader, out var polyComp) && TryComp<CosmicCultLeadComponent>(polyComp.Parent, out var polyLeaderComp))
                 _actions.RemoveAction(polyComp.Parent, polyLeaderComp.CosmicMonumentMoveActionEntity);
         }
-
+        if (uid.Comp.CurrentGlyph is EntityUid glyphEnt && TryComp<VisibilityComponent>(glyphEnt, out var visibility)) // Make the currently scribed glyph visible to everyone
```
deltanedas, line 445: {} instead of EntityUid

Content.Server/_DV/CosmicCult/MonumentSystem.cs
```diff
@@ -462,4 +472,42 @@ public void ReadyFinale(Entity<MonumentComponent> uid, CosmicFinaleComponent fin
...
+
+        var glyphEnt = Spawn(proto.Entity, _map.ToCenterCoordinates(xform.GridUid.Value, targetIndices, grid));
+        ent.Comp.CurrentGlyph = glyphEnt;
+        var evt = new CosmicCultAssociateRuleEvent(ent, glyphEnt);
+        RaiseLocalEvent(ref evt);
+
+        var objectiveQuery = EntityQueryEnumerator<CosmicTierConditionComponent>();
+        while (objectiveQuery.MoveNext(out var _, out var objectiveComp))
+        {
+            if (objectiveComp.Tier <= 2) // If we're below stage 3, make the glyph invisible to non-cultists
```
deltanedas, line 501: you already have the monument with ent dont need to check objectives for tier

Content.Server/_DV/CosmicCult/MonumentSystem.cs
```diff
@@ -462,4 +472,42 @@ public void ReadyFinale(Entity<MonumentComponent> uid, CosmicFinaleComponent fin
 
         _popup.PopupCoordinates(Loc.GetString("cosmiccult-finale-ready"), Transform(uid).Coordinates, PopupType.Large);
     }
+
+    protected override void OnGlyphSelected(Entity<MonumentComponent> ent, ref GlyphSelectedMessage args)
```
deltanedas, line 476: you can still have it in shared by checking the monument for tier and using predicted spawn

deltanedas (changes requested):

Content.Server/_DV/CosmicCult/MonumentSystem.cs
NoElkaTheGod: MonumentComponent doesn't store the current tier for some reason. Though, now that I think about it, I could totally add that.

Also I've heard that predicted spawn doesn't work well with animations.

Proxysseia:
Direction 24h

Content.Server/_DV/CosmicCult/MonumentSystem.cs
deltanedas: try it and see
if it looks bad then `if (_net.IsClient) return;` and comment that animated spawns being predicted looks bad

NoElkaTheGod: As stated in the current comment, it didn't work on the client anyway, probably because one of the checks already called return. I may add this additional check for clarity though.

NoElkaTheGod:
Doesn't seem like 24h to me

Proxysseia:
This kills the unfun coinflip of 'someone saw you paint a glyph early on and now sec is swarming your cult, 50 50 our steward can move the monument'
continues matching what the player knows with what the character knows, which should lead to less cult related frustration.

Direction approved.

Content.Server/_DV/CosmicCult/MonumentSystem.cs
NoElkaTheGod: Ran into another issue, VisibilitySystem is in server namespace, and I don't think I can just set the values on the component either.

PR #4284, opened 2025-08-28
Some loosely connected cosmic cult tweaks

NoElkaTheGod:
## About the PR
Malign rifts now "pulse" twice as often on average. "Pulses" now affect all mobs, not just humanoids (except for cultsits, chaplains, and colossi). Pulses deal ion damage now (same amount as asphyx).
Purging malign rifts as chaplain now takes 20 seconds. Non-chaplains still need 35 seconds. Absorbing rifts also takes 20 seconds now (down from 35).
Vacuous chantries, vacuous spires and multispectral inhibitors are no longer destructible. Instead, they can be purged with a bible, just like a malign rift (the latter two take less time: 5 seconds for chaplains, 10 seconds for non-chaplains). Oh, yeah, cult glyphs can also be removed in the same way. Because why not. Kinda makes sense doesn't it?

## Why / Balance
Damage from malign rifts was way too low, especially on IPCs who don't asphyxiate, and did not have the intended effect. Hopefuly now they should be scarier, and not only for humanoids, but pretty much any creature. Piles of dead mice around rifts may also become a thing.
Time difference for purging rifts for chaplains was already in the component, but it didn't work as intended. Absorbtion time was also lowered, because "staring at the rift for almost a minute" isn't really engaging gameplay for anyone involved.
Getting rid of cult objects is currently very unintuitive. Rifts need the chaplain to shout at them, chantries need to be broken by damaging them, and don't even get me started on effigies (not changing those ones, though, that's for a different PR). Now, if you see a weird cult thing, the solution is pretty much always to call the chaplain, or just smack it with a bible yourself if you have one.
Chantries especially are quite problematic right now. Cultists don't even bother defending them, and instead just start the ritual and skebaddle. Security then gets to the chantry, but due to either the metashield, or literally not knowing what to do (it's really tanky, so you can just magdump it, see no effect, and think that it's not how you destroy it), end up just staring at it until the ritual completes.

## Technical details
"Kill this thing with a bible and a long do-after" has been moved from the CosmicMalignRiftComponent to a new CosmicPurifiableComponent. Said component was applied to a few other entities. Aside from that, it's just numbers and other minor tweaks.

## Media
There isn't really that much to show here.

## Requirements

**Changelog**
:cl:
- tweak: Malign rifts are more dangerous now.
- tweak: Chaplains and mystagogues now have an easier time purging malign rifts. Cultists also absorb them faster now.
- tweak: Vacuous chantry, multispectral inhibitor and vacuous spire are no longer destructible. Instead, you have to purge them with a bible, just like a malign rift.
- tweak: Cosmic cult's glyphs can now be purged with a bible.

### DIFF
--- a/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
+++ b/Content.Server/_DV/CosmicCult/Components/CosmicMalignRiftComponent.cs
@@ -14,29 +14,17 @@ public sealed partial class CosmicMalignRiftComponent : Component
     [DataField]
     public bool Occupied;
 
-    [DataField]
-    public EntProtoId PurgeVFX = "CleanseEffectVFX";
-
-    [DataField]
-    public SoundSpecifier PurgeSound = new SoundPathSpecifier("/Audio/_DV/CosmicCult/cleanse_deconversion.ogg");
-
     // [DataField]
     // public EntProtoId GrailID = "NullRodGrail"; // Not implemented at this time
 
     [DataField]
-    public TimeSpan BibleTime = TimeSpan.FromSeconds(35);
+    public TimeSpan AbsorbTime = TimeSpan.FromSeconds(20);
 
     [DataField]
-    public TimeSpan ChaplainTime = TimeSpan.FromSeconds(20);
+    public TimeSpan MinPulseTime = TimeSpan.FromSeconds(10);
 
     [DataField]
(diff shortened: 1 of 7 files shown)


Proxysseia:
24h

Proxysseia:
makes rifts more dangerous, cool. gives chaps more to do on the chap gamemode, cool.
direction approved

deltanedas (approved):

PR #4291, opened 2025-08-30
Make mindwiping illegal again

kotobdev:
## About the PR
Mindwiping is now *explicitly* stated to be illegal in Space Law.
`To maliciously tamper with or create noöspheric anomalies, entities, aberrations, or other elements; Or to otherwise deliberately perform actions that would raise Glimmer or other anomalous reading levels in a way that threatens normative reality. `
to 
`To maliciously tamper with or create noöspheric anomalies, entities, aberrations, or other elements; to unlawfully mindwipe an entity; or to otherwise deliberately perform actions that would raise Glimmer or other anomalous reading levels in a way that threatens normative reality.`

## Why / Balance
current grammar makes it pretty easy to argue that the mantis shooting every psionic on code green is fine because it isn't malicious, psionics aren't really explicitly protected under the current wording, and that since it lowers glimmer its good and contradicts the last part

im using "mindwiping" and not the old wording "To unlawfully and maliciously rid a psionic of their powers" since mindwiping is previously defined in SL and includes more then *just* removing psionics

## Requirements

no!!!!!!!!!

**Changelog**

:cl:
- tweak: Mindwiping is now more explicitly illegal under Space Law.

### DIFF
--- a/.wiki/_DV/Laws/SpaceLaw.txt
+++ b/.wiki/_DV/Laws/SpaceLaw.txt
@@ -221,7 +221,7 @@ Sentencing modifiers are to be applied by the sentencing officer, judge, or arbi
 | style="border:1px solid black;" | [[File:SL_Mindbreaking.png]]
 | style="border:1px solid black;" | Noöspheric Tampering
 | style="border:1px solid black;" | 14 minutes
-| style="border:1px solid black;" | To maliciously tamper with or create noöspheric anomalies, entities, aberrations, or other elements; Or to otherwise deliberately perform actions that would raise Glimmer or other anomalous reading levels in a way that threatens normative reality.
+| style="border:1px solid black;" | To maliciously tamper with or create noöspheric anomalies, entities, aberrations, or other elements; to unlawfully rid a psionic of their powers; or to otherwise deliberately perform actions that would raise Glimmer or other anomalous reading levels in a way that threatens normative reality.
 |-
 | style="border:1px solid black;" | 305
 | style="border:1px solid black;" | [[File:SL_Sabotage.png]]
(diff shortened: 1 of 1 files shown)


Proxysseia:
currently a mantis mindwiping everyone on code green is 'only' abuse of power and assault. 
I'll take this in with direction, just keep in mind that old mindwiping was just a more specific version of assault.
24h

kotobdev:
> currently a mantis mindwiping everyone on code green is 'only' abuse of power and assault. I'll take this in with direction, just keep in mind that old mindwiping was just a more specific version of assault. 24h

okay yeah fair the "shooting everybody" example is perhaps a bit extreme
point is: since code white mindwipe ROE lift exists, its very much implied that mindwiping is supposed to be it's own crime and not assault, especially considering the actual act causes no/minimal harm. best to clarify that the removal of psionics/mindwiping still isn't something you can do randomly

Proxysseia:
Sorry for the delay.

Noöspheric tampering is currently without contest the longest law we have. We want to have any aspect of video game legalese to be as short, concise, and simple to read and remember as possible, because frankly we cant expect jane secoff to perfectly learn and remember 40 pages of all of this.

The redundancy that this PR adds would be nice in a vacuum, but we do not find sufficient reason to make the longest law we have even longer.
Mindwiping is done exclusively via applying poison to someone, through more or less violent means (the least violent is still forced ingestion) which means that any instance of mindwiping consists of unwanted contact that leads to both physical and psyonic? mental? harm. Mindwiping always falls under assault, except under code white, at which point you may mindwipe at your leisure.

For those reasons, the PR is denied.

PR #4295, opened 2025-08-30
Add examine to IdentityBlocker

NoElkaTheGod:
## About the PR
You can now see if an item covers your face, eyes, or mouth

## Why / Balance
With some items it's really not so intuitive. Medical HUDs, for example, do not cover eyes, while chemical goggles do. This can be rather frustrating at times, especially with the cosmic cult, where you want your eyes covered BEFORE they start glowing, but have no way of confirming that they are covered until they start glowing.

## Technical details
There are switch statements in fluent, apparently. Neat.
Direction review probably not needed since this is a small QoL addition.

## Media
Not like it has literally any effect on gameplay, but it's the same component.
изображение

Yes, they aren't just useless vanity, they're the only eyewear that covers your whole face.
изображение

The most important one.
изображение

## Requirements

**Changelog**
:cl:
- add: You can now examine items to see if they cover your face (or parts of your face, like eyes).


deltanedas (changes requested):
put the code in its own system gamer

NoElkaTheGod:
I sure hope this is the correct way to do it, because I don't exactly have time to test it right now

Proxysseia (approved):

deltanedas (approved):

PR #3871, opened 2025-05-30
Remove the LO knuckledusters from the locker

Radezolid:
## About the PR
Title.

## Why / Balance
Why would we give weaponry to the head of logistics. They shouldn't be carrying weapons unless there is a proper threat, which is security's duty to handle.

## Technical details
N/A

## Media
N/A

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**

:cl:
- remove: Removed LO's knuckledusters from their locker.

PR #4313, opened 2025-09-02
Revert "Remove the LO knuckledusters from the locker"

SirWarock:
## About the PR

Reverts https://github.com/DeltaV-Station/Delta-v/pull/3871
## Why / Balance

I think we have grown and DeltaV has healed - And there's no better way to show this than reverting the most unpopular change that has been made. This will singlehandedly aurafarm, goodwillfarm, imagefarm better than anything else the server could do for absolutely zero downsides or costs.

Let's show the community that we're prioritizing fun again.
## Technical details

Just a revert.
## Requirements

## Breaking changes

**Changelog**

:cl:
- add: added the LO knuckle dusters back to the LO's locker! Life is good again

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
@@ -20,7 +20,7 @@
     - id: RubberStampDenied
     - id: RubberStampQm
     - id: AstroNavCartridge
-    #- id: ClothingHandsKnuckleDustersQM # DeltaV - No weaponry for LO.
+    - id: ClothingHandsKnuckleDustersQM
     #- id: MailTeleporterMachineCircuitboard # DeltaV - upstream deliveries not ported yet
 
 - type: entity


Quanteey (approved):
Real

Proxysseia:
24h

NoElkaTheGod:
Obligatory "DeltaV is healing" comment

Proxysseia:
Knuckledusters come with two problems, one of them is fixable, one of them is not.
The first, and lesser, problem is that they deal a lot of damage. They occupy the glove slot, which is not used for anything except insuls, they cannot be disarmed, they have good dps. all in all they are an outstanding melee weapon. This is part of why people enjoyed the duster: they could beat people to a pulp very easily and without risk of the weapon being turned against them. 
This is fixable by nerfing the weapon down to just being fancy brass knuckle dusters.

The second problem, and the actual crux of the issue, is that knuckledusters are a special item for the LO. It comes in the locker, its a prestige object you get for being the LO; theyre the *LO's* knuckledusters. What that communicates to the LO is that they are authorized and able to beat people for various reasons. If something is in your locker, then because you get to use it, if something spawns on your person, then because you get to use it. Cadets spawn with a baton and are expected to use it. They are not expected to use lethal ammunitions and hence do not spawn with them. The HoS spawns with their special sidearm, and they are expected to use it, etc. 
Non security generally, but especially Command (with their heightened RP standards) are expected to tell security when a threat appears instead of trying to handle the threat itself. If someone is trespassing into your department you are supposed to call sec, and to only beat them if they are an immediate risk and fall under self defence, or if sec is not showing up.
This is not an issue of 'LO can get weapons anyways, so why not give them one' We know that LO has access to swords, and knives, and even guns on some maps. We know that LO can with very little effort order a brass crate and craft a set of knuckle dusters. Aquiring this weaponry is a choice that the LO has to opt into. It is not the default assumption of the role. The LO has as easy access to weaponry as does the HoP, and still we see neither run around with guns and knives and whatever else they can find. This is a good thing. And this is because those weapons you have to go out of your way to aquire. 
Putting a weapon, no matter how bad, into the LO's (or any non security command members) locker, communicates they are expected to beat people. 
And that is not something we want from our command.

This PR is denied.

NoElkaTheGod:
DeltaV is not, in fact, healing. Boowomp.

SirWarock:
> DeltaV is not, in fact, healing. Boowomp.

It's Joever. Perhaps another day.

HTMLSystem:
We wanted to suggest that you could maybe make it salv loot. Though it'd need to be renamed (don't think it needs a resprite since they're just gold, but we could be wrong).

SirWarock:
There was also the suggestion by @Proxysseia to make them craftable instead.

Which makes me wonder.
What if I give the LO a plastic version of it into their locker? So, they still look as shiny, people can waltz around with the golden knuckles as a status symbol - But LOs that'd like to rule with a shiny fist of actual metal could take like 5 gold bars and a diamond to then craft it with the plastic version to make the real deal? It'd be the LO's knuckle dusters, but they wouldn't spawn with it

PR #4318, opened 2025-09-02
Fix Loto Oil

SirWarock:
## About the PR

Fixed Lotooil slurred speech persisting after it wore off.
## Why / Balance

Bugfix.
## Technical details

YML change
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Lotophago Oil sometimes making you permanently slurring your words!


Stop-Signs:
Since this is a bugfix it does not require direction review

Bonktrauma:
This has existed since prebase, kinda sad to see it go.

SirWarock:
This doesn't require comments, right? This is just Nyanotrasen code, do we comment that?

deltanedas:
no

SirWarock:
Beep boop

Resources/Prototypes/Nyanotrasen/Reagents/psionic.yml
```diff
@@ -108,4 +105,4 @@
         probability: 0.4
         damage:
           groups:
-            Toxin: 2
\ No newline at end of file
+            Toxin: 2
```
TehFlaminTaco: Can we untouch the EOF?

TehFlaminTaco (changes requested):

SirWarock:
Done!

TehFlaminTaco:
This PR also seems to remove the effect of Loto causing you to repeat what anyone says in Telepathy, is this part not intentional?

SirWarock:
> This PR also seems to remove the effect of Loto causing you to repeat what anyone says in Telepathy, is this part not intentional?

I can reintroduce it. I don't think it ever worked.
I also made a separate fix for being drunk, since being drunk and under loto oil caused the same issue

TehFlaminTaco:
> I can reintroduce it. I don't think it ever worked. I also made a separate fix for being drunk, since being drunk and under loto oil caused the same issue

It definitely works in live. If you can re-add it to maintain functionality that'd be preferred, otherwise we should consult Direction to see if they're okay with losing that.

SirWarock:
Replaced the keys from ScrambledAccent and TelepathicRepeater, because those are bugtraps. Anything that uses the same keys will cause issues.

TehFlaminTaco (approved):
Good fix, lgtm

PR #4319, opened 2025-09-02
Added Scanners to Modules (#1268)

SirWarock:
## About the PR

Added Medical Scanners to Surgery Modules.
Also gave them a speed modifier of 2x.
CMO with 1.5x role and 2x gloves modifier is still better, which I dislike.
## Why / Balance

Seeing the damage on the patient while working on them is a good thing.
## Technical details

YML changes.
## Media

grafik

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added Health Scanners to Surgery Cyborg Modules!


SirWarock:
Also, while I'm here, a question for direction - Can we just add a medical multitool for the Advanced Surgery Module? It's not like they need to switch hands while working on a patient, since they're holding all tools.

Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml
```diff
@@ -72,6 +72,8 @@
   components:
   - type: BorgSwitchableType
     selectedBorgType: medical
+  - type: SurgerySpeedModifier
+    speedModifier: 2 # DeltaV - +0.5 than CMO, Borgs are specialized.
```
Quanteey:
```suggestion
  - type: SurgerySpeedModifier # DeltaV - +0.5 than CMO, Borgs are specialized.
    speedModifier: 2
```

Proxysseia:
direction 8h

Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml
```diff
@@ -72,6 +72,8 @@
   components:
   - type: BorgSwitchableType
     selectedBorgType: medical
+  - type: SurgerySpeedModifier # DeltaV - +0.5 than CMO, Borgs are specialized.
+    speedModifier: 2
```
Proxysseia: Remove the speed modifier.

Proxysseia (changes requested):

Quanteey (dismissed):

Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml
SirWarock: Is that change inspired by the removal of modifiers from CMO and Surgeon?
If so, I'd like to point out that Cyborgs cannot wear the mechanical surgical gloves.
Cyborgs will always be slower, as well as unable to do organ transplants, which I think is wrong given how Cyborgs are meant to be specialized and effective at their niche.

Proxysseia: its related to the sped modifier removal from CMO and Surgeon, it follows the same line of thought.
Borgs not being able to do organ transplants is something you could work on if you want. You could look into the way upstream handles borg hands? free hand with a very extensive whitelist.

SirWarock: I think simply giving organs an organ, and limbs a limb tag would make it easily achievable.
Unless Upstream Borg hands work with proto IDs and not tags, oh Jesus.

Also, I'm gonna remove the speedmodifier as you wish, but I dislike the change. Medborgs will now be 50% slower than anyone with Mechanical gloves. Which may be intended, but I believe that Borgs should be better than Humanoids at their job, for the price of being a borg

SirWarock:
DIREEEECTIOOON

Proxysseia (approved):

SirWarock:
Approvedops

SirWarock:
@TehFlaminTaco PRAISED BE OUR LORD AND SAVIOR

TehFlaminTaco (approved):
Yamlops

PR #4323, opened 2025-09-03
Rebalance desword some more

NoElkaTheGod:
## About the PR
Made desword even better at cutting people, even worse at cutting everything else. It now deals 45 damage per second (up from 36), 50% more than the esword, but has no structural damage at all.

## Why / Balance
People, like, stopped using the normal esword altogether, because desword was pretty good at dealing with structures as well, and, quote, "it looks cool". Thus, I made it even worse at dealing with structures, and made it better at it's primary role of killing people. Maybe 45DPS is a bit too much, but eh, it's still like two times worse than any actual gun. Give us some melee options that actually don't suck.

## Requirements

**Changelog**
:cl:
- tweak: Double-bladed energy sword is now better at killing people, but worse at dealing with structures,

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
@@ -312,9 +312,8 @@
         variation: 0.250
     activatedDamage:
         types: # DeltaV - was 12 slash, 12 heat, 15 structural
-            Slash: 9
-            Heat: 9
-            Structural: 8
+            Slash: 15
+            Heat: 10
   - type: ItemToggleActiveSound
     activeSound:
       path: /Audio/Weapons/ebladehum.ogg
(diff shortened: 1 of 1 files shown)


Axianamos:
Make it 25 DPS one handed, 50 when wielded. Its a double bladed energy sword. You dont use those one handed. That way you could keep the structural damage when wielded, make it an actual trade off for the normal esword

NoElkaTheGod:
I see no reason why you wouldn't be able to use it in one hand just because it's double-bladed. The blade is weightless, so if you can hold the handle in one hand, you can still hold it in one hand when lit. Also you can't really have it lit but not weilded, so there's that.

Proxysseia:
24h

NoElkaTheGod:
Welp, I guess we chose discord drama over game development. Not big surprise. Time to leave and slam the door then. I will not return, as you clearly don't want me to.

HTMLSystem:
Apolgizes that we don't take to kindly to people who are transphobic. Remember buddy, you digged your own grave.

NoElkaTheGod:
I hope you're happy that I will be remembered for being "transphobe" and "bigot", and not for my contributions. Not like this ever was the case, but you guys would antagonize anyone who does not 100% agree with you. If you have anything else to say, feel free to DM me in discord (but reeber that I can block peolpe, so don't waste your time trying to prove how wrong I was). This is not the right place to discuss that.

Quanteey:
@NoElkaTheGod @HTMLSystem Please refrain from discussing this here, or anywhere else on this repo for that matter. Keep stuff on-topic

PR #4324, opened 2025-09-03
Changes the default white shoes for Reporter and Zookeeper

mnva0:
## About the PR
The white shoes are horribly ugly as placeholders. Reporter and Zookeeper have them, and this PR replaces them. Also cleans up forgotten traces of the default white Atmospheric Tech shoes in the AtmosDrobe.

I'm not sure if this needs a changelog? I'll go with yes for now and I can just trim that off if I don't

Putting black shoes on the Zookeeper is a bit ehh but it's better than the white shoes
## Media
image
image
image

## Requirements

**Changelog**
:cl: Minerva
- tweak: Reporter and Zookeeper now have black shoes instead of white shoes.
- fix: The AtmosDrobe now correctly stocks workboots instead of white shoes.


mnva0:
Actually this should probably go upstream

PR #4326, opened 2025-09-03
Fix missing locale for undergarment

mnva0:
## About the PR
Fixes the missing FTL entries for undergarments. It was apparently caused by having "-(blah blah)" appended to each, which no longer works or something? Whatever. Works after this PR.

I think I added the comment right? I'm not sure what's the best way of doing DV comments for, like... filewide stuff
## Media
image
image
image
image

## Requirements

### DIFF
--- a/Resources/Locale/en-US/markings/undergarment.ftl
+++ b/Resources/Locale/en-US/markings/undergarment.ftl
@@ -1,5 +1,3 @@
-# DeltaV: Fixed missing FTL entries for undergarments. About all lines edited
-
 marking-UndergarmentTopTanktop = Tanktop
 marking-UndergarmentTopBinder = Binder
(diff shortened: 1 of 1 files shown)


Quanteey (changes requested):
https://github.com/space-wizards/space-station-14/pull/38720
This is the same thing you did upstream that is already merged there, right? If so, I don't think you need the # deltaV comments

mnva0:
Oh, right, forgot I used to contribute to Wizden. Do I just remove those comments, @Quanteey? Or do I not have to worry about the PR? Not sure if DV does upstream merges in any way, or if those are even possible

Quanteey (approved):
I think this is good but I'll wait for deltanedas to double check

TehFlaminTaco (approved):

PR #4340, opened 2025-09-07
Nerf prober explosion lowering glimmer

kotobdev:
## About the PR
Prober explosions now lower glimmer by 6%-8%, down from 10%-15%.

## Why / Balance
quote proxy in the discord in regards to the [xenoarch exponential point output thing](https://github.com/DeltaV-Station/Delta-v/pull/4164):
> If the RA wants to extract at 900, he should be able to afford speedboots in the 90 seconds he has left in his body.

i agree with this statement. which is why it's a bit lame that when the prober blows, it automatically blasts glimmer down below mindswap levels instantly. nine times out of ten, the prober will blow before the mindswap event gets a chance to trigger. 
this change makes it so if the prober blows between roughly 950-1000, glimmer won't always be instantly reduced back to safe levels. epi will have to rely on their other methods of lowering glimmer (see: mindbreaking) to get things safe again
## Technical details
nyano code no comments bweeeeeh

## Media
nope

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Prober explosions now lower glimmer substantially less.

### DIFF
--- a/Content.Server/Nyanotrasen/Psionics/Glimmer/GlimmerReactiveSystem.cs
+++ b/Content.Server/Nyanotrasen/Psionics/Glimmer/GlimmerReactiveSystem.cs
@@ -222,7 +222,7 @@ namespace Content.Server.Psionics.Glimmer
             var slope = (float) (11 - _glimmerSystem.Glimmer / 100);
             var maxIntensity = 20;
 
-            var removed = (float) _glimmerSystem.Glimmer * _random.NextFloat(0.1f, 0.15f);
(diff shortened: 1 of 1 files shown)


Quanteey (approved):

Proxysseia:
24h

Proxysseia:
blowing up the prober should not be the be all end all for managing glimmer. while we are wary of nipping at sources of lowering glimmer while there are still so few in the game, we ulitimately think this will lead players to engage with our other (better) sources of managing the glimmer level on station.
direction approved.

PR #4352, opened 2025-09-09
Bugfix microwaves staying locked out after cooking finishes

### LINKED ISSUE #4201: Microwaves stop up when they finish cooking
## Description
Exactly as the description says. After a microwave is started at any time, it will pretty much freeze up when it finishes.

**Reproduction**
Step 1: Toggle microwave.
Step 2: Weep.

**Screenshots**
https://github.com/user-attachments/assets/d39b7afa-06b5-44fa-911f-666d86d85962

cmorley191:
## About the PR

Microwaves currently block you from ejecting their contents after cooking finishes. This is annoying for recipes like Baguette, where you cook Dough with a Beaker of salt and pepper. The now-empty beaker gets stuck in the microwave after cooking.

(see Media for demonstration)

The current workaround is to insert an item, which then lets you eject again. This is annoying.

Bugfix makes it so microwaves now eject all their contents when cooking completes (which is what upstream Wizden and other forks do / is the original behavior before factorio code).

## Why / Balance

fixes #4201 

## Technical details

- Added StopCooking() calls to various points in MicrowaveSystem where ejection or UI updates occur, so the microwave is stopped before "is microwave active" checks are performed.
- Because StopCooking() works by *deferring* a removal of the ActiveMicrowaveComponent, I also adjusted the two places that check for an ActiveMicrowaveComponent to check not only for the presence of the component but also whether it has a deferred removal queued up (these checks are in UpdateUserInterfaceState()'s IsMicrowaveBusy, and the MicrowaveEventsSystem factorio failsafe)

## Media

Aside from the typical microwave behavior, there are 4 edge cases which I've tested:

### Behavior 1: normal microwave behavior

**Current behavior**: microwave stays locked out after cooking, until you "kick" it (insert something, cut power, etc.)
old_normal

**New behavior**: microwave ejects all contents after cooking (matches wizden)
new_normal

### Behavior 2: unanchoring a running microwave

**Current behavior**: items stuck inside, UI is weird
old_wrench

**New behavior**: all contents eject when unanchored (matches wizden)
new_wrench

### Behavior 3: explosion from microwaving metal

**Current behavior**: items stuck inside, UI is weird
old_explosion

**New behavior**: all contents eject when exploded (matches wizden)
new_explosion

### Behavior 4: robotic arm extraction while cooking

**Behavior maintained**: robotic arm is not allowed to pull items out of a running microwave

Before fix:
old_robotic_active

Still works after fix:
new_robotic_active

### Behavior 5: robotic arm extraction while off

**Behavior maintained**: robotic arm can pull out items when microwave is *not* running

Before fix:
old_robotic_inactive

Still works after fix:
new_robotic_inactive

## Requirements

## Breaking changes

N/A: no renames, etc.

**Changelog**

:cl:
- fix: microwaves now eject their contents after cooking, instead of locking them in

### DIFF
--- a/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
+++ b/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
@@ -444,8 +444,14 @@ namespace Content.Server.Kitchen.EntitySystems
 
         private void OnAnchorChanged(EntityUid uid, MicrowaveComponent component, ref AnchorStateChangedEvent args)
         {
-            if (!args.Anchored)
+            if (!args.Anchored) {
+                // DeltaV - start of microwave ejection bugfix
+                // DeltaV's MicrowaveEventsSystem changes prevent ejection from active microwave, so stop cooking first
+                StopCooking((uid, component));
                 _container.EmptyContainer(component.Storage);
+                UpdateUserInterfaceState(uid, component);
+                // DeltaV - end of microwave ejection bugfix
+            }
         }
 
         private void OnSignalReceived(Entity<MicrowaveComponent> ent, ref SignalReceivedEvent args)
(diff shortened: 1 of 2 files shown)


cmorley191:
Forgot to mention, this PR fixes this issue: https://github.com/DeltaV-Station/Delta-v/issues/4201

HTMLSystem:
In the "Why/Balance" section, please add "fixes [issue number]"! It'll automatically close the issue once this is merged!

Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
```diff
@@ -444,8 +444,14 @@ private void OnPowerChanged(Entity<MicrowaveComponent> ent, ref PowerChangedEven
 
         private void OnAnchorChanged(EntityUid uid, MicrowaveComponent component, ref AnchorStateChangedEvent args)
         {
-            if (!args.Anchored)
+            if (!args.Anchored) {
```
Quanteey: The opening { should be on a new line:
```
if (...)
{
    ....
```

Quanteey (changes requested):
Awesome PR, thank you!

Quanteey (approved):

deltanedas (approved):

PR #4356, opened 2025-09-09
IPC - Pressurized Micro Pumps

EmberAstra:
## About the PR
Added the pressurized micro pump, a new IPC organ unlocked from the Cybernetic Enhancements research. It can be printed at the Exosuit Fabricator, and with surgery can replace the standard IPC micro pump to give them pressure immunity.

Also, made it so you can actually insert micro pumps in IPCs, you could only remove them before.

## Why / Balance
The IPC spacewalking nerf was... pretty controversial, and I thought this would be a nice compromise.
You'll still be able show off and go into space with just a coat, but it'll no longer be free and you'll have to rely on Epistemics doing their work and the roboticist having some time for surgery.
Organics can get pressure/breathing immunity in a similar way through space animal organs, and silicons can now have an equivalent to that.

## Technical details
Just .yml changes.

## Media
pump

## Requirements

**Changelog**

:cl:
- add: Added pressurized micro pumps, allowing IPCs to be pressure immune. Ask your local roboticist for one!
- fix: micro pumps can now be inserted in IPCs and not just removed

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
+++ b/Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
@@ -170,9 +170,23 @@
     - type: AugmentArm
 
 - type: entity
+  parent: SurgeryRemoveHeart
+  id: SurgeryRemovePump
+  name: Remove Pump
+  components:
+  - type: SurgeryOrganSlotCondition
+    organSlot: pump
+
+- type: entity
   parent: SurgeryInsertHeart
   id: SurgeryInsertPump
-  name: Insert Heart
+  name: Insert Pump
   components:
+  - type: Surgery
+    requirement: SurgeryOpenRibcage
+    steps:
+    - SurgeryStepSawBones
+    - SurgeryStepInsertPump
+    - SurgeryStepSealOrganWound
   - type: SurgeryOrganSlotCondition
     organSlot: pump
--- a/Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
+++ b/Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
@@ -168,3 +168,11 @@
   - type: SurgeryOrganCondition
     organ:
     - type: AugmentArm
+
+- type: entity
+  parent: SurgeryInsertHeart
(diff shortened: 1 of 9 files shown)


SirWarock:
I like this, but it should cost gold

EmberAstra:
I left the recipe as the default one for now, can make it more expensive if it's needed!

Proxysseia:
24h

Proxysseia (changes requested):
change cost to 20 steel, 10 plasma, 10 plastic, 5 gold

EmberAstra:
Done!

Proxysseia (approved):
Puts IPC surgery on par with organic surgery. Direction approved.

Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
```diff
@@ -168,3 +168,11 @@
...
     organ:
     - type: AugmentArm
+
+- type: entity
+  parent: SurgeryInsertHeart
+  id: SurgeryInsertPump
+  name: Insert Heart
+  components:
+  - type: SurgeryOrganSlotCondition
+    organSlot: pump
```
deltanedas: insert pump not heart

deltanedas (changes requested):

Resources/Prototypes/_DV/Entities/Surgery/surgeries.yml
EmberAstra: I kept it as Heart due to the removal operation being the same as organics and so also saying Heart.
Should I change that one too?

EmberAstra: ...went ahead and just did that actually. No more hearts.

deltanedas (approved):

AveriV2:
i think this should be mapped in the nukie hideout

AveriV2:
i think this should be mapped in the nukie hideout

SirWarock:
I don't think so. They aren't mapped there for the same reason space lungs aren't mapped there, no?

PR #4375, opened 2025-09-11
[PORT] Anomalites from Impstation

AeraAuling:
## About the PR
Ports over Anomalites from impstation, little sprite creatures summoned from anomaly cores that are loyal to whoever holds the core. 

(forgot to include the PRs ported)

https://github.com/impstation/imp-station-14/pull/2313
https://github.com/impstation/imp-station-14/pull/2337
https://github.com/impstation/imp-station-14/pull/2390
https://github.com/impstation/imp-station-14/pull/2408

## Why / Balance
they're cute. also I've been wanting to see more familiars for forever. I still plan on making delta-v specific familiars, but these things fit so well with the direction we're going for things.

## Technical details
Modifies some of the anomaly code and the pinpointer code. All ported changes have the IMP comment style right now because I'm not actually sure on the specifics of how we comment ported changes to upstream code. Otherwise the majority of it is YAML code which is hopefully commented correctly. Also, the anomalites have their own pinpointer which was using Impstation's dragon pinpointer sprite, so I just went ahead and made them one of their own. I thought about PRing that to impstation first, but honestly it's not all that great and they'll probably make something better so consider it a placeholder until they do that. Also the pinpointer doesn't actually work right now I don't think, I don't know why, but I've left the code in to keep the port relatively simple.

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl: WidgetBeck, FaelingDraws, TheGrimbeeper
- add: Anomalites from Impstation are making their appearance in the delta sector. Hold on to their cores if you want to keep a hold on the anomaly sprites.


Proxysseia:
23h

Proxysseia:
very cute little guys, nice gameplay flavour

direction approved.

Content.Server/Pinpointer/PinpointerSystem.cs
```diff
@@ -5,7 +5,7 @@
 using Robust.Shared.Utility;
 using Content.Server.Shuttles.Events;
 using Content.Shared.IdentityManagement;
-
+using Content.Server.Bible.Components; // #IMP
```
TehFlaminTaco: Is this import needed?

TehFlaminTaco (changes requested):
Fix this and lgtm

Content.Server/Pinpointer/PinpointerSystem.cs
AeraAuling: apparently not, removed

TehFlaminTaco (approved):

PR #4379, opened 2025-09-12
make meteor swarm event actually hit the station

### LINKED ISSUE #4148: Meteors don't exist
## Description
Since the last upstream merge (roughly) the meteor event hasn't been working. It's all anecdotal of course but I haven't seen a single meteor hit a station in months and everyone I've asked has said the same thing.

**Reproduction**
Hear meteor announcement
Look inside
No meteors

**Screenshots**

**Additional context**
I'm not 100% on it being related to the last upstream merge and I haven't heard about meteors being broken upstream at any point recently.

cmorley191:
## About the PR

- Meteors in the swarm event will _actually_ be thrown at the station now (previously their target area was frequently very far from the station)
- Meteors don't experience "space friction" anymore (previously they slowed to a stop quite quickly after spawning)
- ~~New admin logs help admins realize meteors are a thing now, and debug any issues~~

## Why

Fixes https://github.com/DeltaV-Station/Delta-v/issues/4148 . This is an iconic SS14 event - it's weird that it doesn't work on DeltaV.

## Balance

This is a very reasonable (gentle) re-introduction of the feature, in my opinion.
- Meteors spawn at the edge of the playable area and make their way to the station over the course of 1-2 minutes.
- There's also the 1-minute delay between the announcement and the event starting.
- There are lots of things floating around in space in DeltaV, so a good chunk of the meteors may hit something else before they reach the station.
- Meteors have a coded-in targeting radius that is slightly wider than the station, so they intentionally miss sometimes (to create fun heart-attacks for EVA crew)
- The result of all of this is that there's only around a dozen actual strikes on the station per swarm.
- DeltaV maps have very good meteor shielding so very little actual damage is done (on the first swarm of a round).

## Technical details

`playableArea` was the culprit for the problem -- we have a lot more "stuff" in space than upstream does (they only have the station and the ATS, both very close to eachother). The code was using the center of the `playableArea` to target meteors; now it uses the Observer spawn point (which is basically the center of the station on all the maps).

Tested on all in-rotation maps, including glacier. (glacier has "dud" meteors that immediately evaporate - the event is a nothing-burger on glacier)

## Media

Full event example on Byoin: https://www.youtube.com/watch?v=fmSOHRj1gG0

First meteor impact is at 2:00

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Meteors will *actually* be on collision course with the station now! Brace for impact.


Content.Server/_DV/StationEvents/Events/MeteorSwarmRule.cs
```diff
@@ -15,38 +16,67 @@ public sealed class MeteorSwarmRule : StationEventSystem<MeteorSwarmRuleComponen
     {
         [Dependency] private readonly SharedMapSystem _map = default!;
         [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+        [Dependency] private readonly SharedTransformSystem _transform = default!;
+        [Dependency] private readonly IChatManager _chatManager = default!;
```
TehFlaminTaco: Should just be `_chat`

pootslap:
Awesome, I'm so glad to see meteors making a comeback. I'm still confused as to why it was an issue in the first place since the huge maps have existed for years (having room for the listening post etc.).

I do have one worry though. I tested it myself a week ago and the meteor swarm event DOES work but it only seems to work in the dev env. https://github.com/DeltaV-Station/Delta-v/issues/4148#issuecomment-3245401425
I guess there's no way of knowing without trying it in an actual round though.

cmorley191:
> I tested it myself a week ago and the meteor swarm event DOES work but it only seems to work in the dev env. [#4148 (comment)](https://github.com/DeltaV-Station/Delta-v/issues/4148#issuecomment-3245401425) I guess there's no way of knowing without trying it in an actual round though.

Yeah, I believe you. All I can say is on my thorough dev env testing they were rarely fast enough to reach the station before slowing down, and almost never targeted the right spot. I did have one round where the playable area was small enough that one meteor did hit, but the rest had targeting off by about 200 tiles and just flew by. You might've just gotten lucky with your test. It could be map-dependent, too.

Content.Server/_DV/StationEvents/Events/MeteorSwarmRule.cs
cmorley191: There's both a ChatManager and a ChatSystem; naming it `_chatManager` is what a lot of other files do (even files that don't use both). I think this is more clear - do you still prefer `_chat`?

TehFlaminTaco: I have no idea which is more correct, I was relaying what I've been informed of in the past.

cmorley191: Gonna leave this as is unless anyone super prefers `_chat`

TehFlaminTaco: I'd prefer if you removed the chat part entirely. Add logs if you have to, but we shouldn't bombard admins with this

TehFlaminTaco (changes requested):

TehFlaminTaco:
Is this still being worked on? I'd love to see this fix.

Toby222:
If this isn't gonna be finished I'll make a new updated PR soon :^)

cmorley191:
Updated to master, tested again on Academy, Arena, Asterisk, Byoin, Division, Edge, Elegance, Glacier (duds), Lighthouse, Pebble, Shoko, Terra, Hive, and Tortuga. Still works; reasonable damage, the youtube video in the PR description is still representative. Between 0-2 rooms tend to get spaced.

Content.Server/_DV/StationEvents/Events/MeteorSwarmRule.cs
cmorley191: Removed the admin logs.

Bonktrauma:
Neat

cmorley191:
Ready for re-review @TehFlaminTaco 

Also a couple things for the record:
- you trigger this with `addgamerule MeteorSwarm`
- if we need to tweak the balance of this event, the best way to do that is probably change the number of waves. That's `MinimumWaves` and `MaximumWaves` in `MeteorSwarmRuleComponent.cs`, or you can change it in the yaml: `minimumWaves` and `maximumWaves` in `Resources/Prototypes/_DV/GameRules/events.yml` `MeteorSwarmRule`

TehFlaminTaco (approved):
Finally fix these rocks, thank you!

PR #4389, opened 2025-09-14
escape pod doors open when evac shuttle arrives

### LINKED ISSUE #4256: Evac pod airlocks do not open when the shuttle is called / arrives
## Description

Evac pod doors are supposed to stay open the whole time, aside when evac departs. However, when a map starts up they will always start close. 
This is a big issue because the only way to open them is with externals access. Meaning anyone who doesn't have externals can't get in an evac pod. 
From talking to other, this is probably a DeltaV only issue. 
**Reproduction**

1. Start up any map
2. Walk to an evac pod
**Screenshots**

Image
Screenshot froma  round a few days ago. 
**Additional context**

cmorley191:
## About the PR

Fixes https://github.com/DeltaV-Station/Delta-v/issues/4256 (assuming people want it fixed with this approach)

Escape pod doors remain closed all shift, but now with this PR they will bolt open automatically when the evac shuttle arrives. They close automatically when the shuttle leaves - that part was already working and hasn't changed.

## Why / Balance

The closed pod doors are locked with Externals access requirement. This is a problem, because _all_ crew really ought to be able to use the escape pods to evacuate.

An alternative approach could've been to bolt open the doors for the entire shift, but I think this approach is better. Please discuss on https://github.com/DeltaV-Station/Delta-v/issues/4256 .

## Technical details

There's a part of the docking sequence that triggers the doors to bolt open after two crafts dock. This PR simply calls that piece of the code again when the evac shuttle arrives. It filters for the pod doors by looking for the EscapePodComponent, which is already part of the pod in `escape_pod_small.yml`.

## Media

escape_pod_arrive

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Escape pods can now be used by all crew, not just those with Externals access; all pod doors now open automatically when the evac shuttle arrives.

### DIFF
--- a/Content.Server/Shuttles/Systems/DockingSystem.cs
+++ b/Content.Server/Shuttles/Systems/DockingSystem.cs
@@ -278,6 +278,42 @@ namespace Content.Server.Shuttles.Systems
             dockA.Comp.DockedWith = dockBUid;
             dockB.Comp.DockedWith = dockAUid;
 
+            // DeltaV - start of escape pod docking
+            TryOpenDockedDoors(dockA);
+            // DeltaV - end of escape pod docking
+
+            if (_pathfinding.TryCreatePortal(dockAXform.Coordinates, dockBXform.Coordinates, out var handle))
+            {
+                dockA.Comp.PathfindHandle = handle;
+                dockB.Comp.PathfindHandle = handle;
+            }
+
+            var msg = new DockEvent
+            {
+                DockA = dockA,
+                DockB = dockB,
+                GridAUid = gridA,
+                GridBUid = gridB,
+            };
+
+            _console.RefreshShuttleConsoles();
+            RaiseLocalEvent(dockAUid, msg);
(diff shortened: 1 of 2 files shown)


Proxysseia:
neat

Content.Server/Shuttles/Systems/DockingSystem.cs
```diff
@@ -446,7 +464,7 @@ public bool CanDock(Entity<DockingComponent> dockA, Entity<DockingComponent> doc
 
             var xformA = Transform(dockA);
             var xformB = Transform(dockB);
-
+
```
TehFlaminTaco, line 467: Untouch this please

TehFlaminTaco (changes requested):

TehFlaminTaco:
Is this still being worked on?

TehFlaminTaco:
Closing for now. Re-open at a later date if this is being worked on.

PR #4393, opened 2025-09-15
allow ghosting if mind swap or telegnosis goes wrong

### LINKED ISSUE #4195: Unable to /ghost after dying while in telegnosis
## Description
stuck in telegnosis, cant ghost
things tried

relogging, ghosting, letting someone kill the body again, insul cap, removing the brain with surgery

related round #23417 apoapsis, happened to Foxington London

cmorley191:
## About the PR

When you use Telegnosis or Mind Swap and your original body dies, you become "trapped in this new vessel".

Now, with this PR, being in that state allows you to `/ghost`. (you couldn't ghost before, it wouldn't let you, meaning you were _literally_ trapped forever)

## Why / Balance

Fixes https://github.com/DeltaV-Station/Delta-v/issues/4195

Everyone should be able to ghost at basically all times -- now you can, you just need to use the "return to your original body" power first, if it's available to you. I think the idea behind that limitation was just so that you don't grief by mind swapping someone and then immediately `/ghost` to trap them.

## Technical details

I think this is the simplest approach - only block ghosting if you're able to use the "return to original body" power.

I considered touching other code in the mind swap system, but it's a spaghetti mess. The whole mind system (`.TransferTo`, `MindSwappedComponent`, `MassMindSwap`, etc.) needs a refactor.

## Media

https://github.com/user-attachments/assets/f871fadb-3424-4d25-afa8-21792d37b60a

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Dying in telegnosis no longer prevents you from ghosting

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
@@ -145,8 +145,8 @@ namespace Content.Server.Abilities.Psionics
             // DeltaV - start of trapped ghost fix
             // If you're able to swap back to your original body, you should swap back before you ghost.
             if (TryComp<MindSwappedComponent>(args.Mind.CurrentEntity, out var component)
-                && _actions.GetAction(component.MindSwapReturnActionEntity) is {} action
-                && action.Comp.AttachedEntity is {} user)
+                && _actions.GetAction(component.MindSwapReturnActionEntity) is { } action
+                && action.Comp.AttachedEntity is not null)
             {
                 args.Result = false;
                 args.Handled = true;
--- a/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
@@ -138,15 +138,20 @@ namespace Content.Server.Abilities.Psionics
             if (args.Handled)
                 return;
 
-            if (!HasComp<MindSwappedComponent>(args.Mind.CurrentEntity))
-                return;
-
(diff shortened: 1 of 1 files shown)


Proxysseia:
neat

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
```diff
@@ -138,15 +138,20 @@ private void OnGhostAttempt(GhostAttemptHandleEvent args)
...
             /*if (!args.viaCommand)
                 return;*/
 
-            args.Result = false;
-            args.Handled = true;
+            // DeltaV - start of trapped ghost fix
+            // If you're able to swap back to your original body, you should swap back before you ghost.
+            if (TryComp<MindSwappedComponent>(args.Mind.CurrentEntity, out var component)
+                && _actions.GetAction(component.MindSwapReturnActionEntity) is {} action
+                && action.Comp.AttachedEntity is {} user)
```
Toby222:
```suggestion
                && _actions.GetAction(component.MindSwapReturnActionEntity) is { } action
                && action.Comp.AttachedEntity is not null)
```

Toby222 (changes requested):
Minor formatting

Content.Server/Nyanotrasen/Abilities/Psionics/Abilities/MindSwapPowerSystem.cs
cmorley191: I need to learn these fancy C# patterns, been out of the loop for a bit

cmorley191:
Re-tested the fix behavior, still works.

cmorley191:
@Toby222 ready for re-review

Toby222 (approved):
seems fine

PR #4394, opened 2025-09-15
make teslas evil again

kotobdev:
## About the PR
Teslas once again hone in on energy sources, instead of random-walking.

## Why / Balance
teslooses right now are just substantially less scary then singulos, 3/4 times they'll just randomwalk into space and barely even damage the containment area. while there is some fun as a ghost spamming "YES YES YES" when it moves toward station and "NO NO NO" when it moves away, it's a bit anticlimatic overall.

one suggestion would be to make the tesla not gain energy from absorbing things (or gain NEGATIVE energy)- so it'll destroy a good amount of things and then decay- but 🤷 a tesloose should be fairly station-ending. still, if direction doesn't want the tesla to eat absolutely *everything*, that is something that could be done fairly easily

## Technical details
yamlops

## Media
nah

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: TESTMERGE: Teslas will once again move toward energy sources.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/energyball.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/energyball.yml
@@ -8,11 +8,11 @@
     bodyStatus: InAir
     sleepingAllowed: false
   - type: CanMoveInAir
-  # - type: ChasingWalk # DeltaV - Replaced with random pathfinding
-  #   minSpeed: 1
-  #   maxSpeed: 3
-  #   chasingComponent:
-  #   - type: LightningTarget
+  - type: ChasingWalk
+    minSpeed: 1
+    maxSpeed: 3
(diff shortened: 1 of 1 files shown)


Bonktrauma:
Return of the tesla..

Stop-Signs:
Direction 3 days

rosieposieeee:
...do IPCs count as energy sources?

SirWarock:
> ...do IPCs count as energy sources?

This is the single greatest reason this PR should be speed merged.

AirlineFood001:
An IPC running around looking for other power sources to distract the tesla sounds really funny to watch for some reason, love that image.

Stop-Signs:
Direction approved for 1 week (or until we actually see a tesloose :trollface: ) testmerge

TehFlaminTaco (approved):
lgtm

PR #4416, opened 2025-09-18
do not reset ai store when intellicarded

### LINKED ISSUE #4361: AI being installed into a core causes its program store to reset
## Description

If the AI is carded and then installed into an AI core again, their program store points are reset back to 3 no matter what, and allows them to buy abiltiies they've already bought (So basically to how it was at the start of the round). This lets the AI have an unlimited amount of identical abilities in the actions menu that all have separate cooldowns.

**Reproduction**

Be Station AI, buy any abilities you want to from the program store, be intellicarded and installed back. This can be repeated an unlimited amount of times.
**Screenshots**

Image

Each of those functions as a separate ability, and each has its own cooldown

**Additional context**

cmorley191:
## About the PR

Fixes https://github.com/DeltaV-Station/Delta-v/issues/4361

With this PR, the AI's money and available merchandise (for their AI action store) no longer resets whenever they are transferred into an intellicard and back. (previously the AI would get 3 new "memory" credits and could purchase duplicates of their previous purchases)

## Why / Balance

Remove infinite purchases glitch.

## Technical details

Moved the "Store" component off of the "AiHeld" entity and on to the "StationAiBrain" entity.

StationAiBrain is an unseen positronic brain contained inside the station's AI machine. This is the station AI player's "mob" -- you can see the InputMover component on it (those inputs are relayed to the AI Eye entity, but I digress). When the AI is moved to an intellicard, it's actually this StationAiBrain entity that's moving from one container to the other.

AiHeld is a dummy entity that just has components that should be added/removed whenever the StationAiBrain is inside the station's AI machine.

With this new setup, the AI player always has the "Store" component; the StationAiBrain has it so it's not getting removed and reset anymore. The "StationAiShop" component, however, remains on the "AiHeld" entity -- this other component provides the action button to open the shop, so it's important that it's added/removed every time.

## Media

https://github.com/user-attachments/assets/9c00126d-eabe-4356-ba9d-f55ea18c645b

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
@@ -83,14 +83,6 @@
   - type: ShowJobIcons
   # Begin DeltaV - AI Shop System
   - type: StationAiShop
-  - type: Store
-    categories:
-    - AiAbilities
-    currencyWhitelist:
-    - AiMemory
-    balance:
-      AiMemory: 3
-    grantActionsToMind: false
   # End DeltaV - AI Shop System
 
 - type: entity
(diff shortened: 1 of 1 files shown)


TehFlaminTaco (approved):
Good fix

PR #4419, opened 2025-09-18
Fix blunt stamina double dipping

SirWarock:
## About the PR

Made Stamina Damage resulting from Blunt Damage AP.
Buffed Oni's prying speed to 50% and included their 25% bonus stamina damage in the .xml species file.
## Why / Balance

Currently, Stamina Melee Resistances double dip. Someone wearing a Captain's Carapace has 50% blunt resistance and 50% stamina resistance.
This results in the truncheon doing 10 blunt damage instead of 20. The 10 blunt damage is then multiplied by it's stun damage factor, which is 1.5x, resulting in 15 stamina damage.
However, that stamina damage is ALSO reduced by the stamina damage, resulting in 7.5 Stamina damage.
This is equal to 75% melee stamina resistance, because both blunt and stamina resistance reduce it. This fixes this.
In my eyes, this is an oversight.

Oni's do more stamina damage, so I updated their .xml. I also made their prying time more accurate to the .xml. Why say 50% when it's 40%? I upgraded it to 50%.

The blunt to stamina factor damage does NOT benefit from Oni's inherent 1.25x multiplier to stamina damage.
I wanted to add it in, but given that Oni's do 1.35x more blunt damage, which results in 1.35x more stamina damage, it felt like it was double dipping, so I refrained from that. Otherwise, it would be ~1.69x more stamina damage.
## Technical details

added literally one boolean.
Changed a bit of YML.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Increased Oni's prying speed to 50% to match their species guidebook page.
- fix: Fixed Truncheons and similar weapons dealing only half their stamina damage on armored targets!


Proxysseia:
Direction 24h

Proxysseia:
the fix is approved, next time please push the fix and the oni buff in seperate PRs.
the oni buff is also approved because oni deserve a little love and oh my that is quant of love

SirWarock:
> the oni buff is also approved because oni deserve a little love and oh my that is quant of love

Real. Will do so

SirWarock:
@TehFlaminTaco Another already Approved PR nod nod

TehFlaminTaco (approved):
No issues here

PR #4429, opened 2025-09-19
make mindbreaker toxin grant a psionic reroll

cmorley191:
## About the PR

With this PR, mindbreaker toxin now resets your "Already Rerolled" psionic flag, allowing you to do another Oracle quest and drink LotophagolOil to roll for a new power.

## Why / Balance

Primary purpose of this is to make mindbreaking a less painful method to help reduce glimmer. Code Whites are pretty frequent right now (somewhat annoyingly) - this helps deal with them.

Mindbreaking is currently the most viable method to reduce glimmer -- it's an instant reduction of 50-70 glimmer (the other methods are less viable / even more painful: getting lucky with mites/wisps, or asking botany to focus entirely on it for quite a long time). But mindbreaking to reduce glimmer is _rough_: the mantis is basically telling the psionic crew member, "hey our artifact researcher had too much fun, so you don't get to use your power anymore".

By adding this PR, mindbreaking is less of a hard sell / something you have to force on people now -- **psionic crew that dislike their current power can volunteer for mindbreaking to help reduce glimmer as a win-win**, and reroll again afterwards with Lotophagol Oil.

## Technical details

I think the diff is pretty self-explanatory. I can explain if you have questions.

## Media

N/A, simple change (can provide on request)

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Mindbreaker Toxin now resets your psionic reroll, so you can use the Oracle to roll for a power again. (helps make Code White mindbreaking for instant glimmer reductions less painful for the victims)

PR #4430, opened 2025-09-19
oracle no longer requests items that are far away in the research tree

cmorley191:
## About the PR

The Oracle now only requests items from these categories:
- unlocked techs
- "available" techs -- techs in the current research tier (the ones that show up as "researchable technologies" in the console)
- ~~techs in the *next* research tier, but only if we're very close to reaching that tier (specifically: the research discipline only has 1 tech remaining to research in the current tier)~~

In other words: the Oracle will no longer request these difficult items that could require 50k-100k research points to reach:
- items from a tech that is 2 tiers up (i.e. we're tier 1 now, and the item is in tier 3)
- items from a tech that is 1 tier up, ~~but we've only recently reached the current tier (e.g., the round just started, and the item is in tier 2)~~

## Why / Balance

There are already quite a few barriers to completing oracle quests:
- botany quests can be hit or miss in general; mutated botany quests are _very_ difficult (almost impossible) if botany has not already happened to start it
- getting materials for items can be difficult
- the refresh period on the oracle is long, so "impossible" (extremely hard) items block the oracle for quite a long time
- research

This PR lessens the final barrier - research - without entirely removing it. You still have to spend research points for these oracle quests, and in that process, you can still get quite unlucky with your research rolls (you still might need to spend 10-30k points or more to reach a tech in the *current* tier).

#### Why/Balance for the underlying motivation of this PR: "oracle quests should be easier"
- Psionics and bluespace technologies are two of the most unique features people see in DeltaV (at least compared to vanilla/Wizden). They should be a little more common, and the Oracle is a great way to get them -- the Chaplain or Mysta is often involved in administering Lotophagic Oil; it's an opportunity to RP discussions about the noosphere a bit.
- Code Whites are a little overly common at the moment, and mindbreaking psionics is the most viable approach for lowering glimmer right now (50-70 instant reduction). More psionics means more volunteers for mindbreaking, so this indirectly helps with the code white problem. (especially alongside https://github.com/DeltaV-Station/Delta-v/pull/4429 )

## Technical details

Primarily just added an if-statement around the "add tech's items to oracle pool" step, checking the conditions:
- add all techs if no research servers are present
- add a tech if it's unlocked
- add a tech if it's available
- ~~add a tech if both "the tech's tier == current research tier + 1"  AND  "techs remaining in current tier <= 1"~~

The newly added-function (for the last condition) borrowed the logic from the function directly above it.

## Media

This PR is hard to demonstrate.

I primarily tested it with log statements in the changed code area, showing which portion of the added if-statement was hit / the variables used in the if-statement.

As an example:

1. Roundstart: tier 2 medical technology is not close to being reached

image

2. Researched THREE out of FOUR tier 1 medical technologies -- we are nearing tier 2:

image

3. Researched ALL FOUR tier 1 medical technologies -- we are now in tier 2:

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The Oracle will no longer request items that are "very" far away from being researched.


kotobdev:
this is peam, can't wait for the chaplain's job to actually be awesome and cool :pray:

Proxysseia:
neat

TehFlaminTaco:
@DeltaV-Station/direction @Proxysseia Direction review required for this one

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -34,6 +36,9 @@ public sealed class OracleSystem : EntitySystem
     [Dependency] private readonly GlimmerSystem _glimmerSystem = default!;
     [Dependency] private readonly PuddleSystem _puddleSystem = default!;
     [Dependency] private readonly IAdminLogManager _adminLog = default!;
+    // DeltaV - start of oracle tech tier check
```
TehFlaminTaco: Nyanotrasen files don't need DV comments

Content.Shared/Research/Systems/SharedResearchSystem.cs
```diff
@@ -146,6 +146,31 @@ public int GetHighestDisciplineTier(TechnologyDatabaseComponent component, TechD
         return tier - 1;
     }
 
+    // DeltaV - start of oracle tech tier check
```
TehFlaminTaco: Move to a _DV system and out of upstream

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -234,15 +239,41 @@ public List<string> GetAllProtos(OracleComponent component)
         var allTechs = _prototypeManager.EnumeratePrototypes<TechnologyPrototype>();
         var allRecipes = new List<string>();
 
+        // DeltaV - start of oracle tech tier check
```
TehFlaminTaco: Nyanotrasen files don't need DV comments

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -234,15 +239,41 @@ public List<string> GetAllProtos(OracleComponent component)
...
+        var query = EntityQueryEnumerator<ResearchServerComponent, TechnologyDatabaseComponent>();
+        while (query.MoveNext(out var serverUid, out var server, out var database))
+        {
+            researchServers.Add((serverUid, database, _research.GetDisciplineTiers(database)));
+        }
+
         foreach (var tech in allTechs)
         {
-            foreach (var recipe in tech.RecipeUnlocks)
+            if (
```
TehFlaminTaco, line 292: The Time complexity for this function is cooked. For Each Tech, You iterate through each server, and then call a function that iterates through each tech. I'd really appreciate if you could avoid all this recalculation.

TehFlaminTaco (dismissed):
See previous comments

TehFlaminTaco:
Is this still being worked on? Changes have been requested

Content.Shared/Research/Systems/SharedResearchSystem.cs
cmorley191: I'd love your thoughts on this. This function is actually a mirror of the upstream function that's right above it (`GetHighestDisciplineTier`). The `GetHighestDisciplineTier` upstream function is _already_ calculating what this new function is calculating, as an intermediate step (it is calculating `unlockedTierTech.Count` and `allTierTech.Count`). But `GetHighestDisciplineTier` only _returns_ the highest discipline tier, rather than this intermediate bit that I needed. I considered modifying the upstream function to return _both_ pieces of information, since it already has both, but thought that might be more intrusive than what I have here.

I settled on putting this function right here below the inspiring function, since if (on the very off chance) upstream changes the research system, we should probably also consider how it should be mirrored in this new function.

So that's why I'm hesitant to move this further away from the code it's supposed to be in parity with. Do you think I should just bite the bullet and modify upstream's `GetHighestDisciplineTier` to return this information (unlocked/all current tier tech counts) as well?

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
cmorley191: Updated. Now it calls the function in a loop over each discipline, rather than over each tech.

By my count, the entire GetAllProtos algorithm complexity is now basically the same as the complexity of SharedResearchSystem.GetAvailableTechnologies, which is called every time the research console updates its list.

GetAvailableTechnologies does:
- a GetDisciplineTiers call
- an IsTechnologyAvailable call for each tech

The new code here does:
- a GetDisciplineTiers call
- a GetTechsRemainingToTierUp call for each discipline. As you pointed out, GetTechsRemainingToTierUp loops over all techs, but it does so in the same way GetHighestDisciplineTier does. Therefore, this whole loop is equivalent to a second GetDisciplineTiers call.
- an IsTechnologyAvailable call for each tech
- an IsTechnologyUnlocked call for each tech, which is roughly as expensive as an IsTechnologyAvailable call

So the new OracleSystem code is basically as expensive as calling GetAvailableTechnologies twice in a row. This should be fine for the infrequent Oracle requested item update, imo.

Iterating over each server shouldn't matter, as we normally only have 1 server.

Content.Shared/Research/Systems/SharedResearchSystem.cs
cmorley191: To be honest if people are up for it, we could just make Oracle quests even easier by ditching this whole complex condition (this condition of "allow Oracle to request techs from tiers that are 1 tech-unlock away")

Balance wise, that would make oracle requests _even easier_ than I originally planned, but if it saves us some maintenance headache maybe that's fine.

cmorley191:
@TehFlaminTaco this is ready for re-review

Content.Shared/Research/Systems/SharedResearchSystem.cs
TehFlaminTaco: Honestly? I suggest doing it that way, as that's how I assumed it was it was going to be anyway.

cmorley191: Let's do that. simplified this PR so much

cmorley191:
@TehFlaminTaco ready for re-review.

Removed the whole "allow oracle to request higher tiers based on how many techs away we are" thing. I'll keep my branch open; some passionate coder can always pick it up at a later time.

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -234,13 +237,33 @@ public List<EntProtoId> GetAllProtos(OracleComponent component)
...
+                )
+            )
             {
-                var recipeProto = _prototypeManager.Index(recipe);
-                if (recipeProto.Result is {} result)
-                    allRecipes.Add(result);
+                foreach (var recipe in tech.RecipeUnlocks)
+                {
+                    var recipeProto = _prototypeManager.Index(recipe);
+                    if (recipeProto.Result is {} result)
```
Toby222:
```suggestion
                    if (recipeProto.Result is { } result)
```

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -234,13 +237,33 @@ public List<EntProtoId> GetAllProtos(OracleComponent component)
         var allTechs = _prototypeManager.EnumeratePrototypes<TechnologyPrototype>();
         var allRecipes = new List<EntProtoId>();
 
+        var researchServers = new List<(
+            EntityUid serverUid, 
+            TechnologyDatabaseComponent database, 
+            Dictionary<string, int> disciplineTiers
+        )>();
```
Toby222, line 283: Use `Entity<TComponent>` instead of separate `EntityUid` and `database`
```suggestion
        var researchServers = new List<(Entity<TechnologyDatabaseComponent> database, Dictionary<string, int> disciplineTiers)>();
```

Toby222 (dismissed):

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
cmorley191: I'm curious, by the way. Do you know what the point of this Resolve check is? That's the only reason I keep the uid

`SharedResearchSystem.cs` line 215
```
    public bool IsTechnologyUnlocked(EntityUid uid, string technologyId, TechnologyDatabaseComponent? component = null)
    {
        return Resolve(uid, ref component, false) && component.UnlockedTechnologies.Contains(technologyId);
    }
```

Is this just like, checking that the server exists? Could you ever have just a detached TechnologyDatabaseComponent on accident?

Toby222: The component parameter is optional
And it ensures that the component belongs to that entity

cmorley191:
@Toby222 ready for re-review

cmorley191:
NukeOps flaky test failure, retriggering

ShepardToTheStars (approved):

PR #4434, opened 2025-09-20
Increase colossus minimum spawn time

kotobdev:
## About the PR
~~Colossus spawn weight has been lowered from 4.5 to 3.0, now being the same as a Loneop. Additionally, the earliest start has been increased to 45 minutes and the minimum players to 30, also the same as Loneop.~~

Colossus earliest start is now 45 minutes (was 30)

## Why / Balance
~~Colossus, in it's initial form, was not balanced or seemingly intended to be a station-ending antagonist, and the spawn weight was set accordingly. Now, it is balanced that way, so I'm updating the weight accordingly.~~

~~Ideally the colossus just. Wouldn't be that powerful. But I don't have the motivation to go implement an entire rework right now, so consider this more of a band-aid fix until somebody does.~~

## Technical details
yaml op

## Media
nah

## Requirements

:trollface: 

## Breaking changes

**Changelog**

:cl:
- tweak: Colossuses will now only spawn 45 minutes into the round.

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/events.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/events.yml
@@ -3,10 +3,10 @@
   id: ColossusSpawn
   components:
   - type: StationEvent
-    weight: 4.5
-    earliestStart: 30
+    weight: 3.0
+    earliestStart: 45
     reoccurrenceDelay: 20
-    minimumPlayers: 25
+    minimumPlayers: 30
     duration: null
   - type: PrecognitionResult
     message: psionic-power-precognition-colossus-spawn-result-message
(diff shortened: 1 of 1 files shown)


Proxysseia:
24h

Proxysseia:
discussion is still ongoing, will take another day

Proxysseia (changes requested):
Colossi are vastly less damaging than a loneop. a Loneop forces evac, a colossus might. Loneop easily gets enough tools to blow up entire departments, a colossus spaces some times and kills some people. With time people have also gotten better at not walking into the colossus to get smacked. In combination with some recent nerfs to colossi we are not willing to equate them to a loneop.
we are willing to delay their earliest start to 45, that part is fine. What we really would like to see is removing the chance for random colossi to spawn during cult rounds, they are too much of a swing. If that is out of scope for this PR then don't worry, just the earliestStart change is fine.

Proxysseia (approved):

TehFlaminTaco (approved):
Yamlops

PR #4437, opened 2025-09-21
Makes tape recorders signal-compatible

mnva0:
## About the PR
Makes tape recorders signal-compatible. There are four ports in the exact order the buttons display on the tape recorder
1. Pause
2. Record
3. Playback
4. Rewind

## Why / Balance
I think this'd open up a lot of cool stuff for antagonists, Security, or just anyone. You could...
1. Set a recorder trap that activates as soon as a door opens. You'd need to check to make sure the room isn't bugged
5. Make a greeter for your department.
6. Make something to ~~yell at~~ advertise to passers-by in the hallways.
7. Leave anonymous `(this is possible even as a non-syndie. damage tapes and it hides the name)` threatening messages that play as soon as a door opens instead of leaving a tape on the ground and hoping it gets played.
8. Set up an intercom with a tape recorder to send alerts (like a door opening, or batteries discharging), or a delayed message to divert attention
9. Just generally use it as a signal output. Right now, signals don't have enough good, noticeable outputs. Best noticeable output I can think of is a figurine
10. Use a dead mans signaler for blackmail (true or voice changered) using an intercom and a hidden tape recorder

I am a bit worried about someone... automatically recursing these to spam chat or whatever? But you could do that anyway with tape recorders to an extent. I've never seen it happen once

## Technical details
I'm terrible at C# so please scrutinize this PR like you're certain I don't know what I'm doing. It is a surprise to me every time I successfully do something like this

Closing the game and reopening it seems weird? If this is clientsided to an extent, it might desync from the rest of the players. I'm not sure if this is just a tape recorder issue. Or if it even is an issue

I used the tape recorder UI FTL entries for the port names. They're meant to say the same things anyway

I have tested...
- Controlling it with signals in-hand. UI updates instantly
- Removing the tape, switching mode via signal, and then inserting the tape. Starts in pause
- Using it without a tape. Faint, distinct clicking noise. Doesn't do anything
- Recording or playing when the tape is at 100%. Clicks twice (likely the mode switching and then immediately switching back), pauses. Double-clicking is weird but minor enough for me to not care

## Media
image

(these arent the best examples, but these are connected to play or record. these would wait as long as they'd have to for a trigger)

https://github.com/user-attachments/assets/cdaedb52-9298-42e0-88b9-ed9319ee318f

https://github.com/user-attachments/assets/bc1a5217-07cc-40ec-b0d6-270ccaf53f10

(this one isnt even a feature the PR added. just showing off the fact you can damage tapes to hide the name)
image

## Requirements

**Changelog**
:cl: Minerva
- add: Tape recorders can now be controlled using four signal ports corresponding to their buttons.

### DIFF
--- a/Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
+++ b/Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
@@ -10,6 +10,8 @@ using Content.Shared.Tag;
 using Content.Shared.Toggleable;
 using Content.Shared.UserInterface;
 using Content.Shared.Whitelist;
+using Content.Shared.DeviceLinking;
+using Content.Shared.DeviceLinking.Events;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Containers;
 using Robust.Shared.Random;
@@ -43,6 +45,7 @@ public abstract class SharedTapeRecorderSystem : EntitySystem
         SubscribeLocalEvent<TapeRecorderComponent, ExaminedEvent>(OnRecorderExamined);
         SubscribeLocalEvent<TapeRecorderComponent, ChangeModeTapeRecorderMessage>(OnChangeModeMessage);
         SubscribeLocalEvent<TapeRecorderComponent, AfterActivatableUIOpenEvent>(OnUIOpened);
+        SubscribeLocalEvent<TapeRecorderComponent, SignalReceivedEvent>(OnSignalReceived);
 
         SubscribeLocalEvent<TapeCassetteComponent, ExaminedEvent>(OnTapeExamined);
         SubscribeLocalEvent<TapeCassetteComponent, DamageChangedEvent>(OnDamagedChanged);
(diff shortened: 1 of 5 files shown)


Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
```diff
@@ -411,6 +414,27 @@ private void UpdateUI(Entity<TapeRecorderComponent> ent)
...
+            SetMode(ent, TapeRecorderMode.Stopped);
+        }
+        else if (args.Port == ent.Comp.RecordPort)
+        {
+            SetMode(ent, TapeRecorderMode.Recording);
+        }
+        else if (args.Port == ent.Comp.PlaybackPort)
+        {
+            SetMode(ent, TapeRecorderMode.Playing);
+
```
TehFlaminTaco: Remove empty line

TehFlaminTaco (changes requested):

Content.Shared/_DV/TapeRecorder/Systems/SharedTapeRecorderSystem.cs
mnva0: Good catch. Thank you!

TehFlaminTaco (approved):
lgtm

PR #4441, opened 2025-09-21
id card computer now sets access when selecting a job preset

### LINKED ISSUE #3803: Selecting CE from the drop down doesn't give CE or Command access
## Description
Promoted a CE and selected Chief Engineer from the drop down under the assumption it would do the same as any other drop down and give the right accesses but instead it just changed the title only.

**Reproduction**
Have Cap ID
Get another ID(in this case it was an engineer ID)
put IDs in computer
select chief engineer
no accesses beyond engineering

**Screenshots**
I didn't take one as I didn't even realize it happened until I was told.

**Additional context**
It was an engi ID i changed so I dunno if that has any bearing on this

cmorley191:
## About the PR

Fixes https://github.com/DeltaV-Station/Delta-v/issues/3803

Previously the job preset's access were just flashing up and then immediately being erased -- they weren't actually being set, due to our changes to how the client-server messaging works in PR https://github.com/deltaV-Station/Delta-v/pull/3238

## Why / Balance

UI bug fix

## Technical details

Just extended the client-server messaging approach we added in the previous PR I linked: since accesses don't get sent in `SubmitData()` anymore, we should send the job preset ones in the new way as well (`OnToggleAccess`).

## Media

id_preset

You can see I also verified that accesses you can't change aren't changed by the preset dropdown:
1. centcomm grants AA ("Captain")
2. mystagogue changes preset to "Assistant"
3. centcomm checks the ID -- sees that it still mostly has AA, just minus all the accesses that Mysta could remove

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/Access/UI/IdCardConsoleWindow.xaml.cs
+++ b/Content.Client/Access/UI/IdCardConsoleWindow.xaml.cs
@@ -92,16 +92,8 @@ namespace Content.Client.Access.UI
             }
         }
 
-        private void ClearAllAccess()
-        {
-            foreach (var button in _accessButtons.ButtonsList.Values)
-            {
-                if (button.Pressed)
-                {
-                    button.Pressed = false;
-                }
-            }
-        }
+        // DeltaV - removed as part of job preset access fix
+        // private void ClearAllAccess()
 
         private void SelectJobPreset(OptionButton.ItemSelectedEventArgs args)
         {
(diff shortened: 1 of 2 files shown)


kotobdev:
this might literally be the best PR opened in the last six months

deltanedas (approved):

PR #4476, opened 2025-10-03
Stale PR/Issue Automation

Vapetastic-Gaming:
Github only issue, Addresses Stale PR's and Issues, Marks them for deletion after 90 days of inactivity

Ignore-stale and Priority 1 labels are ignored from this automation

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,30 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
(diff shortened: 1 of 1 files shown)


Vapetastic-Gaming:
delta has enough on his plate, Yaml warriors, the beacons have been lit

BarryNorfolk (approved):
LGTM, only note is that this will run against draft PRs as well, not sure if that's intended or not.
Only other Nit: would be a lack of `close-{issue,pr}-message` for when they are closed but I think that's probably an incredibly minor thing. Hopefully people are aware of the 2 day limit here and then see their PRs closed.

Toby222 (approved):

PR #4477, opened 2025-10-03
forgot to give them epi headsets oops

AeraAuling:
## About the PR
gives archivists science headsets since they're in science now

## Why / Balance
basically a bugfix since they're supposed to have them

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Archivists now have the right headset


MantasDab360:
I love how you noticed that after having the first shift as archivist :blunt:

deltanedas (approved):

PR #4481, opened 2025-10-03
Demetashield epsilon part 2

Cepelinas1:
## About the PR
Adds the Epsilon gamerule event. Was moved into part 2 due to excessive rewriting on the event itself and I really didn't want to delay direction CC roadmap stuff

## Why / Balance
Cool event, scary even.

## Technical details
Code is based on powerswitch off event, but coded in a different way

## Media
[](https://github.com/user-attachments/assets/87d78eef-43e7-4279-9a24-1f126ef2c0f6)

## Requirements


sheepism:
Now we need to make deathsquad those elites from trepang2

cmorley191:
Taking a look at this

cmorley191 (approved):
Holy shit this is terrifying

Issue #4396, opened 2025-09-16
EMP Implant does 130 Ion


Eillyrea:
## Description
Using an EMP Implant does 130 Ion damage to an IPC instead of the listed 45 it's meant to do in the PR
https://github.com/DeltaV-Station/Delta-v/pull/4255

**Reproduction**
Use an EMP Implant

**Screenshots**
https://github.com/user-attachments/assets/aa49edc2-5dd1-4f3a-bcc7-afc30e218413

**Additional context**

Proxysseia:
its meant to do 80

Eillyrea:
Also it doesn't show Ion damage in admin logs for some reason

<img width="803" height="333" alt="Image" src="https://github.com/user-attachments/assets/018c71dc-0209-46d0-888f-58f9165e2402" />

Eillyrea:
> its meant to do 80

Thank you princess <3

Lonestar827:
Ok i think i figured out why the damage is being really weird and it has to do with the SiliconEmpSystem, specifically on the OnEMP part

<img width="1205" height="254" alt="Image" src="https://github.com/user-attachments/assets/6b1ef092-df4b-4c9d-b216-102d9ecb381d" />

Specifically the fact that the event is being raised twice and the dividing there. I think the damage from the emp is being put through some weird ass math that needs to be untangled

Lonestar827:
poked at it again, it appears the EmpSystem.cs also has an effect on it, meaning nothing in the EmpOnTriggerComponent.cs or subdermal_implants.yml has an effect on how much damage the implant is doing. 
For example I changed the values in EmpSystem.cs and it worked as shown here.

https://github.com/user-attachments/assets/5a47b078-23ce-4933-857f-93d103cf9f13

Lonestar827:
made a draft pull request to keep track of whats going on.

Proxysseia:
@Vapetastic-Gaming #4484 merged

PR #4484, opened 2025-10-03
Fixes damage caused by EMP blast from EMP implanter

Lonestar827:
## About the PR

I changed the damage value of what EMP blasts do to silicons. 

## Why / Balance

According to https://github.com/DeltaV-Station/Delta-v/issues/4396 the Ion damage is doing 130 when it should be doing 80, thus the change

## Technical details

had to go into EMPSystem.cs to get the right damage, but will require review on why the EmpOnTriggerComponent.cs of the EMP implanter has no effect, thus the draft merge

## Media

https://github.com/user-attachments/assets/41732525-b123-47e5-9c3a-048531576106

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: fixed the EMP damage caused by EMP implants to do the intended 80 ion instead of 130 ion


NoElkaTheGod:
The weirdest part is that it worked as intended when I was testing it. So yeah, don't point your fingers at me, I have no idea either.

TehFlaminTaco (approved):
Lgtm

PR #4506, opened 2025-10-07
Fractured Form - New Psionic Power

TehFlaminTaco:
## About the PR

Adds a new Psionic Power, "Fractured Form", which gives the user a second body to switch between

## Why / Balance

More interesting psionic powers good 👍

## Technical details

**Important Code Changes**
 * Made `MindSwapped` component shared and networked, so its state can be tracked on the client
 * Made `ShowSSDIndicator` and related an event, to allow for more modular hiding of the indiciator.
 * Networked `TelegnosisPowerComponent`, for the above.

**Fractured Form**
Added "Fractured Form" power with `0.3` weight, giving it the same roll chance as Mass Sleep and Noospheric Zap.
Upon gaining the power, a second "Body" suddenly appears beneath the psion.
The second body has a 40% chance to be a clone of the user, otherwise, it's a random humanoid.
The second body has a 40% chance to spawn with equipment. If it would spawn as a clone of the user, the equipment is copied as per Paradox Clone. Otherwise, it is given Passenger equipment.

The user can switch between their bodies through one of four mechanisms:
 * The user can sleep, typically in a bed. They cannot return to a body that is forcefully slept (Such as via noct or anesthetic)
 * The user can be incapacitated. Being knocked crit or dead forces a body change. Much like sleeping, a crit/dead body cannot be returned to. If the body is gibbed, they are unable to switch in time.
 * The user can activate the ability, which will cause them to fall asleep after a 5 second do-after (And thus, allowing them to switch) This ability has a 60 second cooldown.
 * Randomly swap. The user will randomly pass out at some point from 5 to 20 minutes after their last swap. They will receive a 5 second warning and a yawn emote similar to Narcolepsy, but users should be aware of this and change frequently to avoid it.

**Balance Considerations**
Some strategies this ability opens are:
 * Having a fall-guy for crimes that you control (Assuming the body isn't a clone of you)
 * Escaping brig (And then breaking yourself out)
 * Having a backup life (It's worth noting, teach a Lesson and similar objectives are counted per body, so this may cause confusion or make it easier for the antag depending on the situation)

The potential extra information gained is considerable, though less consistent than Telegnosis. An unfortunate syndicate could tag a Fractured Form user, and then suddenly they are yelled about over Common. This is less of a threat than a Telegnosis user randomly seeing them, however.
The additional body is as much a risk as it is a bonus. If the user leaves their body in a bed, then it becomes an easy target to pick and rob or kill (They are not marked as SSD, thus are a valid target). And randomly swapping can happen at inconvenient times.

The roleplay potential is diverse, users may play into the personal struggles of having two identities, or the difficulties with interpersonal relationships. (Do I ask HoP for the same job/id as my main body, even though I'm a different species? What do my friends think of my other form? Do I use my other form to see if my friends are talking about me behind my back?)

Whether or not this ability is a boon or a hindrance is very situational, but I think that makes for a more interesting power.

**#### TL;DR watch video ####**

## Media

[Fractured Form Video](https://youtu.be/XfUgKj-CMlA "Fractured Form Video")
image

## Requirements

## Breaking changes

`MindswappedComponent` has been moved from `Server` to `Shared`.

**Changelog**

:cl:
- add: Added a new psionic ability, "Fractured Form"


Vapetastic-Gaming:
Direction review 24h minimum

Vapetastic-Gaming:
Direction Appwoved

TehFlaminTaco:
@deltanedas Requesting your eyes on this one, I touch a lot of systems and I'm sure I've missed something

deltanedas:
the test fail is real

TehFlaminTaco:
@deltanedas That test is haunting me, shit works fine when running. Should be good now.

Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
```diff
@@ -0,0 +1,261 @@
...
+            var validSpecies = new List<ProtoId<SpeciesPrototype>>();
+            var speciesPrototypes = _prototype.EnumeratePrototypes<SpeciesPrototype>();
+            foreach (var proto in speciesPrototypes)
+            {
+                if (proto.RoundStart)
+                    validSpecies.Add(proto.ID);
+            }
+            var species = _random.Pick(validSpecies);
+            var character = HumanoidCharacterProfile.RandomWithSpecies(species);
+            newBody = _stationSpawning.SpawnPlayerMob(xform.Coordinates, hasClothes ? original.Comp.VisitorJob : "", character, _station.GetCurrentStation(original.Owner));
```
deltanedas: empty string giga sus

Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
```diff
@@ -0,0 +1,261 @@
...
+    {
+        var bodies = entity.Comp.Bodies
+            .Where(c =>
+                c != entity.Owner                                 // Not the current body
+             && TryComp<MindContainerComponent>(c, out var cmind) // Has a mindcontainer still
+             && !cmind.HasMind                                    // Is not current possessed somehow
+             && !HasComp<ForcedSleepingComponent>(c)              // Not forcefully unconsious either
+             && !_mobState.IsIncapacitated(c))                    // Isn't bleeding out somewhere (Or dead)
+            .ToList();
+        return bodies.Any();
```
deltanedas: literally no reason to linq or make a list its just checking if any one matches these conditions
loop and if cond return true

Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
```diff
@@ -0,0 +1,261 @@
...
+
+    public bool CanSwap(Entity<FracturedFormPowerComponent> entity)
+    {
+        var bodies = entity.Comp.Bodies
+            .Where(c =>
+                c != entity.Owner                                 // Not the current body
+             && TryComp<MindContainerComponent>(c, out var cmind) // Has a mindcontainer still
+             && !cmind.HasMind                                    // Is not current possessed somehow
+             && !HasComp<ForcedSleepingComponent>(c)              // Not forcefully unconsious either
+             && !_mobState.IsIncapacitated(c))                    // Isn't bleeding out somewhere (Or dead)
```
deltanedas: put the conditions into separate method

Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
```diff
@@ -0,0 +1,261 @@
...
+        // Only picks bodies which don't have actives minds (In case of mindswap or something stupid)
+        var bodies = entity.Comp.Bodies
+            .Where(c =>
+                c != entity.Owner                                 // Not the current body
+             && TryComp<MindContainerComponent>(c, out var cmind) // Has a mindcontainer still
+             && !cmind.HasMind                                    // Is not current possessed somehow
+             && !HasComp<ForcedSleepingComponent>(c)              // Not forcefully unconsious either
+             && !_mobState.IsIncapacitated(c))                    // Isn't bleeding out somewhere (Or dead)
+            .ToList();
+        if (!bodies.Any())
```
deltanedas: no linq

deltanedas (changes requested):

Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
TehFlaminTaco: My love for Linq needs to be beaten out of me, will correct

TehFlaminTaco:
@deltanedas Corrected. Thanks for spotting

TehFlaminTaco:
@deltanedas All tests are passing & requires changes are in, lmk how you go

deltanedas (approved):

PR #4275, opened 2025-08-27
Species - Avali

kotobdev:
## About the PR

Avali have been added, "ported" from RMC. Ported in quotes because there's aspects from about six different servers incorporated here oh god help me
image

## Why
okay direction hear me out on this one
Avali are a species that's becoming more and more prevalent across SS14. RMC, Starlight, CD, Frontier/Monolith, even Harmony has them now! And Harmony's Harmony! I'm not even listing the servers with Resomi, which might as well be the same thing (albeit requiring more spritework). Point is, it's a rather popular species in the spess game community. By adding them here too, it lets incoming gamers transfer their existing characters from other servers, which helps make this place a bit more appealing.

While this species isn't a ss14/ss13 original species like most others, I don't see why that should really be an issue in the first place. Pre-existing lore is made to be ignored. Worst-case scenario, we can just *call* them Resomi and pretend they are original, like what Goob does :trollface: 

Aside from that, the species has a unique look and there's lots of room to put a DeltaV Original:tm: spin on it mechanically. I'm, like, pretty sure we're overdue for a new species anyways :godo:

## Balance
(I am very much willing to modify everything and anything here per direction feedback.)
Avali are a bit of a "glass cannon" archetype, basically the polar opposite of an Oni. Here's the rundown of their mechanics.
#### Benefits
- 🟢Less affected by cold temperatures, and is only slightly slowed by the coldest of temperatures.
	- (this was way cooler before upstream stole my swag and did basically the same thing to every other species. avali are still slowed much less but it's not quite as drastic)
- 🟢~~As a species of predators (with big eyes!), Avali are naturally skilled with ranged weapons, having 35% less spread when using them.~~
	- ~~I balanced this species with the assumption that this may be insanely powerful. I'm not entirely sure how useful this actually is in gameplay. If it turns out to be NOT that powerful, I would loooooove to throw on a thematically-fitting 10% speed boost for funsies :)~~
	- ~~example, L6 and Vulcan respectively (old sprites):~~
image
image

- 🟢Due to their excellent eyesight, they can see further then other species, being able to toggle binocular-like vision at will.

#### Neutral
- 🔶Bleeds ammonia-based blood, same as Vox. Doesn't matter mechanically unless a chemist is really really smart.
- 🔶Their claws deal Slash damage.
- 🔶Slightly lighter then other species.
	- 160 instead of 185, not felinid/kitsune levels of light
- 🔶Lower natural body temperature.
[...]

PR #4510, opened 2025-10-08
Fix Avali zoom distance

kotobdev:
## About the PR
Avali zoom distance, due to a very annoying bug, was 3 instead of 5. This is fixed now.

## Why / Balance
This was the originally direction-approved distance (see https://github.com/DeltaV-Station/Delta-v/pull/4275#issuecomment-3290674723), I accidentally broke it without realizing when I moved the relevant components to an organ instead of the base mob

## Technical details
the shared OrganEffectSystem adds components by copying them, slapping them on the parent entity, and dirtying them. MaxOffset and PvsIncrease weren't AutoNetworked, so they weren't getting updated on the client. To fix this, I had to make EyeCursorOffsetComponent entirely Shared, I guess you can't use AutoGenerateComponentState on abstract, non-registered components.

## Media
avali see two tiles further now :shrug:

## Requirements

## Breaking changes
SharedEyeCursorOffsetComponent has been moved into Shared, and as such, is now EyeCursorOffsetComponent.

**Changelog**

:cl:
- fix: Avali can now see slightly further.


TehFlaminTaco (approved):
About the only solution you had to fix this, and it's technically a bugfix with pre-direction approval. Go ham

kotobdev:
@TehFlaminTaco test fails seem? unrelated

PR #4527, opened 2025-10-14
Elegance Late Join Spawns Hotfix

IAmNotGray:
## About the PR
Adds late join spawns to the cryosleep room. That's literally it.

## Why / Balance
People are spawning in the paramedic room for some reason..? That's no good.

## Technical details
Map YAML

## Requirements


FieldCommand (approved):
Good to go 👍

PR #4540, opened 2025-10-17
Archivist closet

AeraAuling:
## About the PR
adds a closet for archivist. requested by mappers. also now fixes an exploit with display cases that lets people get cheap artifacts by blacklisting the display case board from the flatpacker.

## Why / Balance
mapping stuff. is just a closet, has a spare book restoration kit, a spare pen, a change of clothes, some paper, and some books, as well as 1-3 display case boards.

## Requirements

**Changelog**
:cl:
- add: Archivists now have a place to store their stuff.


Vapetastic-Gaming:
I mean, this is standard for roles to have (A closet) I would assume no direction revieb

Stop-Signs (approved):

PR #4546, opened 2025-10-18
Ensure grappled entities cleanup grapples when they crit/die

BarryNorfolk:
## About the PR
Fixes an issue where killing the target currently grappled by Laika forces her to be stuck in a grappled state.

## Why / Balance
Laika shouldn't get stuck

## Technical details
Somehow forgot to check the mob state of the grappled target, silly me

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Killing the grappled target of a grappler no longer causes them to be stuck forever


Toby222 (approved):
This looks sane

PR #4579, opened 2025-10-29
fix mantis time requirements

AeraAuling:
## About the PR
Changed the time requirements for Mantis because they were uh... there is no way it was intended.

## Why / Balance
so, before this, mantis only required a single hour in epi, and I'm told that's broken right now so *actually* you only needed to play 5 hours overall before being allowed to play mantis. In other words, someone could go through, play passenger and just sit around doing nothing, and then be handed a gun, knife, and the ability to shove toxins down people's throat. This fixes that because you should probably have to play epi to be an advanced epi role.

## Technical details
changes playtime trackers on mantis.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Psionic mantis time requirement has been changed. It now requires 5 hours of scientist and 5 hours of chaplain.


AeraAuling:
I have no idea why the yaml linter is failing, I used the exact same thing for the role time requirement as other jobs that works fine

SirWarock:
Rerun it? Or run it on your PC? There should be a test file in your folder

AeraAuling:
all code has been fixed by the way

ShepardToTheStars (approved):
LGTM

PR #4582, opened 2025-10-30
Fractured Form Patches

TehFlaminTaco:
## About the PR

Fixes two missed interactions with Fractured Form

## Why / Balance

Bugfix x2

## Technical details

Manually blacklisted IPC from the list of valid species to spawn when spawning a fracturedform
Keep track of the host body when switching fractured forms so we can detect when the power is removed
Only sleep bodies that have no mind, so no player is sleep-locked.

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- fix: Fractured Form can now be mindbreaker'd without perma-sleeping
- tweak: Fractured Form will no-longer spawn an IPC body.


Content.Server/_DV/Abilities/Psionics/FracturedFormPowerSystem.cs
```diff
@@ -149,7 +154,7 @@ private EntityUid GenerateForm(Entity<FracturedFormPowerComponent> original)
             var speciesPrototypes = _prototype.EnumeratePrototypes<SpeciesPrototype>();
             foreach (var proto in speciesPrototypes)
             {
-                if (proto.RoundStart)
+                if (proto.RoundStart && proto.ID != "IPC")
```
Toby222: Hardcoding this feels awful, is there no better way?

TehFlaminTaco: @Toby222 `SpeciesPrototype` does not provide an easy way to check if the species is Silicon. It's this, or add a vestigial flag to the species prototype.

Toby222:
```suggestion
            foreach (var proto in speciesPrototypes)
            {
                var speciesEntityPrototype = _prototype.Index<EntityPrototype>(proto.Prototype);

                if (proto.RoundStart && speciesEntityPrototype.TryGetComponent<PotentialPsionicComponent>(out var canBePsionic, Factory))
                {
                    var chance = canBePsionic.Chance;

                    if (speciesEntityPrototype.TryGetComponent<PsionicBonusChanceComponent>(out var bonusChance, Factory))
                        chance = (chance * bonusChance.Multiplier) + bonusChance.FlatBonus;

                    if (chance > 0)
                        validSpecies.Add(proto.ID);
                }
            }
```
Thing I hammered together, seems to work fine
Generates `Vox, SlimePerson, Diona, Felinid, Human, Reptilian, Chitinid, Moth, Feroxi, Vulpkanin, Oni, Harpy, Thaven, Avali, Kitsune, Rodentia, Arachnid` as potential species list, which seems correct
I believe the intent of the bugfix was to not spawn a non-psionic body, not specifically a non-Silicon one, right?

TehFlaminTaco: Yeah that should work fine

ShepardToTheStars (approved):
Was going to make a comment about what if `ControllingForm` is an entity that doesn't exist, but its been accounted for.

PR #4584, opened 2025-10-30
Standardize PartDamage

SirWarock:
## About the PR

I set the standard part damage multiplier to 0.5x.
## Why / Balance

Every healing option and a few select damage options only heal/deal 0.5x of their damage amount to parts.
Medicine, with just HealthChange or EvenHealthChange, healed only 0.5x of the damage to the parts.
Surgery only heals 0.5x of it's damage to the parts.
Wideswings only deal 0.5x of its damage to the parts.

This resulted in massive amounts of damage on the persons limbs, which resulted in heavy chemical usage even after revival of the patient. When you need 40u Aloxadone to revive a patient, you will then need an additional 40u aloxadone to return their limbs to working condition.

The only outline that I know off is precision swings, aka, Left Clicks with weapons. Those will still deal their respective damage equally to parts and general health. But that's intended.

This also fixes zombie plague fucking up your limbs. The poison damage on the limbs was accumulated on a 1x modifier, but chemicals healed only 0.5x of it. So when you cured a patient of 40 poison damage, they retained 20 poison on their limbs, which eventually lead to the limbs shitting bricks.
## Technical details

A few lines c# warrior?
Just kinda removed the mentions of a partDamage multiplier from HealthChange and EvenHealthChange, and returned the default PartDamageMultiplier to 0.5 down from 1.
## Media

grafik

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Standardized Part Damage, so the crispy Salvager won't need twice the dosage of Aloxadone to finally walk again.
- tweak: Non-curing Medical Treatment now effictively reverts the damage of the zombification on your limbs.


SirWarock:
If the standard damage is preferred to remain at 1x, then I suggest changing every other instance of partdamage to be also 1x, except for melee damage instances, which would remain at 0.5x for wideswings, 1x for precision swings.

If the necrosis part of a zombie infection is intended, then the partdamage of that infection can simply be raised to 1x or perhaps lower or higher.
Alternatively, the damage for disabling a limb can be reduced.

Cepelinas1:
direction 24h

Cepelinas1:
approved

Toby222 (commented):
I think the word you mean is "normalize", not "standardize"?

SirWarock:
> I think the word you mean is "normalize", not "standardize"?

I guess? I kinda just mean to make it all conform to one number, really. Is that not what both mean? Do you want me to rename it to Normalize?

Content.Shared/Damage/Systems/DamageableSystem.cs
```diff
@@ -181,7 +181,7 @@ public void DamageChanged(EntityUid uid, DamageableComponent component, DamageSp
         public DamageSpecifier? TryChangeDamage(EntityUid? uid, DamageSpecifier damage, bool ignoreResistances = false,
             bool interruptsDoAfters = true, DamageableComponent? damageable = null, EntityUid? origin = null,
             // Shitmed Change
-            bool? canSever = true, bool? canEvade = false, float? partMultiplier = 1.00f, TargetBodyPart? targetPart = null, bool doPartDamage = true, bool onlyDamageParts = false) // DeltaV - Fix EvenHealing on Limbs.
+            bool? canSever = true, bool? canEvade = false, float? partMultiplier = 0.5f, TargetBodyPart? targetPart = null, bool doPartDamage = true, bool onlyDamageParts = false) // DeltaV - Fix EvenHealing on Limbs && Standardize PartDamage.
```
Toby222, line 184: Probably also want to set the null-coalescing value on line 208 to 0.5?

Toby222 (commented):
Tested with romerol and poison auto-injectors and seems to work as intended

Toby222 (dismissed):

Content.Shared/Damage/Systems/DamageableSystem.cs
SirWarock: I think the second one is irrelevant, due to partMultiplier never being null at that point, no?
I'd rather not touch lines I don't have for merge conflicts.
Or do you want me to change that?

Toby222: > I think the second one is irrelevant, due to partMultiplier never being null at that point, no?

If you explicitly pass in `null`, it won't use the default parameter value, but actually be null
That being said I'm not sure if that's actually relevant, hence me giving approval despite the comment :⁾
If you look into it and think it's fine, I'll believe you and let you resolve the comment and merge

SirWarock: Oh, I see what you mean

Toby222 (approved):

PR #4599, opened 2025-11-02
Made Zombie bites deal only one DamageType

SirWarock:
## About the PR

Shifted the 7 Piercing Damage from Zombies to Slash.
Added 5 Structural Damage.
Previous Damage: 13 Slash, 7 Piercing, 10 Structural.
Proposed Damage: 20 Slash, 15 Structural.
## Why / Balance

A mixture of Slash/Blunt & Piercing is hard to properly protect without just putting on the hardsuit with a 40% all-round brute protection.
Shields suffer from this especially.

Wearing just a Durathread vest, a bite originally dealt 14.1 Damage.
Now, it'll only deal 12 Damage. 

However, Slash is better as a Damage Type, causing more bleeding stacks than Piercing. In addition, to counteract the nerf, I've chosen to up their Structural by 50%. 
## Technical details

Simple C# Changes.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Zombies now deal solely Slash with their bites!

### DIFF
--- a/Content.Shared/Zombies/ZombieComponent.cs
+++ b/Content.Shared/Zombies/ZombieComponent.cs
@@ -148,9 +148,11 @@ public sealed partial class ZombieComponent : Component
     {
         DamageDict = new()
         {
-            { "Slash", 13 },
-            { "Piercing", 7 },
-            { "Structural", 10 }
+            // Delta-V Start - Shifting Piercing Damage to Slash.
+            { "Slash", 20 }, // Was 13.
+            //{ "Piercing", 7 }, // Removed, so a durathread vest properly protects against it now.
+            { "Structural", 15 } // Was 10.
+            // Delta-V End - Shifting Piercing Damage to Slash.
         }
     };
 


Stop-Signs:
I dunno about this one....

SirWarock:
We could actually also convert some to Blunt. That causes less bleeding damage.

Stop-Signs:
> We could actually also convert some to Blunt. That causes less bleeding damage.

how would a bite be blunt??? bites pierce and tear which matches slash and pierce pretty well. I don't really see the reason for this PR other than marginal balance.

(not direction review)

Stop-Signs:
Oh also direction review 24 hours

SirWarock:
> > We could actually also convert some to Blunt. That causes less bleeding damage.
> 
> how would a bite be blunt??? bites pierce and tear which matches slash and pierce pretty well. I don't really see the reason for this PR other than marginal balance.
> 
> (not direction review)

There's a thread on discord complaining that the bite of a zombie is too strong. There was also said that it'd be better if the damage was only one type, since otherwise, it cannot be properly resisted.
Though it can be blunt and slash, given that armors that resist slash typically also resist blunt.

And blunt can just be from the zombie decking you with their fist. Or they just smash their head so hard into your torso to bite it that they bludgeon it.

Cepelinas1:
wow direction super late
anyways we had a big talk (unsurprising) and we ultimately decided to deny it

PR #4634, opened 2025-11-11
Revert upstream nerf to flashes

Woljif:
## About the PR

I mostly reverted the random upstream nerf to flashes (the handheld one) that gave them a 4 second cooldown. I say mostly because I'm pretty sure flashes used to have _no_ cooldown if used in-hand, so if you wanted to spam and burn out your flash instantly you could, but you're never reasonably flashing more than once per second. Plus that's kind of pointless anyway since you'd just burn out your charges, but 4 seconds is jaw-droppingly long. I've reduced the cooldown to 2 seconds so it doesn't become (even more) useless upon its first use.

## Why / Balance

There was a quiet nerf from upstream that nobody asked for and wasn't needed either (flashes are already very bad). Flashes are really only useful against crew, and even then they're only useful if crew isn't trying since flash protection is so easy to get. Flashes were already balanced in that the flash duration they give is short, and they have only 5 charges before permanently burning out.
## Technical details

- Changed Flash component's delay variable from 4 seconds to 2 in Security.yml
- Added a line of yml that defines FlashingTime at 2 second, to match the delay variable 

## Requirements

## Breaking changes

**Changelog**

- tweak: reduce handheld flash's ginormous 4-second cooldown to 2 seconds, toning down the upstream nerf.


Content.Shared/Flash/Components/FlashComponent.cs
```diff
@@ -28,7 +28,7 @@ public sealed partial class FlashComponent : Component
     /// These two times should be the same.
     /// </summary>
     [DataField, AutoNetworkedField]
-    public TimeSpan FlashingTime = TimeSpan.FromSeconds(4);
+    public TimeSpan FlashingTime = TimeSpan.FromSeconds(1); // DeltaV - Reduce from 4 to 1 to match Security.yml
```
ShepardToTheStars: Undo these changes and add this change on the `Flash` component in the yml.

ShepardToTheStars (dismissed):

Stop-Signs:
Personally I see spam flashes used all the time, against cult and just regular dissidents. I'm thankful for the nerf.

Cepelinas1:
direction review 24h

Woljif:
> Personally I see spam flashes used all the time, against cult and just regular dissidents. I'm thankful for the nerf.

Cult literally has built in flash protection in phase 3 and a host of ways to achieve it at any tier below. If it was really a problem people would have said something, which no one really ever has

kotobdev:
Agree with Woljif. Flashes can be easily countered by any crewmember and 90% of antags already, they didn't need a further nerf.

SirWarock:
I need my flash crit

Vova567810:
Five. Hundred. Burned Flashes.

Content.Shared/Flash/Components/FlashComponent.cs
Woljif: Done

Niomi0:
Borgs > AHHHHHHHHHHHHHHHHHHHHHHHHHHHH my EYES

Woljif:
> direction review 24h

Is this still happening

Vapetastic-Gaming:
Apologies, this has been a big debate, give us 24ish hours, ill get back to you by tomorrow, if not, Ping me

Woljif:
> Apologies, this has been a big debate, give us 24ish hours, ill get back to you by tomorrow, if not, Ping me

Get pinged

Vapetastic-Gaming:
@Woljif So, the consensus is, we do not want to remove the delay completely, although we are satisfied with a 2 second delay instead of the 4 (It seems a bit excessive) however we dont want it dropped to 1, to prevent spam

Vapetastic-Gaming:
So, changes reqquested, both to 2 instead of 1, once those are in, its approved

Woljif:
> So, changes reqquested, both to 2 instead of 1, once those are in, its approved

Done

Toby222 (approved):

PR #4636, opened 2025-11-11
Remove BeamDev

SirWarock:
## About the PR

I removed the Beam Devastator from Spawn Pools, as well as the safe.
## Why / Balance

The original Creator is unhappy with it, and therefore, wants them gone.
This isn't a balance discussion, therefore do not reply with balancing arguments or suggestions.
This is also not really something for Direction to review, because the only one to review this is @Stop-Signs as the Creator.
So, uh. Yeah.

I personally don't want it gone, but seeing the original creator saying their PR was denied when they wanted them gone just sits wrong with me.
## Technical details

Two lines of code.
## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removed the Beam Devastator from random spawn tables.


SirWarock:
Should I add the Beamdev to migration?

ShepardToTheStars:
> Should I add the Beamdev to migration?

Nah. If maps have it actually mapped, the maintainers of those maps can replace it. This is fine.

ShepardToTheStars (approved):
LGTM

PR #4641, opened 2025-11-12
MASSIVELY buff Random Animation

kotobdev:
## About the PR
Random animation minimum/maximum objects raised from 7/10 -> **40/60** (!!!)
Animation time raised from 7/15 -> 15/30

## Why / Balance
I have literally never seen this event trigger. Ever. Nobody even knows it exists. My current theory is that 99% of the time it triggers, it's on Elegance or something and it just picks up random trash in maints and nobody sees it and it immediately dies out with nobody the wiser.

With this change, there's a GOOD chance that multiple people will get attacked on large maps like Elegance. On SMALLER maps, every department will have a small nightmare for about 15-30 seconds, which I believe is ideal.

On Dev, you get hunted by basically everything that's not nailed down :)

(in theory, at least, on release config it's a bit wacky)

## Technical details
could probably update it in the yaml but it's never ever going to be used anywhere else so might as well just mess with the component

have only tested on debug, not sure if release spawns trash differently, will test Later:tm:
ok on Release even on Glacier it's a bit rare to see anything actually happen from any singular location but. still a hell of a lot better then it was

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The random animation glimmer event is now a LOT more scary.

### DIFF
--- a/Content.Server/_DV/StationEvents/Components/RandomAnimationRuleComponent.cs
+++ b/Content.Server/_DV/StationEvents/Components/RandomAnimationRuleComponent.cs
@@ -6,14 +6,14 @@ namespace Content.Server._DV.StationEvents.Components;
 public sealed partial class RandomAnimationRuleComponent : Component
 {
     [DataField]
-    public int MinAnimates = 7;
+    public int MinAnimates = 40;
 
     [DataField]
-    public int MaxAnimates = 10;
+    public int MaxAnimates = 60;
 
     [DataField]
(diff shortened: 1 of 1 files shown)


TehFlaminTaco:
Direction Review 24h

Toby222 (approved):
yea no need to update yaml I don't think

Vapetastic-Gaming:
Dir Approved

PR #4659, opened 2025-11-15
remove particle impermenance

AeraAuling:
## About the PR
I don't know what specifically changed it, but particle impermanence now uses the effect of the _rerolled_ particle as opposed to the particle effect that was there when it was hit. when that is then applied to an infectious anomaly that is currently inside of someone, it basically gives them two options to either roll the dice and explode, or to get themselves killed. getting themselves killed, in this case, actually ends up being the safer option now. suicide roleplay is against the rules for a reason, but rather then punishing people for taking the method that is safest for everyone, I'm just disabling it until I figure out how to change it back.

## Why / Balance
not a good mechanic for our server, we have that kind of roleplay banned for a reason

## Technical details
yaml

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: anomalies don't seem to be randomly changing their particles anymore, for now at least.

### DIFF
--- a/Resources/Prototypes/Anomaly/behaviours.yml
+++ b/Resources/Prototypes/Anomaly/behaviours.yml
@@ -16,15 +16,15 @@
   #Hard
     Fast: 0.5
     Strenght: 0.5
-    Inconstancy: 0.5
-    InconstancyParticle: 0.5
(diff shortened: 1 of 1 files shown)


Toby222:
Closing in favor of new PR which should fix the problem in lieu of turning off impermanence

PR #4570, opened 2025-10-25
Return of the Ifrit

AeraAuling:
## About the PR
Returns the Ifrit to the delta sector. A guardian from the noosphere that defends the mystagogue and the station. 

## Why / Balance
A few reasons. Firstly, Ifrit is cool. it's one of the neatest things we've had on the server. unfortunately, it got removed because pyrokinesis was horrifically overpowered, and when that was removed it became an incredibly underwhelming familiar for what it was supposed to be. What I have done is rebalanced the main problem aspects of the Ifrit to bring it to a more reasonable average. The main problems identified beforehand being that

1. Ifrit was far too easy to kill
2. Ifrit had too little melee damage
3. Pyrokinesis was way, way, way too powerful

so as part of this, firstly ifrit have had their health increased by 30 points, giving them a death threshold of 150 up from 120. I've also given them a custom damage modifier set, which keeps their immunity to cold, but adds an immunity to heat(they're fire spirit, why were they weak to heat), makes them resistant to slash and shadow, but makes them weak to holy and blunt damage. Second, their melee damage has increased from 10 heat to 15. Against a nukie, an ifrit will have to hit them 14 times, completely unarmored targets +/- 7 times depending on resistances, and any sort of armor bringing that up. Along with their swing speed being average at best this gives them a time to kill roughly in line with other "dedicated" weapons. Finally, the biggest pain point and what ultimately led to their removal, Pyrokinesis. Pyrokinesis still works as it did before, that hasn't changed. Rather instead of just giving someone an immediate full 5 fire stacks, I've lowered this to 3 fire stacks. Fire stacks seem to work on magnitude rather then having a linear progression in damage, so whereas before pyrokinesis could do 300+ damage in 12 seconds if someone didn't roll, it is now down to a more reasonable max of 130 damage against a completely unarmored diona in 12 seconds. Or in other words, one of the three species weakest to fire can actually survive being set directly on fire even without protection by just rolling. (this applies to moths as well, but not avali. sorry avali mains) This makes it a good distraction and suppressive tool that *can* be dangerous if ignored, but is less effective then an incendiary grenade at the bonus of having a cooldown instead of using resources.

## Technical details
mostly yaml. changed a variable in the pyrokinesis power system from 5 to 3, so technically a little C#, but like... the most minimal amount. none of this touches upstream code, does touch a little bit of nyano code but no worries there considering.

## Media
no video today am tired

## Requirements

- [x] I have tested all added content and changes.
[...]

PR #4670, opened 2025-11-16
Buff ifrit for real this time

Dorragon:
## About the PR
Buffed ifrit but for real this time

## Why / Balance
So the #4570 Ifrit buff wasn't real because they:
Had a damage trigger to destroy at 120 damage
Didn't use actual entity damage to attack, they use ifrit hands (resprited welders) to do everything

So both health buff or damage buff were a lie
Also we can indeed make it have a different hit sound when it's on now, so i did that and removed the comment from ancient times (rip)

## Media
Lazy

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/NPCs/familiars.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/NPCs/familiars.yml
@@ -49,7 +49,7 @@
     thresholds:
     - trigger:
         !type:DamageTrigger
-        damage: 120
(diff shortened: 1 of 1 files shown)


dekigangsta:
Ifrit love is SO back

Toby222 (commented):
lgtm
unrelatedly I noticed that the Ifrit's normal body temp is actually below freezing which is hilarious and should probably be changed at some other point

Toby222 (approved):

PR #4674, opened 2025-11-16
Nerf footprints.

pootslap:
## About the PR

~Removed the ability to create footprints from puddles.~ 
Footprints remove far less fluid from puddles. Dragging bodies through puddles still leaves a smear behind.
## Why / Balance

There's been some discussion for a while now about how footprints take away from the janitor role. Since anyone can clean a puddle by walking through it, a lot of players will opt to do that rather than call a janitor to come clean it. This leaves the janitor only getting brought in to chuck cleanades at huge spills and nothing else.

~I think the footprints are very cool and immersive but it cuts into janitor gameplay a lot. Somebody smarter than me could probably keep the footprints without making puddles too easy to clean but that's a bit beyond me.~
Just kidding I did it.

## Technical details

Footprints now contain 0.01u instead of 1u.
Stepping in a puddle removes 0.01u instead of 1u.
## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: Walking through puddles no longer makes the puddle smaller. Call your local janitor!

### DIFF
--- a/Content.Client/_EE/FootPrint/FootPrintsVisualizerSystem.cs
+++ b/Content.Client/_EE/FootPrint/FootPrintsVisualizerSystem.cs
@@ -44,9 +44,9 @@ public sealed class FootPrintsVisualizerSystem : VisualizerSystem<FootPrintCompo
 
         sprite.LayerSetState(layer, new RSI.StateId(printVisuals switch
         {
-            FootPrintVisuals.BareFootPrint => printsComponent.RightStep ? printsComponent.RightBarePrint : printsComponent.LeftBarePrint,
-            FootPrintVisuals.ShoesPrint => printsComponent.ShoesPrint,
-            FootPrintVisuals.SuitPrint => printsComponent.SuitPrint,
+            FootPrintVisuals.BareFootPrint => null,
+            FootPrintVisuals.ShoesPrint => null,
+            FootPrintVisuals.SuitPrint => null,
             FootPrintVisuals.Dragging => _random.Pick(printsComponent.DraggingPrint),
             _ => throw new ArgumentOutOfRangeException($"Unknown {printVisuals} parameter.")
         }), printsComponent.RsiPath);
--- a/Content.Client/_EE/FootPrint/FootPrintsVisualizerSystem.cs
+++ b/Content.Client/_EE/FootPrint/FootPrintsVisualizerSystem.cs
@@ -44,9 +44,9 @@ public sealed class FootPrintsVisualizerSystem : VisualizerSystem<FootPrintCompo
 
         sprite.LayerSetState(layer, new RSI.StateId(printVisuals switch
         {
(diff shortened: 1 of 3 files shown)


Spielern:
so based...

IamVelcroboy:
Just a thought, but since you reduced the amount of reagent in a footprint could you also just reduce the amount the puddle is reduced by rather than zeroing it out? Make it more of a balance thing, like instead taking 20 seconds to clean a puddle by walking through it it takes 20 mins or whatever. Just make it stupid to use that method. 

Another thought I had was to have a threshold, like a puddle can only be reduced to X units by footprints. From there it would remain, no matter how much you walk through it.

Either way, this is an improvement.

pootslap:
I did try out reducing the reagent loss rather than just removing it but it very quickly turns one big connected puddle into a bunch of smaller ones. Purely an aesthetic choice, I could definitely change it to maintain balance a bit better.

Vapetastic-Gaming:
This has been denied from direction. We like footprints too much. they add a bit of a unique element to the game, someone died, and the culprit walked through the blood puddle. Or the chef spills some oil in the kitchen, and walking about just drags out the oil. So I would much rather there be a fix to it, where the footprints take far less quantity to make, as I do not think that walking over the puddle should be the solution to cleaning the puddle so that the janitor can once again have something to do.

In other words, if you want to make footprints cost far less "Fluid" to be produced, we will back that, however direction is against removal

pootslap:
I'm kind of glad because the more I thought about it, the more I wanted to keep the footprints.
Made some changes, footprints exist again and remove 0.01u from the puddle instead of 1u. If someone is still willing to clean a puddle by running in circles for 15 minutes, there's no helping them at that point.

Toby222:
> I'm kind of glad because the more I thought about it, the more I wanted to keep the footprints. Made some changes, footprints exist again and remove 0.01u from the puddle instead of 1u. If someone is still willing to clean a puddle by running in circles for 15 minutes, there's no helping them at that point.

The insane janiborg who couldn't find any water:

SirWarock:
what if we just make footprints not evaporate?

Toby222:
> what if we just make footprints not evaporate?

Common wisdom is that this will cause extreme performance issues

SirWarock:
Could we increase the amount of units that each footprint takes out of a puddle to keep the number lower to combat performance issues?

pootslap:
It's possible but I can only see performance being an issue in very extreme cases since prints still despawn after 60 seconds. AFAIK the performance issues that goob had with this was due to footprints that never despawned and puddles that didn't deplete at all.

Bonktrauma:
> > what if we just make footprints not evaporate?
> 
> Common wisdom is that this will cause extreme performance issues

I haven't seen any performance issues on ee servers with them, or ee based servers, such as den or floof, although they may also have changes of their own

ShepardToTheStars (approved):
Looks fine.

PR #4688, opened 2025-11-18
Fixed colossus stamina issues

### LINKED ISSUE #4559: Colossus can be stunned and cannot recover stamina.
## Description
Stun batong a colossus, and now you're permanently slowed :(

**Reproduction**
> enable stun baton
> smack colossus in the face multiple times
> no stamina recovery ever, get a permanent slowness

**Screenshots**
happens to me on round #24477, in which in attempting to raise the monument in security an officer batons me and i lose my stamina forever

**Additional context**
john colossus needs to destroy curtains i fuckign hate those curtains

ShepardToTheStars:
## About the PR
Added additional melee resistance property to `StaminaResistanceComponent`, since we ignore stamina resistance on melee hits due to #4419. If this property is true, it basically double-dips stamina resistance on melee hits. But since this property is false, it should only affect entities that require additional stamina resistance.

## Why / Balance
Fixes #4559

## Technical details
Added a `MeleeResistance` boolean to `StaminaResistanceComponent` that is false by default (so things work like normal). SharedStaminaSystem will check that if its true and apply the resistance value.

Then I tested with a baton that was off and saw that stamina damage was still happening (from the blunt damage), so I had to modify if we ignore stamina resistance during the `TakeStaminaDamage` call. It checks for the `StaminaResistanceComponent`, and if it finds it, it will use the inverse of the `MeleeResistance` component. 

## Media
(Colossus POV btw)

https://github.com/user-attachments/assets/5b4b5339-c9c6-4925-8a0f-358f471cb806

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Colossi no longer take stamina damage from melee weapons.

### DIFF
--- a/Content.Shared/Damage/Components/StaminaResistanceComponent.cs
+++ b/Content.Shared/Damage/Components/StaminaResistanceComponent.cs
@@ -35,4 +35,16 @@ public sealed partial class StaminaResistanceComponent : Component
     /// </summary>
     [DataField]
     public LocId Examine = "armor-stamina-projectile-coefficient-value"; // DeltaV
+
+
+    /// <summary>
+    /// DeltaV - Whether or not this includes melee resistance. By default, DeltaV assigns melee stamina resistance to
+    /// blunt damage, but if this is set to true, it was override that and use the stamina resistance value.
+    ///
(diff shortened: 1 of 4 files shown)


Toby222 (approved):
seems fine but wow our stamina damage system is a mess

Issue #4692, opened 2025-11-19
Rat kings come with a free double


Dorragon:
## Description

When you take the rat king role, it spawns another rat king controlled by AI that can be taken as a ghostrole too

**Reproduction**

addgamerule RatKingSpawn 
take the ghostrole
ghost
There are two ratkings

**Screenshots**

https://github.com/user-attachments/assets/ab9e39d5-c3e8-4369-9b60-f448eb20a08c

pootslap:
Do they share the same name/number? I've had a couple of times where I saw Trash Lord (735) die and then see Trash Lord (735) running by a couple of minutes later.
Also explains the weird AI rat kings on the station, plus the stupidly high spawn rate.

ShepardToTheStars:
Fix merged

PR #4702, opened 2025-11-20
Fix Ratking Double Spawn

Coryler:
## About the PR

Make the Ratking only spawn once instead of twice upon taking the ghostrole. #4692 

## Why / Balance

Halves the Rats along station, to the intended amount.

## Technical details

Changed the BaseAntagSpawner to MarkerBase.... which fixed that. 

## Media

Super small change...

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Ratking Double Spawn

### DIFF
--- a/.wiki/_DV/Laws/CompanyPolicy.txt
+++ b/.wiki/_DV/Laws/CompanyPolicy.txt
@@ -135,7 +135,7 @@ While working aboard NanoTrasen stations, you may encounter the highly unlikely 
 '''NANOTRASEN DOES NOT TAKE RESPONSIBILITY FOR CREW INJURY OR DEATH IF CREW MEMBERS DISREGARD THE INFORMATION BELOW.'''
 
 === Hostile Fauna ===
-Be it from Epistemics experiments, corpses brought in by salvage, or something living in the vents, you may be attacked by non-sophont creatures. NanoTrasen does not encourage befriending such animals, unless it’s an approved experiment for research. Below are the procedures of how to deal with such a threat:
+Be it from Epistemics experiments, corpses brought in by salvage, or something living in the vents, you may be attacked by hostile non-sophont entities. NanoTrasen does not encourage befriending such animals, unless it’s an approved experiment for research. Below are the procedures of how to deal with such a threat:
 
 *Do not engage with the threat yourself.
 *Evacuate the department or the zone.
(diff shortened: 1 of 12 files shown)


Coryler:
The Failed test seems unrelated to my changes but I can take a look if needed... or can it be run again?

Toby222:
random fail, rerunning

cmorley191:
The whole antag system code is mind boggling, there are so many different components.

It seems the duplication was coming from there being two spawner components:
1. the `type: GhostRoleMobSpawner` component on the `id: SpawnPointGhostRatKing` prototype in `ghost_roles.yml`
2. the `type: AntagSpawner` on the `id: RatKingSpawn` gamerule prototype in `events.yml`

You opted to remove component 1, but I'd actually recommend we remove component 2 instead, for these reasons:
- `SpawnPointGhostRatKing` is used in the `KingRatMigration` event. We have this event disabled on DeltaV, but if we ever decide to re-enable it later, then it'd be nice if right now we aren't modifying its pieces.
- `RatKingSpawn` is better to edit because it's completely our own custom `_DV` event - we don't have to touch any upstream files to modify it

Then again, I really don't have a good grasp on all the different approaches here to really argue this. Our MenaceSkeleton gamerule takes a page out of the "remove component 1" book (though that gamerule has no connection to upstream, to be fair).

cmorley191:
It seems some of the relevant components were made last year as part of this "antag refactor"

https://github.com/space-wizards/space-station-14/pull/28217

Coryler:
Hmm ok. What is the Rat King Migration Event? I saw it being commented out though no actual thing which belongs to that id.

cmorley191:
> What is the Rat King Migration Event? I saw it being commented out though no actual thing which belongs to that id.

It is found in `Resources/Prototypes/GameRules/pests.yml`

`id: KingRatMigration`

It seems to me like it immediately spawns an AI rat king and some other lesser AI rodents. And it opens the spawned rat king to be taken over as a ghost role.

Coryler:
Ah thank you! From what I get, it's basically just another Ratking Spawn through the vents event?

> It seems to me like it immediately spawns an AI rat king and some other lesser AI rodents. And it opens the spawned rat king to be taken over as a ghost role.

Yeah. From what I get, it queues two spawns in one ghostrole. From what I've tested, either removal work.

ShepardToTheStars (approved):

Issue #4645, opened 2025-11-12
Borgs cant use @emote that have an associated sound


Niomi0:
## Description
If you @ laughs or type lol simply nothing will happen

**Reproduction**
(I've not tested this in the dev environment but it is the case in regular games)
chat "lol"
Notice there is no "'borgname' laughs"
other emote chat will work

**Screenshots**

**Additional context**
Probably upstream

kotobdev:
maaaaaybe fixed by https://github.com/DeltaV-Station/Delta-v/pull/4621 ? once the server updates (about 12 hours from now), see if it's still an issue

ShepardToTheStars:
> maaaaaybe fixed by [#4621](https://github.com/DeltaV-Station/Delta-v/pull/4621) ? once the server updates (about 12 hours from now), see if it's still an issue

The code was changed to still follow the blacklist/whitelist for emotes. It looks like borgs are [blacklisted](https://github.com/ShepardToTheStars/Delta-V/blob/allow-all-emotes/Resources/Prototypes/Voice/speech_emotes.yml#L39) for that emote?

Toby222:
Citation needed but I recall there being a borg laugh that was intentionally turned off at some point for some reason I don't recall

Niomi0:
Not yet fixed by any merges

MsEpic59:
Related
Can't do *boop or boops with any non-borg/ipc species.

ShepardToTheStars:
Again, this is entirely related to the whitelist and blacklist properties on each emote. 

If an emote has a blacklist or whitelist property, the current system will look at that. The fix would probably be to comment out those blacklist/whitelist properties.

MsEpic59:
I don't have dev environment set up, otherwise I'd comment it out, since the sounds only play anyways for the proper species.

PR #4703, opened 2025-11-20
Allow Borgs to use emotes without sound

Coryler:
## About the PR

Allow Silicons to laugh without sound, and allow organics to boops. Based on #4645 

## Why / Balance

There was already a PR dealing with this though it doesn't affect Silicons and some emotes only for 'em.

## Technical details

Commented out the Blacklist and Whitelist check in ChatSystem.Emote.cs... since it has only been mostly been used to block Silicons afaik. Felt more appropriate (or smaller) to just comment out that instead several line in a yml.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Cyborgs can now laugh silently again

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.Emote.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.Emote.cs
@@ -235,12 +235,13 @@ public partial class ChatSystem
             return true;
         }
 
-        // Check the whitelist and blacklist
-        if (_whitelistSystem.IsWhitelistFail(emote.Whitelist, source) ||
-            _whitelistSystem.IsBlacklistPass(emote.Blacklist, source))
-        {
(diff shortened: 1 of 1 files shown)


Coryler:
Can the test be run again? The test failed, if I read it correctly, due to a hat which I didn't touch 🤣

ShepardToTheStars (changes requested):
A simpler change would be to just comment out this if statement and the continue in `TryEmoteChatInput`. Less chance of a merge conflict.

```c#
if (!AllowedToUseEmote(uid, emote))
                continue;
```

Coryler:
Yea, that makes more sense :D. I've made the change as requested.

ShepardToTheStars (approved):

PR #4709, opened 2025-11-21
Tweak Distance Calculation for Shuttle/Radar Console

Coryler:
## About the PR

Changed the way distance is calculated for the Nav Computer, so that it shows the right distance again.

## Why / Balance

Previously it always calculated from the Local Offset plus the World Origin. towards the distance of the shuttle. Resulting in rather weird distances. Now it calculated from the World Position of the Grid plus the Local Center Offset, towards the Shuttle.

## Technical details

Changed the calculation to Add LocalCenter onto the World Position, then used Distance instead of the Length call.

Simply due to how the Handheld Scanner is working, it still shows up the wrong distances to the Grids. (Since it also doesn't do distance updates)

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Nav/Radar Console Distance calculation more correct

### DIFF
--- a/Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
+++ b/Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
@@ -1,5 +1,4 @@
 using System.Numerics;
-using Content.Shared.Coordinates;
 using Content.Shared.Shuttles.BUIStates;
 using Content.Shared.Shuttles.Components;
 using Content.Shared.Shuttles.Systems;
@@ -221,7 +220,8 @@ public sealed partial class ShuttleNavControl : BaseShuttleControl
 
                 var gridCentre = Vector2.Transform(gridBody.LocalCenter, curGridToView);
 
-                var gridCenterPos = _transform.GetWorldPosition(gUid) + gridBody.LocalCenter;
+                
+                var gridCenterPos = _transform.GetWorldPosition(gUid) + (_transform.GetWorldRotation(gUid).ToVec() * gridBody.LocalCenter);
 
                 var gridDistance = Vector2.Distance(gridCenterPos, xform.LocalPosition);
                 var labelText = Loc.GetString("shuttle-console-iff-label", ("name", labelName),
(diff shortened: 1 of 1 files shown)


MantasDab360:
Salvage major!!

Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
```diff
@@ -220,7 +220,10 @@ protected override void Draw(DrawingHandleScreen handle)
 
                 var gridCentre = Vector2.Transform(gridBody.LocalCenter, curGridToView);
 
-                var gridDistance = (gridBody.LocalCenter - xform.LocalPosition).Length();
+                
+                var gridCenterPos = _transform.GetWorldPosition(gUid) + (_transform.GetWorldRotation(gUid).ToVec() * gridBody.LocalCenter);
+
+                var gridDistance = Vector2.Distance(gridCenterPos, xform.LocalPosition);
```
ShepardToTheStars: Requires `// DeltaV` comments.

ShepardToTheStars (changes requested):
Have you considered making a PR to upstream for this?

Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
Coryler: I can write up an upstream pr for this (that's the one from wizden right?) Just hoped I'd find a solution for the handheld mass scanner... it irks me a bit that the numbers don't update :').

Still want this PR then meanwhile?

ShepardToTheStars: I'd still approve it, yea. If it also gets accepted into upstream, then I'd just remove the `// DeltaV` comments once I get to it. :)

ShepardToTheStars (approved):

Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
Coryler: Alright, just made another commit with the DeltaV comments. o7. Will take a look into making an upstream pr tommorow when I got time.

PR #4719, opened 2025-11-23
Improve calculation... again

Coryler:
## About the PR

The calculation again, to not use math which doesn't work completely, and the comment of the actual difference to upstream once their fix is pushed.

## Why / Balance

Consistent Distances which don't change with the angle.

## Technical details

Math is hard at night. Multiplying your Vector with Directional Vector doesn't make sense and gives inconsistent results.
Changed it to Transform.

## Media

Super small change.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: distance to fit more


Vapetastic-Gaming:
are you fixing math ?

Coryler:
@Vapetastic-Gaming Yes basically... the one I put in before will give different results depending on the rotation and is wrong.

ShepardToTheStars (approved):

PR #4728, opened 2025-11-23
Add Minor Mass Mind Swap

Coryler:
## About the PR

Adding a Minor Version of the Mass Mind Swap at a lower glimmer level with a ton of time (60s) to avoid it, whilst also only affecting active players.

## Why / Balance

Two reasons:
- I want to give a new reason for Epistemics paranoia
- Give a smaller version of the big one so people know what to avoid (only affecting a small amount of people)

## Technical details

- Made two new Rule Classes for the Minor Swap
- Editted the glimmer_events.yml and adding it to the list

## Media

https://github.com/user-attachments/assets/e55ea8e1-882f-4d08-ae27-543e6b8bc054

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The minor mass mind swap glimmer event has been added. If glimmer is above 750, it can mindswap up to 6 people temporarily.


Coryler:
Test error seems unrelated.

Vapetastic-Gaming:
direction review 48h. Apologies, but due to us having so many Content reviews, Im going to delay this for an extra 24h. This is a good problem to have :)

Vapetastic-Gaming:
could you define what you mean by active players

Content.Server/Nyanotrasen/StationEvents/Components/_DV/MinorMassMindSwapRuleComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Content.Server.Nyanotrasen.StationEvents.Components._DV;
+
+[RegisterComponent, Access(typeof(MinorMassMindSwapRule))]
+public sealed partial class MinorMassMindSwapRuleComponent : Component
+{
+    /// <summary>
+    /// The mind swap is only temporary if true.
+    /// </summary>
+    [DataField("isTemporary")]
+    public bool IsTemporary;
```
ShepardToTheStars: Set a default value.

kotobdev:
anecdotal comment: for something starting at 500, ideally this would be temporary. but i do know that the IsTemporary parameter is rather buggy on the existing mass mindswap (last I checked)

code-wise: any reason this can't be integrated into the existing MassMindSwapRule, perhaps just as a set of parameters in the component? duplicating this much code seems wacky, considering the minor and normal mindswaps seem to essentially do the same thing, just with minor differences

ShepardToTheStars (changes requested):
Please move all components to the _DV namespace and out of the Nyanotrasen namespace.

ShepardToTheStars:
Lemme look at the current MindSwap rule. You bring up a good point.

Coryler:
> direction review 48h. Apologies, but due to us having so many Content reviews, Im going to delay this for an extra 24h. This is a good problem to have :)

All right no problem :D. Take your time.

> could you define what you mean by active players

The actual mass mind swap, which happens at 900, also includes all other Psionic Entities (including NPCs) as targets the player can swap into. Active Players, means that only player controlled actors minds are swapped. (Player <=> Player, Instead of Player => NPC).

~~May I add that I find it scary that the Rat King is a viable target for those swaps?~~

> anecdotal comment: for something starting at 500, ideally this would be temporary. but i do know that the IsTemporary parameter is rather buggy on the existing mass mindswap (last I checked)

Honestly had a bit of a discussion with someone and was thinking to move this up to 750... so the moment the High Glimmer Visual effect starts, so it's both later and makes more sense.

> code-wise: any reason this can't be integrated into the existing MassMindSwapRule, perhaps just as a set of parameters in the component? duplicating this much code seems wacky, considering the minor and normal mindswaps seem to essentially do the same thing, just with minor differences

I can do and rewrite it, just thought it might be easier to seek and find and fits with the other glimmer events (as seperate files). Was rather thinking pulling out some functions from both (Like pooling the players, utility stuff) and putting them in a seperate util class.

> Please move all components to the _DV namespace and out of the Nyanotrasen namespace

> Set a default value.

Will do1

mmmrooster:
Would definitely say moving it to 750 is better, 500 is way too low
makes it feel kind of like a precursor event to the mass mindswap at 900, letting you see a smaller form of what's going to go down *soon* if you dont get things under control

Woljif:
I agree 500 is pretty low, epi is basically hitting 500 every shift whether you like it or not unfortunately. Who exactly and how many people does it affect, and is it permanent?

Coryler:
> Would definitely say moving it to 750 is better, 500 is way too low
makes it feel kind of like a precursor event to the mass mindswap at 900, letting you see a smaller form of what's going to go down soon if you dont get things under control

Yes. With the Glimmer Effects starting then, would be cool.

> I agree 500 is pretty low, epi is basically hitting 500 every shift whether you like it or not unfortunately. 

The initial 500 was me, a Epistemic guy who thinks up to 750 Glimmer is fine, was still ok :'D.

> Who exactly and how many people does it affect, and is it permanent?

It randomly decides/rolls a pair amount of 1 to 5 (of which I'm thinking lowering to 3). Then from all actively played characters from the psychic pool, randomly pulls out 1-5 pairs. Which, in most cases, is an amount of people which can be fixed in a timely manner with surgery (thus 3 pairs maximum, to make that more possible).

Coryler:
> Permanent

I'm... not really aware what the bug was for the mass mind swap, that temporary doesn't work.
Technically temporary seems to work for this... at least on the dev environment.

So can be considered... just feels a bit weak at 750.

https://github.com/user-attachments/assets/7cdb8607-2581-4050-92e6-4a8576819c03

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
```diff
@@ -0,0 +1,102 @@
...
+                && HasComp<ActorComponent>(psion))
+            {
+                // Only a list of Players
+                psionicActors.Add(psion);
+            }
+        }
+
+        // We go with 4 pairs for now
+        List<EntityUid> actorsToSwap = new();
+        var swapPairCount = _random.Next(1, 3);
```
ShepardToTheStars: Add number of pairs to the rule component and set 3 as the default. Also add a bit of logic that if its 0 or less, it sets it to 1 (to prevent people being idiots in the code)

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        var query = EntityQueryEnumerator<PotentialPsionicComponent, MobStateComponent>();
+        while (query.MoveNext(out var psion, out _, out _))
+        {
+            if (_mobstateSystem.IsAlive(psion) && !HasComp<PsionicInsulationComponent>(psion)
+                && HasComp<ActorComponent>(psion))
+            {
+                // Only a list of Players
+                psionicActors.Add(psion);
+            }
```
ShepardToTheStars: nit: Unnecessary brackets.

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
```diff
@@ -0,0 +1,102 @@
...
+        while (query.MoveNext(out var uid, out var comp, out var ruleComp))
+        {
+            if (comp.SoundTime != null && comp.SoundTime <= Timing.CurTime)
+            {
+                _audio.PlayGlobal(_resolvedWarningSound, Filter.Broadcast(), true);
+                comp.SoundTime = null;
+            }
+
+            if (comp.SwapTime == null || comp.SwapTime > Timing.CurTime)
+                continue;
```
ShepardToTheStars, line 68: Do the null check first in the while loop, and then you won't have to check `comp.SoundTime != null` when you check `comp.SoundTime <= Timing.CurTime`.

```
        while (query.MoveNext(out var uid, out var comp, out var ruleComp))
        {
            if (comp.SwapTime == null || comp.SwapTime > Timing.CurTime)
                continue;

            if (comp.SoundTime <= Timing.CurTime)
            {
                _audio.PlayGlobal(_resolvedWarningSound, Filter.Broadcast(), true);
                comp.SoundTime = null;
            }
```

Content.Server/_DV/StationEvents/Components/MinorMassMindSwapRuleComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Content.Server._DV.StationEvents.Components;
+
+[RegisterComponent, Access(typeof(MinorMassMindSwapRule))]
+public sealed partial class MinorMassMindSwapRuleComponent : Component
+{
+    /// <summary>
+    /// The mind swap is only temporary if true.
+    /// </summary>
+    [DataField("isTemporary")]
+    public bool IsTemporary = false;
```
ShepardToTheStars, line 14: If temporary doesn't work at the moment, please note that in the summary.

ShepardToTheStars (changes requested):

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: Ah will change that. Somewhere else I kept being screamed at where I don't do exactly that :3

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: > Do the null check first in the while loop, and then you won't have to check `comp.SoundTime != null` when you check `comp.SoundTime <= Timing.CurTime`.
> 
> ```
>         while (query.MoveNext(out var uid, out var comp, out var ruleComp))
>         {
>             if (comp.SwapTime == null || comp.SwapTime > Timing.CurTime)
>                 continue;
> 
>             if (comp.SoundTime <= Timing.CurTime)
>             {
>                 _audio.PlayGlobal(_resolvedWarningSound, Filter.Broadcast(), true);
>                 comp.SoundTime = null;
>             }
> ```

Ok uh...
Have tried a bit, and this is originally from the actual MassSwap code. It's intended to be that way so the sound starts playing 15s before the actual swaps. 

Changing the order will make it a bit wonky and buggy. Espescially since the comp.xxxTime = null delcartion in the if condition is to prevent it going back int.

If it 100% has to be changed I can probably find something, though this seems to be the simplest solution for now.

Toby222: The suggestion is not accurate, only the `== null` part should be moved up
And the continue after `SoundTime = null` should stay

Coryler:
> direction review 48h. Apologies, but due to us having so many Content reviews, Im going to delay this for an extra 24h. This is a good problem to have :)

Has there there been an update from direction or is there more time needed?

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: Ok. Made a change which works and moves `SwapTime == null` up.

Vapetastic-Gaming:
So the update from direction, The only way this can be approved is if it is temporary, at a glimmer count of 750 or greater, and lasts less than 120 seconds

Coryler:
Ok, @Vapetastic-Gaming. 

> So the update from direction, The only way this can be approved is if it is temporary, at a glimmer count of 750 or greater, and lasts less than 120 seconds

Status now: 
- It's now temporary, with a 120s delay before you can click to swap back
- Glimmer Event starting at 750.
- Max people involved: 6

Vapetastic-Gaming:
Sounds good

Content.Server/_DV/StationEvents/Components/MinorMassMindSwapRuleComponent.cs
Coryler: After changing a bit of code it works as it intended. (It works from what I get, though was only buggy on the bigger one).

ShepardToTheStars (approved):

PR #4734, opened 2025-11-24
Holofan Rebalance (and standardizing Holoprojectors)

ShepardToTheStars:
## About the PR
I changed all holoprojectors to use a limited changes system, that produce a temporary holoprojection. This makes all holoprojectors function the same - with the exception of the force field projector. That is the only one that cannot pick up its holograms (I figured it would be too strong).

It is also similar to the current holofan system where you can pick up holofans. That has not changed. The projectors recharge at the same rate their projections last.

Holofans: Last 300 seconds. Recharges one charge in 300 seconds. 6 charges.
Security Holobarriers: Last 180 seconds. Recharges one charge in 180 seconds. 6 charges.
Janitor Holosign: Lasts 90 seconds. Recharges in 90 seconds. 6 charges. 
Holopeel: Lasts 30 seconds. Recharges in 30 seconds. 3 charges.
Engineering Caution Sign: Lasts 180 seconds. Recharges in 180 seconds. 6 charges.
Force Field Projector: Lasts 180 seconds. Recharges in 180 seconds. 6 charges. ~~Cannot be picked up.~~

## Why / Balance
With the Delta Pressure change, infinite duration holofans are too overpowered. Also, people cannot just put holofans over hyper-convection fabricators and cheese the system anymore.

Figured I'd fix that and also bring up all other projectors up to the same level.

## Technical details
I dunno. Code. Added a few components. Port a component from [Den](https://github.com/TheDenSS14/TheDen/pull/1394), but modified a good portion of it.

## Media

image
image
image

https://github.com/user-attachments/assets/347b7897-3809-4dd3-9a02-c2be1424a4fd

## Requirements

## Breaking changes

**Changelog**

:cl: ShepardToTheStars, Jakumba
- tweak: Holofans are no longer infinite duration. They last 5 minutes but also recharge over time. They can still be picked up.
- tweak: All holoprojectors function the same way now. You can pick up janitor holosigns, security holobarriers, and they all recharge over time.


ShepardToTheStars:
Direction approved.

Toby222:
Test fail seems real

ShepardToTheStars:
> Test fail seems real

Only cause that test specifically does not include entities with TimedDespawn components, so makes sense to exclude TimedDespawnDetailed as well.

Content.Server/_DEN/TimedDespawn/TimedDespawnDetailSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+        if (remaining == null || remaining.Value.TotalSeconds > 0)
+            return;
+
+        if (ent.Comp.EndSound != null)
+        {
+            var entCoords = _transformSystem.GetMoverCoordinates(ent);
+            _audioSystem.PlayPredicted(ent.Comp.EndSound, entCoords, null, ent.Comp.EndSoundParams);
+        }
+
+        EntityManager.QueueDeleteEntity(ent);
```
cmorley191, line 93: I know this is _DEN code so feel free to ignore.

Is it a problem that StopTimer never gets called? If the timer runs out naturally, doesn't this ent still remain in `_timedDespawns`? (even though it's being deleted from the entitymanager)

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
```diff
@@ -36,24 +36,13 @@ private void OnInit(Entity<ChargeHolosignProjectorComponent> ent, ref ComponentI
...
 
-        ent.Comp.Container = _container.EnsureContainer<Container>(ent, ent.Comp.ContainerId);
         ent.Comp.SignComponent = EntityManager.ComponentFactory.GetRegistration(ent.Comp.SignComponentName).Type;
     }
 
     private void OnMapInit(Entity<ChargeHolosignProjectorComponent> ent, ref MapInitEvent args)
     {
         if (!TryComp<LimitedChargesComponent>(ent, out var charges))
             return;
-
```
cmorley191, line 47: Looks like you could just remove this whole function now, and its constructor subscription

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
```diff
@@ -66,54 +55,54 @@ private void OnBeforeInteract(Entity<ChargeHolosignProjectorComponent> ent, ref
...
         var coords = args.ClickLocation.SnapToGrid(EntityManager);
         var mapCoords = _transform.ToMapCoordinates(coords);
+
         _signs.Clear();
+
         _lookup.GetEntitiesInRange(ent.Comp.SignComponent, mapCoords, 0.25f, _signs);
+
         if (_signs.Count == 0)
-            TryPlaceSign((ent, ent, charges), coords, args.User);
+            TryPlaceSign((ent, ent, charges), args);
```
cmorley191, line 54: As an alternative you could do:

```if (!ent.Comp.CanPickup || _signs.Count == 0)```

That would allow you to place duplicate signs on a tile, if the projector type does not allow pickups.

Rational minds may disagree on whether that's a good feature. Wizden has that feature, since their holofans aren't pick-up-able. (it can be useful if you know a fan's about to expire)

Comment (cmorley191) [reply Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs:54]: If you do go this route I would actually flip the if statement around so the logic reads a bit more smoothly

```
if (ent.Comp.CanPickup && _signs.Count >= 1)  // if we can remove, remove!
  TryRemoveSign...
else  // otherwise place
  TryPlaceSign...
```

File: Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs [hunk 4]
```diff
@@ -66,54 +55,54 @@ private void OnBeforeInteract(Entity<ChargeHolosignProjectorComponent> ent, ref
...
         }
 
-        var placed = container.ContainedEntities.First(); // checked Count beforehand so this won't fail
-        _transform.SetCoordinates(placed, coords);
-        _transform.AnchorEntity(placed);
+        var holoUid = EntityManager.PredictedSpawnAtPosition(ent.Comp1.SignProto, args.ClickLocation.SnapToGrid(EntityManager));
+        var xform = Transform(holoUid);
+        if (!xform.Anchored)
+            _transform.AnchorEntity(holoUid, xform); // anchor to prevent any tempering with (don't know what could even interact with it)
+
```
cmorley191, line 73: Hm, I wonder why they went this "ContainedEntities" route in the first place... seems a bit roundabout compared to what you have here...

Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
```diff
@@ -6,51 +6,61 @@
...
-  components:
-  - type: ItemSlots
-    slots:
-      cell_slot:
-        name: power-cell-slot-component-slot-name-default
+  #components: # DeltaV - Holoprojectors don't need batteries
+  #- type: ItemSlots
+  #  slots:
+  #    cell_slot:
+  #      name: power-cell-slot-component-slot-name-default
```
cmorley191, line 46: You could make this actually empty:

```
- type: LimitedCharges
  maxCharges: 6
  lastCharges: 0
```

cmorley191: Although there's no point to mapping this entity now -- it'll just recharge over the first few minutes of the shift...

Should we just remove this entity entirely?

Edit: ah it seems `Recipes/Lathes/janitorial.yml` produces this still

Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
```diff
@@ -105,13 +114,21 @@
...
+  # End DeltaV Additions
+
 
 - type: entity
   parent: HolofanProjector
   id: HolofanProjectorEmpty
   categories: [ HideSpawnMenu ] # DeltaV - this is identical to the normal one
   suffix: Empty
-  #components: # DeltaV - no cell slot for empty one
+  #components: # DeltaV - Holoprojectors dont need batteries
```
cmorley191, line 132: Same comment as above. Lathes produce the empty version, could make them actually empty with `lastCharges: 0` if you want

Or just remove these empty versions entirely, make the lathe produce full versions (again, if you want, this is minor)

Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
ShepardToTheStars: These entities are more-or-less holdovers from Wizden’s prototypes. I opted to just hide from the spawn menu and leave them since removing them and running a migration seemed more intrusive.

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
ShepardToTheStars: Unsure. I’m pretty sure its the reason why Skias die when sitting next to a holofan. Since the entities inside have the PointLight component

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
```diff
@@ -29,10 +25,19 @@
...
             acts: [ "Destruction" ]
+  # DeltaV Additions
+  - type: TimedDespawnDetailed
+    lifetime: 90
+  - type: Holosign
+  - type: PointLight # DeltaV - lit holograms
+    enabled: true
+    radius: 3
+    color: yellow
+  # END DeltaV
```
cmorley191, line 23: \<super_minor>

In future I'd consider just putting new components like this right up there below the commented-out old one. Might save you some `# DeltaV` comment lines and also save readers some time scrolling around looking for components.

</super_minor>

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
ShepardToTheStars: Can do that. Should I still include the CanPickup check in the TryRemoveSign in that case?

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
ShepardToTheStars: True. Ill fix that.

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
```diff
@@ -45,17 +50,15 @@
...
     noAirWhenFullyAirBlocked: false
-  - type: Holofan # DeltaV - for whitelisting
-  - type: PointLight # DeltaV - lit holograms
-    enabled: true
-    radius: 3
-    color: yellow
+  # DeltaV Additions
+  - type: TimedDespawnDetailed
+    lifetime: 300
+  - type: Holofan
```
cmorley191: Pretty sure this entity is still inheriting `- type: Holosign` from its parent as well.

I'm guessing that means:
- holofan projector can pick up only holofans
- but wet floor sign projector can pick up like, everything (wet floor signs, holofans, etc.)

Not sure if that's a major problem. I mean maybe a janitor can grief a bit by removing a safety-critical holofan that an atmos tech put down...

cmorley191: If you wanted to fix this, simplest way would probably be to turn the HolosignWetFloor entity code above into just a BaseHoloprojection abstract entity *without* the Holosign component, and add a new little HolosignWetFloor entity child that adds the component

Resources/Prototypes/_DV/Entities/Objects/Specific/Security/security.yml
```diff
@@ -68,10 +68,3 @@
   id: HoloprojectorSecurityRecharging
   suffix: Recharging, Do Not Map
   categories: [ HideSpawnMenu ]
```
cmorley191: Maybe just remove this entity, swap out the BorgModuleSecurityPatrol for a normal one (that's the only place this is used). Would reduce clutter is all

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
cmorley191: TryRemoveSign isn't called anywhere else, so it isn't *necessary* to have it. But the function *is* public (might call it from elsewhere in the future?) so may as well keep it. It ain't hurting anything.

Either way works for me.

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
ShepardToTheStars: You’re right. God, the thought of making a new base HoloProjection abstract/base prototype did cross my mind at one point but since these are upstream prototypes, I opted to keep the inheritance structure the same.

I may just create a new file in the DV namespace and do this. Possibly submit the inheritance fix to upstream

cmorley191:
Think I'm done leaving comments. I realize now that you can't Resolve any of them which is extremely annoying -- I'll just do a normal review next time instead of Single Comments. Did not know how that works.

You'll need a maintainer with write access to review (i.e. not me... yet)

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
cmorley191: I don't think it would be terrible to do in this upstream file. Basically any new upstream changes to HolosignWetFloor would now *actually* be changes to our BaseHoloprojection. I think git actually would find that automatically too! (since it was basically just a rename and adding `abstract: true`) Might actually be cleaner for merges than moving it out to _DV

Resources/Prototypes/_DV/Entities/Objects/Specific/Security/security.yml
Toby222: Seconded, ditch this and only use HoloprojectorSecurity, it's all in _DV anyway

Resources/Prototypes/_DV/Entities/Objects/Devices/holoprojectors.yml
```diff
@@ -1,9 +1,5 @@
 - type: entity
```
Toby222, line 1: This prototype is also redundant

Content.Shared/_DV/Holosign/ChargeHolosignProjectorComponent.cs
```diff
@@ -15,23 +14,17 @@ public sealed partial class ChargeHolosignProjectorComponent : Component
...
-    public EntProtoId SignProto;
+    [DataField]
+    public EntProtoId SignProto = "HolosignWetFloor";
 
-    /// <summary>
-    /// Component on <see cref="SignProto"/> to check for duplicates.
-    /// </summary>
-    [DataField(required: true)]
-    public string SignComponentName;
+    [DataField]
```
Toby222, line 24: Don't completely remove the comment explaining what SignComponentName is/does

Content.Server/_DEN/TimedDespawn/TimedDespawnDetailSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+/// <summary>
+/// This handles the <see cref="TimedDespawnDetailedComponent"/>
+/// </summary>
+public sealed class TimedDespawnDetailedSystem : EntitySystem
+{
+    [Dependency] private readonly AudioSystem _audioSystem = default!;
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
+    [Dependency] private readonly TransformSystem _transformSystem = default!;
+
+    private HashSet<EntityUid> _timedDespawns = new();
```
Toby222: Make field readonly
```suggestion
    private readonly HashSet<EntityUid> _timedDespawns = new();
```

Toby222 (changes requested):
Very cursory review

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
ShepardToTheStars: Done.

Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
ShepardToTheStars: I could, but with the original implementation from Wizden, empty means it spawns without a battery in it, rather than no changes. Things like Welding Tools spawn with maximum fuel, and if we print an empty holofan, it will take 5 minutes to get a charge, making it useless for that time. It might be better for players just to give them a full one, as if they were putting in a fresh battery into a newly printed holofan.

Resources/Prototypes/_DV/Entities/Objects/Specific/Security/security.yml
ShepardToTheStars: Done.

Content.Server/_DEN/TimedDespawn/TimedDespawnDetailSystem.cs
ShepardToTheStars: Good point. The array would just keep getting bigger and bigger I think. I'll add it to the `TryDelete` method.

Content.Shared/_DV/Holosign/ChargeHolosignProjectorComponent.cs
ShepardToTheStars: I'll re-add it but more accurate on what it actually does.

Resources/Prototypes/_DV/Entities/Objects/Devices/holoprojectors.yml
ShepardToTheStars: Removed and migrated since it was ours.

ShepardToTheStars:
@Toby222 Ready for review (btw)

Content.Shared/_DEN/TimedDespawn/TimedDespawnDetailedComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+    [DataField]
+    public AudioParams StartSoundParams { get; set; } = AudioParams.Default;
+
+    [DataField]
+    public SoundSpecifier? EndSound { get; set; }
+
+    [DataField]
+    public AudioParams EndSoundParams { get; set; } = AudioParams.Default;
+
+}
```
Toby222, line 34: Remove extra blank line
```suggestion
    public AudioParams EndSoundParams { get; set; } = AudioParams.Default;
}
```

Content.Server/_DEN/TimedDespawn/TimedDespawnDetailSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+using Content.Shared.Examine;
+using Robust.Server.Audio;
+using Robust.Server.GameObjects;
+using Robust.Shared.Timing;
+
+
+namespace Content.Server._DEN.TimedDespawn;
+
+
+/// <summary>
```
Toby222, line 9: Remove extra blank lines
```suggestion
using Robust.Shared.Timing;

namespace Content.Server._DEN.TimedDespawn;

/// <summary>
```

Content.Shared/_DV/Holosign/ChargeHolosignProjectorComponent.cs
```diff
@@ -15,23 +14,22 @@ public sealed partial class ChargeHolosignProjectorComponent : Component
...
-    /// Container to store sign entities in before they are "spawned" on use.
+    /// If true, the holosign projector can pick up the entity whitelisted in the SignComponentName variable.
     /// </summary>
     [DataField]
-    public string ContainerId = "signs";
-
-    [ViewVariables]
-    public Container Container = default!;
+    public bool CanPickup = true;
 }
```
Toby222, line 34: Removal of Container makes `using Robust.Shared.Containers` unused

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
```diff
@@ -19,14 +19,13 @@ public sealed class ChargeHolosignSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
 
-    private HashSet<Entity<IComponent>> _signs = new();
+    private HashSet<Entity<IComponent>> _placedSigns = new();
```
Toby222: Make field readonly
```suggestion
    private readonly HashSet<Entity<IComponent>> _placedSigns = new();
```

Separately: `_container` is unused and can be removed
`using Robust.Shared.Map;` is unused and can be removed

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
```diff
@@ -66,54 +48,54 @@ private void OnBeforeInteract(Entity<ChargeHolosignProjectorComponent> ent, ref
...
-        }
 
-        if (!_container.Insert(sign, ent.Comp1.Container, force: true))
+        // don't overfill
+        if (_charges.GetCurrentCharges((ent, ent.Comp2)) < ent.Comp2.MaxCharges)
         {
-            Log.Error($"Failed to insert holosign {ToPrettyString(sign)} back into {ToPrettyString(ent)}!");
-            return false;
+            _charges.AddCharges((ent, ent.Comp2), 1);
         }
```
Toby222: Pretty sure AddCharges already handles overfill?

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
```diff
@@ -1,38 +1,39 @@
+
```
Toby222: Remove blank line at file begin

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
```diff
@@ -45,21 +46,20 @@
...
   - type: Airtight
     noAirWhenFullyAirBlocked: false
-  - type: Holofan # DeltaV - for whitelisting
-  - type: PointLight # DeltaV - lit holograms
-    enabled: true
-    radius: 3
-    color: yellow
+  # DeltaV Additions
+
+  # END DeltaV
```
Toby222: Bunged up End comment?

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
```diff
@@ -117,18 +120,22 @@
...
+    - type: Holobarrier # DeltaV - Whitelist for holoprojector pickup
     - type: PointLight
       enabled: true
       radius: 3
       color: red
     - type: Climbable
     - type: Clickable
 
+
 - type: entity
```
Toby222, line 140: Remove new blank line
```suggestion

- type: entity
```

Toby222 (changes requested):

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
ShepardToTheStars: Yea

Content.Shared/_DV/Holosign/ChargeHolosignSystem.cs
ShepardToTheStars: Looks like it does. The old charge system might not have handled that

Resources/Prototypes/Entities/Structures/Holographic/projections.yml
ShepardToTheStars: Fixed all comments in the file.

ShepardToTheStars:
Alright, ready for re-review.

Toby222 (approved):
seems good enough

PR #4640, opened 2025-11-12
Revert New Xenoarch

kotobdev:
## About the PR
Reverts https://github.com/DeltaV-Station/Delta-v/pull/4470/, bringing xenoarch back to it's pre-rework state.

## Why / Balance
The general reaction I've seen to new xenoarch is mixed at *best*, and I'm inclined to agree. On a technical level, it feels buggy, with some triggers being wonky to obtain alongside other triggers, some nodes being entirely impossible to trigger (without artifexium), some triggers being needlessly confusing in their wording and/or mechanical activation, and some deltav-specific mechanics still missing/not properly functional. On a conceptual level, the new system is extremely predictable and extremely safe, lacking the dangerous aspects of old xenoarch. There is a literal ten second timer you can use to escape the area before it does anything, and you can see exactly what the upcoming nodes do before you trigger them. 

New xenoarch was a valiant attempt at improving the system, and it does succeed in some ways, namely the UI. However, I believe it didn't hit the mark.

## Technical details
If we wanted to revert this cleanly, we should have done it before the upstream merge. Unfortunately, we're now post-upstream-merge, meaning I had to just let God take the wheel when resolving the merge conflicts. A lot of this code doesn't even exist on upstream anymore, and I get the impression the original new xenoarch PR didn't care too much about preserving deltav comments on imports and whatnot, so I'm just gonna leave it as-is for the sake of my own sanity.

I "tested" this in the sense that I confirmed it ran, and confirmed that the xenoarch interactions I could think of (artifexium, crushing, just normal xenoarch stuff) are still functional and don't crash the server. Something felt weird with the analyzer picking up new artifacts, but I couldn't reproduce anything so it's probably fine. CI/CD shall judge my soul. **if anybody else wants to pull this PR and test it on their own, that would be awesome.**

Notably, the UnlockNodeCommand has been nuked and I might've fucked up and nuked XenoArtifactUnlockNodeCommand too, but you can just right-click activate artifacts anyways so who care

## Media

image

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- remove: The new artifact system has been reverted. Get ready to print a lot of artifact reports again.

PR #4740, opened 2025-11-24
xenoarch: remove impossible artifact trigger combinations

cmorley191:
## About the PR

Blacklisted artifacts from having undesired trigger combinations, such as "Low Pressure + High Pressure", or "Low Temperature + High Temperature".

## Why / Balance

This was a common criticism of the new xenoarch system, when discussing the system in https://github.com/DeltaV-Station/Delta-v/pull/4640

While technically doable (frezon+welding will give both temperature triggers, and just "be fast" to do the pressure combination), these are impractical trigger combinations.

## Merging

## Technical details

The way this *actually* works is it blacklists the two triggers from appearing on the same artifact, not just on the same node. Limiting it to the same node would be doable in the code, but way more complicated / hard to review and maintain. Happy to discuss that approach though.

Recommended review order:
- `XenoArchTriggerPrototype.cs`
- `triggers.yml`
- `XenoArtifactSystem.ProcGen.cs`

## Media

Difficult to demonstrate visually. I tested it by adding a log-statement in the if-statement, giving details when the blacklist triggered. It does work as intended, blocking one or the other trigger when one of the pair is already present (spawned lots of artifacts to test).

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: "Impossible" arti trigger combinations (like low+high pressure) will no longer appear on artifacts.

### DIFF
--- a/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
+++ b/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
@@ -53,8 +53,14 @@ public sealed partial class XenoArtifactSystem
             var triggerId = RobustRandom.Pick(weightsByTriggersLeft);
             weightsByTriggersLeft.Remove(triggerId);
             var trigger = PrototypeManager.Index<XenoArchTriggerPrototype>(triggerId);
-            if (_entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent))
+            // DeltaV - start of TriggerBlacklist
+            if (
+                _entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent)
+                || (trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
+                || (triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId)))
+            )
                 continue;
+            // DeltaV - end of TriggerBlacklist
 
             triggerPool.Add(trigger);
         }
--- a/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
+++ b/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
@@ -53,10 +53,11 @@ public sealed partial class XenoArtifactSystem
             var triggerId = RobustRandom.Pick(weightsByTriggersLeft);
             weightsByTriggersLeft.Remove(triggerId);
             var trigger = PrototypeManager.Index<XenoArchTriggerPrototype>(triggerId);
+            if (_entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent))
+                continue;
             // DeltaV - start of TriggerBlacklist
             if (
-                _entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent)
-                || (trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
+                (trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
                 || triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId))
             )
                 continue;
--- a/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
+++ b/Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
@@ -57,7 +57,7 @@ public sealed partial class XenoArtifactSystem
             if (
                 _entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent)
                 || (trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
-                || (triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId)))
+                || triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId))
             )
                 continue;
             // DeltaV - end of TriggerBlacklist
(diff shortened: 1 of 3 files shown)


dekigangsta:
1 step closer to peak epi gameplay...

Bonktrauma:
So actually,  this is very possible, aslong as there's multiple atmospheric chambers and you've got a holoprojector and hardsuit

EDIT:HOLOPROJECTOR, NOT HOLOCAUST

AveriV2:
> EDIT:HOLOPROJECTOR, NOT HOLOCAUST

that is such a unfortunate auto-correct, my god

StarboundSolaris:
> So actually, this is very possible, aslong as there's multiple atmospheric chambers and you've got a holoprojector and hardsuit
> 
> EDIT:HOLOPROJECTOR, NOT HOLOCAUST

jarvis, set the "days since last autocorrect fuckup" counter back to 0

Content.Shared/Xenoarchaeology/Artifact/Prototypes/XenoArchTriggerPrototype.cs
```diff
@@ -25,6 +25,14 @@ public sealed partial class XenoArchTriggerPrototype : IPrototype
     [DataField]
     public EntityWhitelist? Whitelist;
 
+    // DeltaV - start of TriggerBlacklist
+    /// <summary>
+    /// Triggers that CANNOT appear on the same artifact as this trigger.
+    /// </summary>
+    [DataField]
+    public List<ProtoId<XenoArchTriggerPrototype>>? TriggerBlacklist;
```
Toby222:
```suggestion
    public HashSet<ProtoId<XenoArchTriggerPrototype>>? TriggerBlacklist;
```

Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
```diff
@@ -53,8 +53,14 @@ private List<XenoArchTriggerPrototype> CreateTriggerPool(Entity<XenoArtifactComp
...
             var trigger = PrototypeManager.Index<XenoArchTriggerPrototype>(triggerId);
-            if (_entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent))
+            // DeltaV - start of TriggerBlacklist
+            if (
+                _entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent)
+                || (trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
+                || (triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId)))
+            )
                 continue;
+            // DeltaV - end of TriggerBlacklist
```
Toby222, line 65: Remove extra parentheses, move comments so we're not marking the original line as part of the dV block
```suggestion
            var trigger = PrototypeManager.Index<XenoArchTriggerPrototype>(triggerId);
            if (
                _entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent)
                || trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID))     // DeltaV - add blacklist of conflicting triggers
                || triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId)) // DeltaV - add blacklist of conflicting triggers
            )
                continue;

```

Resources/Prototypes/XenoArch/triggers.yml
```diff
@@ -40,6 +40,8 @@
 - type: xenoArchTrigger
   id: TriggerHeat
   tip: xenoarch-trigger-tip-heat
+  triggerBlacklist: # DeltaV
```
Toby222: Change the dV comments here to explain what was added
```suggestion
  triggerBlacklist: # DeltaV - add blacklist of conflicting triggers
```

Toby222 (dismissed):

Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
cmorley191: On past PRs I have been told to lean towards a wider comment net in cases like this (an if-statement condition), rather than dissecting units of code.

I gather that the reason is to make it less likely that the comment gets lost/missed when glancing over it, and also to reduce clutter. In reality, anyone doing merge conflicts (which is mainly why we add these comments) is also going to be considering this if-statement as a whole, and will have the upstream code side-by-side to compare.

So I'd prefer not to do this, since I'm receiving mixed signals.

If you really want me to, I could split this into a separate `if(new_conditions) continue;` statement, right after the original one.

Resources/Prototypes/XenoArch/triggers.yml
cmorley191: Isn't "add blacklist of conflicting triggers" already obvious when you read the property name "triggerBlacklist" that this comment is on? (and then also see the context of the line below it)

I don't mean to be argumentative, but I see blank `# DeltaV` comments *everywhere* in places where the entire pointed-at feature has been added (in VSCode, search `*.yml` files for regex `# DeltaV(?!.)`  ). I take review requests like this seriously because I try to integrate them into my habits / update my coding style to minimize review issues in future PRs.

cmorley191: I'm a huge proponent of self-documenting code, too, so my main point here is the property name is already "triggerBlacklist". If there's something unclear about that, do you have a suggestion on how I could change the property name?

cmorley191:
@Toby222 ready for re-review. changes have been re-tested

Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
ShepardToTheStars: Alternatively, you can do this as well. It totally separates your changes from anything originally there.

EDIT: I should have finished reading your entire comment. Yea, this is basically what you said at the end.

```c#
if (_entityWhitelist.IsWhitelistFail(trigger.Whitelist, ent))
    continue;

// DeltaV - TriggerBlacklist
if ((trigger.TriggerBlacklist != null && triggerPool.Any(otherTrigger => trigger.TriggerBlacklist.Contains(otherTrigger.ID)))
  || triggerPool.Any(otherTrigger => otherTrigger.TriggerBlacklist != null && otherTrigger.TriggerBlacklist.Contains(triggerId)))
    continue;
// END DeltaV

```

Resources/Prototypes/XenoArch/triggers.yml
Toby222: > Isn't "add blacklist of conflicting triggers" already obvious when you read the property name "triggerBlacklist" that this comment is on? (and then also see the context of the line below it)

Without clarification it's unclear if it's just added to this item, newly implemented, or changed from parent, unless you look further

Content.Server/Xenoarchaeology/Artifact/XenoArtifactSystem.ProcGen.cs
Toby222: Splitting into separate if probably makes more sense yea

ShepardToTheStars (approved):

PR #4743, opened 2025-11-25
xenoarch: arti effect happens faster, for more fun danger

cmorley191:
## About the PR

Shortened the delay between 
1. a scientist triggering the final trigger of an arti node, and 
2. the node unlocking/the effect happening.

The delay before this PR was ~5-10 seconds (it varied depending on how fast you did the triggers). Now, it is a fixed 0.7 seconds.

## Why / Balance

This was a common criticism of the new xenoarch system, when discussing the system in https://github.com/DeltaV-Station/Delta-v/pull/4640

The delay was very long, meaning you had plenty of time to run away before the effect triggered. If the effect was "spawn dangerous fauna" or "explode" or something else, you'd have plenty of time to dodge it. People liked the high danger of the old xenoarch system, so let's bring some of it back!

## Merging

## Technical details

How it worked before: every trigger that occurs adds time to the "unlocking phase" timer. When that timer runs out, the unlocking checks occur; if they find a node that was triggered, it's unlocked.

How it works now: The same, but also every time a trigger occurs, all the standard unlocking checks are run as well. If they succeed, then the remaining time in the unlocking phase is set to the short 0.7 seconds time.

0.7 seconds was chosen because it's:
- Short enough that you can't run away
- Not instant, so there isn't too much happening visually on the screen at once
- Long enough that other conflicting triggers present (e.g. low oxygen + low pressure) can still occur consistently. (setting the timer to zero would probably cause intermittent / glitchy triggering)

## Media

Demo: artifact will `Create Hostile Fish` after triggers `Water` and `Blood`. See:
- Splash with Water, the first trigger
- Popup says "artifact begins to shift", indicating the start of the unlocking time
- Plenty of time is given to do the second trigger
- Splash with Blood
- A very short time later (0.7s), "the artifact is permanently changed", and the fish spawns
  - This would've been a lot longer before this PR; you'd have to wait for the rest of the unlocking phase timer to drain

https://github.com/user-attachments/assets/6d8fa5fc-737e-43ef-8194-e45c7d72397d

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Artifacts now activate much faster after you successfully perform their triggers. More danger = more fun

### DIFF
--- a/Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
+++ b/Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
@@ -78,18 +78,14 @@ public sealed partial class XenoArtifactComponent : Component
     [DataField]
     public TimeSpan UnlockStateRefractory = TimeSpan.FromSeconds(5);
 
-    // DeltaV - start of faster unlock effect
-
     /// <summary>
-    /// If set, will overwrite the unlocking state time remaining as soon as a node is ready to unlock (triggers are met).
+    /// DeltaV - If set, will overwrite the unlocking state time remaining as soon as a node is ready to unlock (triggers are met).
     ///   (why: letting the timer run out naturally makes artifacts less dangerous, since scientists have lots of time to run away)
     /// If null, the unlock time will just run its course naturally.
     /// </summary>
     [DataField]
     public TimeSpan? UnlockCompleteDuration = TimeSpan.FromSeconds(0.7);
     
-    // DeltaV - end of faster unlock effect
-
     /// <summary>
     /// When next unlock session can be triggered.
     /// </summary>
--- a/Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
+++ b/Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
@@ -78,6 +78,18 @@ public sealed partial class XenoArtifactComponent : Component
     [DataField]
     public TimeSpan UnlockStateRefractory = TimeSpan.FromSeconds(5);
 
+    // DeltaV - start of faster unlock effect
+
+    /// <summary>
+    /// If set, will overwrite the unlocking state time remaining as soon as a node is ready to unlock (triggers are met).
+    ///   (why: letting the timer run out naturally makes artifacts less dangerous, since scientists have lots of time to run away)
+    /// If null, the unlock time will just run its course naturally.
+    /// </summary>
+    [DataField]
+    public TimeSpan? UnlockCompleteDuration = TimeSpan.FromSeconds(0.7);
+    
+    // DeltaV - end of faster unlock effect
+
     /// <summary>
     /// When next unlock session can be triggered.
     /// </summary>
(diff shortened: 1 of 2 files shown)


dekigangsta:
dude this is amazing
to up more of fun-aspect could aswell not make description so detailed and bring back the old "enviromental disruption", or "biochemical disruption" etc etc

this is sick +1+1+1+1+1

cmorley191:
> could as well not make description so detailed and bring back the old "enviromental disruption", or "biochemical disruption"

There are conflicting schools of thought on this. On the one hand, mysterious trigger/effect tips are fun because they're mysterious, but they can also lead to frustration due to confusion about the exact mechanics. Some people have requested that they actually be clarified *more*.

I'd encourage you to add this to the discussion in https://github.com/DeltaV-Station/Delta-v/pull/4640, since that seems to be the de-facto "discuss how new xenoarch could be better like old xenoarch" forum at the moment.

Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
```diff
@@ -78,6 +78,18 @@ public sealed partial class XenoArtifactComponent : Component
     [DataField]
     public TimeSpan UnlockStateRefractory = TimeSpan.FromSeconds(5);
 
+    // DeltaV - start of faster unlock effect
+
+    /// <summary>
+    /// If set, will overwrite the unlocking state time remaining as soon as a node is ready to unlock (triggers are met).
+    ///   (why: letting the timer run out naturally makes artifacts less dangerous, since scientists have lots of time to run away)
+    /// If null, the unlock time will just run its course naturally.
+    /// </summary>
```
ShepardToTheStars: You can remove the `// DeltaV - start of faster unlock effect`, and just add it to the summary block like so.

```
    /// <summary>
    /// DeltaV - If set, will over...
```

Content.Shared/Xenoarchaeology/Artifact/SharedXenoArtifactSystem.XAT.cs
```diff
@@ -90,6 +90,13 @@ public void TriggerXenoArtifact(Entity<XenoArtifactComponent> ent, Entity<XenoAr
 
         if (node != null && unlockingComp.TriggeredNodeIndexes.Add(GetIndex(ent, node.Value)))
         {
+            // DeltaV - start of faster unlock effect
+            if (ent.Comp.UnlockCompleteDuration != null && TryGetNodeFromUnlockState((ent.Owner, unlockingComp, ent.Comp), out var unlockingNode))
+            {
+                unlockingComp.EndTime = _timing.CurTime + ent.Comp.UnlockCompleteDuration.Value;
```
ShepardToTheStars: Change the `ent.Comp.UnlockCompleteDuration != null` to `ent.Comp.UnlockCompleteDuration is {} unlockDuration` (or whatever local variable you wanna set), and then you can use that variable in place of the `ent.Comp.UnlockCompleteDuration.Value`.

ShepardToTheStars (changes requested):

cmorley191:
@ShepardToTheStars ready for re-review. changes have been re-tested

ShepardToTheStars (approved):

PR #4744, opened 2025-11-25
xenoarch: make commonly-conflicting atmos triggers appear less often

cmorley191:
## About the PR

Reduced the likelihood of the following artifact node triggers by ~half:
- `Low-oxygen environment` (reduce oxygen presence to ~10% or less)
- `High Pressure` (> 385 kPa)
- `Low Pressure` (< 50 kPa)
- `Heat` (> 100 C, or deal Heat damage - hit with a welding tool)
- `Cold` (< -20 C, or deal Cold damage - not sure how you do that)

That is, each of the above are roughly half as less likely to appear in any given artifact.

Related: a separate PR makes sure that certain "impossible" combinations of these don't show up. That's a separate thing: https://github.com/DeltaV-Station/Delta-v/pull/4740

## Why / Balance

This is related to a common criticism of the new xenoarch system, when discussing the system in https://github.com/DeltaV-Station/Delta-v/pull/4640

The new xenoarch system has the brand new feature that: **simultaneously activating two different, unrelated nodes in the tree will cause NEITHER of them to unlock**. The idea is that you have to create isolated conditions: trigger one, but avoid triggering others. It's an intended challenge.

However, people have been really struggling with these atmos ones. (e.g. opening the arti room blast doors creates low-oxygen AND low pressure AND Cold)

The broader concept here is that "scientists are generally not expected to have a detailed understanding of the atmos system and pumping infrastructure". For some of these triggers, a good atmos understanding (or an engi department that is free enough to send an atmos tech over) is necessary to isolate the effects.

Reducing the difficulty of these triggers (by reducing their frequency) may or may not be something you agree with. This is a balance change, and I defer to Direction to make the decision about where this should land.

## Technical details

`Weights = Weights / 2`

## Merging

## Media

N/A

Only test I did was spawning a bunch of artifacts and verified no error messages or anything popped up.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The more difficult atmos triggers on artifacts (low O2, heat, cold, low pressure, high pressure) will show up less frequently.

### DIFF
--- a/Resources/Prototypes/XenoArch/triggers.yml
+++ b/Resources/Prototypes/XenoArch/triggers.yml
@@ -3,6 +3,10 @@
   weights:
     TriggerMetapsionic: 1 # DeltaV
     TriggerMusic: 1
+    # DeltaV - make less frequent
+    #TriggerHeat: 1
+    #TriggerCold: 0.5
+    #TriggerNoOxygen: 1
     TriggerHeat: 0.5
     TriggerCold: 0.25
     TriggerNoOxygen: 0.5
@@ -14,6 +18,9 @@
     TriggerAmmonia: 0.1
     TriggerFrezon: 0.1
     TriggerRadiation: 1
+    # DeltaV - make less frequent
+    #TriggerPressureHigh: 0.5
+    #TriggerPressureLow: 1
     TriggerPressureHigh: 0.25
     TriggerPressureLow: 0.5
     TriggerExamine: 1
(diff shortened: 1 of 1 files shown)


cmorley191:
This requires direction review.

Cepelinas1:
so true bestie, so we're doing a direction review 24h

ShepardToTheStars (changes requested):
Add `# DeltaV` comments

cmorley191:
@ShepardToTheStars ready for re-review

ShepardToTheStars (approved):

PR #4747, opened 2025-11-25
make slips scary again

kotobdev:
## About the PR
upstream merge giga-nerfed slips so that they don't actually stun you for longer then 0.5 seconds and you can just immediately pick your items back up and it sucks. now you get stunned for however long you get knocked down. awesome

## Why / Balance
clown major

friction is kinda a bit weird right now with it, but :idk:

## Technical details
either I do this, or I go to every single prototype that's slippery and modify the values here. each solution is equally mediocre, so up to the maints

## Media

https://github.com/user-attachments/assets/a876c958-51bd-4cee-b8c1-6b6f4cf32508

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Slipping is now scary again, and will stun you for a while. Beware of the clown.

### DIFF
--- a/Content.Shared/Slippery/SlipperySystem.cs
+++ b/Content.Shared/Slippery/SlipperySystem.cs
@@ -132,7 +132,7 @@ public sealed class SlipperySystem : EntitySystem
         if (!knockedDown)
         {
             // Status effects should handle a TimeSpan of 0 properly...
-            _stun.TryUpdateStunDuration(other, component.SlipData.StunTime);
+            _stun.TryUpdateStunDuration(other, component.SlipData.KnockdownTime); // DeltaV - buff slips
 
             // Don't make a new status effect entity if the entity wouldn't do anything
             if (!MathHelper.CloseTo(component.SlipData.SlipFriction, 1f))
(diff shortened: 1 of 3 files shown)


Cepelinas1:
direction 24h

Content.Shared/Slippery/SlipperySystem.cs
```diff
@@ -132,7 +132,7 @@ public void TrySlip(EntityUid uid, SlipperyComponent component, EntityUid other,
         if (!knockedDown)
         {
             // Status effects should handle a TimeSpan of 0 properly...
-            _stun.TryUpdateStunDuration(other, component.SlipData.StunTime);
+            _stun.TryUpdateStunDuration(other, component.SlipData.KnockdownTime); // DeltaV - buff slips
```
ShepardToTheStars: > either I do this, or I go to every single prototype that's slippery and modify the values here. each solution is equally shitty, so up to the maints

Or you could just modify the default value in the [SlipperyComponent](https://github.com/DeltaV-Station/Delta-v/blob/e9c914ba5921511716b14cd647ae89bdffdaa7e7/Content.Shared/Slippery/SlipperyComponent.cs#L59C25-L59C33) to be the same default as KnockdownTime. This is preferable since we can still configure the stun and time knocked down to be different values if need be.

The vast majority of prototypes with the `SlipperyComponent` use default values anyways, so you would not need to update them.

ShepardToTheStars (changes requested):

ShepardToTheStars (approved):
Thanks for the changes. Direction has approved this as well.

PR #4748, opened 2025-11-25
xenoarch: hide some node effects until unlocked

cmorley191:
## About the PR

Artifact node effects are now sometimes hidden (in the analysis console), until you unlock them.

- there is a 10% chance for a node's effect to be hidden completely (it says "Unable to detect (unlock to discover)")
- there is a 60% chance for only a vague description of the effect to be shown (e.g. "Strong energy release (unable to detect details)")
- otherwise (30%), the detailed description is shown (the descriptions we already were showing)
- after a node is unlocked, its detailed description is shown forever after

## Why / Balance

This is related to a common criticism of the new xenoarch system, when discussing the system in https://github.com/DeltaV-Station/Delta-v/pull/4640

A general feedback point of the new xenoarch system is that it is missing the mystery and danger of the old system. The old system had vague descriptions of what effect will actually occur, while the new system always shows a unique and precisely-worded effect description.

With this PR, some nodes will be reminiscent of the old system, where many possible effects fall under the vague descriptions. The description narrows it down a little, but not much.

## Technical details

Description of the changed files:
- `XenoArtifactComponent.cs` defines the probability rates of the hiding to occur
- With those probabilities, the `SharedXenoArtifactSystem.Node.cs` decides whether each node will hide, vaguely-describe, or specifically-describe the effect until unlocked (this is tracked in `XenoArtifactNodeComponent.cs`)
- `AnalysisConsoleMenu.xaml.cs` does the rendering of this decision, via the localization in `artifact-analyzer.ftl`
- To facilitate each effect having multiple descriptions (a specific one, and a vague one shared with other effects), the descriptions were moved out of `effects.yml` and into `artifact-hints.ftl`. 
  - Each effect specifies its descriptions via the new `XAEDetailsComponent.cs`

### Testing

This is hard to test extensively. Effects are lists of components to add to the Node entity, so there's no single "EffectComponent" I can edit to procedurally check every single effect is working in-game.

Here's all I did for testing:
- did the xenoarch gameplay a while, just making sure I didn't encounter any issues from a while of normal gameplay
- spawned a ton of artifacts, watch console log for errors (and also logged to see if the `-unknown` default tip ever got used, which it didn't, which is good)
- thoroughly double- and triple-checked I didn't have any typos in effects.yml (regex and other methods)

## Media

Example of a vaguely-described effect (33% of nodes). After the node is triggered for the first time (unlocking it), the text in parentheses is replaced with the more specific description.

image

---

Most nodes (60%) still just immediately show the specific description, when they're next to be unlocked.

image

---

Some effects are hidden entirely (10% of nodes), until you unlock them.

image

## Merging

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Artifact node effects are sometimes hidden or unpredictable until after you trigger them for the first time. Be careful!


cmorley191:
This PR is code-complete. Opening it as a draft because:
- I want feedback on the idea
- I need to do more extensive testing

AirlineFood001:
these changes you're doing are genuinely peak and I PRAY they get approved, new xenoarch is missing the real risk and danger that the old system had and I LOVE seeing things like what you're doing bringing it back.

Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
```diff
@@ -166,6 +166,18 @@ public sealed partial class XenoArtifactComponent : Component
...
+    /// Probability for each node that its effect will be hidden in the analysis console until it is unlocked.
+    /// </summary>
+    [DataField]
+    public float LockedEffectHiddenProbability = 0.1f;
+
+    /// <summary>
+    /// Probability for each node that its effect will give a generic hint, rather than a specific one (until it is unlocked).
+    /// </summary>
+    [DataField]
+    public float LockedEffectVagueProbability = 0.33f;
```
ShepardToTheStars: Add `DeltaV` to the beginning of the Summary blocks, in case of upstream conflicts, I'll know that these are DeltaV-specific variables.

Alternatively, since this Component is a partial class, you can create a file in the _DV directory and keep these totally separate. Either one works for me.

Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactNodeComponent.cs
```diff
@@ -28,6 +28,30 @@ public sealed partial class XenoArtifactNodeComponent : Component
...
+    /// Whether the effect of the node is hidden in the analysis console until unlocked.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool LockedEffectTipHidden = false;
+
+    /// <summary>
+    /// Whether the effect of the node is described only vaguely in the analysis console until unlocked.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool LockedEffectTipVague = false;
```
ShepardToTheStars, line 53: Add `DeltaV` to the beginning of the Summary blocks, in case of upstream conflicts, I'll know that these are DeltaV-specific variables.

Alternatively, since this Component is a partial class, you can create a file in the _DV directory and keep these totally separate. Either one works for me.

Content.Shared/Xenoarchaeology/Artifact/SharedXenoArtifactSystem.Node.cs
```diff
@@ -111,6 +113,23 @@ public Entity<XenoArtifactNodeComponent> CreateNode(Entity<XenoArtifactComponent
...
+        nodeComponent.LockedEffectTipHidden = RobustRandom.Prob(ent.Comp.LockedEffectHiddenProbability);
+        nodeComponent.LockedEffectTipVague = RobustRandom.Prob(ent.Comp.LockedEffectVagueProbability);
+        if (TryComp<XAEDetailsComponent>(nodeEnt.Value.Owner, out var details))
+        {
+            nodeComponent.EffectTipSpecific = details.SpecificTip;
+            nodeComponent.EffectTipVague = details.VagueTip;
+            if (!details.AllowLockedEffectHiding)
+            {
+                nodeComponent.LockedEffectTipHidden = false;
+            }
```
ShepardToTheStars: nit: take out the parentheses since its a single-line if statement

ShepardToTheStars (changes requested):
I somewhat dislike how the localization string is created in `AnalysisConsoleMenu.xaml.cs` but I have no advice on how to improve it at the moment. It's probably fine for now, but we may want to revisit it if we do end up keeping new XenoArch in the long-term.

cmorley191:
> dislike how the localization string is created in AnalysisConsoleMenu.xaml.cs

this is prevalent throughout upstream code in the file, so if we want to change that we should probably do it in upstream

(my understanding is a smaller DeltaV code footprint == better DeltaV code footprint)

Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
cmorley191: The only components in the entire codebase that I see taking advantage of partial classes is `PhysicsComponent.Physics.cs` and `PhysicsComponent.Collision.cs`. So I'm gonna follow the trend here and just leave it. I like the idea though

cmorley191:
@ShepardToTheStars ready for re-review

Content.Shared/Xenoarchaeology/Artifact/Components/XenoArtifactComponent.cs
ShepardToTheStars: Either way is fine. If it doesn't have a DeltaV marker, I usually end up clobbering it during upstream merges as it might just be a part of the code that has missed a patch or something.

ShepardToTheStars:
> > dislike how the localization string is created in AnalysisConsoleMenu.xaml.cs
> 
> this is prevalent throughout upstream code in the file, so if we want to change that we should probably do it in upstream
> 
> (my understanding is a smaller DeltaV code footprint == better DeltaV code footprint)

No worries. I understand how complex this stuff can be.

Is this ready to be taken out of draft?

cmorley191:
Nah I still gotta figure out how to test... every single effect to make sure I didn't mess up

Or something

cmorley191:
This is probably ready to go. I tested it as best I could (which wasn't a great amount of testing). See the PR description (Testing section) for details.

Direction might want to have an opinion about this PR.

ShepardToTheStars:
Very relevant test fails.

ShepardToTheStars (commented):
~~Clearing my requested changes status.~~

Or not.

cmorley191:
My first TryStopNukeOpsFromConstantlyFailing failure! :)

cmorley191:
Changed percentages to the current direction consensus:

Was: 10% total hiding, 30% vague hiding, 60% specific

Now: 10% total hiding, 60% vague hiding, 30% specific

ShepardToTheStars (approved):
Maintainer approval.

PR #4768, opened 2025-11-26
Fixes mime survival box spawns

### LINKED ISSUE #4766: Mimes don't start with a survival box
It's really that simple. Mimes don't get a survival box. It's actually been this way for months but I don't think anyone has actually cared enough to report it. But I figure it should PROBABLY get reported eventually. It may just be felinid mimes somehow, I've heard reports from other mimes that they get theirs, but both Clumsy and myself have confirmed that we at least aren't getting them and it's not a loadout issue that I can tell.

pootslap:
## About the PR

Fixed some mimes not spawning with survival boxes.
## Why / Balance

Fixes #4766 I hope.
Delta-V specific species that breathe oxygen weren't getting their survival boxes after the upstream merge, every other species was fine.
## Technical details

Don't know what changed behind the scenes so I copied the OxygenBreather stuff to OxygenBreatherMime and it worked.
## Media

N/A
## Requirements

## Breaking changes

God I hope not.
**Changelog**

:cl:

- fix: Fixed certain species not starting with a survival box when joining as a mime.

### DIFF
--- a/Resources/Prototypes/Loadouts/Miscellaneous/survival.yml
+++ b/Resources/Prototypes/Loadouts/Miscellaneous/survival.yml
@@ -97,7 +97,7 @@
 - type: loadout
   id: EmergencyOxygenMime
   effects:
-  - !type:GroupLoadoutEffect
+  - !type:SpeciesLoadoutEffect
     proto: OxygenBreatherMime
   storage:
     back:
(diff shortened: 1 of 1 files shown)


ShepardToTheStars (approved):
Looks good.

PR #4783, opened 2025-11-27
Change age appearance

pootslap:
## About the PR

Changed the thresholds for being middle-aged from 30 -> 40 and old from 60 -> 70.
## Why / Balance

30 is not middle-aged and characters can be as old as 120, there could be a bit more leeway with what's considered old. 
Currently there are only 12 years where you're considered young and 60 where you're old, this shortens the gulf slightly.

This will result in some characters changing category but the numbers should be close enough that they are fixable without completely rewriting any backstories.

## Technical details

Modified YoungAge and OldAge
## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: Characters must now be 40 to be considered middle-aged and 70 to be considered old.

### DIFF
--- a/Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
+++ b/Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
@@ -108,7 +108,7 @@ public sealed partial class SpeciesPrototype : IPrototype
     ///     Characters younger than this appear young.
     /// </summary>
     [DataField]
-    public int YoungAge = 40;
+    public int YoungAge = 39;
 
     /// <summary>
     ///     Characters older than this appear old. Characters in between young and old age appear middle aged.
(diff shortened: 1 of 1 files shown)


Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
```diff
@@ -108,13 +108,13 @@ public sealed partial class SpeciesPrototype : IPrototype
     ///     Characters younger than this appear young.
     /// </summary>
     [DataField]
-    public int YoungAge = 30;
+    public int YoungAge = 40;
```
Toby222: Add `// DeltaV` comments explaining the change
Also would it not be better to override these on the individual species and give them appropriate values? (It's fine to leave this as is, just asking)

Toby222 (changes requested):

Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
pootslap: I was considering something like that, maybe even giving species like IPCs a lower minimum or higher maximum age e.g. An IPC built two years ago actually being two years old. I might try something like that in a separate PR, this was just to fix a weird inconsistency some people pointed out.

Toby222 (approved):

weskified:
diretion review 24h

ShepardToTheStars:
Direction approved.

PR #4746, opened 2025-11-25
Bring crawling values more in-line with shitmed crawling

kotobdev:
## About the PR
Stand-up time reduced to 0.5, crawling speed halved to 0.2.

## Why / Balance
We've been playing with those values since January, and they've worked pretty well. Feels familiar to everyone, and IMO just works better then the current speeds.

This does increase how fast you stand up after your stunned status expires. However, at that point you can already pick items up, and overall movement speed is still drastically lowered by stamina anyways.

Semi-related notes regarding crawling in general:
It's worth noting that upstream crawling killed the slip meta. After slipping, you can immediately pick items back up. This might have been the biggest clown nerf in SS14 history.
<sub>Anecdotal thing: the new crawling sprite rotations are nowhere near as pretty as WD/shitmed's crawling system was, but it might be a pain to modify. i'll have to look into that later</sub>

## Technical details
component ops

## Media

https://github.com/user-attachments/assets/7f88f100-b0e8-4b5d-aebd-1b95dfb8ea86

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Crawling should feel more familiar now.

PR #4802, opened 2025-11-28
Added double-tapping stand-up bind to push yourself up.

ShepardToTheStars:
## About the PR

If you are trying to get up from crawling or being knocked down, if you have a free hand and press your stand-up again, it will get you up quicker at the cost of stamina.

This is not a new mechanic. You can already double-click the Downed icon on the side, or shove yourself while down to do this, but this makes it a bit more accessible to push yourself up at the cost of stamina.

Supersedes part of https://github.com/DeltaV-Station/Delta-v/pull/4746
## Why / Balance

Just an extra bind for this feature. Feels more fluid imo.

## Technical details

None really. I just check if we're down, and already in the middle of trying to stand up, and if we are, do a call to `ForceStand`.

## Media

https://github.com/user-attachments/assets/13ce5dfc-7e13-4358-a41b-b67fdc373bd6

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: While trying to stand-up, you can now press your Toggle Crawling keybind to get up faster at the cost of stamina. Of course, you still still need a free hand to do this.

### DIFF
--- a/Content.Shared/Stunnable/SharedStunSystem.Knockdown.cs
+++ b/Content.Shared/Stunnable/SharedStunSystem.Knockdown.cs
@@ -231,7 +231,14 @@ public abstract partial class SharedStunSystem
         if (playerSession.AttachedEntity is not { Valid: true } playerEnt || !Exists(playerEnt))
             return;
 
-        ToggleKnockdown(playerEnt);
+        // DeltaV - Double-tap Standup bind forces standup (unless hands full)
+        if (_standingState.IsDown(playerEnt)
+            && TryComp<KnockedDownComponent>(playerEnt, out var knockedDown)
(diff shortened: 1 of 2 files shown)


dekigangsta:
..? why not bring back what we had this entire time

kotobdev:
i am a true Free Stand Up believer

ShepardToTheStars:
I mean, I can just not do this and you can do this move the hard way - shoving your character while you're on the ground.

But what we had was the precursor to what we have now. It was basic. This promotes at least a little bit of skill expression.

weskified:
direction review 24h

Toby222 (approved):

ShepardToTheStars:
Direction approved.

PR #4804, opened 2025-11-29
CMO Hardsuit Buff

snowywinters:
## About the PR
Fix cmo suit so its not the worst suit in the game

## Why / Balance
Cmo suit only protected from caustic and infection that being only one good thing the infection (no one uses caustic damige) but even then the infection is only 60% for a suit that should be more 

## Technical details
caustic is 70% (lowered from 90%)
blunt 30% (why medbay fights with unruly patients)
slash 30% (why medbay fights with unruly patients)
piercing 15% 
heat 20% (dont know if i should increase more since caustic and heat are under the same damage type) 
rads 20% (should be fine maybe)
explosion 20% (almost all suits have this and so cmo is not intsa gibbed by any bomb)
infection resistance 80% (cmo should have more then only 60% for a bio suit)

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Made cmo hardsuit better

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -435,12 +435,12 @@
   - type: Armor
     modifiers:
       coefficients:
-        Caustic: 0.3 #DeltaV (was 0.1)
         Blunt: 0.70 #DeltaV
         Slash: 0.70 #DeltaV
         Piercing: 0.85 #DeltaV
         Heat: 0.80 #DeltaV
         Radiation: 0.80 #DeltaV
+        Caustic: 0.3 #DeltaV (was 0.1)
   - type: ExplosionResistance #DeltaV
     damageCoefficient: 0.80 #deltaV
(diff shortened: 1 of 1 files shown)


weskified:
direction review 24h

ShepardToTheStars (approved):
Looks fine from a maintainer standpoint

ShepardToTheStars:
Direction Denied.

There's a few reasons why Direction (as a whole) denied it.
* The CMO's hardsuit only has a 5% speed decrease. This is already a huge boon to the suit. Increasing resistances would make it a very capable combat suit, which direction doesn't really want.
* We did chat about giving the CMO's suit something in terms of resistances, but ultimately, it was hard to come to a consensus on what should be given, given its already innate speed bonus (for a hardsuit). We talked about giving them at least the paramedic's suits resistances, but we also don't want them being a paramedic+. The CMO shouldn't be rushing out to save people like a paramed would.
* Zombification resistance is inversely balanced against brute resistance. Its one of the main reasons why there isn't any zombie resistance modifiers on hardsuits, with the exception being the CMO's hardsuit.

PR #4816, opened 2025-11-30
psychokinetic scream LogPowerUsed

cmorley191:
## About the PR

Psychokinetic scream fixes:
- now adds glimmer like all the other powers do
- now mantis sees a popup when used nearby, like all other powers
- now works with xenoarch "psionic disturbance" artifact trigger

Skia ability was left untouched - I tested that it still worked.

## Why / Balance

Not having these elements was an oversight, I think. I briefly looked over the skia PR's; doesn't seem like this was intentional.

## Technical details

LogPowerUsed is what all the power systems call -- does an admin log, increases glimmer, and does the "power used" event for metapsionics and artifacts

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: psychokinetic scream now works with "psionic disturbance" arti trigger


MantasDab360:
Bro is singlehandedly carrying xenoarch and psionics.

ShepardToTheStars (approved):

PR #4807, opened 2025-11-29
Pebble's Reincarnation!

FieldCommand:
## About the PR

I hold many fond memories of Pebble, but over time it has fallen behind. This PR is the culmination of this effort to make it new and up-to-date.

## Why / Balance

Pebble was old and needed updating.

## Technical details

Too numerous to list, treat this PR as if it were a new map.

## Media

PebbleV2

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: FieldCommand
MAPS: 
- tweak: Pebble: Pebble's reincarnation, Rejoice. Pebble has now been fully overhauled and should be up to the latest standards.

PR #4820, opened 2025-12-01
Protect arrivals and evac from meteors (slightly)

cmorley191:
## About the PR

Adds a slight bias for meteors to miraculously avoid arrivals and evac.

- The Arrivals area will be hit by (roughly) 67% fewer meteors
- The Evac area will be hit by (roughly) 25% fewer meteors

Note: this PR will not take effect until each maptainer updates their particular map to include the feature.

## Why / Balance

People are generally liking that [meteors are back](https://github.com/DeltaV-Station/Delta-v/pull/4379), but arrivals and evac get damaged by them a LOT. This is not because they are being targeted more than other places, but rather because basically none of the maps provide grille shielding for them (the idea is ships are supposed to be "flying in" to there). For the same reason, the cargo dock is often damaged.

I think we should compensate by reducing the number of meteors attacking Arrivals and Evac, because:
- brand new players that join DeltaV are currently quite likely to spawn in a spaced arrivals, which is probably very confusing to people who only just installed the game
- it's boring for engi to fix the same areas over and over -- variety is better

## Technical details

~~Added a component to the arrivals/evac station map beacons, with average settings that should work well with most maps. Mappers can probably customize this if they want different settings.~~

New approach: added a marker that's only visible in the `mapping` tool, very similar to `AntiAnomalyZone` marker.

The point of the StandardRate versions of the marker is making it so we can change a single number in one file to change the balance of this PR (i.e. the 67% / 25% numbers) across all maps simultaneously. I tested that this works.

## Media

Demo on Byoin with exaggerated settings, to make the feature visually clearer:
- Turned up the meteor spawn rate
- Adjusted protection settings
  - Arrivals (N) 75%
  - Evac (S) 100% (no hits)
  - Cargo (SW) 100% (got hit by only 1, which is probably due to luck / the targetingAttempts loop running out)
    - note: this PR is not protecting cargo, I just manually did this for the demo

https://github.com/user-attachments/assets/de044736-5cad-4261-adcc-c03a04d66aed

## Requirements

## Breaking changes

**Changelog**

Intentionally no changelog -- this is a behind-the-scenes / meta thing.


ShepardToTheStars:
Direction review 24h

ShepardToTheStars:
God, this video just shows me how much of a bitch our meteors are lol

ShepardToTheStars:
How does this affect maps where evac and arrivals are the same place? Will meteors basically never hit there?

cmorley191:
> God, this video just shows me how much of a bitch our meteors are lol

Just to reiterate: the video is "faked". I turned up the spawn rate, and I messed with protection rates to show the feature more clearly. See the text above the video

> How does this affect maps where evac and arrivals are the same place? Will meteors basically never hit there?

Glacier is the only map I can see where there is true overlap (and note: glacier meteors burn up, don't hit the station).

But yeah the effect would stack like so:
`total protection = protectionA + ( protectionB * (1-protectionA) )`

With the current numbers, a combined evac+arrivals area (with both beacons present) will be hit by a combined 75% less meteors.

Other cases:

- Hive: it's one long hallway - the arrivals beacon at one end, evac beacon at the other end. so asteroids will hit one side more than the other lol (this is probably a good candidate for customized map settings of this feature)
- Pebble: does not have a station beacon for neither arrivals nor evac, so the area is not protected (mapper could fix this by adding the component onto some other entity in arrivals, like the holopad)

cmorley191:
I should move `ProtectionRadius` out to its own component.

That way mappers can override it without messing with or having to re-state the `ProtectionRate` and `DisablePermanentlyIfUnanchored` parts

IamVelcroboy:
Just a thought but instead of beacons, which can be moved by players...what if it was on the airlocks instead? We have specific evac and arrivals airlocks

Edit: I suppose those can technically be moved as well, but a little more work than unanchoring a beacon

FieldCommand:
> > God, this video just shows me how much of a bitch our meteors are lol
> 
> Just to reiterate: the video is "faked". I turned up the spawn rate, and I messed with protection rates to show the feature more clearly. See the text above the video
> 
> > How does this affect maps where evac and arrivals are the same place? Will meteors basically never hit there?
> 
> Glacier is the only map I can see where there is true overlap (and note: glacier meteors burn up, don't hit the station).
> 
> But yeah the effect would stack like so: `total protection = protectionA + ( protectionB * (1-protectionA) )`
> 
> With the current numbers, a combined evac+arrivals area (with both beacons present) will be hit by a combined 75% less meteors.
> 
> Other cases:
> 
> * Hive: it's one long hallway - the arrivals beacon at one end, evac beacon at the other end. so asteroids will hit one side more than the other lol (this is probably a good candidate for customized map settings of this feature)
> * Pebble: does not have a station beacon for neither arrivals nor evac, so the area is not protected (mapper could fix this by adding the component onto some other entity in arrivals, like the holopad)

Pebble have a PR #4807 ready for review, that addresses that. Well its a total rework really xD

cmorley191:
> Just a thought but instead of beacons, which can be moved by players...what if it was on the airlocks instead? We have specific evac and arrivals airlocks

Brilliant idea

ShepardToTheStars:
Should probably be its own entity tbh. Like parent from MarkerBase

IamVelcroboy:
> Should probably be its own entity tbh. Like parent from MarkerBase

We *do* have the anti-anomaly markers which are similar.

ShepardToTheStars:
Yea, thats exactly what I was thinking of.

cmorley191:
> Should probably be its own entity tbh. Like parent from MarkerBase

Unfortunately that would mean this feature doesn't take effect until mappers manually put them in :(

Which might be okay...?

Could also mean mappers forget to put them in...

I am not experienced with the whole mapping ecosystem/procedures so I defer to others' judgement

ShepardToTheStars:
Airlocks don't always signify a place where people spawn imo. Take Shoko for example. The cryo pods are close to arrivals, but asteroids wouldn't avoid hitting that area, even though its a common area where people spawn.

FieldCommand:
> > Should probably be its own entity tbh. Like parent from MarkerBase
> 
> We _do_ have the anti-anomaly markers which are similar.

That was my recommendation, this is a entity that cant be eddied or removed by anyone/anything during a live round. and I think that's a far better idea then having it be a beacon or on the airlocks.

> > Should probably be its own entity tbh. Like parent from MarkerBase
> 
> Unfortunately that would mean this feature doesn't take effect until mappers manually put them in :(
> 
> Which might be okay...?
> 
> Could also mean mappers forget to put them in...
> 
> I am not experienced with the whole mapping ecosystem/procedures so I defer to others' judgement

If it this PR is merged, then I don't have a problem with this. it will take a bit of time, but I think its the better approach.

IamVelcroboy:
> Unfortunately that would mean this feature doesn't take effect until mappers manually put them in :(
> I am not experienced with the whole mapping ecosystem/procedures so I defer to others' judgement

We will make an announcement for maptainers to include in their next update

ShepardToTheStars:
Direction approved. Though I think we’ll buff meteors a bit but thats not part of this.

@FieldCommand you can also add it to ghost role ships too :)

cmorley191:
The feature is now an invisible marker like we discussed. Feels way better!

This just needs maintainer review.

cmorley191:
@FieldCommand Does this workflow sound good to you from a mapping perspective? I thoroughly tested that it all behaves as follows:
- in Tools `mapping` mode, add the arrivals or evac version of the entity
image

- Once placed, edit the radius of it in Right Click -> View Variables
  - The View Variables window will *not* let you edit the strength / protection rate
- The `savemap` command will save *just* your radius edit. The rate will just be inherited from the prototype so that it's the same on all maps
- If we ever want to change the protection rates, we change them in `meteors.yml`, which will automatically take effect on all the entities across all the maps.

And finally, there's a third version of the entity that lets you customize the strength *and* radius in View Variables -- in case mappers want to protect other places that aren't evac or arrivals.

ShepardToTheStars (approved):

ShepardToTheStars:
We can always tweak it if need be, but I think its good to go at this point. :)

PR #4821, opened 2025-12-01
Small Borg ID Chips Rework

ShepardToTheStars:
## About the PR

* Borgs made mid-round no longer have the bare amount of access.
  * ID chips printed from the Exofab have the same access as a round-start borg
  * Syndicate Exofabs can no longer print normal borg ID chips
    * The extra cost to make Syndicate Borg ID chips has been lowered slightly (2 gold, 2 plasma -> 0.5 gold, 0.5 plasma)
  * Borg ID chip access can no longer be copied by agent IDs (Sorry Roboticists)
* The new derelict borgs now start with the same access as the regular derelict borg (Bug fix)
  * However, the derelict syndicate borg does not, but it has at least Externals and Maints access. Talk to the command for more access!

## Why / Balance

Its a pretty big pain point if you take a borg ghost role. You spend the next 20-30 minutes trying to get access from captain/ACO/heads and its just a pain.

See Niomi's [rant](https://discord.com/channels/968983104247185448/1010560161888350340/1440868920482598986) and [design doc](https://discord.com/channels/968983104247185448/1441420403804344461).

## Technical details

* Added ID chips to derelict borgs.
* Added Borg ID chip component and added it to the parent of all borg chips.
  * Check for that component when agent IDs try to copy and return a popup that you can't copy.
* Fixed some funky inheritance in the upstream exosuit fab's file.

### Copying Access

https://github.com/user-attachments/assets/5ad79634-860a-4d4c-a2e7-c01c3d8ea064

### Exofab versus Syndicate Exofab ID Cards
image

### Regular Derelict Borg Access (like the original derelict has)
image

### Syndicate Borg Access
image
(Not pictured - Syndicate Agent and Nuclear Operative - though not like it matters)

## Requirements

## Breaking changes

**Changelog**

:cl: ShepardToTheStars, Niomi
- tweak: Borg ID chips from the exofab now have the same access as round-start borgs.
- tweak: SRNs can no longer print normal borg ID chips.
- tweak: Syndicate borg ID chips now cost a bit less to make.
- tweak: Agent IDs can no longer copy access from borg ID chips.
- fix: The new derelict borgs now have the intended borg access to the station! 
- tweak: Derelict syndicate borgs don't have usual borg access, but has enough to get into the station.
- tweak: Derelict syndicate borgs can be unlocked by people with Robotics access (so they can get their access updated).


ShepardToTheStars:
Direction review 24h

ShepardToTheStars:
Direction approved.

Content.Shared/_DV/Access/Components/BorgIdCardComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._DV.Access.Components;
+
+/// <summary>
+/// Component to denote the ID card as one specific to a borg.
+/// Used to prevent agent IDs from copying access from borg IDs chips.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class BorgIdCardComponent : Component { }
```
Toby222:
```suggestion
public sealed partial class BorgIdCardComponent : Component;
```

Toby222 (changes requested):
For consideration:
Instead of a marker component, a new field `CanBeCopied` on IdCardComponent would be more re-usable
Could be added to e.g. CC IDs?

ShepardToTheStars:
> For consideration: Instead of a marker component, a new field `CanBeCopied` on IdCardComponent would be more re-usable Could be added to e.g. CC IDs?

I did consider that first, but then I found the agent ID system and saw how it determines if an ID can be copied.

```c#
if (!TryComp<AccessComponent>(uid, out var access) || !HasComp<IdCardComponent>(uid))
     return;
```

I’d also have to add it to `AccessComponent` as well in the case that something has both `IdComponent` and `AccessComponent`, otherwise CanBeCopied wouldn’t matter. Also didnt want to touch upstream access code too much if I could help it. 

Instead, what do you think about making `BorgIdComponent` more generic (i.e. `AccessUncopyableComponent`, `AgentIdUncopyableComponent`, or whatever), and added the LocId for the popup to that component.

Toby222:
> Instead, what do you think about making `BorgIdComponent` more generic (i.e. `AccessUncopyableComponent`, `AgentIdUncopyableComponent`, or whatever), and added the LocId for the popup to that component.
> 

Also sounds good :^)
Could make the loc be smth like "This ID seems hardened against copying" instead of just saying "You can't copy borg IDs because it'd break game balance :^)"

ShepardToTheStars:
Yea, I’ll do that.

ShepardToTheStars:
Changed made @Toby222

Toby222 (approved):

PR #4822, opened 2025-12-01
add MeteorLargeDeltaV to entity spawn menu

cmorley191:
## About the PR

Not sure why these were hidden.

These are the meteors that are *actually* used in the meteor event. The other meteors in the list are from upstream's version of the meteor event, which we have disabled.

## Why / Balance

Admemes & testing

## Technical details

Note: to tether gun these, you have to use the right-click context menu to target it. Simply left clicking on the meteor doesn't work.

I don't know why this is the case.

## Media

Bottom: upstream's meteors

Top: our meteors

image

## Requirements

## Breaking changes

**Changelog**

N/A

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Projectiles/meteors.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Projectiles/meteors.yml
@@ -2,6 +2,7 @@
   id: MeteorLargeDeltaV
   name: meteor
   description: Big.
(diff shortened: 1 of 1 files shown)


ShepardToTheStars (changes requested):
Can you add `suffix: DeltaV` to these two meteor types? Just so its clear which ones we use.

cmorley191:
@ShepardToTheStars done

image

image

ShepardToTheStars (approved):

Issue #4810, opened 2025-11-29
Stasicilin not turning the metabolizer unconcious


apeiroeder:
## Description
Despite more than 11u stasicilin being present in the bloodstream, the metabolizer didnt turn unconcious. However, the healing aspect of it still worked without issue 

**Reproduction**
 Stasicillin was given through 20u pills, not sure if relevant 

**Screenshots**
No screenshots

**Additional context**
Stasicillin, an exotic chemical, is supposed to turn the metabolizer unconcious and start healing them once there is at least 11u of it in the bloodstream

HTMLSystem:
Same as what I asked on the other issue, can you provide some details on how to reproduce the bug and what it actually is?

verybigman311:
<img width="1367" height="196" alt="Image" src="https://github.com/user-attachments/assets/5e59df0a-7496-46f6-ab0e-780b2e481bfc" />

I actually tried this a couple days ago. No matter how many units I injected, it never produced unconsciousness.

apeiroeder:
okay, edited this issue as well. Im terribly sorry for the text being hidden for some reason Dx

ShepardToTheStars:
Likely do to the ForcedSleeping component being changed in the last upstream patch. I'll fix it soon.

ShepardToTheStars:
Should be fixed now.

PR #4823, opened 2025-12-01
Stasicillin Forced Sleeping Fix

ShepardToTheStars:
## About the PR

Kinda just copied nocturine. Need to test still.

## Why / Balance

Fix. Addresses https://github.com/DeltaV-Station/Delta-v/issues/4810

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Stasicillin should once again force the target to fall asleep if they have over 11u.

### DIFF
--- a/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
+++ b/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
@@ -32,13 +32,13 @@
             Toxin: -4
             Brute: -4.5
             Burn: -6
-      - !type:GenericStatusEffect
+      - !type:ModifyStatusEffect
         conditions:
         - !type:ReagentThreshold
+          reagent: Stasicillin
           min: 11
-        key: ForcedSleep
-        component: ForcedSleeping
-        refresh: false
+        effectProto: StatusEffectForcedSleeping
+        time: 3
         type: Add
 
 - type: reagent
--- a/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
+++ b/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
@@ -38,7 +38,7 @@
           reagent: Stasicillin
           min: 11
         effectProto: StatusEffectForcedSleeping
-        time: 3
+        time: 1.5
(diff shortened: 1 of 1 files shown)


cmorley191:
Note: Stasicillin's `metabolismRate` is `0.2`, while Nocturine uses the default of `0.5`. That means the same volume will stay in the bloodstream 2.5x longer, and thus will generate 2.5x more sleep time compared to Nocturine. (the Add 3 Seconds gets triggered 2.5x more times in total)

No opinions here just making sure you know about that part

ShepardToTheStars:
Yea, I hadn't gotten around to testing this yet, but I assumed the value probably needed changing.

Toby222 (dismissed):
Can confirm it sends you to honk-sho-mimimi land
Merge yourself if you think it's good, or ask direction if need changes, unclear if you wanted the metabolismRate to be changed

ShepardToTheStars:
@Toby222 ill give this a quick test today and see if it 1) keeps the user asleep while its over 11u and 2) wears off once under 11u

Toby222:
> @Toby222 ill give this a quick test today and see if it 1) keeps the user asleep while its over 11u and 2) wears off once under 11u

I forgor how to see chemicals in blood, but it does keep the user asleep for some time, and allows them to wake up eventually

cmorley191:
> unclear if you wanted the metabolismRate to be changed

If you want to change the sleep duration, then reduce the effect `time: ` added per metabolic update. (changing from 3 to 1.2 would make it match the duration of nocturine:  Nocturine adds 3 every 0.5u; this would add 1.2 every 0.2u)

Just changing the metabolismRate would mess with the total healing the drug does, which I imagine we don't want to change

ShepardToTheStars:
@cmorley191 Tested with 1.2 but I could wake up for very brief moments if you spammed wake up. 1.5 seemed like the sweet spot. You wake up when you have around around 10.6u of it in your system.

Toby222 (approved):

ShepardToTheStars:
Test fails fake. Re-running.

PR #4862, opened 2025-12-04
Fishing Vendor and Fish Basket

Stxcking:
## About the PR

Add a fishing vendor similar to the salvage vendor that uses fishing points which can be earn by inserting fishes into the vendor. Also, adds a 6x6 container that holds fish and can be equipped in the belt slot.
## Why / Balance

People have been enjoying fishing and there was no way to get the golden fishing rod as well as the hawaiian shirts. This also makes it so fish can be exchange for spesos which as of this PR, they still can't be sold at ATS. 
## Technical details

It's the mining point and salvage vendor codes (credit to: Deltanedas) essentially minus the lathe component so points are inserted into players' IDs instantly. 
`FishingPointsComponent.cs` Stores fishing points in ID.
`FishingVendorComponent.cs` Fishing points awarded on type of fish (10 for base fish, 25 for rare fish).
`FishingPointsSystem.cs` Add and remove fishing points from ID.
`FishingVendorSystem.cs` Fish to fishing points exchange.
## Media

image

## Requirements

## Breaking changes

**Changelog**

🆑 
- add: Fishing vendor: Exchange fish for fishing points!
- add: Fishing basket to easily store caught fish.


Apachit0:
fishing grows 👍

ShepardToTheStars:
Direction review 24h

ShepardToTheStars:
Direction Approved.

ShepardToTheStars (approved):
Im a bit worried about using unsigned ints for the points but the mining points system uses them, and it *should* never be negative so I guess its fine.

PR #4835, opened 2025-12-02
Give chitnids moth eye sprites

HTMLSystem:
## About the PR

With just literally one new thing added to their species file, chitnids will now use moth sprites for their eyes. Yippee!
## Why / Balance

Sprites matching is good and awesome.
## Technical details

Added
- type: Inventory
    speciesId: moth
to species/chitnid.yml
## Media

I have a migraine and just closed the test server so you're not getting it, sorry.
## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Chitnids will now use moth sprites for their glasses

PR #4884, opened 2025-12-05
Give dummy avali and chitnid moth eye sprites

HTMLSystem:
## About the PR

In my last pr (#4835) I forgot to give the chitnid dummy `speciesId: moth`. This does that. We also gave it to avali.
## Why / Balance

This makes it where whenever the dummy is used for these species (in the character creator menu and probably other places) they will also have moth sprites for eyewear. Consistently. 
## Requirements

## Breaking changes

**Changelog**


cmorley191 (approved):
Love it

PR #4895, opened 2025-12-06
Fix respawn timer

beck-thompson:
## About the PR

Mald pr... This bug really drove me nuts!

## Technical details

The `OnSystemLoaded` and unloaded functions were never registered (`IOnSystemChanged` was already set for the ghost system) and it was an entity system for some reason as well so I just nuked it and replaced it with a normal subscription. I think the system in general needs to be redone to be a little less sussy but this is an improvement.

**Changelog**

:cl:
- fix: Respawn timer now works when observing


SirWarock (commented):
I playtested it, set the respawntimer to 40 and went into the round observing.
The timer correctly displayed the time that was left for my respawn.
It also worked when I clicked on respawn, correctly booting me to the menu.

This was done in Dev environment though, I have no clue how to get release mode on.
But playwise, the issue seems fixed.

SirWarock:
tested it in release mode now too, it works there too.

cmorley191:
What exactly does this fix? I am not familiar with this bug

beck-thompson:
> What exactly does this fix? I am not familiar with this bug

Try joining a game as observer, it says you can respawn but its not true you still have a full timer

Toby222 (approved):
Bug as old as the respawn mechanic being added to dV in the first place

PR #4905, opened 2025-12-07
AI HUD and Comms Update

snowywinters:
## About the PR
This will give the ai access to traffic coms now witch was never added after it was added and diagnostic hud\mind shield hud

## Why / Balance
To give the ai access to trafic coms without them needing to use the intercom all the time and since cap\lo have it why should the station ai not have access to it

diagnostic hud is for the ai to see there borgs health and make sure there all ok under there watch 

mind shield hud is to let the ai see who is and is not mindshield so they can better go about there work 

## Technical details
added trafic coms to the ai coms list
added the diagnostic component and mind shield component to the list of the player ai component

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Station AI can now use traffic comms. Your secrets are no longer safe.
- add: Station AI can now see silicon health, just like diagnostic HUDs.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
@@ -19,6 +19,7 @@
     # Begin DeltaV additions
     - Justice
     - Prison
+    - Traffic
     # End DeltaV additions
   - type: ActiveRadio
     channels:
@@ -34,6 +35,7 @@
     # Begin DeltaV additions
     - Justice
     - Prison
+    - Traffic
     # End DeltaV additions
   - type: IgnoreUIRange
   - type: StationAiHeld
@@ -322,6 +324,14 @@
       state: ai_female
   - type: NameIdentifier
     group: StationAi
+  # Begin DeltaV additions
+  - type: ShowMindShieldIcons
+  - type: ShowHealthBars
+    damageContainers:
+    - Inorganic
+    - Silicon
+    - HumanoidSilicon
+  # End DeltaV additions
 
 # Hologram projection that the AI's eye tracks.
 - type: entity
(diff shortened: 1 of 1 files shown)


snowywinters:
I could add sec and med huds but i dont think direction would like that sadly if they do i can add them

Proxysseia:
Direction Review 24h

ShepardToTheStars (dismissed):
Looks fine, maintainer-wise.

Resources/Prototypes/Entities/Mobs/Player/silicon.yml
```diff
@@ -322,6 +324,14 @@
       state: ai_female
   - type: NameIdentifier
     group: StationAi
+  # Begin DeltaV additions
+  - type: ShowMindShieldIcons
```
Proxysseia: Mindshield vision is exclusive to sec and will encourage even more vunting bevaviour from ai. Same reason it doesnt see wanted status. Remove.
Rest is good.

Proxysseia (changes requested):

Resources/Prototypes/Entities/Mobs/Player/silicon.yml
snowywinters: I mainly wanted to add that for the ai so people cant trick it by puting on an eva suit/command eva and using there id to trick and subvert the ai and any one can buy mind shields but i will remove it 

what about med hud?
reason is for the ai to see crew health to help them better or prioritize the weaker to save or to see intruders health so they can see how long they are to death to ether fully kill them or help them and alert med there health amount but this can ether be abused with zombies and electric door but it can also alert the ai they are going to far with the shocking (i just want to ask about this part if its a yes i then can add it to this pr instead of the mind shield)

Proxysseia: ai does not need to see any organic healthbars. if they want to check their crews health they can look at the crew monitor. we also do not want to pretend that the frankly insane micromanagement of conveying which nukie has what hp value is remotely intended. Ai gets to see borg health because it has an ability to heal borgs.

Proxysseia (approved):

ShepardToTheStars (approved):

Issue #4495, opened 2025-10-05
Security Hardsuit is causing Constant cold damage on terra station


Vapetastic-Gaming:
Unsure how to reproduce, Sec hardsuit equipped with internals is causing constant cold damage

Using lepo causes a small temp fix, but it returns

Using debug "Rejuvinate" temp fixes and it doesnt return

Unable to reproduce in admin testing arena during round with same equipment and species

HTMLSystem:
What species was the player playing? Just incase that would help.

Vapetastic-Gaming:
Both vulp, human, and avali are affected

kotobdev:
same issue popped up randomly when I was testing asakim on glacier a few days ago. 0.5 cold damage a tick, no matter where I stood

Toby222:
Probably fixed with 4931, reopen or re-report if issue happens again

PR #4931, opened 2025-12-10
EntropicDegen Fix

### LINKED ISSUE #4929: Malign rifts HORRENDOUSLY bugged (Constant random cold damage)
## Description
Malign rifts, on the **first round post-restart**, occasionally apply their status effect to nearby mobs in a way that makes it deal 0.05 cold and 0.15 cold damage, once per tick, forever. As far as I know, this isn't intentional.

If this happens, malign rifts will fail to apply *any* status effect in subsequent rounds, and will throw an error to the server console every time they pulse.

Purging the rift *appears to* stop the constant damage from being inflicted. However, it won't stop the errors after restart.

**Reproduction**
this one's a doozy. somewhat inconsistent so I'm including the exact actions I took

- load into dev. make a note of your character's NetID.
- place a Malign Rift in the bottom-right corner of the empty dev area, past the airlock
- stand next to rift for a while while keeping an eye on your health, wait for the pulse/status effect to apply
- walk to the top-left corner near the fireaxe. set a timer for 40 seconds. 
- can't reproduce 100% of the time, but watch as the damage (0.05 cold 0.15 asphyx) continues to be applied past the maximum 40 seconds, sometimes stopping for a bit before resuming. 
- if it doesn't happen the first time, walk up to the rift, wait for the pulse to happen, and walk away again

-----

- after following the previous steps, golobby
- startround
- place a malign rift anywhere
- wait for it to pulse, check server console:
```
[ERRO] runtime: Caught exception in "entsys"
System.Collections.Generic.KeyNotFoundException: Entity 6781 does not have a component of type Robust.Shared.GameObjects.TransformComponent
   at Content.Server._DV.CosmicCult.EntitySystems.CosmicRiftSystem.Update(Single frameTime) in <root>\Delta-v\Content.Server\_DV\CosmicCult\EntitySystems\CosmicRiftSystem.cs:line 55
   at Robust.Shared.GameObjects.EntitySystemManager.TickUpdate(Single frameTime, Boolean noPredictions) in <root>\Delta-v\RobustToolbox\Robust.Shared\GameObjects\EntitySystemManager.cs:line 318
```
(6781 was my character's NetID in the previous round, and that entity no longer exists. This is thrown every time the rift pulses.)
- stand near the rift while it pulses, notice how no status effect is applied to you and no damage is taken.

**Screenshots**
Image

**Additional context**
This has been reported in various forms since August, although nobody had any idea what was going on. 100% related to PR https://github.com/DeltaV-Station/Delta-v/pull/4216
This is probably the same issue as https://github.com/DeltaV-Station/Delta-v/issues/4495, although, since i'm reporting effectively two issues, I think it's probably best to make a new issue altogether.

ShepardToTheStars:
## About the PR

This might fix #4929. In theory, it does. But the code needs to be re-worked at some point.

## Why / Balance

Hotfix.

## Technical details

GetEntitiesInRange only ever adds to the hashset. It won't clear the entries. These entries will persist between rounds and cause server errors. It will also reapply the debuff even if they aren't near for EACH rift that is currently pulsing, giving them an infinite debuff as long as there's at least one rift alive anywhere. lmao.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Entropic Degen should no longer last forever.

### DIFF
--- a/Content.Server/_DV/CosmicCult/EntitySystems/CosmicRiftSystem.cs
+++ b/Content.Server/_DV/CosmicCult/EntitySystems/CosmicRiftSystem.cs
@@ -62,6 +62,7 @@ public sealed class CosmicRiftSystem : EntitySystem
 
             var pos = Transform(uid).Coordinates;
             Spawn(comp.PulseVFX, pos);
+            _mobs.Clear();
             _lookup.GetEntitiesInRange<MobStateComponent>(pos, comp.PulseRange, _mobs);
             _mobs.RemoveWhere(target => _chaplainsQuery.HasComp(target) || _cultistsQuery.HasComp(target) || _colossiQuery.HasComp(target));
             foreach(var mob in _mobs)


ShepardToTheStars:
Might also address https://github.com/DeltaV-Station/Delta-v/issues/4495

Toby222 (approved):
Me when I accidentally store data in unserialized system fields

Toby222:
I also saw the error message mentioned in 4929 in some random testfails so this might fix one of those cases, too

PR #4933, opened 2025-12-10
One line YML warrior

SirWarock:
## About the PR

Made Rollerbeds immune to steptriggers, such as mousetraps.
## Why / Balance

Borers require medical to use mousetraps. Mousetraps kill rollerbeds. Not good.
## Technical details

One LINE YML WARRIOR
## Media

grafik

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Rollerbeds getting annihilated by mouse traps!

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Furniture/rollerbeds.yml
+++ b/Resources/Prototypes/Entities/Structures/Furniture/rollerbeds.yml
@@ -80,6 +80,7 @@
             True: {visible: false}
     - type: StaticPrice
       price: 120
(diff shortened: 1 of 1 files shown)


Toby222:
Shouldn't this only be on the hover version?

SirWarock:
> Shouldn't this only be on the hover version?

Both rollerbeds and the hoverbeds were being demolished by a mousetrap. While I partly agree that only the hover version would ignore steptriggers, I can't let the normal ones get destroyed by a measly mousetrap

ShepardToTheStars (approved):
This doesn't need a direction review says direction me.

PR #4941, opened 2025-12-11
meteors: balance changes pt. 1 (bias/targeting)

cmorley191:
## About the PR

- **Reworked meteors to target specific points as a swarm, rather than spreading impacts evenly across the station**
  - 20% of swarm events will still be totally random like before
  - 20% of swarms will come from a single direction, like "rain" spreading out across that side of the station
    - these uni-directional swarms have a slightly better chance of taking down grille shielding and actually reaching the station walls
  - 60% of swarms will specifically target a small patch of the station exterior
    - these targeted swarms have a **much** better chance of getting through grille shielding and causing spacings
    - randomness is still at play -- as a victim, it (hopefully) does not feel like the system is artificially/clunkily targeting you
- Reduced the frequency of meteor swarms from every 14 minutes to every 20 minutes.
- Increased the number of meteors per swarm slightly
  - (not enough to overcome the reduced frequency, imo. In other words, I think overall this PR is a very slight meteors nerf)

## Why / Balance

Per feedback from #community-feedback and other channels:
- meteors are a bit too strong / annoying
- *lots* of areas are hit with *superficial* damage. As an engineer, running from one tiny 1-wall hole to the next while a bunch of different people yell at you over radio to hurry up is not very fun. Having the chance to "sink your teeth" into only a *couple* concentrated, *big* repair projects sounds like more fun
- docks like arrivals, evac, cargo are always victims of every meteor swarm, due to not having shielding. This is kind of lame (repairing the same areas over and over is boring). https://github.com/DeltaV-Station/Delta-v/pull/4820 is poised to address this (slowly as mappers implement it), but this PR also helps out by actually creating chances for other areas of the station to have their grille shielding breached and the deeper rooms actually spaced
  - prior to this PR, it really feels like the first few meteor swarms have absolutely no chance of spacing any grille-shielded part of the station (i.e. anywhere but arrivals or evac or cargo). that's kinda lame

This PR makes each meteor swarm slightly stronger, but hopefully less of an annoyance as damage will be more concentrated -- engineers are spending less time walking across the station finding holes, and spending more time actually fixing the holes.

Victim departments will probably have a good and more interesting experience, since bigger damage means teams of engineers can focus on getting a temporary fix in place faster:
- use holofans or inflatable barricades to get a temporary seal in place
  - allows victims to resume work in the area while repairs continue
- plenty of time to design/work together on permanent repairs to the damage
  - concentrated damage keeps engineers working together, hopefully improving the community's skills and intern teaching
- finally wrap up by fixing the absolutely wrecked grille shields nearby -- an underutilized element of the potential engineering experience since previously grille damage is so light / spread out that it hardly matters

In summary, I see this PR as a *light* step in the right direction towards engaging meteor events. I'd like to see *even stronger* swarms in the future, with mitigating advantages like advance warnings ("meteors incoming in *this* direction, ETA 5 minutes, please set up grilles!"), or point defense, so that it feels more interactive as an engineer. I'd love to see more swarm variety as well.

## Technical details

Refactored 90% of `MeteorSwarmRule.cs`. Sorry. The diff sucks; I'd just download the new version and look through the file as a whole. It should be way more readable now, and we should have smaller diffs in future changes since we don't have everything in one function anymore.

Made two new variations of the `id: MeteorSwarm` gamerule, and set up the weights as described at the top. They use new variables that control the focused targeting / make sure it still feels roughly "random" and not too overwhelming.

Honestly I hope that this code speaks for itself. If you are confused, please point it out so I can improve the self-documentation of the code. The goal is for anyone to be able to work on this code, just from the variable names and comments.

## Media

Three videos.

### 1. Targeted Swarm (60% of swarms)

This one randomly decided to target the south side of Byoin station. It also randomly decided to have the meteors approach from the south-east (not an important detail in this case, but can be important on some other stations with more unique geometry! Coordinating the approach direction gives them the potential to "thread the needle" in certain other cases) 

The swarm spaced the east part of medical, and a bit of the nearby maints. **Notice how the first wave(s) of meteors hit the grille shielding, clearing the way for the second wave(s) of meteors to break the actual windows/walls.** This kind of outcome rarely happens with the existing non-targeted swarms -- stations are just so big that random chance almost never targets the same spot twice.

https://github.com/user-attachments/assets/fe29f0f3-ff8a-4081-8890-1c8fd9b7f813

Also worth noting that Byoin is a pretty unique station for having outside-facing departments like Medical. Most other maps have maints tunnels going around basically the entire perimeter of the station.

### 2. Directional Swarm (20% of swarms)

This one randomly decided to approach Byoin from the south-west. Directional swarms don't target a particular area. The meteors spread out all the way from the north-west salvage dock and solars, to the south-east solars and maints near the AI core.

https://github.com/user-attachments/assets/56a5636c-bcec-4bc9-b7f9-2d13eb2d7064

Directional swarms also intentionally create more missed meteors in an area around the station, too. I was going for a "passing through a large, stable asteroid field" kind of vibe, to differentiate from the more chaotic swarms. Not sure I succeeded / if that will come through to players but whatever lol

### 3. Random Swarm (20% of swarms)

The existing fully random swarms.

https://github.com/user-attachments/assets/723ce53e-60f1-4c33-b91c-19b7c9caa041

Note, I mapped in the anti-meteor arrivals/evac markers from https://github.com/DeltaV-Station/Delta-v/pull/4820 for these demos, since Byoin hasn't been updated with those yet.

## Requirements

## Breaking changes

**Changelog**

Stealth change. I think changes like this are more fun when they are discovered by players organically.


cmorley191:
Direction 3d

cmorley191:
~~I am doing a decent amount of refactoring on my changes right now so I would not recommend reviewing at this moment.~~

All done.

cmorley191:
Ok, ready for review.

Resources/Prototypes/GameRules/meteorswarms.yml
```diff
@@ -55,6 +55,16 @@
...
 
+# Begin DeltaV - add meteor rules table
+- type: entityTable
+  id: MeteorSwarmRulesTableDeltaV
+  table: !type:AllSelector
+    children:
+    - id: MeteorSwarm
+    - id: MeteorSwarmUnidirectional
+    - id: MeteorSwarmTargeted
+# End DeltaV - add meteor rules table
```
ShepardToTheStars: Move to DV namespaced file.

Resources/Prototypes/_DV/GameRules/events.yml
```diff
@@ -24,7 +24,7 @@
   components:
   - type: StationEvent
     earliestStart: 30
-    weight: 7.5
+    weight: 1.0
```
ShepardToTheStars, line 35: 1.0 might be a little low? I think mid-round zombies is a also 1.0, unless the fact that meteors is on a separate scheduler makes this a moot point?

ShepardToTheStars (commented):
Always love how thorough your summary comments are.

Resources/Prototypes/_DV/GameRules/events.yml
cmorley191: It's a moot point - these events are only part of the meteor swarm scheduler. 

Despite this, I of course pointlessly obsessed about it, briefly. :)

The pointlessly-obsessed reason it's 1.0 here is:
- the total, general "meteor swarm" weight (was 7.5) has now been divided up among three event types, with weights 1.0, 1.0, and 3.0. For a total of 5.0
- The total was reduced from 7.5 to 5.0 because we're making these events stronger, and less common.
   - again, this is a moot point. The real way we're making these less common is by slowing down the scheduler. But I reduced this number just in case we or upstream adds meteors to some other scheduler in the future

cmorley191:
> Always love how thorough your summary comments are.

Thank you :)

It's a double-edged sword to watch out for, though. I've seen code changes with highly detailed comments that basically "cause" reviewers to miss mistakes in the code itself, since they're mostly focusing on reading the comments. In this case there's a lot of weird math going on so I thought comments would be appropriate. Really obscure optimizations are another good case for heavy commenting.

ShepardToTheStars:
If you resolve the conflicts @cmorley191, I'll re-review this week.

PR #4945, opened 2025-12-11
Fix beacon map colors

IamVelcroboy:
## About the PR
- Courtroom and Law office use Justice colors
- Library uses Epistemics colors
- Theater, Boxing Ring, Barber Shop, Zoo use Service colors
- Restroom uses General colors

## Why / Balance
Consistency

## Technical details
n/a

## Media
n/a

## Requirements

## Breaking changes
n/a

**Changelog**
n/a

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/station_beacon.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/station_beacon.yml
@@ -221,7 +221,7 @@
     defaultText: station-beacon-detective
 
 - type: entity
-  parent: DefaultStationBeaconSecurity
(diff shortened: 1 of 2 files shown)


cmorley191 (approved):

cmorley191:
I honestly have no idea if direction wants to have an opinion on something like this

Seems like a no-brainer but idk

weskified:
> I honestly have no idea if direction wants to have an opinion on something like this
> 
> Seems like a no-brainer but idk

there's no direction review needed for this tbh, its more of just a fix to take account for the recent changes from jobs and stuff.
it would be very pointless to do a review on this

PR #4948, opened 2025-12-12
Remove Harbinger Bundle job restriction

kotobdev:
## About the PR
The Harbinger Bundle/Accursed Scythe can now be purchased by all traitors.

## Why / Balance
Stats-wise, this is easily the best melee weapon in the game for combat.
- Deals 30 damage a swing
- Attack rate of 1.5 (esword is 1.0), effective 1.5x DPS of the esword
- 2/3 of it's damage is armor piercing and difficult to heal (shadow, combat hardsuit has 15% resist but that's pretty inconsequential)
- The remaining 1/3 is slash, which causes bleed, and isn't cauterized immediately (unlike esword)

Additionally, the fact that it gives you a ninja blink makes it one of the most useful traitor utility items in the game. This plus an emag means you can teleport into T3 armory (or vault, cap room, ai upload, what have you) on many maps, take 30 seconds to steal things and be evil, and then teleport right out. I'd spend 12 TC for that alone.

Having this item with insane overall utility be available only to chaplains and botanists feels a bit unfair.

## Technical details
yamlops

## Media
semi-unrelated but did you know blink stacks with avali zoom? fun!

https://github.com/user-attachments/assets/92639aca-01af-479c-8d00-575e228baee2

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The Harbinger Bundle is now available to all traitors.

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/Uplink/weaponry.yml
+++ b/Resources/Prototypes/_DV/Catalog/Uplink/weaponry.yml
@@ -69,9 +69,4 @@
   cost:
     Telecrystal: 12
   categories:
-  - UplinkJob
-  conditions:
-  - !type:BuyerJobCondition
-    whitelist:
-    - Chaplain
-    - Botanist
+  - UplinkWeaponry


Proxysseia:
direction 1day

Proxysseia:
Direction here
You are correct in your understanding of this weapon being extremely strong, even without accounting for the armor that comes with the purchase (that lets be honest youre tossing if youre looking for stats, but its free armor nontheless). Part of the uniqueness of the harbinger set stems from its strength, that being both the very good damage spread of shadow/slash, and of the absurdly high utility of a mid cooldown instant, LoS blink. 
If we are to lift the job restrictions on the Harbinger Set, we would have to also tone down the weapons strength, either by hitting the damage spread, the damage numbers, the utility it offers, or the cost. 
We are unwilling to adjust both the damage spread, and the utility, because those are what make the item unique. We are unwilling to increase the cost or the damage, because in order to account for the insane utility the harbingers scythe offers, we would have to either massively increase cost, which massively cuts down on the options you have besides the item, or we would have to decrease the damage by a lot.
While both a heavy decrease to the damage or a massive hike in TC cost are feasible options, we have decided that we do not at this point want to balance this item around every job being able to access it. We are however open to adding additional jobs (with however much of a flimsy justification you'd like) to the list of jobs allowed to purchase the Harbinger Set.

In its current form this PR is denied.

SirWarock:
Isn't the harbinger set available to the botanist? I can't think of any role that would be stronger than a botanist with this weapon.

Proxysseia:
> Isn't the harbinger set available to the botanist? I can't think of any role that would be stronger than a botanist with this weapon.

A chemist, a salvager, an atmosian, a paramedic.
All the generically strong syndies are strong with this weapon, because its damage and utility are obvious, universal and fairly self contained. We prefer to have this set strong and limited rather than underwhelming and everywhere.

SirWarock:
I don't see any of those being stronger than the botanist, unless you're atmosian maxcapping, but then you don't really need a melee scythe.
Nothing beats having jugs of hyperzine and omnizine, really.

PR #4949, opened 2025-12-12
Fix cut-off replays

### LINKED ISSUE #4926: Replays cutting off VERY early on highpop
## Description
As of late, replays appear to be cutting off early more often then they used to, sometimes cutting off upwards of **half an hour** early.

To the best of my knowledge, this is due to the maximum file size for replays- any replay that reaches about 540 megabytes is automatically cut off. Additionally, round-end text/player lists aren't saved in the metadata if this happens. 

This is probably a really big issue for administration, as it makes viewing the events of a round exactly as they happened impossible if the events in question happened after the cutoff.

Could probably be fixed by increasing the filesize limit, somehow, maybe? ~~It's not a cvar, not sure where that's stored.~~ Anecdotally, I think this became more of an issue when we increased maxpop to 100.

**Reproduction**
- Go to https://deltav.gay/replays/apoapsis/
- Notice how quite a few replays cut off at 539 megabytes
- Check those replays on https://replays.iterator.systems/
- Notice how the round-end text and player lists are null
- Cross-reference the replay site's end times with the webhook's end times

**Screenshots**
take a look at this awesome table instead

| Round # | Replay end time | Actual end time | Time Lost |
| --- | --- | --- | --- |
| 25225   | 1:18     | 1:35    | 0:17 |
| 25226    | 1:31       | 2:10     | 0:39 |
| 25238    | 1:46       | 2:16     | 0:30 |

kotobdev:
## About the PR
Replay maximum compressed/uncompressed size has been doubled. Hopefully, this will fix replays getting cut off.
Fixes https://github.com/DeltaV-Station/Delta-v/issues/4926, hopefully

## Why / Balance
Priority one issue, admins not being able to reliably use replays is Bad. Sucks for normal players, too.

## Technical details
Changes two cvars on the default deltav preset.

I can't test this beyond confirming that adding the cvar to the debug preset doesn't immediately crash the server. Even if I went through the effort to reconfigure my own dedicated server to test this config, the replay wouldn't reach the limit anyways (and hetzner would kill me). Not sure if CI will test it or not. Couldn't find any other servers that override this. Somebody sanity-check this please 🥺 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Replays should no longer get cut off early.

### DIFF
--- a/Resources/ConfigPresets/DeltaV/deltav.toml
+++ b/Resources/ConfigPresets/DeltaV/deltav.toml
@@ -101,3 +101,7 @@ storage_limit = 2
 
 [nf14.respawn]
 time = 600
+
+[replay]
+max_compressed_size = 1048576 # 1024 x 1024 (in kb), double the default, approx. 1gb
+max_uncompressed_size = 2097152 # 1024 x 2048 (in kb), double the default, approx. 2gb


Woljif:
Very good fix but this does double the amount of storage delta uses on wherever it hosts its replays right

kotobdev:
> Very good fix but this does double the amount of storage delta uses on wherever it hosts its replays right

It shouldn't. Replays should still take up the same amount of space relative to how long they are, they just won't be cut off prematurely anymore.

Axionyxx:
> > Very good fix but this does double the amount of storage delta uses on wherever it hosts its replays right
> 
> It shouldn't. Replays should still take up the same amount of space relative to how long they are, they just won't be cut off prematurely anymore.

It will. And this value should already be set unless it was manually changed, since it should've been in the config file I provided when the server changed hands.

This also doesn't fix the issue entirely. Replays will still be cut off at some point because the way they increase in size isn't linear.

kotobdev:
> It will. And this value should already be set unless it was manually changed, since it should've been in the config file I provided when the server changed hands.

Oh, is the value already being set in the instance config then? I assume it'd end up overriding what it's set to in deltav.toml in that case

Toby222:
> > It will. And this value should already be set unless it was manually changed, since it should've been in the config file I provided when the server changed hands.
> 
> Oh, is the value already being set in the instance config then? I assume it'd end up overriding what it's set to in deltav.toml in that case

You'd have to ask current host what they're actually set to

Axionyxx:
> Oh, is the value already being set in the instance config then? I assume it'd end up overriding what it's set to in deltav.toml in that case

Yes, the watchdog configuration will always take priority over anything set by the repo.

But I also can't give you 100% certainty on this, it *was* when I handed the config file over, but that was months ago. So who knows what changed. You might be able to check by just seeing what size a cut off replay is presenting in the replay_final.yml(iirc) file

kotobdev:
One of the cut off replays I tested has a compressed size of 537032702 bytes per that yaml file (and ALL of the cut off ones display a compressed size of 539mb on the replay index), while the [default RT max compressed size](https://github.com/space-wizards/RobustToolbox/blob/fe35a24e88f7aa55a6b07aeb31cea82d6bfb6e5f/Robust.Shared/CVars.cs#L1710) is a pretty-close 524288000, which led me to believe that it's *probably* not being set elsewhere. Uncompressed size is 2765662340 while RT default max is 1048576000, however, so that's *probably* being set already. 

unfortunately it's all pretty much just speculation without knowing what the watchdog config exactly looks like now, yeah. but worst-case scenario, this is merged and does precisely nothing

Toby222:
> unfortunately it's all pretty much just speculation without knowing what the watchdog config exactly looks like now, yeah. but worst-case scenario, this is merged and does precisely nothing

I've messaged the host about this PR

Axionyxx:
> One of the cut off replays I tested has a compressed size of 537032702 bytes per that yaml file (and ALL of the cut off ones display a compressed size of 539mb on the replay index), while the [default RT max compressed size](https://github.com/space-wizards/RobustToolbox/blob/fe35a24e88f7aa55a6b07aeb31cea82d6bfb6e5f/Robust.Shared/CVars.cs#L1710) is a pretty-close 524288000, which led me to believe that it's _probably_ not being set elsewhere. Uncompressed size is 2765662340 while RT default max is 1048576000, however, so that's _probably_ being set already.

When I was dealing with this issue it was always the uncompressed limit being hit, not the compressed one. It's very likely this is already set somewhere else, but I'm sure Toby will get that answer for you

makyo:
Host here. These values are indeed being overwritten in the instance config:

```toml
max_compressed_size = 524288
max_uncompressed_size = 10485760
```

This was set when there was a push to use admin logs over replays. I can clear these before the branch lands, but we probably won't be able to maintain quite so many replays in that case. Per discussion with heads, we currently only maintain the last 30 days.

cmorley191:
If it's being overriden elsewhere, let's not set this in the repo's `deltav.toml`. Would only cause more confusion.

Can continue discussion on changing the overriden values, but discord might be a better place for that.

PR #4974, opened 2025-12-14
Trait - Redshirt

kotobdev:
Get in, get out again, and no one gets hurt.
    Something is pulling me up the hill.
    I look down in my red shirt.
    I look down in my red shirt.](https://youtu.be/xQIuapbeh0I?si=MTctJYxFknBxCqVA&t=47)

The Redshirt trait has made its way to DeltaV. When chosen, you immediately die when reaching 100 damage - no crit.

## Why / Balance
Fun trait, can't wait to play as an unrevivable unborgable redshirt secoff and be irreversibly murdered by a traitor.

Health analyzers get a line of text indicating when somebody has the trait. This way, they'll be able to know if they need to heal the person back to under 100 damage or not.

This will also affect IPCs by effectively removing their extra 20 health. So maybe just don't pick it as an IPC. Nothing will break, it'll just make your experience slightly worse.

## Technical details
Bit more complicated then it ought to be, for the following reasons:
1) Needs to modify MobThresholds. Ended up doing it essentially the same way Devil does with its contracts, overwriting the existing MobThreshold component. While this could also probably be done through RedshirtSystem, leaving a way to let traits override components in the future might be useful.

2) Needs to tell DeathgaspComponent to still deathgasp when you die, since it won't normally if you go directly from alive to dead. Hence having a RedshirtSystem/Component in the first place

3) the medical indicator :(

## Media

https://github.com/user-attachments/assets/e7e39a10-b38b-4661-a189-55a6fc3ff30c

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The Redshirt trait has been added, which causes you to immediately die instead of going crit.

### DIFF
--- a/Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
+++ b/Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
@@ -217,7 +217,7 @@ namespace Content.Client.HealthAnalyzer.UI
             // Alerts
 
             var unborgable = _unborgable.IsUnborgable(_target.Value); // DeltaV
-            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable;
+            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable || msg.Redshirt == true; // DeltaV - Redshirt
 
             AlertsDivider.Visible = showAlerts;
             AlertsContainer.Visible = showAlerts;
@@ -245,6 +245,14 @@ namespace Content.Client.HealthAnalyzer.UI
                 AlertsContainer.AddChild(new RichTextLabel
                 {
                     Text = Loc.GetString("health-analyzer-window-entity-unborgable-text"),
+                    Margin = new Thickness(0, 4),
+                    MaxWidth = 300
+                });
+
+            if (msg.Redshirt == true) // DeltaV
+                AlertsContainer.AddChild(new RichTextLabel
+                {
+                    Text = Loc.GetString("health-analyzer-window-entity-redshirt-text"),
                     Margin = new Thickness(0, 4),
                     MaxWidth = 300
                 });
--- a/Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
+++ b/Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using System.Numerics;
+using Content.Client._DV.Traits.Assorted; // DeltaV
 using Content.Client.Message;
 using Content.Shared._DV.Traits.Assorted; // DeltaV
 using Content.Shared.Atmos;
@@ -39,6 +40,7 @@ namespace Content.Client.HealthAnalyzer.UI
         private readonly IPrototypeManager _prototypes;
         private readonly IResourceCache _cache;
         private readonly UnborgableSystem _unborgable; // DeltaV
+        private readonly RedshirtSystem _redshirt;
 
         // Shitmed Change Start
         public event Action<TargetBodyPart?, EntityUid>? OnBodyPartSelected;
@@ -69,6 +71,7 @@ namespace Content.Client.HealthAnalyzer.UI
             _prototypes = dependencies.Resolve<IPrototypeManager>();
             _cache = dependencies.Resolve<IResourceCache>();
             _unborgable = _entityManager.System<UnborgableSystem>(); // DeltaV
+            _redshirt = _entityManager.System<RedshirtSystem>(); // DeltaV
             // Shitmed Change Start
             _bodyPartControls = new Dictionary<TargetBodyPart, TextureButton>
             {
@@ -217,7 +220,8 @@ namespace Content.Client.HealthAnalyzer.UI
             // Alerts
 
             var unborgable = _unborgable.IsUnborgable(_target.Value); // DeltaV
-            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable || msg.Redshirt == true; // DeltaV - Redshirt
+            var redshirt = _redshirt.IsRedshirt(_target.Value) && mobStateComponent?.CurrentState == MobState.Dead; // DeltaV - Redshirt
+            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable || redshirt; // DeltaV - Unborgable/Redshirt
 
             AlertsDivider.Visible = showAlerts;
             AlertsContainer.Visible = showAlerts;
@@ -249,7 +253,7 @@ namespace Content.Client.HealthAnalyzer.UI
                     MaxWidth = 300
                 });
 
-            if (msg.Redshirt == true) // DeltaV
+            if (redshirt) // DeltaV
                 AlertsContainer.AddChild(new RichTextLabel
                 {
                     Text = Loc.GetString("health-analyzer-window-entity-redshirt-text"),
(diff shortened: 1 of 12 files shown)


AirlineFood001:
environmental storytelling character shall be done, and it shall be beautiful

Proxysseia:
direction 24h

Proxysseia:
Might lead to some confusion in the medbay and accidental RRs, but if you dont want to die to 100 poison then dont click the trait that makes you die at 100 poison.
direction approved.

Proxysseia (dismissed):

ShepardToTheStars:
@kotobdev One small change. When someone is dead and has this trait, send a message with the health analyzer payload that says that they have this trait (but like in a way that makes sense and isnt "They have the redshirt trait".)

Only ever really need to see it when they are dead, and the code that constructs the health analyzer payload will already contain current health info so it shouldn't be too involved.

SirWarock:
Imo we should just show on the analyzer at which damage they crit or die.
We already have damage numbers, why not show those?

Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
```diff
@@ -249,6 +249,14 @@ public void Populate(HealthAnalyzerScannedUserMessage msg)
                     MaxWidth = 300
                 });
 
+            if (msg.Redshirt == true) // DeltaV
+                AlertsContainer.AddChild(new RichTextLabel
+                {
+                    Text = Loc.GetString("health-analyzer-window-entity-redshirt-text"),
+                    Margin = new Thickness(0, 4),
+                    MaxWidth = 300
+                });
```
ShepardToTheStars:
```suggestion
            // DeltaV - Redshirt
            if (msg.Redshirt == true && mobStateComponent?.CurrentState == MobState.Dead) // DeltaV
                AlertsContainer.AddChild(new RichTextLabel
                {
                    Text = Loc.GetString("health-analyzer-window-entity-redshirt-text"),
                    Margin = new Thickness(0, 4),
                    MaxWidth = 300
                });
            // END DeltaV
```

Content.Server/_DV/Traits/Assorted/RedshirtSystem.cs
```diff
@@ -0,0 +1,19 @@
+using Content.Server.Mobs;
+using Content.Shared._DV.Traits.Assorted;
+
+namespace Content.Server._DV.Traits.Assorted;
+
+public sealed class RedshirtSystem : EntitySystem
```
ShepardToTheStars:
```suggestion
public sealed class RedshirtSystem : SharedRedshirtSystem
```

Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
```diff
@@ -217,7 +217,7 @@ public void Populate(HealthAnalyzerScannedUserMessage msg)
             // Alerts
 
             var unborgable = _unborgable.IsUnborgable(_target.Value); // DeltaV
-            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable;
+            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable || msg.Redshirt == true; // DeltaV - Redshirt
```
ShepardToTheStars:
```suggestion
            var redshirt = _redshirt.IsRedshirt(_target.Value) && mobStateComponent?.CurrentState == MobState.Dead; // DeltaV - Redshirt
            var showAlerts = msg.Unrevivable == true || msg.Bleeding == true || unborgable || redshirt; // DeltaV - Unborgable/Redshirt
```

Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs
```diff
@@ -249,6 +249,14 @@ public void Populate(HealthAnalyzerScannedUserMessage msg)
                     MaxWidth = 300
                 });
 
+            if (msg.Redshirt == true) // DeltaV
```
ShepardToTheStars:
```suggestion
            if (redshirt) // DeltaV
```

ShepardToTheStars (commented):
Alright, I'm going to have you make some fairly significant changes to this for the following reasons.

1. Upstream is going away from checking for random components and moving towards accessing if an entity has something through relevant systems (so we should check if a entity has Redshirt through the system, not just TryComp'ing it)
2. You can get rid of all of the health analyzer payload logic, and just use the System to query if they have redshirt.
3. (optional) Feel free to get rid of the `Assorted` namespace for Redshirt. Its pointless. Very optional though, since there are other components and systems in the namespace.

* [ ] Create the following classes:
### `Content.Shared/_DV/Traits/Assorted/SharedRedshirtSystem.cs`
```c#
namespace Content.Shared._DV.Traits.Assorted;

public abstract class SharedRedshirtSystem : EntitySystem
{
    public bool IsRedshirt(Entity<RedshirtComponent?> entity)
    {
        if (!Resolve(entity, ref entity.Comp, false))
            return false;

        return true;
    }
}
```

### `Content.Client/_DV/Traits/Assorted/RedshirtSystem.cs`
```c#
using Content.Shared._DV.Traits.Assorted;

namespace Content.Client._DV.Traits.Assorted;

public sealed class RedshirtSystem : SharedRedshirtSystem;
```

* [ ] Undo changes in `Content.Server/Medical/HealthAnalyzerSystem.cs`
* [ ] Undo changes in `Content.Shared/MedicalScanner/HealthAnalyzerScannedUserMessage.cs`
* [ ] Add the following to `Content.Client/HealthAnalyzer/UI/HealthAnalyzerWindow.xaml.cs`

```C#
using Content.Client._DV.Traits.Assorted;

       // ...

        private readonly RedshirtSystem _redshirt; // Below the unborgable system
        
        public HealthAnalyzerWindow()
        {
            // ... other systems being initialized
            _redshirt = _entityManager.System<RedshirtSystem>(); // DeltaV
        }
```

Resources/Locale/en-US/_DV/traits/traits.ftl
```diff
@@ -41,3 +41,12 @@ trait-depression-desc = No mechanical effect. The world is dark but there is a l
...
 trait-dysgraphia-desc = You cannot write, whether because of a disability, your history, or by choice.
+
+trait-redshirt-name = Redshirt
+trait-redshirt-desc =
+    "They said this air would be breathable.
+    Get in, get out again, and no one gets hurt.
+    Something is pulling me up the hill.
+    I look down in my red shirt.
+    I look down in my red shirt."
+    Sets your death threshold to 100 damage.
```
ShepardToTheStars: Readability and clarity.

```suggestion
trait-redshirt-desc =
    Sets your death threshold to 100 damage. You no longer have a critical state.

    "They said this air would be breathable.
    Get in, get out again, and no one gets hurt.
    Something is pulling me up the hill.
    I look down in my red shirt.
    I look down in my red shirt."
   
```

Image

Resources/Prototypes/_DV/Traits/disabilities.yml
```diff
@@ -38,3 +38,17 @@
...
+  name: trait-redshirt-name
+  description: trait-redshirt-desc
+  category: Disabilities
+  components:
+  - type: Redshirt
+  overriddenComponents:
+  - type: MobThresholds
+    thresholds:
+      0: Alive
+      99.9: Critical
```
ShepardToTheStars, line 53: Is `Critical` needed?

ShepardToTheStars (changes requested):

Resources/Prototypes/_DV/Traits/disabilities.yml
kotobdev: Initially tried testing without it, which caused MobThresholds to have an Invalid state initially. Things seem to work fine despite that, but IPCs use this same trick, and I'm assuming there's a reason why.

image

ShepardToTheStars (approved):
Awesome. Much cleaner. :)

PR #4976, opened 2025-12-14
xenoarch: artifacts no longer irradiate themselves (much)

cmorley191:
## About the PR

Artifacts take 90% less radiation from themselves.

i.e.
Your artifact has two nodes. Node 1 has an *effect* of "become radioactive". Node 2 has a *trigger* of "radiation". If you activate node 1, then, from then on Node 2 will trigger over and over, due to the artifact's radiation irradiating itself.

## Why / Balance

Right now an artifact in this situation will trigger the radiation trigger about every 10 seconds. That really inhibits your ability to unlock non-radiation nodes since the radiation trigger will trigger and conflict very often.

I considered removing the self irradiation entirely but I think it could be fun/useful to keep in. Just need it to be infrequent enough so that scientists can work around it. (with this PR it will trigger about every 2 minutes; or you can still use an outside radiation source like a uranium spear to make it faster)

## Technical details

This was an unintended side-effect of https://github.com/DeltaV-Station/Delta-v/pull/4817 -- radiation is aggregated each frame before applying, so it can't be attributed to a single source, so the more general "you can't damage yourself" system wasn't able to take effect.

Tested that this PR works by putting two artifacts with "radiation" triggers next to eachother. Turned one of them into a radiation source -- it triggered way slower than the other one, as intended.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Radiation/Components/RadiationReceiverComponent.cs
+++ b/Content.Server/Radiation/Components/RadiationReceiverComponent.cs
@@ -16,5 +16,11 @@ public sealed partial class RadiationReceiverComponent : Component
     /// </summary>
     [ViewVariables(VVAccess.ReadOnly)]
     public float CurrentRadiation;
+
+    /// <summary>
+    /// DeltaV - multiplier on the rads received if the RadiationSource is the same entity as this RadiationReceiver
+    /// </summary>
+    [DataField]
+    public float SelfSourceMultiplier = 1.0f;
 }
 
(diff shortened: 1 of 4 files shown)


cmorley191:
I consider this a bug fix, unless direction has an opinion about the "keeping it in at 10% strength" part

ShepardToTheStars (approved):

PR #4980, opened 2025-12-14
Gibbing (via nymphing, artifact crusher, psionic eruption, and anomaly critting) now drops items

### LINKED ISSUE #4478: Diona self-gib deleting items.
## Description
When self-gibbing as a diona (and possibly being gibbed by another player, untested.), all items on the character are fully deleted from the playing field. This could be exploited to delete mission critical items, and is just generally a massive pain to deal with as a diona player especially with resources that are limited per station such as a handheld crew monitor.

**Reproduction**
Die as a diona and gib yourself, all equipped items will be completely round removed from the playing field, including PDA and bags.

Stxcking:
## About the PR

Diona drops all items before nymphing (self-gibbing). Artifact crusher used primary by SRNs now drops items of those being crushed. Psionic eruptions user drops items upon gibbing. Anomaly holders drop items when anomaly goes critical. (Basically, when getting gibbed by something, character's items will drop.)
## Why / Balance

I believe it was intentional that these mechanics delete items, but it makes sense for items to drops and Diona nymphing deleting items is considered a bug:
Fixes https://github.com/DeltaV-Station/Delta-v/issues/4478
## Technical details

Change `acidify` parameters for respective systems to false.
## Media

https://github.com/user-attachments/assets/4dd11557-4394-48db-899f-791b9527f43c
## Requirements

## Breaking changes

**Changelog**

- fix: Nymphing, crushing, psionic eruption, and anomaly critting drops items now! (Items are no longer deleted on gibbing)


ShepardToTheStars:
This is a decent solution to this issue. However, I think there are some general issues with gibbing in general where this could be applied. For example, when someone is crushed in a crusher (because of an SRN or something), I believe there is also an issue where the items are not all dropped. If we fix the issue there as well, it should fix it for Diona's gibbing themselves (probably).

Stxcking:
I will look into it more since I didn't want to mess with the gibbing systems, but I probably have to if that's the case!

Content.Shared/_DV/Species/Systems/DionaNymphingSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+using Content.Shared.Hands.EntitySystems;
+using Content.Shared.Inventory;
+using Content.Shared.Species.Components;
+
+namespace Content.Shared._DV.Species.Systems;
+
+/// <summary>
+/// Handles item dropping for diona nymphing to prevent items from being deleted during the gib transformation.
+/// </summary>
+public sealed class DionaNymphingSystem : EntitySystem
```
Toby222: The system is generic for all entities that do something when gibbed, no? Can rename it to not be specific to Diona in the name

Content.Shared/_DV/Species/Systems/DionaNymphingSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+        base.Initialize();
+
+        // Subscribe to BeingGibbedEvent which is raised before the body is deleted
+        // This only triggers for entities with GibActionComponent (dionas)
+        SubscribeLocalEvent<GibActionComponent, BeingGibbedEvent>(OnBeingGibbed);
+    }
+
+    private void OnBeingGibbed(EntityUid uid, GibActionComponent comp, ref BeingGibbedEvent args)
+    {
+        // Drop all items before the body is deleted to prevent item loss during nymphing
```
Toby222: Should check for args.Cancelled here probably, though I'm not sure when that would ever happen

Content.Shared/_DV/Species/Systems/DionaNymphingSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        // Subscribe to BeingGibbedEvent which is raised before the body is deleted
+        // This only triggers for entities with GibActionComponent (dionas)
+        SubscribeLocalEvent<GibActionComponent, BeingGibbedEvent>(OnBeingGibbed);
+    }
+
+    private void OnBeingGibbed(EntityUid uid, GibActionComponent comp, ref BeingGibbedEvent args)
```
Toby222: Use `Entity<GibActionComponent>` instead
```suggestion
    private void OnBeingGibbed(Entity<GibActionComponent> ent, ref BeingGibbedEvent args)
```

Toby222 (dismissed):

ShepardToTheStars:
> I will look into it more since I didn't want to mess with the gibbing systems, but I probably have to if that's the case!

It may be as easy as looking at at the `GibContentsOption` and changing references from `GibContentsOption.Skip` to `GibContentsOption.Drop` in some of the shared body systems that deal with gibbing body parts.

ShepardToTheStars:
If its not, we'll go with this solution.

Stxcking:
> > I will look into it more since I didn't want to mess with the gibbing systems, but I probably have to if that's the case!
> 
> It may be as easy as looking at at the `GibContentsOption` and changing references from `GibContentsOption.Skip` to `GibContentsOption.Drop` in some of the shared body systems that deal with gibbing body parts.

Okay! So, it seems like it was really simple as changing the `acidify` parameter for nymphing and artifact crusher and psionic eruption (which also had the problem of not dropping items). I will go ahead and turn this PR to just fix other gibbing methods to drop items.

ShepardToTheStars:
> Okay! So, it seems like it was really simple as changing the acidify parameter for nymphing and artifact crusher and psionic eruption (which also had the problem of not dropping items). I will go ahead and turn this PR to just fix other gibbing methods to drop items.

Aweome! Feel free!

Stxcking:
I also just thought of anomaly critting not dropping items, I wonder if there's any other gibbing mechanics that we want to have drop items...otherwise, this should be good!

ShepardToTheStars (approved):

PR #4982, opened 2025-12-14
[K9] Add ability to control entities via remote controls

BarryNorfolk:
## About the PR
Add remote control system which enables users to control NPCs via pointing. This is intended to be a generic system, hopefully enabling other usages for other types of play (What if the dragon could control all the fish they spawn?).

Notes:
- Do we need to have the control in hand to be able to use it?
  - No, it's usable in both the hands and as an equipment on the neck. It is a toggled action to have the point orders active or not. 
- How do I bind a creature to a control?
  - Have the control (or whistle in the dog's case) in your hands and right click the dog to see "Bind to Control".
- Where do I get the whistle from?
  - Each secdog/laika starts with one dog whistle in their pocket slot. Simply take it off them, bind them, and start sending orders.

## Why / Balance
Part of the K9 rework described in the design-documents forum.

## Technical details
Like the grapple system, the remote controls are designed to be generic and used in multiple areas. They have direct bind/unbind mechanics and try to cover everything neatly so it's extensible. You can set the kind of AI interactions you want for the orders via the HTNCompounds, like I've done for secdogs.

The length of time the popup message stays up on the screen, for short messages, like "Sit" and "Catch" is annoying but set by the Popup system and would be quite intrusive to change at this point.

A player who disconnects/ghosts from Laika will leave an NPC that can't respond to orders due to the HTN system (from what I can see) still seeing the non-null mind left over. I haven't seen a nice way to clear it and am just assuming that it's a separate system and the intended effect. I'm hoping that there's some timeout here that forces the mind to null out, which would return the entity to a controllable state.

## Media

Controls are able to bind/unbind from receivers.
Screenshot From 2025-12-16 18-52-40
image

View from the player as a Secdog

[2025-12-16.19-49-06.mp4](https://github.com/user-attachments/assets/52b9e1dc-73da-4e13-a4f8-fc6d2e4da71c)

Orders let you control the doggo!

https://github.com/user-attachments/assets/d9a6ec9d-4969-4a9d-94cb-134f25156336

Screenshot From 2025-12-21 20-18-00

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Secdogs can now be controlled via the dog whistles they spawn with! Finally you can tell Laika what to do!


Cepelinas1:
direction 24h

cmorley191 (commented):
I'll review this on behalf of maintainers shortly.

Resources/Locale/en-US/_DV/remotecontrol/remotecontrol.ftl
```diff
@@ -0,0 +1,9 @@
+remote-control-order-default = Incomprehensible screeching
+
+remote-control-bind-verb-text = Bind to control
+remote-control-bind-verb-message = Bind this entity to the control.
+remote-control-bind-popup-start = {$user} attempting to bind {$target}.
```
cmorley191:
```suggestion
remote-control-bind-popup-start = {$user} is attempting to bind {$target}.
```

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// Handles when an entity using an active remote control points at an entity.
+    /// </summary>
+    /// <param name="holder">Entity holding/wearing the remote control.</param>
+    /// <param name="args">Args for the event, notably the entity pointed at.</param>
+    private void OnPointedAtEntity(Entity<RemoteControlHolderComponent> holder, ref AfterPointedAtEvent args)
+    {
+        if (holder.Owner == args.Pointed)
+            SendSelfPointOrder(holder.AsNullable());
+        else
+            SendEntityPointOrder(holder.AsNullable(), args.Pointed);
```
cmorley191, line 193: May be a good spot for the free order?

```suggestion
        else if (
          TryComp<RemoteControlReceiverComponent>(args.Pointed, out var receiver)
          && receiver.BoundController == holder.Comp.Control
        )
            SendFreeUnitOrder(holder.AsNullable());
        else
            SendEntityPointOrder(holder.AsNullable(), args.Pointed);
```

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            RaiseLocalEvent(recipient.Value, ref ev);
+            _useDelay.TryResetDelay(holder.Comp.Control);
+        }
+        else
+            SendOrderToReceivers(control, ref ev);
+
+        return true;
+    }
+
```
cmorley191, line 178: I think it'd be worthwhile to consolidate the functions even more; less copy-paste often means fewer chances for mistakes. 

It actually made me realize we should check `recipient`'s channel name too

```
    private bool TrySendOrder<TEvent>(
        Entity<RemoteControlHolderComponent?> holder, 
        Func<Entity<RemoteControlComponent>, TEvent> getEvent, 
        EntityUid? recipient = null
    ) where TEvent : notnull
    {
        if (!Resolve(holder, ref holder.Comp))
            return false;

        if (!TryComp<RemoteControlComponent>(holder.Comp.Control, out var controlComp))
            return false;

        Entity<RemoteControlComponent> control = (holder.Comp.Control, controlComp);

        if (_useDelay.IsDelayed(control.Owner)) // prevent recipient from being spammed with messages
            return false;

        _useDelay.TryResetDelay(control.Owner);

        var ev = getEvent(control);
        if (recipient.HasValue)
        {
            if (
                TryComp<RemoteControlReceiverComponent>(recipient.Value, out var receiverComp)
                && receiverComp.ChannelName == control.Comp.ChannelName
            )
            {
                RaiseLocalEvent(recipient.Value, ref ev);
            }
        }
        else
        {
            var query = EntityQueryEnumerator<RemoteControlReceiverComponent>();
            while (query.MoveNext(out var ent, out var receiverComp))
            {
                if (receiverComp.ChannelName == control.Comp.ChannelName)
                {
                    RaiseLocalEvent(ent, ref ev);
                }                
            }
        }

        return true;
    }

    public bool SendEntityPointOrder(Entity<RemoteControlHolderComponent?> holder, EntityUid pointed, EntityUid? recipient = null)
    {
        return TrySendOrder(
            holder,
            (control) => new RemoteControlEntityPointOrderEvent(holder, control.Owner, control.Comp.BoundEntities, pointed),
            recipient
        );
    }

    public bool SendTilePointOrder(Entity<RemoteControlHolderComponent?> holder, MapCoordinates tile, EntityUid? recipient = null)
    {
// etc.
```

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.Receiver.cs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// <summary>
+    /// Updates the NPCs current orders and runs the supplied action to populate any blackboard
+    /// details that are required.
+    /// </summary>
+    /// <param name="npc">NPC to update orders for.</param>
+    /// <param name="order">The current order enum they should follow.</param>
+    /// <param name="orderAction">A lambda that updates any blackboard information for the current order.</param>
+    private void UpdateNPCOrders(EntityUid npc, RemoteControlOrderType order, Func<bool> orderAction)
```
cmorley191: If we're not using the `bool` output of `orderAction`, let's just make it an `Action orderAction`

Content.Shared/Pointing/PointingEvents.cs
```diff
@@ -31,3 +32,12 @@ public PointingAttemptEvent(NetEntity target)
...
 public readonly record struct AfterGotPointedAtEvent(EntityUid Pointer);
+
+// Begin DeltaV Additions - Interactions with tile pointing
+/// <summary>
+/// Raised on the entity who is pointing after they point at at a tile.
+/// </summary>
+/// <param name="Pointed">MapCoordinates pointed at by the entity.</param>
+[ByRefEvent]
+public readonly record struct AfterPointedAtTileEvent(MapCoordinates Pointed);
+// End DeltaV Additions - Interactions with tile pointing
```
cmorley191: Summary comments are a special case you can consolidate

```suggestion
/// <summary>
/// DeltaV - Raised on the entity who is pointing after they point at at a tile.
/// </summary>
/// <param name="Pointed">MapCoordinates pointed at by the entity.</param>
[ByRefEvent]
public readonly record struct AfterPointedAtTileEvent(MapCoordinates Pointed);
```

cmorley191 (commented):
I'm still looking through this, but here are some initial comments.

Overall this is looking awesome so far! Thanks for the contribution!

Content.Shared/_DV/RemoteControl/Components/RemoteControlHolderComponent.cs
cmorley191, line 1: Rename file to RemoteControlHolderComponent.cs

cmorley191:
I had some ideas while playing around with the whistle. These are totally optional though! I'm just throwing them out there, feel free to ignore all of these:

- Players might have a better experience if the "Toggle remote control orders" button felt more like the harm-mode button
  - call it "Enter/Exit Order Mode"
  - make it have a slightly brighter shade like the harm mode button does
  - have it change color like the harm mode button does
    - this is the main thing. I think an "Am I currently in Order Mode?" visual would be a bit clearer than the current "What would pressing the button do right now?" visual
- Pressing Z to move the whistle from your hand to your neck would be nice (like most clothing)
- I would enjoy seeing the order popups as the whistle user. Having some feedback is always good on tools. I imagine it could help people realize if the whistle is working vs. if the dog is just a player that's ignoring them lol
  - Could put the user's popups over themselves, or alternatively over the spot that they point at?
- Can vulpkanin players hear the whistle sound? Could be a fun addition
- Split the loadout into a normal sec dog loadout (vest+whistle) and a Laika loadout (vest+whistle+sunglasses), because Laika is cool
- imo a very short whistle audio clip would be better; this is probably gonna get spammed a good amount.

Content.Shared/Pointing/PointingEvents.cs
BarryNorfolk: Oh? Even if I've added the record struct to the file?

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.cs
BarryNorfolk: Good catch, I can apply this to the receiver end as well so there's less duplication of the checks.

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.cs
BarryNorfolk: Sure, I think it's a decent way of giving freedom back to the unit.

BarryNorfolk:
For the order mode, I think that's a good idea but I definitely lack the spriting skills to make something look nice and workable. :C 

Pressing Z usually makes the noises for other whistles, but here it defo makes sense to allow quickEquip so I've done that.
**Done**

For showing the orders, that could be a nice addition yeah. I think it could work just fine I'll just have to add the same localised strings to the control as the receiver. At the moment it's the receiver who "Perceives" the orders so just need to duplicate that across.
    **Done**.

Vulps understanding this is easy to do yeah, I made it with that in mind. We just add the receiver, channel, and make it so they can't understand. As long as the bound entities and stuff don't make it more annoying...
**Done**

I can split the inventory, sure. Don't mind!
      **Done**.

I can have a look for a shorter whistle sound.

Content.Shared/Pointing/PointingEvents.cs
cmorley191: Yeah, one-line declarations -- this, or new variables on upstream components -- you can just put DeltaV in the summary comment.

Although now that I think of it, our upstream mergers probably would appreciate this one just being moved to a _DV/Pointing/PointingEvents.cs file

Content.Server/_DV/RemoteControl/EntitySystems/RemoteControlSystem.cs
cmorley191: The Receive code looked a bit more complicated to factor out, but if you can find a way to do it then more power to ya! :)

cmorley191:
Like I said all those ideas are totally optional, feel free to ignore them if you don't feel like adding them.

We could do them in followup PRs too.

Content.Shared/Pointing/PointingEvents.cs
BarryNorfolk: Yeah, that's fair and easier to do. Cheers.

cmorley191:
@BarryNorfolk I have some more review comments for you.

In summary... Currently this PR is sort of duplicating the functionality of the ItemToggle and ToggleClothing components. Instead I'd prefer we just use those components and systems directly.

Magboots are a great example of ItemToggle/ToggleClothing. They have an on-off switch, and because it's inconvenient to rightclick+activate them when you're wearing them, we've added an action for it! The dog whistle is in a similar situation to that, so I think we can model it as toggleable clothing.

Originally I was going to write a bunch of comments on your PR to describe all this, but it's sort of a big multi-file change.

I've actually implemented it for you so you can see a diff of the changes you'd need to make. I didn't touch your branch, just did this on my own fork so I could show you a diff -- take a look at the Files Changed in this fake PR: https://github.com/cmorley191/Delta-v/pull/1

cmorley191:
Sorry for the confusing presentation of my review; happy to answer any questions, you can ping/DM me on the DeltaV discord too if that's easier for you. Or just post your comments here; either way.

My goal with this suggestion is to prevent us from duplicating functionality -- these ItemToggle/ToggleClothing systems have been in use for a while so theoretically they'll already be more robust to weird edge cases we might not've thought of. Much easier to maintain than implementing it all with a new ToggleRemoteControlEvent

cmorley191:
I think that's the last of my requests from a maintainer perspective. Thanks for bearing with me; things slowing down over the holidays.

BarryNorfolk:
Thanks for the in-depth review and even PR to look at for changes. If I can get away with not duplicating code then that's all good by me, I'll check out the PR and see what the changes look like :)
Doubly appreciate it since you're doing it over the holidays, but there's no rush and no hurry for me on this so we can take our time with it.

Content.Shared/_DV/RemoteControl/Components/RemoteControlRecieverComponent.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// If no order string exists for a given order, a default one will be applied.
+    /// </summary>
+    [DataField]
+    public Dictionary<RemoteControlOrderType, LocId> OrderStrings = new();
+
+    /// <summary>
+    /// Controller that can send orders to this entity.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? BoundController = null;
```
cmorley191: \<minor>

I think this would help future readers (I wasn't sure if this was the remote control or the control holder until I looked at the System code)

```suggestion
    /// <summary>
    /// Control that a holder can use to send orders to this entity.
    /// </summary>
    [DataField, AutoNetworkedField]
    public EntityUid? BoundControl = null;
```

\</minor>

Content.Shared/_DV/RemoteControl/EntitySystems/SharedRemoteControlSystem.cs
```diff
@@ -0,0 +1,284 @@
...
+        if (!Resolve(control, ref control.Comp) ||
+            !TryComp<RemoteControlReceiverComponent>(entity, out var receiverComponent))
+            return;
+
+        if (!control.Comp.AllowMultiple && control.Comp.BoundEntities.Count > 0)
+        {
+            while (control.Comp.BoundEntities.Count > 0)
+            {
+                if (!UnbindEntity(control, control.Comp.BoundEntities.First()))
+                    break;
```
cmorley191: Did you want a return false here?

Or maybe just ignore the return value of `UnbindEntity` entirely? (doesn't look like the `return false` conditions are too concerning)

Having `break` here seems odd -- has the potential to leave other entities in the list bound, and then continue with the function binding this new one.

Maybe even something like this would be best:

```
foreach (var boundEntity in control.Comp.BoundEntities.ToList()) // make a copy of the list since this loop will modify it
{
    if (!UnbindEntity(control, boundEntity))
        return;
}
if (control.Comp.BoundEntities.Count > 0) // sanity check -- make sure they all did, in fact, unbind
    return;
```

Content.Shared/_DV/RemoteControl/EntitySystems/SharedRemoteControlSystem.cs
```diff
@@ -0,0 +1,284 @@
...
+    /// Binds an entity to a remote control.
+    /// </summary>
+    /// <param name="control">The control the bind to.</param>
+    /// <param name="entity">The entity to bind.</param>
+    public void BindEntity(Entity<RemoteControlComponent?> control, EntityUid entity)
+    {
+        if (!Resolve(control, ref control.Comp) ||
+            !TryComp<RemoteControlReceiverComponent>(entity, out var receiverComponent))
+            return;
+
```
cmorley191, line 68: Let's add

```
if (receiverComponent.BoundController is { } existingControl)
{
    if (existingControl.Owner == control.Owner)
        return;

    if (!UnbindEntity((existingControl, null), entity))
        return;
}
```

This stops the potential case of Officer1 binding Laika, and then Officer2 binding Laika with a different control -- Laika's BoundController would change further down in this function, but Laika would remain in Officer1's control's BoundEntities list.

Don't think that case would've caused any real issues but it seems like a good thing to prevent.

Content.Shared/_DV/RemoteControl/Events/RemoteControlOrderEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Content.Shared._DV.RemoteControl.Events;
+
+/// <summary>
+/// Base event for all orders available to remote controls
+/// </summary>
+/// <param name="user">The user who activated the remote control.</param>
+/// <param name="control">The remote control the activation came from.</param>
+/// <param name="boundNPC">Which NPC entities, if any, this control is bound to.</param>
+public abstract class BaseRemoteControlOrderEvent(EntityUid user, EntityUid control, List<EntityUid> boundEntities)
```
cmorley191, line 11: Wow, TIL about [primary constructors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors). Incredible!

cmorley191 (changes requested):
A few last things, this is looking great!!

I should be more responsive now that the holidays are over

cmorley191:
Direction review kind of trailed off during the holidays but we're getting it back in to gear.

It seems like the Direction response so far is that this is an amazing PR, but that the AI instantly grappling you is pretty over-powered.

One suggestion was to make Laika only grapple once the target has taken 50 stamina damage, but it's still being debated -- I pieced together that in my local environment (it was pretty messy), and even that still seems fairly strong to me:

laika_bite

If you have any ideas for nerfs to AI Laika, feel free to share them!

Otherwise, someone actually from the Direction team (I'm not) will have an actual response for you soon. This is just... me trying to keep you in the loop.

cmorley191:
At some future point I have some more AI ideas:
- Laika defends herself if attacked, regardless of orders (currently she just sits and takes it)
- Laika defends her controller (and maybe even other secoffs or nanotrasen faction members) during Follow mode (and maybe during other modes too? could get pretty complicated)

But those definitely can wait until a future PR imo

BarryNorfolk:
No worries on being responsive and such, holidays are coming to an end for me and so I'll be back at work and probably a bit more drained :P 
I can fix up the changes mentioned which look fine, but in terms of gameplay I have no idea. I wasn't the person who designed this/thought this up, just the person implementing it. I wonder if @SirWarock has any suggestions here.

In my opinion, yeah it's a strong thing and definitely buffs securities ability to deal with just about any humanoid threats on the station. Obv the Nukies can just shoot the dog coming blindly at them with no dodging, but in a fire fight it might be a bit harder to control.

- We could limit the distance the AI can actually handle that order, such that grappling only occurs if you're within X units of the target. Else it's just an attack.
- We could make it temporary, so limiting it a few seconds before the AI automatically drops and returns.
- We could make it a threshold with stamina, as you've shown.
- Finally, we could just drop the grappling part and make it an attack only setup.

BarryNorfolk:
After talking with @ShepardToTheStars , I think I'll try and make the grapple drain stamina (but not knock down) until some certain point. 
The idea being that the dog has attached to the player and is actively draining their stamina but doesn't stop them from moving directly.

Content.Shared/_DV/RemoteControl/EntitySystems/SharedRemoteControlSystem.cs
BarryNorfolk: That's a fair point and a logic error on my part. Your suggestion here + an actual log message might be better here.

Content.Shared/_DV/RemoteControl/EntitySystems/SharedRemoteControlSystem.cs
BarryNorfolk: Great catch and good idea.

BarryNorfolk:
I'll be reworking the grapple a little in a separate PR so we don't muddy this one, then I'll come back and update it.

Stop-Signs:
This is approved contingently upon a grapple rework PR also being made and merged.

BarryNorfolk:
Closing this as I don't have the energy to work on it at the moment.

PR #4988, opened 2025-12-15
Trait - Arm Amputee

kotobdev:
## About the PR
You can now pick a trait to start with either your left or right arm missing.

## Why / Balance
My character's arm got cut off in a round like thirty minutes ago. I want to do character development. Direction speedmerge pls :)

## Technical details
Most of the logic is pretty much the same as the Den/EE system for replacing limbs with cybernetic ones, albeit adapted to a normal system/component instead of EE's trait system thing. Den's amputee traits just utilize that system without defining any replacement limb, so I did the same thing. Left in the code for replacement limbs though, in case we want that down the line.

Currently, choosing both the traits will cause only one to be amputated, because of the age-old annoyance of duplicate components not being possible. if https://github.com/DeltaV-Station/Delta-v/pull/4974 is merged, I can make them both overrides so both can apply sequentially. But it's a minor issue, and I don't want it holding up this PR. Might be for the best, anyways.

Also you drop gloves. You can put them right back on. 🤷 

## Media
image
image

## Requirements

**Changelog**

:cl:
- add: Arm amputee traits have been added. They're not very *hand*y.


Content.Server/_DV/Traits/Assorted/AmputeeComponent.cs
```diff
@@ -0,0 +1,23 @@
...
+    public BodyPartSymmetry PartSymmetry { get; private set; } = BodyPartSymmetry.Left;
+
+    /// <summary>
+    /// Body part prototype to use as a replacement limb, if applicable.
+    /// </summary>
+    [DataField]
+    public EntProtoId? ProtoId { get; private set; }
+
+    [DataField]
+    public string SlotId { get; private set; } = "right arm";
```
Toby222: This should be nullable just like ProtoId
This defaults to right arm, but the removed body part is left arm by default

apeiroeder:
direction review 24h

Content.Server/_DV/Traits/Assorted/AmputeeComponent.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Content.Server._DV.Traits.Assorted;
+
+[RegisterComponent]
+public sealed partial class AmputeeComponent : Component
+{
+    [DataField]
+    public BodyPartType RemoveBodyPart { get; private set; } = BodyPartType.Arm;
+
+    [DataField]
+    public BodyPartSymmetry PartSymmetry { get; private set; } = BodyPartSymmetry.Left;
```
Toby222, line 13: instead of a single body part, this should probably be a collection of `(BodyPartType, BodyPartSymmetry)` to allow removing multiple body parts?
Might also be able to put the protoId in there instead so you can have all functions in one field

Resources/Locale/en-US/_DV/traits/traits.ftl
```diff
@@ -41,3 +41,13 @@ trait-depression-desc = No mechanical effect. The world is dark but there is a l
 
 trait-dysgraphia-name = Unable to write
 trait-dysgraphia-desc = You cannot write, whether because of a disability, your history, or by choice.
+
+trait-amputee-left-arm-name = Left Arm Amputee
```
Toby222: Maybe change to something like `Amputee (Arm, Left)` for better sorting in character editor?

Resources/Prototypes/_DV/Traits/disabilities.yml
```diff
@@ -38,3 +38,23 @@
...
+  id: ArmAmputeeL
+  name: trait-amputee-left-arm-name
+  description: trait-amputee-left-arm-desc
+  category: Disabilities
+  components:
+  - type: Amputee
+    removeBodyPart: Arm
+    partSymmetry: Left
+
+- type: trait
```
Toby222: Picking both of these traits only triggers one of them, either make it so you can only pick one, remove one of the traits, or ideally make them properly merge

Toby222 (dismissed):

Resources/Prototypes/_DV/Traits/disabilities.yml
ShepardToTheStars: I would probably make them mutually exclusive to be honest.

Content.Server/_DV/Traits/Assorted/AmputeeComponent.cs
kotobdev: If done that way, there would need to be individual traits for every combination of removed limbs. Not really an issue for just two arms, granted, but I'm not sure if it's really a good idea to give people the ability to chop off both their arms roundstart anyways, and still wouldn't be very extensible if we ever added legs.
<sub>should have just made it so traits can use JobSpecials smh</sub>
 Probably easiest to just limit them to picking one of the traits?

AveriV2:
add legs too so i can be miller mgsv phantom pain

kotobdev:
> add legs too so i can be miller mgsv phantom pain

we already have impaired mobility trait just use that smh

kotobdev:
Limited it via giving it its own category, so you can only pick one at a time.

image

Content.Server/_DV/Traits/Assorted/AmputeeComponent.cs
Toby222: works for me ig given we don't have the good trait system

apeiroeder:
direction here, we have decided to approve this PR!

thanks for the contribution, personally i think its an amazing trait

Content.Server/_DV/Traits/Assorted/AmputeeSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                QueueDel(child.Id);
+            }
+
+            _transform.AttachToGridOrMap(part.Id);
+            QueueDel(part.Id);
+
+            // apparently chopping off limbs makes people bleed a lot. Who would have guessed?
+            if (TryComp(ent, out BloodstreamComponent? bloodstream))
+                _bloodstream.TryModifyBleedAmount((ent, bloodstream), -10f);
```
ShepardToTheStars: TryComp'ing the bloodstream component isn't needed since `TryModifyBleedAmount` will resolve that component if the entity has it.

```suggestion
            _bloodstream.TryModifyBleedAmount(ent.Owner, -10f);
```

Content.Server/_DV/Traits/Assorted/AmputeeSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            // goes unused for the purposes of the arm amputee traits, but might as well keep it in
+            if (ent.Comp.ProtoId is null || ent.Comp.SlotId == null)
+                continue;
+
+            var newLimb = SpawnAtPosition(ent.Comp.ProtoId, xform.Coordinates);
+            if (TryComp(newLimb, out BodyPartComponent? limbComp))
```
ShepardToTheStars: I know this isn't used but we should probably include an additional check in case the new limb proto's `Symmetry` property does not match the `PartSymmetry` of the Amputee comp. (I think the suggestion works but you probably want to see if its syntactically correct lol)


```suggestion
            if (TryComp<BodyPartComponent>(newLimb, out var limbComp) && limbComp.Symmetry == ent.Comp.PartSymmetry)
```

ShepardToTheStars (changes requested):

ShepardToTheStars (approved):

PR #4994, opened 2025-12-17
admin logs for research, glimmer, artifacts

cmorley191:
## About the PR

Added new epistemics admin logs. I intend to use them to analyze generally how players are generating research points, and how players are interacting with glimmer, anomalies, and xenoarch.

We should start collecting data ASAP. Grafana visualizations or one-off analyses based on these logs will come later.

To be clear: these are logs that appear in the \`adminlogs\` menu, or the `admin_log` database table. They are "LogSeverity.Low", so they do *not* spam/appear in the chat window. Most admins won't know they're there

## Why / Balance

Xenoarch is busted. The recent [glimmer scaling PR](https://github.com/DeltaV-Station/Delta-v/pull/4486) has combined poorly with other scalings, and [discord users are reporting](https://discord.com/channels/968983104247185448/968983104662409244/1450391143916376115) that xenoarcheologists are finishing the research tree very early in the shift, due to nodes producing insane amounts of research points.

It'd be awesome if we start collecting data *now*, so that we will have solid before & after statistics on whatever balance changes we make in the near future.

## Technical details

There are 11 new types of log messages being generated. In total, I estimate ~20,000 more logs per round to be generated. The current average ([as discussed on discord](https://discord.com/channels/968983104247185448/1449276473650450432)) is probably about ~150,000 per round, so this is a fairly significant increase! 

However, I think these are okay to add for now, and we can look at slimming down our logs over the next couple months. A month or two of 15% more logs shouldn't break the bank on server disk storage. (we're currently storing the past year of rounds)

Examples of the 11 new logs:

### New `Research` category
1. **Points report:** (1 log every 60s, ~120 logs per round)
> current points in R&D server (10255/n10255, ResearchAndDevelopmentServer): 4592
2. **Passive research point generation:** (1 log every 1s, ~7200 logs per round)
> R&D server (10255/n10255, ResearchAndDevelopmentServer) collected 236 points from 3 sources over 1 seconds: 179 points from glimmer prober (7922/n7922, GlimmerProber), 30 points from anomaly vessel (8732/n8732, MachineAnomalyVessel), 27 points from anomaly vessel (8733/n8733, MachineAnomalyVessel), 

### New `ArtifactDetails` category

(originally had these in the existing `ArtifactNode` but I'd like to keep them separate in case we want to revert this PR & purge these logs)

3. **Each node's details logged when artifact spawns:** (~100 logs per round)
[...]

PR #4995, opened 2025-12-17
Xenoarch Research Mult Balance

SirSmith148:
## About the PR

I created a new formula for the glimmer/research points multiplier for the new xenoarch system as the formula from the old system breaks the balance. This new formula will now scale glimmer multiplier to 1.5x, 2x and 3x at the glimmer values of 500, 750, and 1000 respectively and will clamp at the max of 3x.  Where as the previous version had the multiplier at 3x for 500 and no cap. I also increased the extract ratio from 750 to 1250, this will decrease the amount of glimmer added to the world per point of research to slow progress even more.

## Why / Balance

This is to address concerns about my previous glimmer changes to xenoarch which brought over research values of nodes that were higher than the old system causing an issue with how the old system handled the glimmer multiplier, and how it can be abused to get research absurdly quick. This can be further adjusted based on feedback and admin logging of the new proposed system in #4994. 

These changes will make it so that it will get exceedingly dangerous to push research to the max while still rewarding and interacting with our unique glimmer system. Perhaps more changes could be made to make high glimmer values even more dangerous in later PRs.

## Technical details

Removed the PointGlimmerMultiplier variable from Artifact Analyzer component as it was obsolete in the new formula. And changed Extract ratio from 750 to 1250

Created the new formula in ArtifactAnalyzerSystem as follows 

multiplier = 1 + (8/3) * (glimmer value/1000)^3 - 2 * (glimmer value/1000)^2 + (4/3) * (glimmer value/1000)

## Requirements

- [ X] I have tested all added content and changes.
- [ X] I have added media to this PR or it does not require an ingame showcase.

## Media

New Curve (Note this is before a +1)
Screenshot 2025-12-17 040245

Old Curve
image

## Breaking changes

**Changelog**

:cl:
- fix: Nanotrasen has found scientists inflating their research evaluations. These scientists have been reprimanded by the company and re-educated on the proper way to fill out the research logs

### DIFF
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -86,7 +86,7 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
     {
         float normalizedGlimmer = _glimmerSystem.Glimmer / 1000f;
         //DeltaV - Prevents extreme glimmer multipliers
-        return 1 + Math.Clamp((8f / 3f) * MathF.Pow(normalizedGlimmer, 3f) - 2f * MathF.Pow(normalizedGlimmer, 2f) + (4f / 3f) * (normalizedGlimmer), 0f, 2f);
+        return (float)(.5f + Math.Clamp(Math.Pow(normalizedGlimmer,.5) + 1.5f * Math.Pow(normalizedGlimmer, 10f), 0f, 2f));
 
     }
 }
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -86,7 +86,7 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
     {
 
         //DeltaV - Prevents extreme glimmer multipliers
-        return 1 + Math.Clamp((MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) * comp.PointGlimmerMultiplier), 0, 2);
+        return Math.Clamp(4f * MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) - 3f * (_glimmerSystem.Glimmer / 1000f) + 2, 0f, 3f);
 
     }
 }
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -84,7 +84,10 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
     // DeltaV
     private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
     {
-        return 1 + (MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) * comp.PointGlimmerMultiplier);
+
+        //DeltaV - Prevents extreme glimmer multipliers
+        return 1 + Math.Clamp((MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) * comp.PointGlimmerMultiplier), 0, 2);
+
     }
 }
 
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -84,9 +84,9 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
     // DeltaV
     private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
     {
-
+        float normalizedGlimmer = _glimmerSystem.Glimmer / 1000f;
         //DeltaV - Prevents extreme glimmer multipliers
-        return Math.Clamp(4f * MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) - 3f * (_glimmerSystem.Glimmer / 1000f) + 2, 0f, 3f);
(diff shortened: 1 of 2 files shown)


ShepardToTheStars:
@cmorley191 I will defer to you on this one.

cmorley191 (approved):
Great work. You were super fast on making this!

cmorley191:
I'm tempted to say we should wait for the admin logging to be added first, so we can get some good before & after data.

If that takes longer than a day or two though, we should probably just merge this

SirSmith148:
Yeah I don't mind that, I figured this could act as a stop gap for the moment to act as a baseline since the current balance is obviously out of whack regardless

cmorley191:
Direction review 24h

(people continue to report the issue; we'll let direction decide if we should wait on data)

cmorley191:
Also @SirSmith148 imo this issue is now high-profile enough to warrant a changelog message on this patch. Could even rp it a bit if you want to. "Glimmer's interaction with artifact research value has been reduced. Unfortunately, NanoTrasen's primary dig site for these ultra-psionic research subjects has been depleted."

It's totally up to you though. I'm not sure if we even often write changelogs for balance changes. I'm still new-ish to DeltaV.

cmorley191:
java pointed out it might be good to have a small glimmer threshold (like 200 or something) where, below that, the multiplier is <1. Make it more high risk high reward.

Multiplying your whole equation by `tanh((glim + 150) / 300)` creates that kind of effect, but you may have a better idea (tanh is kinda gross)

image

image

cmorley191:
ah, me posting those plots has sparked further direction discussion, one moment.

SirSmith148:
That could help balance the higher node rewards instead of changing them, might be something to toy with after some data tracking

cmorley191:
Ok Direction has settled on preferring something that looks roughly like this, for the interim.

image

SirSmith148:
Okay do you have the equation for that line?

cmorley191:
I don't but I'll find one

SirSmith148:
> I don't but I'll find one

image


I got this one


.5+ sqrt(x/1000) +1.5(x/1000)^10

Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
```diff
@@ -84,7 +84,10 @@ private void UpdateClientUI(EntityUid console, ArtifactAnalyzerComponent analyze
     // DeltaV
     private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
     {
-        return 1 + (MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) * comp.PointGlimmerMultiplier);
+        float normalizedGlimmer = _glimmerSystem.Glimmer / 1000f;
+        //DeltaV - Prevents extreme glimmer multipliers
+        return 1 + Math.Clamp((8f / 3f) * MathF.Pow(normalizedGlimmer, 3f) - 2f * MathF.Pow(normalizedGlimmer, 2f) + (4f / 3f) * (normalizedGlimmer), 0f, 2f);
```
cmorley191: Welp, had to add two more terms, but it does the trick.

```suggestion
        return 0.5f + Math.Clamp(26 * MathF.Pow(normalizedGlimmer, 5) - 52 * MathF.Pow(normalizedGlimmer, 4) + 40 * MathF.Pow(normalizedGlimmer, 3) - 16 * MathF.Pow(normalizedGlimmer, 2) + 4.45f * normalizedGlimmer, 0, 2.5f);
```

I have not yet tested this in game, I'll do that now

cmorley191 (changes requested):

Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
cmorley191: Oh wait just saw yours. Yours is way better.

cmorley191:
Let's use your sqrt one. Looks great

SirSmith148:
> Contributor

okay getting it pushed in a sec

SirSmith148:
> Let's use your sqrt one. Looks great

double check my work Im a little busy and might have missed something sorry

Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
```diff
@@ -86,7 +86,7 @@ private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
     {
         float normalizedGlimmer = _glimmerSystem.Glimmer / 1000f;
```
cmorley191: Let's just make super-sure we don't take sqrt of a negative

```suggestion
        float normalizedGlimmer = Math.Clamp(_glimmerSystem.Glimmer / 1000f, 0, 1);
```

Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
```diff
@@ -84,7 +84,10 @@ private void UpdateClientUI(EntityUid console, ArtifactAnalyzerComponent analyze
     // DeltaV
     private float GetGlimmerMultiplier(ArtifactAnalyzerComponent comp)
     {
-        return 1 + (MathF.Pow(_glimmerSystem.Glimmer / 1000f, 2f) * comp.PointGlimmerMultiplier);
+        float normalizedGlimmer = _glimmerSystem.Glimmer / 1000f;
+        //DeltaV - Prevents extreme glimmer multipliers
+        return (float)(.5f + Math.Clamp(Math.Pow(normalizedGlimmer,.5) + 1.5f * Math.Pow(normalizedGlimmer, 10f), 0f, 2f));
```
cmorley191:
```suggestion
        return (float)(.5f + Math.Clamp(Math.Pow(normalizedGlimmer,.5) + 1.5f * Math.Pow(normalizedGlimmer, 10f), 0f, 2.5f));
```

cmorley191:
Ok I'll go stress test this in game, then we'll be good to go.

cmorley191:
Sorry this was a lot of back and forth. Thanks for your work on this

SirSmith148:
> Sorry this was a lot of back and forth. Thanks for your work on this

No worries its fun haha

cmorley191 (approved):
No surprises from testing. Our next changeset/changelog is in 12 hours -- I'll merge it before then, but gonna leave this open for a while longer in case there are any comments.

ShepardToTheStars:
> in case there are any comments.

ship-it

PR #4996, opened 2025-12-17
Fix Duplicate Kill Heads Objectives and Add TargetObjectiveImmune

ShepardToTheStars:
## About the PR

Fixed TargetObjectiveImmune.

Also, traitors will get more variety of teach a lesson objectives (since before it was just choosing heads)

Dedicated to @PureBreadBagel 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed teach a lesson objectives from only choosing heads of staff.

### DIFF
--- a/Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
+++ b/Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
@@ -1,11 +1,7 @@
-using Content.Server._DV.Objectives.Components;
+using Content.Server._DV.Objectives.Components; // DeltaV
 using Content.Server.Objectives.Components;
 using Content.Shared.Mind;
 using Content.Shared.Objectives.Components;
-using Content.Shared.Roles; // DeltaV
-using Content.Server.GameTicking.Rules;
-using Robust.Shared.Prototypes; // DeltaV
-using Robust.Shared.Random;
 
 namespace Content.Server.Objectives.Systems;
 
@@ -96,3 +92,4 @@ public sealed class PickObjectiveTargetSystem : EntitySystem
         _target.SetTarget(ent, picked, target);
     }
 }
+
--- a/Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
+++ b/Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
@@ -28,14 +28,6 @@ public sealed class PickObjectiveTargetSystem : EntitySystem
 
     private void OnSpecificPersonAssigned(Entity<PickSpecificPersonComponent> ent, ref ObjectiveAssignedEvent args)
     {
-        // DeltaV - TargetObjectiveImmune
-        if (HasComp<TargetObjectiveImmuneComponent>(ent.Owner))
-        {
-            args.Cancelled = true;
-            return;
-        }
-        // END DeltaV
-
         // invalid objective prototype
         if (!TryComp<TargetObjectiveComponent>(ent.Owner, out var target))
         {
@@ -53,6 +45,14 @@ public sealed class PickObjectiveTargetSystem : EntitySystem
             return;
         }
 
+        // DeltaV - TargetObjectiveImmune
+        if (HasComp<TargetObjectiveImmuneComponent>(target.Target))
+        {
+            args.Cancelled = true;
+            return;
+        }
+        // END DeltaV
+
         var user = args.Mind.OwnedEntity.Value;
         if (!TryComp<TargetOverrideComponent>(user, out var targetComp) || targetComp.Target == null)
         {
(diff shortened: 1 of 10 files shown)


PureBreadBagel:
This is a peak.

cmorley191 (commented):
The objective/gamerule code seems to always lose me when I try to understand it but...

Isn't your current version of the code checking for `TargetObjectiveImmuneComponent` on the *objective entity*?

I would've expected you'd need to check for that component on the `target.Target`

like

```
        // target already assigned
        if (target.Target != null)
            return;

        // couldn't find a target :(
        if (_mind.PickFromPool(ent.Comp.Pool, ent.Comp.Filters, args.MindId) is not {} picked)
        {
            args.Cancelled = true;
            return;
        }

        // DV - TargetObjectiveImmune
        if (HasComp<TargetObjectiveImmuneComponent>(picked)) return;

        _target.SetTarget(ent, picked, target);
```

ShepardToTheStars:
@cmorley191 You're right. Oops. Good catch.

ShepardToTheStars:
Ready for re-review @cmorley191

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -1,13 +1,7 @@
...
 using Content.Server.Objectives.Components;
 using Content.Shared.Mind;
 using Content.Shared.Objectives.Components;
-using Content.Shared.Roles; // DeltaV
-using Content.Shared.Roles.Jobs; // DeltaV
-using Content.Server.GameTicking.Rules;
-using Content.Server.Revolutionary.Components;
-using Robust.Shared.Prototypes; // DeltaV
-using Robust.Shared.Random;
-using System.Linq;
```
Toby222, line 5: Removing some non-DeltaV stuff, maybe untouch?

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -19,10 +13,6 @@ public sealed class PickObjectiveTargetSystem : EntitySystem
 {
     [Dependency] private readonly TargetObjectiveSystem _target = default!;
     [Dependency] private readonly SharedMindSystem _mind = default!;
-    [Dependency] private readonly SharedRoleSystem _role = default!; // DeltaV
-    [Dependency] private readonly IPrototypeManager _proto = default!; // DeltaV
-    [Dependency] private readonly IRobustRandom _random = default!;
-    [Dependency] private readonly TraitorRuleSystem _traitorRule = default!;
```
Toby222, line 16: Removing some non-DeltaV stuff, maybe untouch?

ShepardToTheStars: A lot of this stuff was removed in an August clean-up PR. https://github.com/space-wizards/space-station-14/blob/438090a5054aabf23237ca37a40c282a7eda4243/Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs

They are unused dependencies.

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
ShepardToTheStars: The one thing that PR did not do was remove the unused namespaces, since including unused namespaces doesn't generate warnings. I just went ahead and did that.

If an upstream merge touches add/removes namespaces in the future here, the person merging can just take incoming changes anyways. One of the easier merge conflicts to resolve.

ShepardToTheStars: Also, adding in unused namespaces seems very dirty lol

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -51,6 +41,14 @@ private void OnSpecificPersonAssigned(Entity<PickSpecificPersonComponent> ent, r
             return;
         }
 
+        // DeltaV - TargetObjectiveImmune
+        if (HasComp<TargetObjectiveImmuneComponent>(target.Target))
```
cmorley191: I think this one might need to be lower, acting on targetComp.Target.Value

I think the point of this function is to take the specific person from the TargetOverrideComponent and put it onto the TargetObjectiveComponent. So the filter should be getting in between that process

cmorley191 (changes requested):

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
cmorley191: Do we even have TargetOverride objectives?

Does it even make sense for the Immune filter to take precedence over a "set this specific person" component like TargetOverride?

ShepardToTheStars: Ive never seen it used. 

Regardless, we never EVER want anyone with the target immune to be chosen. Not even with the target override

ShepardToTheStars: Either way, I switched it.

Resources/Prototypes/Objectives/traitor.yml
```diff
@@ -156,6 +158,7 @@
         components:
         - HelpProgressCondition
         - KeepAliveCondition
+        - TargetObjectiveImmune # DeltaV
```
FaintSpeaker: question: same as wizard, would this prevent getting the protect objective on someone who is targeted by a kill objective?

ShepardToTheStars:
@cmorley191 @Toby222 Ready for re-review by the way.

cmorley191 (approved):

PR #4998, opened 2025-12-17
Sort Accesses and Jukebox UIs

ShepardToTheStars:
## About the PR

I sorted the ID card computer's buttons, the jukebox's song list, and the access config's buttons.

## Why / Balance

Because its annoying when its not sorted.

## Technical details

Added order by methods and did some locale lookups and stuff.

## Media

image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The jukebox, ID card computer, and access config lists are now sorted.


Content.Client/Access/UI/AccessOverriderBoundUserInterface.cs
```diff
@@ -73,5 +74,18 @@ public void SubmitData(List<ProtoId<AccessLevelPrototype>> newAccessList)
         {
             SendMessage(new WriteToTargetAccessReaderIdMessage(newAccessList));
         }
+
+        /// <summary>
+        /// DeltaV - Used to get the localized access name for an access level. Used to sort the list.
+        /// </summary>
+        /// <param name="access">The access prototype to get the name of.</param>
+        /// <returns>The localized name of the access.</returns>
+        private string GetAccessLevelName(AccessLevelPrototype access)
```
cmorley191: Would probably be best to put this in `SharedAccessSystem` or something, instead of duplicating it three times.

cmorley191: \<optional_addition>

If you wanted to move down that ternary from up above and just make this function work on `ProtoId<AccessLevelPrototype>` I'd be all for it. But I'm cool with this approach too.

```
private string GetAccessLevelName(ProtoId<AccessLevelPrototype> accessId)
{
    if (!_prototypeManager.TryIndex<AccessLevelPrototype>(x.Id, out var access)
        return x.Id;

    if (access.Name is { } name)
        return Loc.GetString(name);

    return access.ID;
}
```

\</optional_addition>

cmorley191 (changes requested):

Content.Client/Access/UI/AccessOverriderBoundUserInterface.cs
ShepardToTheStars: The method is actually from `AccessLevelPrototype`'s class but there was some reason I couldn't use that when I first started this branch, but then I changed things a bit more and I could just go back to using the method in `AccessLevelPrototype`. Should be addressed now.

ShepardToTheStars:
Ready for re-review @cmorley191

cmorley191 (approved):

Toby222:
Wrong CL format

cmorley191:
🤦 I always forget to check the changelog, dammit

ShepardToTheStars:
Eh, not a huge changlog loss

Toby222:
> Eh, not a huge changlog loss

Too late I already manually fixed it :⁾

PR #5001, opened 2025-12-17
[K9] Give Laika and other secdogs Access + Tracking

BarryNorfolk:
## About the PR
Adds a Job, tracking, and ID access to Security dogs so they can wander on their own.
Access is limited to entities with a Mind, so should be only players (or disconnected ones at least) that have access. This is to limit the problem of having a dog that has general security access from being abused.

As a possible extension to this, we could make more department animals have mind only access to their departments if Laika's usage of this is generally successful.

## Why / Balance
Part of the K9 additions as outlined in the Design documents forum.
Dogs can be free, dogs should be tracked, dogs are cool.

## Technical details
Generally straightforward, had to make a new system for introducing sub-dermal ID cards, spawning them into the right container, and setting up the right full name on the ID card.

Mind only access is broad and works off of a marker entity for ease of use, otherwise there'd be a lot of work to just get the User entity into the right place.

## Media

image

https://github.com/user-attachments/assets/9f70930d-a78e-43ac-8728-3f87eee32d3a

And they even show up on door logs

image

ID Cards not retrievable via gibbing

image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Laika now has tracking built in!
- add: Security dogs purchased outside of Laika will now have names generated from the list of fellow Space dogs.
- add: Laika now has access to general security areas, this is limited to entities with minds (i.e. Players).


Vapetastic-Gaming:
Dir Review 24h, not sure if this has been reviewed or not

BarryNorfolk:
Not been reviewed directly but this is part of the K9 design doc that I think was approved a decent while ago, just been too busy to actually code the damned things until christmas break.

BarryNorfolk:
~Ah wait, I've just realised they are in the "Unknown" section and should probably be in Security. Dammit.~

Fixed.

Vapetastic-Gaming:
Approved, with some requests. Please remove external Access from the ID's. 

Possibly somewhere down the line, we would like these ID's to be extractable, either with security access like borgs have, or some other way.

BarryNorfolk:
Fair enough, removed the external access from the k9 job.
Will have a think about extractability in the future :)

BarryNorfolk:
There was a request from StopSigns to move this to being an intrinsic slot on the dogs rather than a subdermal ID.
I'll have to check how this works with the suit sensors but I imagine I can move some things around.

ShepardToTheStars:
This was discussed in the direction thread. I thought we landed on since the design document was approved with subdermal access, that we were going to go with that.

The issue with that is that sec dogs can no longer come with security access if its not subdermal.

ShepardToTheStars:
@Stop-Signs

Resources/Prototypes/_DV/Roles/play_time_trackers.yml
```diff
@@ -45,3 +45,6 @@
 
 - type: playTimeTracker
   id: RolePunPun
+
+- type: playTimeTracker
+  id: RoleK9Officer
```
cmorley191: Maybe it's a bit of a weird idea, but should we just make K9s use the RoleLaika play time tracker? There's no significant difference between Laika and other k9's is there?

Pros:
- avoids two separate role trackers depending on if you happen to play the k9 called Laika
- existing laika playtime before this PR is carried over for k9 situations

Cons:
- just weird that it's called RoleLaika, if you're playing a k9 other than laika

Maybe this is all a moot point since Laika/K9 playtime is not a requirement for anything? Is it?

Resources/Prototypes/_DV/Roles/Jobs/Security/k9.yml
cmorley191, line 1: This comment appears at the top of pets.yml:

```
# Ensure that all setPreference are set to false here. Otherwise, they appear on the job select screen.
# These are for ghost role play timers anyways
```

I think that's relevant here too. "K9 Officer" is appearing in the character customize -> Jobs screen right now. Unless you intended that? It's not a round start role ever is it?

Resources/Prototypes/_DV/Entities/Mobs/NPCs/animals.yml
```diff
@@ -68,9 +68,10 @@
 
 - type: entity
   parent: SimpleMobBase
-  id: MobSecDog
+  id: MobSecDogBase
   name: security dog
   description: Funnily enough, this pig is actually a dog.
+  categories: [ HideSpawnMenu ]
```
cmorley191: I think `abstract: true` would fit better?

Pretty sure that implicitly does hidespawnmenu as well as makes sure we don't spawn these in other ways

cmorley191:
> As a possible extension to this, we could make more department animals have mind only access to their departments if Laika's usage of this is generally successful.

Clippy needs this so much. Don't let me forget to implement this (unless you beat me to it!)

Resources/Prototypes/_DV/Entities/Objects/Misc/identification_cards.yml
```diff
@@ -287,3 +287,22 @@
...
+  id: K9OfficerIDCard
+  name: K9 officer ID card
+  components:
+  - type: Sprite
+    layers:
+    - state: default
+    - state: stripe_top
+      color: *color-justice-top
+    - state: stripe_bottom
+      color: *color-justice-bottom
```
cmorley191: Should this be security colors?

Content.Shared/_DV/Access/Components/SubdermalIdCardComponent.cs
```diff
@@ -0,0 +1,24 @@
+using Content.Shared._DV.Access.Systems;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared._DV.Access.Components;
+
+/// <summary>
+///
```
cmorley191:
```suggestion
/// A subdermal id card implant (entities with this component should also have SubdermalImplantComponent).
```

Content.Shared/_DV/Access/Components/SubdermalIdCardComponent.cs
```diff
@@ -0,0 +1,24 @@
...
+/// </summary>
+[RegisterComponent]
+[Access(typeof(SharedSubdermalIdCardSystem))]
+public sealed partial class SubdermalIdCardComponent : Component
+{
+    /// <summary>
+    /// The IDCard to spawn into the implant's container.
```
cmorley191:
```suggestion
    /// The IDCard to spawn into the implant's id card container.
```

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -213,6 +218,11 @@ public bool IsAllowed(EntityUid user, EntityUid target, AccessReaderComponent? r
         if (!_tag.HasTag(user, PreventAccessLoggingTag))
             LogAccess((target, reader), user);
 
+        // Begin DeltaV Additions - Mind only access
+        if (HasComp<MindOnlyAccessComponent>(user) && !_mindSystem.TryGetMind(user, out _, out _))
+            return false;
+        // End DeltaV Additions - Mind only access
```
cmorley191, line 221: Let's put this above the LogAccess call.

If we aren't logging `!IsAllowed(access, ....)` cases, we probably shouldn't log `not allowed due to missing mind` cases

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -852,6 +862,11 @@ public bool FindAccessItemsInventory(EntityUid uid, out HashSet<EntityUid> items
             items.Add(idUid.Value);
         }
 
+        // Begin DeltaV Additions - Subdermal ID chips
+        if (_subdermalId.TryGetIdCard(uid, out var idEntity))
+            items.Add(idEntity.Value);
+        // End DeltaV Additions - Subdermal ID chips
+
```
cmorley191, line 869: I wish these "find the id" functions/use cases were more generic so we don't have to change so many places, but alas...

I did a search of the codebase and there are three other places like this that I think you should consider too.

1. `AdminVerbSystem.Tools.cs` in the `FindActiveId` function. That function is used for the right click -> Tricks -> Grant All Access / Revoke All Accesses which seems like a good idea to make work on these dogs.
2. There's also the IdExaminableSystem, which you might also consider adding support to, but I don't know if subdermal id's *should* be examinable... that seems a bit weird.
3. NanoChatCartridgeSystem.... Laika is showing up as a potential NanoChat recipient which is uh... maybe fine but a little odd. I wouldn't worry too hard about this one. Sending her a message just shows "Failed to deliver"

There are more  places where id cards are searched for, but none that I think really matter for subdermals in their current state.

Content.Server/_DV/Access/Systems/SubdermalIdCardSystem.cs
```diff
@@ -0,0 +1,61 @@
...
+/// <summary>
+/// Server side handling subdermal ID cards.
+/// </summary>
+public sealed class SubdermalIdCardSystem : SharedSubdermalIdCardSystem
+{
+    [Dependency] private readonly ContainerSystem _container = default!;
+    [Dependency] private readonly IdCardSystem _cardSystem = default!;
+
+    private const string IDCardContainerName = "subdermalIdCard";
+    private const string ImplantContainerName = "implant";
```
cmorley191: Let's just use `ImplanterComponent.ImplantSlotId` instead of this variable

And maybe we should move the `subdermalIdCard` const into the component in a similar way? And out of this file and the Shared file

cmorley191 (approved):
you are absolutely killing it with these prs

cmorley191 (changes requested):
Whoops, I didn't mean to approve with the last review.

cmorley191:
I asked direction to officially clarify the confusion regarding "subdermal vs. stealable id".

Feel free to hold off making changes until they post that clarification.

Content.Shared/Access/Systems/AccessReaderSystem.cs
BarryNorfolk: Makes sense!

Resources/Prototypes/_DV/Entities/Objects/Misc/identification_cards.yml
BarryNorfolk: Yeah, must have C+P'd from the wrong set. Oops.

Resources/Prototypes/_DV/Entities/Mobs/NPCs/animals.yml
BarryNorfolk: Yeah, it does hide it. Swapped it out now :)

Resources/Prototypes/_DV/Roles/play_time_trackers.yml
BarryNorfolk: Well you require time in security to be Laika, I think? But they probably give security department time as you play them as well.
I don't have any particularly strong opinion on this since I think most players will be with Laika and not the secdog "general" one that requires a cargo purchase/admeme.

Resources/Prototypes/_DV/Roles/Jobs/Security/k9.yml
BarryNorfolk: Yeah, that's a mistake on my part. Fixed it now, and also added the departmental required time for you to play since that's in line with Laika.

Content.Shared/Access/Systems/AccessReaderSystem.cs
BarryNorfolk: Fixed both the VerbSystem, so access is added/removed now, and the NanoChatCartridge system so secdogs don't show up. I also don't think the subdermal IDs should be examinable right now as they shouldn't ever be present for a player to directly examine.
If we move to having removeable IDs for this then yeah we should look at it again.

Stop-Signs:
Direction is approving this under a few conditions. One, the precedent with these kinds of inherent access will be that functional roundstart ghostrole pets will be given these sorts of accesses. Including but not limited to laika, shiva, clippy and silvia. Two, the access granted to these inherent ID chips should be departmental + maintenance ONLY. If you have any questions feel free to let me know.

BarryNorfolk:
@Stop-Signs is the expectation to add those accesses in this PR or that another one can add and handle them?
@cmorley191 Pls to re-review when you has the time.

Content.Server/_DV/Access/Systems/SubdermalIdCardSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<SubdermalIdCardComponent, EntGotInsertedIntoContainerMessage>(OnImplantInserted);
+    }
+
+    /// <summary>
+    /// Handles when a subdermal ID impant is inserted into a container matching the correct ID.
```
Toby222:
```suggestion
    /// Handles when a subdermal ID implant is inserted into a container matching the correct ID.
```

Content.Shared/_DV/Access/Systems/SharedSubdermalIdCardSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Content.Shared._DV.Access.Systems;
+
+/// <summary>
+/// Shared handling for subdermal ID Cards.
+/// </summary>
+public abstract class SharedSubdermalIdCardSystem : EntitySystem
+{
+    [Dependency] private readonly SharedContainerSystem _container = default!;
+
+    public const string SubdermalContainerId = "subdermalIdCard";
```
Toby222, line 19: duplicate of SubdermanIdCardComponent.IDCardContainerName

Resources/Prototypes/_DV/Roles/play_time_trackers.yml
Toby222: Agreed that all secdogs should share one playtime tracker, but ultimately a CR decision

Toby222 (changes requested):

Toby222 (changes requested):
Also it won't let me comment on them, but there's some added files outside of _DV

Content.Shared/_DV/Access/Systems/SharedSubdermalIdCardSystem.cs
BarryNorfolk: Solved, decided on the Entity system owning the name since there's no way to know the name of the container THAT particular subdermalIDCard would be put into when getting the ID card.

BarryNorfolk:
Good catch on the file, I've rebased and pushed.

BarryNorfolk:
Rebased

sowelipililimute:
this is a lot of code for something that could be an ID card slot with all the usual in-world interactions that brings

rebe83:
> this is a lot of code for something that could be an ID card slot with all the usual in-world interactions that brings

You missed the like 1,000 comments of arguing in CR about this
For some reason there was a large ish camp of people saying "it's lame for sec access to be able to be stolen shift start because someone took it off the dog ghost role" to which I said it's secs fault for not taking the id off or losing their dog
There was also "the design doc got approved already and this was in it" which was also discussed as a faulty line of logic
But the compromise was "we can steal the access if we gib / butcher the animal" because a couple of us did insist the access needs to be stealable

SirWarock:
Yeah, the idea behind it is that Laika has sec access that cannot be taken away from her, because she is a ghostrole that will not always have a player actively defending themselves.

sowelipililimute (approved):
that seems easy to contend with by making laika aggro on someone trying to steal her ID, but the code is fine and i'll sign off on it

SirWarock:
> that seems easy to contend with by making laika aggro on someone trying to steal her ID, but the code is fine and i'll sign off on it

Honestly good idea.

PR #5007, opened 2025-12-20
[K9] Add new K9 shock jaws as a surgery option for Laika

BarryNorfolk:
## About the PR
Added a new augment type for Jaws with a K9 only set of jaws that enable Laika, and other sec dogs, to melee as though they had a stun baton in their mouths! 

## Why / Balance
Part of the K9 rework as discussed in the design-documents form. Laika being able to stun people ~to death~ is important for role playing.

## Technical details
Had to make some changes to surgery to block Laika's head from being a valid surgery target. Getting Laika's body parts to be removable, like a humanoids, might be a bit more tricky and annoying than I would like so it was simpler to just not let players remove the head.

Augment system got a little change to enable forwarding of events to augments that are installed on the entity in the case where the augment wants to take part in some kind of event logic. I think this enables some extra/new augments that could be made if people wanted to.

The BatteryProvider system is basically a more generic version of the Ninja suit and should probably be moved upstream so that more things can power other things. The Shock Jaws themselves are a more specialized version of the StunBaton, copying some of the logic from there.

## Media

https://github.com/user-attachments/assets/d9d98135-ef43-4d23-bdb6-a2b80155a306

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added a new K9 shock jaw augment for Sec dogs from RnD, ensure Laika's battery on her vest is fully charged and watch her stun the janitor for not petting her!


BarryNorfolk:
The numbers here are not final and I highly recommend direction to have an input on them.

BarryNorfolk:
Back to draft due to predicted batteries. Systems need a rework now.

BarryNorfolk:
Now that shitmed is closer and closer, I'm going to close this until such time as newmed or offmed or whatever it's called is available.

PR #5013, opened 2025-12-22
Change how Minor Mass Mind Swap Cooldown starts/Fix Telegnosis

Coryler:
## About the PR

Changed where the cooldown started... cause I'm somewhat baffled that until this time nobody complained about Telegnosis being borked.

## Why / Balance

Two minutes of being a ghost in Telegnosis is hardcore.

## Technical details

Remove the Cooldown start of delay and changes in the psionic.yml and move them as a manual start within the same method where as we do the swap.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Remove Telegnosis having a two minute cooldown to return

### DIFF
--- a/Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
+++ b/Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
@@ -11,6 +11,8 @@ using Robust.Server.Audio;
 using Robust.Shared.Audio;
 using Robust.Shared.Player;
 using Robust.Shared.Random;
+using Content.Shared.Actions;
+using Content.Shared._DV.Abilities.Psionics;
 
 namespace Content.Server._DV.StationEvents.Events;
 
@@ -22,6 +24,7 @@ internal sealed class MinorMassMindSwapRule : StationEventSystem<MinorMassMindSw
     [Dependency] private readonly AudioSystem _audio = default!;
     [Dependency] private readonly MindSwapPowerSystem _mindSwap = default!;
     [Dependency] private readonly ChatSystem _chat = default!;
+    [Dependency] private readonly SharedActionsSystem _action = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly MobStateSystem _mobstateSystem = default!;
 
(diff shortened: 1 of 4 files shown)


Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
```diff
@@ -97,6 +108,9 @@ private void SwapMinds(MinorMassMindSwapRuleComponent component)
 
             _mindSwap.Swap(target01, target02);
 
+            StartSwapMindsCooldown(target01);
+            StartSwapMindsCooldown(target02);
+
```
Toby222: Shouldn't this happen inside of `_mindSwap.Swap` instead?

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
```diff
@@ -73,6 +76,14 @@ public override void Update(float frameTime)
         }
     }
 
+    private void StartSwapMindsCooldown(EntityUid target)
+    {
+        // Cooldown Solution just for this.
+        var targetMindPowerComp = Comp<MindSwappedComponent>(target);
+
+        _action.SetCooldown(targetMindPowerComp.MindSwapReturnActionEntity, TimeSpan.FromSeconds(120));
```
Toby222: Hardcoded delay of 120 seconds is cringe, add it to `MinorMassMindSwapRuleComponent` instead

Toby222 (dismissed):

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: Swap itself doesn't have a parameter or something in which we actually start the cooldown. Once we swap, we get the Return Mind Swapped action, which is a 20 second cooldown which we can change in the yml and actually doesn't start with a cooldown.

I can add it in there though.

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: I've added a delay variable in component.

Content.Server/_DV/StationEvents/Events/MinorMassMindSwapRule.cs
Coryler: Anyway, I've moved that bit in there now.

Cepelinas1:
forgot to say but we started direction review like yesterday
should have results today

Cepelinas1:
approved

ShepardToTheStars (approved):

cmorley191:
for posterity: pretty sure this is a fairly recent bug, hasn't been like this forever.

I know for sure that there was no telegnosis-return cooldown in late September, since I have a video on this PR showing there was no cooldown: https://github.com/DeltaV-Station/Delta-v/pull/4393

PR #5014, opened 2025-12-22
Adjutant Rekitting

Halo3moth:
Adjutant changes

## About the PR

Changes the Adjutant to be magazine fed rather than tube fed lowers the fire rate
## Why / Balance

"Why change it at all?": Right now the adjutant is just a worse enforcer in every way smaller mag size worse dps and cost more from cargo (10k) there is no reason to use it over a enforcer other than that it can fit in a satchel I want to give it a identity that makes it relevant in balancing without putting it in a position where it mogs the existing shotguns (Kammerer Enforcer Bulldog)

"Why lower the dps": Because right now the enforcer has a dps equal with the bulldog in exchange for being tube fed I want this to stand out against the enforcer without completely overshadowing it, the amount of utility being mag fed gives you is insane. You can quickly switch between tranqs and lethals if your hos carrying it on green you don't have to spill beanbags everywhere to load in lethals it lets you give more consistent fire against swarm antags like the ratking and dragon. I want the enforcer to still be some peoples favorite gun after this change so I'm letting the stat difference between the 2 stay rather large (35%)

"How does this stack up to the bulldog?": In general nukie guns are suppose to be around 25-35% stronger than their security version this falls well within that range being 35% less dps (1.3 shots per second) and semi automatic compared to the bulldogs 2 shots per second full auto design (Arguable semi automatic vs full auto means very little when talking about shotguns bc you can just click alot) 

"Do the sprites look good?" I don't know I just work here
## Technical details

All changes made in _DV no need for noting mostly just copying over bulldog code because its basically the same gun removed the shotgun parent from the Adjutant because my debug hated it some sprites modified 1 new sprite added for the magazine
Mostly just YML
## Media

Saves me a bit of effort uploading it as a yt clip hope that doesn't bother yall? Ill leave pictures for those of us too lazy to browse.
https://youtu.be/RpkvrmrWCWU?si=p9UspEz0zW8ACUEl
image
image
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Nanotrasen arms is proud to present their re envisioning of the adjustant, now accepting drum magazines


Toby222 (dismissed):
lgtm

Stop-Signs:
Direction 24 hours
image

ShepardToTheStars:
that awkward moment when you've already merged this into the upstream merge lmao https://github.com/space-wizards/space-station-14/pull/40790

einknusprigestoast:
We do support the idea of giving the Adjutant drum magazines, as this adds a niche to a weapon that is objectively worse than any other shotgun in securities arsenal. But we would like the Adjutant to keep its 1.5 shots per second fire rate as lowering that even more puts this shotgun in a place where it is non-competitive to rifles and SMGs in closest quarters (which is bad for a shotgun)

ShepardToTheStars (approved):

PR #5017, opened 2025-12-23
Port Fentanyl

kotobdev:
## About the PR
Fentanyl has been ported from Monolith. In addition to being a potent and fast-processing narcotic, it can be crystallized, producing physical crystals which are worth decent cash when sold. However, improper crystallization may lead to rather explosive side effects.

Can't link the original PR since the person who made it got banned from GitHub and their PR history got entirely nuked

## Why / Balance
I want to see cargo techs jugmixing at the ATS to bring the station out of crippling debt. Or, well, chemists doing the same. More fun when the techs do it, though
Reverted Mono replacing the Oil requirement with Cogni. Let chem have their fun

## Technical details
yaml op

## Media
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Fentanyl is here. When crystallized, it's worth quite a bit of cash when sold.


apeiroeder:
direction reveiw 48h

MilonPL:
should probably work like the rest of painkillers, see `Delta-v/Resources/Prototypes/_CD/Reagents/medicine.yml` for reference

kotobdev:
> should probably work like the rest of painkillers, see `Delta-v/Resources/Prototypes/_CD/Reagents/medicine.yml` for reference

The CD painkillers only work because they have a **very** low metabolism rate (anywhere from 0.01-0.03u/s), while fentanyl processes at 1u/s. *Maybe* it could work if the `SuppressionTime` is set astronomically high? I'll have to poke at it

MilonPL:
> The CD painkillers only work because they have a **very** low metabolism rate (anywhere from 0.01-0.03u/s), while fentanyl processes at 1u/s. _Maybe_ it could work if the `SuppressionTime` is set astronomically high? I'll have to poke at it

If it metabolizes at 1u/s and you make the time like 200s then a 5u dose should last ~16 minutes which could make it overpowered but in the end chronic pain is an opt-in trait and it's not like this would become "meta", eh might want to make it a bit harder to get if anything
Or like somehow cap the effect at 10 minutes or something along those lines

kotobdev:
> If it metabolizes at 1u/s and you make the time like 200s then a 5u dose should last ~16 minutes which could make it overpowered but in the end chronic pain is an opt-in trait and it's not like this would become "meta", eh might want to make it a bit harder to get if anything Or like somehow cap the effect at 10 minutes or something along those lines

200s sounds good, although honestly there's not much reason somebody would bother med for fentanyl over just taking a roundstart 10u soretizone pill that lasts for 1000s. But regardless, fentanyl suppressing pain just makes sense, no harm in having that on there

Halo3moth:
Could we maybe rename it to be something that isn't literally fent people tend to get pretty Lrp with drugs and especially addictions

apeiroeder:
hi, direction here. apologies for the holdup.

Direction has decided to deny this PR in its current form. the reasons for this are mainly:
-being a drug often associated with drug abuse, addiction, and more IRL.
-its lack of identity and use cases outside of just being a drug

However, there are some interesting factors to this chem, so direction would approve this if this got a rebranding, to align more with deltaV. The following points were the most important ones brought up:

0: change its naming, and flavour text, to something different, ideally non-IRL represented
1: adding more practical usecases outside of just making money. examples of potential usecases which were brought up as ideas (although these were just example ideas) were:
-making it more efficient/better at healing damage/metabolize slower
-making the OD something like 4 or 5u, so you cannot just use a syringe to give it someone in pure form. considering the horrible OD effects, this would be something that has actual consequences ODing, unlike most meds that are just kinda.....harmless more or less, when ODd, and turn this into an interesting but difficult to dose medicine
-make it have long-lasting painkilling effects, so it can be used in surgery too. potentially also making it bad for the person receiving it, aka unconcious/slow/slurred speech/drunk, etc, the whole deal
-giving it a connection to psionic system
3: adjusting the recipe: the current recipe is rather cheap. ideas brought up would be to, for example, replace the oil with diphenylmethylamine, to make it a bit more difficult and requiring more effort to make.
4: the potential to explode during crystallisation was also seen as generally not great, due to the risk of ruining chemistry and thus meds shift due to the explosion spilling chems from the chemmaster/dispenser. although the need to make the crystals one by one is interesting, we decided that explosions seem a bit too disruptive for that. potential alternatives would possibly be cooling the remaining solution to prevent re-reaction, or similar things.

(as this is my first time returning direction feedback on PRs, i would apprechiate any feedback regarding potential improvement. I would also like to apologize for the late reply again.)

kotobdev:
Thanks for the feedback! If I do that, it'll be in another PR.

PR #4962, opened 2025-12-12
Upstream Merge (September and October 2025)

ShepardToTheStars:
#### September 2025
https://github.com/space-wizards/space-station-14/pull/40098 (Inflatable DeltaP)
https://github.com/space-wizards/space-station-14/pull/40053
https://github.com/space-wizards/space-station-14/pull/39897
https://github.com/space-wizards/space-station-14/pull/39886
https://github.com/space-wizards/space-station-14/pull/40099 (Also reverted later)
https://github.com/space-wizards/space-station-14/pull/37585 (Anomaly Scanner Updates)
https://github.com/space-wizards/space-station-14/pull/38928
https://github.com/space-wizards/space-station-14/pull/40096 (Commented out)
https://github.com/space-wizards/space-station-14/pull/40070
https://github.com/space-wizards/space-station-14/pull/40113 (Package win-arm64 and osx-arm64 servers)
https://github.com/space-wizards/space-station-14/pull/40118
https://github.com/space-wizards/space-station-14/pull/40097
https://github.com/space-wizards/space-station-14/pull/39525
https://github.com/space-wizards/space-station-14/pull/39774
https://github.com/space-wizards/space-station-14/pull/39975
https://github.com/space-wizards/space-station-14/pull/40125
https://github.com/space-wizards/space-station-14/pull/40126 (Shutter DeltaP)
https://github.com/space-wizards/space-station-14/pull/39544
https://github.com/space-wizards/space-station-14/pull/40129
https://github.com/space-wizards/space-station-14/pull/30706
https://github.com/space-wizards/space-station-14/pull/36894
https://github.com/space-wizards/space-station-14/pull/39612
https://github.com/space-wizards/space-station-14/pull/40137
https://github.com/space-wizards/space-station-14/pull/40141
https://github.com/space-wizards/space-station-14/pull/39932
https://github.com/space-wizards/space-station-14/pull/40133
https://github.com/space-wizards/space-station-14/pull/40147
https://github.com/space-wizards/space-station-14/pull/40151
https://github.com/space-wizards/space-station-14/pull/39773
https://github.com/space-wizards/space-station-14/pull/40041
https://github.com/space-wizards/space-station-14/pull/40158
https://github.com/space-wizards/space-station-14/pull/31213
https://github.com/space-wizards/space-station-14/pull/40174
https://github.com/space-wizards/space-station-14/pull/40192
https://github.com/space-wizards/space-station-14/pull/40171 (Only for the displacement map fixes)
[...]

PR #5026, opened 2025-12-24
Removed crawling over tables

pootslap:
## About the PR

Made it so crawling no longer lets you move over tables/desks/railings or under flaps.
## Why / Balance

Balance-wise this doesn't change much except making it slightly harder to have melee fights over tables/counters.
Actual reasons why I changed it:

- It always looked silly, doubly so now that everyone slides feet first over tables.
- There was already a feature for hopping tables. Alt-clicking on a surface lets you jump on top of it, this feature was around long before shitmed added crawling. They're around the same speed, alt-clicking is arguably faster than crawling in most cases.
- It's a bit of a moot point because most maps have windoors over the airtight flaps at logistics nowadays but being able to slide under the "airtight" flaps leads to a lot of cargo techs getting annoyed.
- Rodentia having the ability to sneak under tables is undermined by everyone being able to do the almost same thing. The only difference is rodentia go under the table instead of over which lets them hide underneath.

## Technical details

Commented out the section in StandingStateSystem.cs that changes layers when crawling. 
## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: Crawling no longer lets you slide over surfaces.

### DIFF
--- a/Content.Shared/Standing/StandingStateSystem.cs
+++ b/Content.Shared/Standing/StandingStateSystem.cs
@@ -123,7 +123,7 @@ public sealed class StandingStateSystem : EntitySystem
         // Seemed like the best place to put it
         _appearance.SetData(uid, RotationVisuals.RotationState, RotationState.Horizontal, appearance);
         // Change collision masks to allow going under certain entities like flaps and tables
-        ChangeLayers((uid, standingState));
+        // ChangeLayers((uid, standingState));
 
         // check if component was just added or streamed to client
         // if true, no need to play sound - mob was down before player could seen that


kotobdev:
> It always looked silly, doubly so now that everyone slides feet first over tables.

Yeah, that is true, I've wanted to change it around so sprite rotations function the same way they did under shitmed crawling, but it would be a bit of a nightmare to code from what I saw

> There was already a feature for hopping tables. Alt-clicking on a surface lets you jump on top of it, this feature was around long before shitmed added crawling. They're around the same speed, alt-clicking is arguably faster than crawling in most cases.

Worth noting that our crawling is now upstream crawling, and upstream crawling works the same way- that is, you can crawl over things. It's not a shitmed crawling specific thing

> It's a bit of a moot point because most maps have windoors over the airtight flaps at logistics nowadays but being able to slide under the "airtight" flaps leads to a lot of cargo techs getting annoyed.

Being able to slide under airtight flaps is a LIFESAVER for navigating cargo, especially when going between rooms connected by conveyor belts

> Rodentia having the ability to sneak under tables is undermined by everyone being able to do the almost same thing. The only difference is rodentia go under the table instead of over which lets them hide underneath.

Being able to go underneath tables both makes you extremely difficult to spot, and practically impossible to hit. It's still a massive buff for Rodentia

Halo3moth:
Anti tider propaganda, I'm a big fan of our current state of crawling it feels dynamic without punishing gunplay (other than stealth suit craw which kinda sucks and the gamba can't hit crawling players)

kotobdev:
> ...the gamba can't hit crawling players) 

should be fixed now? Last upstream merge fixed the evil scope/recoil sprite clicking issues

ShepardToTheStars:
This is partially being addressed in the upstream merge. #4962 

See https://github.com/space-wizards/space-station-14/pull/40099

Woljif:
Remove glass table stunning you upon vaulting then

ShepardToTheStars:
Going to close this one as a maintainer of the primary reasons - crawling over counters - is being addressed in the upstream merge. If airtight flaps are still an issue, we can add properties to them instead of removing the fixture mask changes in standing state (which may cause other issues as well).

PR #5060, opened 2025-12-30
Fix DV Warp Points

IamVelcroboy:
## About the PR
Parents our WPs to use proper blacklists for wizard teleporting and bomb targets

## Why / Balance
Fix

## Technical details
n/a

## Media
n/a

## Requirements

## Breaking changes
n/a

**Changelog**
n/a

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Markers/warp_point.yml
+++ b/Resources/Prototypes/_DV/Entities/Markers/warp_point.yml
@@ -1,7 +1,16 @@
+- type: entity
+  id: WarpPointNoBomb
+  parent: WarpPoint
+  suffix: No Bomb Target
+  components:
+  - type: Tag
+    tags:
+    - NinjaBombingTargetBlocker
+
 # Off station
 - type: entity
   id: WarpPointEvacShuttle
-  parent: WarpPoint
+  parent: WarpPointNoBomb
   suffix: Evac Shuttle
   components:
   - type: WarpPoint
@@ -9,7 +18,7 @@
 
 - type: entity
   id: WarpPointShuttle
-  parent: WarpPoint
+  parent: GhostWarpPoint
   suffix: Shuttle
   components:
   - type: WarpPoint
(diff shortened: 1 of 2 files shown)


ShepardToTheStars:
Why are epistemics and medical no bomb targets?

IamVelcroboy:
> Why are epistemics and medical no bomb targets?

It's always been our mapping standard to leave them alone when mapping bomb targets. Epi blows themselves up enough and medical is already one of the most frequent targets.

Resources/Prototypes/_DV/Entities/Markers/warp_point.yml
```diff
@@ -1,31 +1,40 @@
...
+  components:
+  - type: Tag
+    tags:
+    - NinjaBombingTargetBlocker
+
 # Off station
 - type: entity
   id: WarpPointEvacShuttle
-  parent: WarpPoint
+  parent: WarpPointNoBomb
```
ShepardToTheStars: Should this be `GhostWarpPoint`? It is technically off-station as well.

ShepardToTheStars (commented):
Kind of odd that that's a mapping standard but alright. Approved after addressing this.

Resources/Prototypes/_DV/Entities/Markers/warp_point.yml
IamVelcroboy: That was my question in Discord, whether we should let the wizard warp to the evac shuttle or not? I never really got a clear answer, so I went with yes

ShepardToTheStars: Oh, I suppose I didn't understand exactly when you asked. I would change it because otherwise wizards can just teleport to the Evac Shuttle before its called, which might be near Midpoint or CC.

image

I'm guessing it would be an admin nightmare if they got to CC with the ability to just open any door.

IamVelcroboy: Roger that, is changed

ShepardToTheStars (approved):

PR #5081, opened 2025-12-30
Advanced plate carrier nerf

Halo3moth:
## About the PR
Moves the APCs off values to be a bit less exessive

## Why / Balance
Right now as hos your armor choices are a coat that does 25% across the board a trench coat that does 35-40% across the board and a plate carrier that makes you 15% slower for 40% across the board and a 60% in pierce resistance which is madness that's 25% faster than the warden hardsuit for nearly the same values. Plate carriers are allowed to be strong against ballistics because they are weak to literally everything else and are slow. I dont think that 5% slowness compared to a normal plate carrier is enough of a nerf for 15% more armor across the board. If we want hos to be a counterable force they should probably be weak to their counters (A energy sword, emagged lathe's energy weapons) 

P.S.: keep in mind that this nerf will effect both warden and ERT as they both spawn with a advanced plate carrier

Hos armor options below to compare it to
## Technical details
3 lines of yml inside _DV changes removed a note bc it was both redundant and factually incorrect

## Media
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Advanced plate carrier's heat, slash, and blunt resistance nerfed from 40% to 30%. Shock nerfed from 30% to 20%

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Clothing/OuterClothing/vests.yml
+++ b/Resources/Prototypes/_DV/Entities/Clothing/OuterClothing/vests.yml
@@ -62,11 +62,11 @@
   - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.75
-        Slash: 0.75
+        Blunt: 0.7
+        Slash: 0.7
         Piercing: 0.4
         Shock: 0.8
-        Heat: 0.75
+        Heat: 0.7
   - type: StaminaResistance
     damageCoefficient: 0.6
   - type: ClothingSpeedModifier
--- a/Resources/Prototypes/_DV/Entities/Clothing/OuterClothing/vests.yml
+++ b/Resources/Prototypes/_DV/Entities/Clothing/OuterClothing/vests.yml
@@ -59,14 +59,14 @@
     sprite: _DV/Clothing/OuterClothing/Vests/advcarrier.rsi
   - type: Clothing
     sprite: _DV/Clothing/OuterClothing/Vests/advcarrier.rsi
-  - type: Armor # intended as Head of Security's and Warden's armour only, hence high values; resists at 40% flat with 60% pierce and 30% shock.
+  - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.6
-        Slash: 0.6
+        Blunt: 0.75
+        Slash: 0.75
         Piercing: 0.4
-        Shock: 0.7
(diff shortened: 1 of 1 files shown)


Cepelinas1:
direction review 24h

SirWarock:
This would nerf their blunt/slash protection below what it once was. 
I do not like this change. I don't mind nerfing the fire resistance on it though, as that is uncharacteristic of NT armor.

Halo3moth:
> This would nerf their blunt/slash protection below what it once was. 

If we are using the normal plate carrier as a base it should be good at its 1 thing and meh and anything else, I could understand wanting to keep 1 of the resistances or moving the blunt and slash resistance back to what it was but I don't believe it makes sense for the plate carrier designed for ballistic combat to so throughly outclass T1 armor and durathread vest especially with heat and slash damage.

If someone were to make the horrible mistake of buying a rigged laser gun they should be able to drop John plate carrier in 1 magazine reliably

Woljif:
Armored trenchcoat stays winning 0% slowdown and better stats than everything

Halo3moth:
Right now an entire mag of a c20 assuming by some miracle you hit every bullet on hos does 148 damage an energy sword does 15ishhh a hit depending on what mask they have I'm fine with ballistics sucking because like don't throw grenades at the mg that's not new but if ballistics suck and heat sucks and melee sucks how is a traitor suppose to interact with hos other than hitting them with a hypo pen and minibombing them.

SirWarock:
The traitor can just gun them down. Their webvest outclasses the APC by 10% pierce and 17% more speed, in addition to their weapons being better than the HoS' arsenal, minus superweapons. Again, I don't mind nerfing the heat resistance.
But it's practically meant to be a Durathread and Plate Carrier combined. Hence there is no Advanced Durathread vest.

If you believe 60% pierce resistance to be too high, then I'd suggest reducing it to 50%, and keeping the blunt/slash resist on 40%, as well as reducing the speeddebuff from 15% to 10%.

Cepelinas1:
direction wants to make it:
blunt: 30%
slash: 30%
piercing: 60%
shock: 20%
heat: 30%

the PC carrier should generally be designated against long range targets, so should resist the long range weapon damage better

cmorley191 (approved):

cmorley191:
I'm updating your changelog message since the numbers have changed

PR #5101, opened 2026-01-01
Adv Truncheon Stamina Damage Tweak. Jug Suit Melee Stamina Damage Resistance.

ShepardToTheStars:
## About the PR

* Reduced blunt stamina damage of advanced from 1.6x to 1.2x.
* Allowed juggernaut suit to actually live up to its name by being very melee stamina damage resistant.

## Why / Balance

Currently, the advanced truncheon does 36 stamina damage normally, 48 while wielded, and if you're an Oni, it does a whopping 64.8 stamina damage per hit.

With the new modifier, it would do 27 stamina damage per hit, 36 while wielded, and if you're an Oni, it would do 48.6 when wielded by an Oni.

Also, just gave the Jugg suit insane melee damage resistance because the literal definition of juggernaut is "someone or something overwhelmingly powerful, dominant, and **unstoppable**", so melee stamina damage shouldn't affect them very much, if at all.

## Requirements

## Breaking changes

**Changelog**

<!--
:cl:
- tweak: The advanced truncheon's stamina damage modified has been reduced from 1.6x to 1.2x.
- tweak: The juggernaut suit now lives up to its name and actually laughs in the face of melee stamina damage.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -782,6 +782,7 @@
     damageCoefficient: 0.3
   - type: StaminaResistance # DeltaV
     damageCoefficient: 0.1
+    meleeResistance: true
   - type: Armor
     modifiers:
       coefficients:
--- a/Resources/Prototypes/_DV/Entities/Objects/Weapons/Melee/advanced_truncheon.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Weapons/Melee/advanced_truncheon.yml
@@ -1,7 +1,7 @@
 - type: entity
   parent: Truncheon
   id: AdvancedTruncheon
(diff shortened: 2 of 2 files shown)


SirWarock:
Truth be told, I am a OniTruncheon enjoyer, so this is biased.
But I dislike both changes.
The normal truncheon has a 1.5x stamina damage modifier, the advanced one shouldn't be nerfed under that.
Secondly, the Juggernaut suit shouldn't be immune to melee stamina damage.
There's two ways to deal stamina damage - Through Blunt Damage, and through weapons that deal specifically stamina damage (baton).
Due to the juggernaut suit having 80% blunt resistance, the stamina damage dealt by blunt damage is then reduced by 80%.
The resulting stamina damage is then counted as AP to not be reduced further by the additional 80% melee stamina damage resistance, which would be a 96% melee stamina damage resistance. It would be genuinely better to just baton the Jugsuit.

PR #5103, opened 2026-01-01
hotfix id card console

### LINKED ISSUE #5100: ID computer bug
## Description
ID computers are now unable to add or subtract access
**Reproduction**
Try changing access to ID's on a ID computer

**Screenshots**

**Additional context**
something something upstream merge

cmorley191:
## About the PR

Fixes https://github.com/DeltaV-Station/Delta-v/issues/5100

by allowing the ID Card Console to change the following accesses:
- GenpopEnter, GenpopLeave
- Emergency Shuttle Repeal All (which is used to reset the early launch authorizations, requiring heads to re-authorize. This is part of all-access -- the captain already has this access)

These three accesses were part of the "AllAccess" preset, but not changeable in the console. Due to quirks in the console, this prevented the console from working when trying to change *other* accesses.

I also added some accesses to the Admin ID Card Console's list, since they were missing.

## Technical details

This is a hotfix; a better fix would make it so the ID Card Console can change accesses even if the target card has access that the console isn't supposed to know about.

For example, the Captain cannot grant accesses to the Wizard at this time. That's still broken; requires a better fix.

Also, we really should just make the ID Card Console's access list pull from the existing AllAccess list. And/or we should add some unit tests that check if there are accesses in the codebase somewhere not covered in the admin id card console list.

## Media

HoP id card can now remove/add Bar access from the Captain id card:

(that was an example of an access change that was broken due to the bug)

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed some id cards that were unable to be changed in the ID Card Console

### DIFF
--- a/Content.Shared/Access/Components/IdCardConsoleComponent.cs
+++ b/Content.Shared/Access/Components/IdCardConsoleComponent.cs
@@ -65,8 +65,8 @@ public sealed partial class IdCardConsoleComponent : Component
         "EmergencyShuttleRepealAll", // DeltaV - fix mismatch with Access/misc.yml
         "Engineering",
         "External",
-        "GenpopEnter", // DeltaV - fix mismatch with Access/misc.yml
-        "GenpopLeave", // DeltaV - fix mismatch with Access/misc.yml
+        "GenpopEnter",
+        "GenpopLeave",
         "HeadOfPersonnel",
         "HeadOfSecurity",
         "Hydroponics",
--- a/Content.Shared/Access/Components/IdCardConsoleComponent.cs
+++ b/Content.Shared/Access/Components/IdCardConsoleComponent.cs
@@ -43,6 +43,9 @@ public sealed partial class IdCardConsoleComponent : Component
     [DataField, AutoNetworkedField]
     public List<ProtoId<AccessLevelPrototype>> AccessLevels = new()
     {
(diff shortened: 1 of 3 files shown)


Content.Shared/Access/Components/IdCardConsoleComponent.cs
```diff
@@ -59,8 +62,11 @@ public WriteToTargetIdMessage(string fullName, string jobTitle, List<ProtoId<Acc
         "Corpsman", // DeltaV - Add Corpsman access
         "Command",
         "Cryogenics",
+        "EmergencyShuttleRepealAll", // DeltaV - fix mismatch with Access/misc.yml
         "Engineering",
         "External",
+        "GenpopEnter", // DeltaV - fix mismatch with Access/misc.yml
+        "GenpopLeave", // DeltaV - fix mismatch with Access/misc.yml
```
ShepardToTheStars: Doesn't need DV comments since its like this in [upstream](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Access/Components/IdCardConsoleComponent.cs). We must have skipped it when we didn't add GenPop right away? idk.

```suggestion
        "GenpopEnter",
        "GenpopLeave",
```

ShepardToTheStars (changes requested):

ShepardToTheStars (approved):

PR #5104, opened 2026-01-02
Spare ID always unlocks during War Ops

ShepardToTheStars:
## About the PR

If War Ops is declared, the spare ID will unlocked 15s after war is declared. (Time can be configured to whatever).

If there is a captain, the spare ID will gain Captain access only. If there is no captain, the spare ID will gain Command access (like normal).

## Why / Balance

This allows a few things:
* If there is no captain, the heads of staff can immediately sort that out without having hope there's an admin online to send them a biscuit. WarOps delay is only 15 minutes so, waiting until the 20 minute mark isn't really feasible.
* If there is a captain, this allows them to secure the spare ID in a better spot.

If you think about it, its almost better to not have a captain and get the spare ID unlocked (or get a thief to emag the spare ID) so that the nukies can't just walk into the bridge and get free AA. This just fixes that discrepancy. Now the captain is tasked with securing access as well.

## Technical details

Just hooked into the WarOpsDeclaration event.

## Media

https://github.com/user-attachments/assets/fdaf4721-bd70-4653-b106-dcde66335d0b

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The spare ID will always unlock 15 seconds after nukies declare war, even if there is a captain.


ShepardToTheStars:
Direction approved.

Toby222 (approved):

PR #4984, opened 2025-12-14
Port Starlight's expanded makeshift items

Apachit0:
## About the PR

Ported over from starlight: https://github.com/ss14Starlight/space-station-14/pull/939, https://github.com/ss14Starlight/space-station-14/pull/1232, https://github.com/ss14Starlight/space-station-14/pull/1170
## Why / Balance

DV can benefit from a gameplay perspective from a makeshift item expansion. This also gave the modular receiver more prevalence since you were only able to make one weapon with it. Now you can craft 3 tiers of wepondry which are makeshift then improvised and finally forged. The makeshift and improvised have a chance to malfunction and explode in your hands. The forged ones don't since they are "better" made but actual guns are still better. This port also includes more craftable armor/shields and swords. Which also have tiers to them. These items are also too cheap to craft, so don't expect people to be pumping out guns. 
## Technical details

Changed the bullets to work with DV since Starlight used hitscan one.
Removed Starlight contraband type for every item since DV does not use it.
Removed the pierceable component from the armor cause DV does not use it.
Removed the reflect chance on the Forged Sword cause it's just a normal sword and made the damage the same as the plasteel arming sword.

## Media

Tools
swords
guns
armor

These are all fully assembled you need to weld the parts together to finish them
## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

🆑 
- add: Ported Starlight's Improvised Crafting of Various Things

PR #5105, opened 2026-01-02
Fix starlight prototype directory case

IRSMsoso:
## About the PR

Move `/Resources/Prototypes/_StarLight` into `/Resources/Prototypes/_Starlight`.

## Why / Balance

#4984 Introduced a secondly cased _Starlight folder in `/Resources/Prototypes`. There's now `_Starlight` as before, and `_StarLight` in addition. Mac and Windows have case insensitive file systems, so I shifted the new files from that PR into the old `_Starlight` directory. An alternative is moving everything from the old `_Starlight` directory to the new `_StarLight` directory. This could be preferable, as it is the casing that starlight uses on their own fork (I'm assuming that's the reason the PR ended up using that case).

## Requirements

## Breaking changes

**Changelog**


ShepardToTheStars:
Get speedmerged.

ShepardToTheStars (approved):

PR #5111, opened 2026-01-03
footprints prediction

MilonPL:
## About the PR

Reworks how footprints work, they are now in shared and use predicted spawning! Removes the timed despawn, but you now have a limit of 2 footprints per tile and a total of 1000 per grid, after which the oldest one will start to despawn. 

Also removes the slowdown on footprints because it really just doesn't make sense? And causes rubberbanding which is Not Great.

Issue: the footprints start with a slightly different color when spawned on the client, probably the Alpha channel is wrong, best visible with blood. It takes someone smarter than I to fix that, so help is appreciated.

## Why / Balance

prediction = good
rubberbanding = bad
and I also want to see the medbay get even more messy

## Technical details

pretty well document and self-explanatory I hope

## Media

https://github.com/user-attachments/assets/5601b8ce-b215-43d8-bcea-5111122a5fdd

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Footprints will no longer despawn after 60 seconds.
- tweak: Footprints are now limited to two per tile.


Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,328 @@
...
+        // Add slight random variation to step size for more natural-looking prints
+        ent.Comp.StepSize = Math.Max(0f, ent.Comp.StepSize + _random.NextFloat(-0.05f, 0.05f));
+        DirtyField(ent.AsNullable(), nameof(FootPrintsComponent.StepSize));
+    }
+
+    private void OnMove(Entity<FootPrintsComponent> ent, ref MoveEvent args)
+    {
+        // Don't create footprints if flying
+        if (_flightQuery.TryComp(ent, out var flight) && _flight.IsFlying((ent, flight)))
+            return;
```
ShepardToTheStars, line 80: Is the query necessary since `_flight.IsFlying` will resolve the component itself? Is it just more efficient?

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,328 @@
...
+
+        if (!_transformQuery.TryComp(ent, out var transform))
+            return;
+
+        if (!_mobThresholdQuery.TryComp(ent, out var mobThresholds))
+            return;
+
+        // Check if entity is being dragged (critical or dead)
+        var isCrit = mobThresholds.CurrentThresholdState is MobState.Critical or MobState.Dead;
+        var dragging = isCrit || _standingQuery.TryComp(ent, out var standingComp) && _standing.IsDown((ent, standingComp));
```
ShepardToTheStars, line 94: If someone is not critical/dead AND either crawling or being dragged by another entity, they should leave a mark according to this, right?

If that's true, do you even need to check for `isCrit`? Critical people will likely be dragged by someone else and will also not be in a standing state.

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,328 @@
...
+            ref footPrintComponent.Solution,
+                out var solution))
+            return;
+
+        // Don't overfill
+        if (solution.Volume >= 1)
+            return;
+
+        // Transfer a small amount of reagent
+        _solution.TryAddReagent(footPrintComponent.Solution.Value, reagentId, 0.01, out _);
```
ShepardToTheStars: Move the transfer amount (0.01) to the FootPrintComponent.

Content.Shared/_EE/Footprint/FootPrintsComponent.cs
```diff
@@ -32,57 +43,77 @@ public sealed partial class FootPrintsComponent : Component
...
     /// </summary>
     [DataField]
     public float ColorReduceAlpha = 0.1f;
 
-    [DataField]
+    /// <summary>
+    /// The reagent ID to transfer to footprint entities, set when stepping in puddles.
+    /// </summary>
+    [DataField, AutoNetworkedField]
     public string? ReagentToTransfer;
```
ShepardToTheStars: Should this be a `ProtoId<ReagentPrototype>`?

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,328 @@
...
+                {
+                    gridFootPrints.DirtyTiles.Add(tile);
+                }
+
+                return;
+            }
+        }
+    }
+
+    private void TryTransferReagent(EntityUid entity, FootPrintComponent footPrintComponent, string reagentId)
```
ShepardToTheStars: nit: Since it looks like entity owns the `FootPrintComponent`, consider:

```C#
private void TryTransferReagent(Entity<FootPrintComponent> ent, string reagentId) // Or ProtoId<ReagentPrototype> if you change the component to that ofc
```

Content.Shared/_EE/Footprint/Systems/PuddleFootPrintsSystem.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            if (colorValue is Color color && volumeValue is float volume)
+            {
+                AddColor(color, volume * ent.Comp.SizeRatio, footPrints);
+                Dirty(tripper, footPrints);
+            }
+        }
+
+        // Remove small amount of reagent from puddle
+        _solutionContainer.RemoveEachReagent(puddleComp.Solution.Value, 0.01);
```
ShepardToTheStars: Kind of a repeat of another comment, but move the transfer amount to the FootPrintsComponent. 

(Since you have the `FootPrintsComponent` as `footPrints`, and you've moved the transfer amount to that component, you should easily be able to get the transfer amount.)

Content.Shared/_EE/Footprint/Systems/PuddleFootPrintsSystem.cs
```diff
@@ -0,0 +1,119 @@
...
+            return;
+
+        // Calculate total solution quantity and water percentage
+        var totalSolutionQuantity = solution.Contents.Sum(sol => (float)sol.Quantity);
+
+        if (totalSolutionQuantity <= 0 || solution.Contents.Count <= 0)
+            return;
+
+        var waterQuantity = solution.Contents
+            .Where(sol => sol.Reagent.Prototype == "Water")
```
ShepardToTheStars: something something `ProtoId<ReagentProtoype>`

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
MilonPL: Resolve is using regular trycomp and this is a hot path

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
MilonPL: Not all mobs would have standing states though probably

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
ShepardToTheStars: Not necessarily in this PR, but is there anything we can do to encapsulate flying logic to the FlightSystem instead of using FlightComponent queries within non-flight related systems?

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
ShepardToTheStars: Touche.

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
MilonPL: you can just put the query on the API system like tag or physics do then it's safer in hot paths

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
ShepardToTheStars: I guess its way more clear that its transferring between two entities, but oh god, I guess I mistook `FootPrintComponent` for `FootPrintsComponent`. I still think its a bit more clear and better now.

ShepardToTheStars (approved):

Content.Shared/_EE/Footprint/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,330 @@
...
+
+        if (!TryComp<FootPrintComponent>(entity, out var footPrintComponent))
+            return;
+
+        // Set the owner reference
+        footPrintComponent.PrintOwner = ent;
+        Dirty(entity, footPrintComponent);
+
+        // Track the footprint on the grid
+        TrackFootPrint(gridUid, GetNetEntity(ent), tilePos);
```
ShepardToTheStars, line 159: So, uh, this should probably be `entity`, not `ent`, right? @MilonPL 

Otherwise, it adds the player and not the footprint to `FootPrintsByTile` in `GridFootPrintsComponent`

PR #5112, opened 2026-01-03
Fix debug assert in OnUpdateAlert of Cosmic Cult System

IRSMsoso:
## About the PR

Fix a debug assert that occurs when making someone a cosmic cultist in debug configuration.

## Technical details

Whenever I antag verbed myself into a cosmic cultist the client crashed. I tracked this down to `Content.Client/_DV/CosmicCult/CosmicCultSystem.cs` line 80 and commit `3019d8201c640585ba7e15ebcf8076e8c83243a4` introduced it.

The assertion is `Unhandled exception. Robust.Shared.Utility.DebugAssertException: Entity 11 is not the owner of the component. Component: SpriteComponent`.

To fix, instead of using the cosmic cult entity, I switched to using the actual sprite entity. I think this is the intended API usage.

## Requirements

## Breaking changes

**Changelog**

None because this is developer facing? (only happens with debug assertions enabled)


ShepardToTheStars (approved):
Thanks for the fix. I had tracked it down to this spot as well, so I know this is the correct spot. Just had no had time to submit the fix. :)

PR #5122, opened 2026-01-04
tweak: nanotask papers don't provide throwing pushback

DisposableCrewmember42:
## About the PR
Stop NanoTask printouts from providing pushback to the thrower when thrown.

## Why / Balance
Being able to infinitely print paper that you can use to traverse space seems unintentional.

## Technical details
Added NoThrowingPushbackComponent, which is checked for in the HandsSystem to set the pushbackRatio to 0.

## Media

https://github.com/user-attachments/assets/68e2c6d9-8946-4414-8516-2c2c94963e17

## Requirements

**Changelog**

:cl: DisposableCrewmember42
- tweak: NanoTask cartridges were the target of budget cuts. Task printouts are now too lightweight to provide throwing pushback in space.

### DIFF
--- a/Content.Server/Hands/Systems/HandsSystem.cs
+++ b/Content.Server/Hands/Systems/HandsSystem.cs
@@ -1,6 +1,7 @@
 using System.Numerics;
 using Content.Server.Stack;
 using Content.Server.Stunnable;
+using Content.Shared._DV.Item;
 using Content.Shared._NF.Standing; // Frontier
 using Content.Shared.ActionBlocker;
 using Content.Shared.Body.Part;
@@ -224,7 +225,8 @@ namespace Content.Server.Hands.Systems
             if (IsHolding((player, hands), throwEnt, out _) && !TryDrop(player, throwEnt.Value))
                 return false;
 
-            _throwingSystem.TryThrow(ev.ItemUid, ev.Direction, ev.ThrowSpeed, ev.PlayerUid, compensateFriction: !HasComp<LandAtCursorComponent>(ev.ItemUid));
+            // DeltaV - Set pushback to 0 if NoThrowingPushback, previously unset
+            _throwingSystem.TryThrow(ev.ItemUid, ev.Direction, ev.ThrowSpeed, ev.PlayerUid, pushbackRatio: HasComp<Shared._DV.Item.Components.NoThrowingPushbackComponent>(ev.ItemUid) ? 0 : ThrowingSystem.PushbackDefault, compensateFriction: !HasComp<LandAtCursorComponent>(ev.ItemUid));
 
             return true;
         }
--- a/Content.Shared/_DV/Item/Components/NoThrowingPushbackComponent.cs
+++ b/Content.Shared/_DV/Item/Components/NoThrowingPushbackComponent.cs
@@ -1,0 +1,4 @@
+namespace Content.Shared._DV.Item.Components;
+
+[RegisterComponent]
+public sealed partial class NoThrowingPushbackComponent : Component;
--- a/Resources/Prototypes/Entities/Objects/Misc/paper.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/paper.yml
@@ -236,6 +236,7 @@
     contentMargin: 12.0, 0.0, 12.0, 0.0
     # This is a narrow piece of paper
     maxWritableArea: 256.0, 0.0
+  - type: NoThrowingPushback # DeltaV - Added comp
 
 - type: entity
   id: PaperCargoBountyManifest


jimmy12or:
Every other thrown thing does this. Even if negligible, it's true to physics. I understand that it's infinite and that alone is illogical, but you're not here to limit how much paper can be printed. There are already other limitless propulsion options such as the PKA, bluespace fire extinguisher, or the cash stashed away in a cargo-tech's pockets. I don't personally see this as an issue, but i'm not a maintainer. I see this as a reasonable emergency way to get back to the station in a panic. I hadn't even thought of this route before to be honest. I often forget about nanotask.

MilonPL:
> Every other thrown thing does this. Even if negligible, it's true to physics. 

Yes, conservation of momentum is annoyingly consistent like that, but the issue here is scale.

The average A6 piece of paper is roughly 1.2 grams. NanoTask printouts would probably be even smaller at a business card size and printed on thermal paper. For a person weighing 80kg (probably average spessman w/ EVA), it would take 3.6 hours to traverse one meter.

This really does not make sense to keep. If you really wanna stay true to physics, don't give PDAs a way to print unlimited paper, and make those printouts have basically zero mass.

cmorley191:
I'd argue we should remove all the other infinite traversal ways mentioned, too. This isn't about realism; it's a gameplay problem. Players aren't supposed to have infinite space traversal. Space is supposed to be scary.

Every crew member is _not_ supposed to have an emergency way to get back to the station in a panic.

The exception is bluespace fire extinguisher -- it's balanced by a big tech & resource investment.

cmorley191:
> I hadn't even thought of this route before

I have seen the nanotask method used quite a bit. I haven't seen the cash method.

jimmy12or:
> I'd argue we should remove all the other infinite traversal ways mentioned, too. This isn't about realism; it's a gameplay problem. Players aren't supposed to have infinite space traversal. Space is supposed to be scary.
> 
> Every crew member is _not_ supposed to have an emergency way to get back to the station in a panic.
> 
> The exception is bluespace fire extinguisher -- it's balanced by a big tech & resource investment.

I think the PKA is intentional though so that a salvager who runs out of jetpack fuel or similar still has a (rather slow) method of getting back to station since it can be very difficult for a rescuer to find them if needed.

kotobdev:
I'm personally a fan of the NanoTask method, largely because it's extremely easy to accidentally end up off-station on Glacier (and still fairly easy on other maps), and there is absolutely no worse method of dying then being lost in space. Dying in space leaves you in a pseudo-state of not knowing if your body will be randomly found or not, it's 100 times worse then being gibbed, where at least you have a brain that'll get found fast. Having an extremely inefficient way to navigate back on-grid is just QOL to me

cmorley191:
I would dispute the idea that the nanotask method is "extremely inefficient". It's pretty strong.

Axianamos:
We've removed nearly every single other form of round removal so why add one in? Realism or not, being lost in space isn't fun. We removed sy dictate kill objectives, removed the metam machine, added a million cryo chems, made reversing rot easier, added surgery. The theme on DV seems to be "round removal bad"

So why intentionally add a way for it to happen now?

ShepardToTheStars:
> So why intentionally add a way for it to happen now?

One of the things I would like to add is crew monitors showing people off station, so no more slightly being 2 feet from the station grid and not being noticed. So this would allow paramedics to actually go and rescue people out in space.

Woljif:
Not sure if this is a mald PR but Ive never seen anyone use the nanotask method for meaningful space traversal besides recovery

Halo3moth:
I have seen this happen a few times it's not my favorite you almost always have something to throw if your have your PDA to print paper you have your PDA to throw, it's a bit silly and could probably be less effective without doing literally nothing after all that much paper fitting in your pda those have to be small strips meaning the energy your exchanging for throwing them probably isn't much

MantasDab360:
> One of the things I would like to add is crew monitors showing people off station, so no more slightly being 2 feet from the station grid and not being noticed. So this would allow paramedics to actually go and rescue people out in space.
We probably just port some of frontier crew server/monitor code

MantasDab360:
> I have seen this happen a few times it's not my favorite you almost always have something to throw if your have your PDA to print paper you have your PDA to throw, it's a bit silly and could probably be less effective without doing literally nothing after all that much paper fitting in your pda those have to be small strips meaning the energy your exchanging for throwing them probably isn't much 

Monolith has this so the propulsion you gain from object is size dependant (unless I'm delusional)

DisposableCrewmember42:
> Not sure if this is a mald PR but Ive never seen anyone use the nanotask method for meaningful space traversal besides recovery

not a mald pr, i just thought it's weird that this is a thing

DisposableCrewmember42:
> I have seen this happen a few times it's not my favorite you almost always have something to throw if your have your PDA to print paper you have your PDA to throw, it's a bit silly and could probably be less effective without doing literally nothing after all that much paper fitting in your pda those have to be small strips meaning the energy your exchanging for throwing them probably isn't much

I could change the comp so we can set how high the ratio is, but I think this is something that should be done automatically based on size/mass instead. 

From a cursory glance, the Physics system already does dynamically set mass somehow, but it doesn't seem to make a meaningful difference (for small items at least). I didn't feel like touching the Physics code because it could mess the whole game up and I don't fully understand the system.

rebe83:
Direction review 24h

rebe83:
So there were some suggestions on making it so nanotask uses a limited paper cartridge instead (making it finite) or making it so printing has a cool down.
Would it be possible to implement either of these?

rebe83:
Due to scope difference in requested changes we're just gonna approve merge it's not that deep

Toby222 (approved):
yeag sure

PR #5123, opened 2026-01-04
feat: adminoverlay command

DisposableCrewmember42:
## About the PR
Added a command to toggle the admin overlay so I can include it in my aghost script :blunt:

## Why / Balance
I'm lazy and scripts are convenient.

## Technical details
Added client-side command that uses AdminSystem to toggle the overlay. Added access to ADMIN perm.

## Media
N/A

## Requirements

**Changelog**
:cl: DisposableCrewmember42
DELTAVADMIN:
- add: Command to toggle the admin overlay: "adminoverlay <true|false>"

### DIFF
--- a/Content.Client/_DV/Administration/Commands/AdminOverlayCommand.cs
+++ b/Content.Client/_DV/Administration/Commands/AdminOverlayCommand.cs
@@ -1,0 +1,32 @@
+using Content.Client.Administration.Systems;
+using Robust.Shared.Console;
+
+namespace Content.Client._DV.Administration.Commands;
+
+public sealed class AdminOverlayCommand : LocalizedEntityCommands
+{
+    [Dependency] private readonly AdminSystem _admin = default!;
+    public override string Command => "adminoverlay";
+    public override void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        if (args.Length != 1 || !Boolean.TryParse(args[0], out var newValue))
+        {
+            shell.WriteLine(Help);
+            return;
+        }
+
+        if (newValue)
+        {
+            _admin.AdminOverlayOn();
+        }
+        else
+        {
+            _admin.AdminOverlayOff();
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


Resources/clientCommandPerms.yml
```diff
@@ -74,6 +74,7 @@
     - dumpentities
     - fuckrules # DeltaV: move fuckrules to DEBUG
     - showaccessreaders
+    - adminoverlay # DeltaV: add adminoverlay command
```
MilonPL: remove this and use flags

Content.Client/_DV/Administration/Commands/AdminOverlayCommand.cs
```diff
@@ -0,0 +1,32 @@
+using Content.Client.Administration.Systems;
+using Robust.Shared.Console;
+
+namespace Content.Client._DV.Administration.Commands;
+
+public sealed class AdminOverlayCommand : LocalizedEntityCommands
```
MilonPL, line 6: `[AdminCommand(AdminFlags.Admin)]` attribute

MilonPL (changes requested):

Content.Client/_DV/Administration/Commands/AdminOverlayCommand.cs
DisposableCrewmember42: the AdminCommand attribute is only a thing in Content.Server

MilonPL: dang it, this should really be a shared attribute
either way it should be being Admin and not Debug since it's the admin overlay which is accessed from the admin panel normally, this would be bad perms-wise if any codebase has Codermins with Debug access like WizDen does

DisposableCrewmember42: gotcha, I'll do that in yml then

Toby222 (approved):

Issue #4131, opened 2025-08-05
Revenant ectoplasm being put in water insta-kills said revenant


userfacethis:
## Description

When a revenant "dies" they get turned into ectoplasm that eventually revived iirc, and has intended ways of being dealt with like the Bible and such. But instead you can dissolve the ecto into liquid like you would a pill and that instantly ends the ecto and revenant.  Making the regen phase completely pointless because it takes 1 second to kill it.
**Reproduction**

Kill revenant
Grab liquid container
Grab the ecto of the revenant and put it in the liquid container
**Screenshots**

No screenshots as what would I even screenshot
**Additional context**

I heard this is a known bug but wasn't in the github which was weird. I hope this gets fixed soon because it completely shuts down revenants.

Vapetastic-Gaming:
I believe this was once a bug, and is now an intended feature, since its been around for AGES. uh Closed for now

PR #5125, opened 2026-01-04
fix: rev ectoplasm can no longer be used to spike drinks

DisposableCrewmember42:
## About the PR
Fixed an oversight allowing the Revenant to be RR'd by spiking a drink using its ectoplasm.

## Why / Balance
This was obviously not intended to happen and completely circumvents the respawn system in place for the Revenant. The argument that this bug has been around too long and should thus be considered a feature is, respectfully, dumb as hell.

## Technical details
~~Add InvalidSpikeSourceComponent and abort spiking if present on source.~~
Adjust YAML to remove ScoopableComponent and SolutionSpikerComponent

## Media
N/A

## Requirements

## Breaking changes

**Changelog**

:cl: DisposableCrewmember42
- fix: The Revenant can no longer be killed by spiking your own water bottle.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
@@ -1,10 +1,12 @@
 - type: entity
-  parent: Ash
+  parent: BaseItem
   id: EctoplasmRevenant
   suffix: DO NOT MAP
   name: ectoplasm
   description: Slimey, vaporous residue of an ancient spirit.
   components:
+    - type: Item
+      size: Tiny
     - type: Sprite
       sprite: _DV/Mobs/Ghosts/revenant.rsi
       state: ectoplasm
@@ -23,6 +25,8 @@
               Quantity: 10
             - ReagentId: Ectoplasm
               Quantity: 30
+    - type: Extractable
+      grindableSolutionName: food
     - type: GuideHelp
       guides:
       - MinorAntagonists
--- a/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
@@ -29,3 +29,4 @@
     - type: StatusEffects
       allowed:
       - Stasis
+    - type: InvalidSpikeSource
--- a/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
@@ -29,4 +29,3 @@
     - type: StatusEffects
       allowed:
       - Stasis
-    - type: InvalidSpikeSource
--- a/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
@@ -2,7 +2,7 @@
   parent: BaseItem
   id: EctoplasmRevenant
   suffix: DO NOT MAP
-  name: ectoplasm
+  name: revenant ectoplasm
   description: Slimey, vaporous residue of an ancient spirit.
   components:
     - type: Item
--- a/Content.Shared/Chemistry/EntitySystems/SolutionSpikerSystem.cs
+++ b/Content.Shared/Chemistry/EntitySystems/SolutionSpikerSystem.cs
@@ -1,4 +1,3 @@
-using Content.Shared._DV.Revenant.Components;
 using Content.Shared.Popups;
 using Content.Shared.Chemistry.Components;
 using Content.Shared.Chemistry.Components.SolutionManager;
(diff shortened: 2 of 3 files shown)


Halo3moth:
If we wanna go through with this it should come with a patch to the grinding system and should increase the dead time so crew can actually find the chaplain in my humble opinion

DisposableCrewmember42:
what kind of patch would you want to see for the grinding system?

apeiroeder:
I believe the grinding sysyem is referring to being able to grind the ectoplasm to kill the revenant, even without salt being needed to be present in the grinder; while the guidebook states that it must be grinded together with salt to fully kill it

Woljif:
> If we wanna go through with this it should come with a patch to the grinding system and should increase the dead time so crew can actually find the chaplain in my humble opinion

Either this, or there needs to be more reliable ways to deal with the ectoplasm. You can't execute it with lasers, you can't butcher it, you can't pour salt on it, and you can't even vanquish it with a bible unless you're the chaplain, and even if you are, it takes like 5 seconds (about 8% of the time it takes a revenant to respawn). I feel like at least one of those methods should be made more doable

cmorley191:
I believe I remember that direction's latest position is that this is intended, not an oversight. https://github.com/DeltaV-Station/Delta-v/issues/4131

Though that was a few months ago; they can reconsider it here.

NoElkaTheGod:
Isn't revenant being hard to kill, like, the entire point of the respawning system? What's the point of it if anyone can just kill the revenant easily?

Could totally extend the timer, though. Better yet, make the timer depend on the amount of essence that the revenant had gathered – the longer you've been haunting the station, the harde ryou are to destroy.

apeiroeder:
Hi, direction here!

Just wanted to add some info:
This was still considered a bug, but classified as "no longer a bug" to prevent a lot of appeal-only bans for a long standing and widely known way to take out a rev, with the bug being over a year old, and no fix in foreseeable future in sight.

TLDR: that "no longer a bug" reclassification was due to administrative reasons.

apeiroeder:
Also, direction here again, just to clarify:

As this is a bugfix, no direction review is needed.

Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
cmorley191, line 1: Making a new component that just blocks SolutionSpikerComponent is a pretty roundabout approach. Let's just remove the SolutionSpiker component from ectoplasm instead, by removing the `parent: Ash` here. The Ash inheritance is hardly doing anything useful anymore -- the only things being inherited are the Item and Extractable components, so we can just add those.

```
- type: entity
  parent: BaseItem
  id: EctoplasmRevenant
  suffix: DO NOT MAP
  name: ectoplasm
  description: Slimey, vaporous residue of an ancient spirit.
  components:
    - type: Item
      size: Tiny
    - type: Sprite
      sprite: _DV/Mobs/Ghosts/revenant.rsi
      state: ectoplasm
    - type: Tag
      tags:
        - Trash
        - Ectoplasm
    - type: SolutionContainerManager
      solutions:
        food:
          maxVol: 50
          reagents:
            - ReagentId: Ash
              Quantity: 10
            - ReagentId: Necrosol
              Quantity: 10
            - ReagentId: Ectoplasm
              Quantity: 30
    - type: Extractable
      grindableSolutionName: food
    - type: GuideHelp
      guides:
      - MinorAntagonists
    - type: StatusEffects
      allowed:
      - Stasis
```

In fact, in looking into this I realized that Ash's ScoopableComponent would've still allowed reverse spiking

cmorley191 (dismissed):

Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
DisposableCrewmember42: Oh, for whatever reason I thought the System was applied to all ents with SolutionContainerManager

DisposableCrewmember42: tested, works!

DisposableCrewmember42:
working on a solution for grinding system, pls don't merge yet

DisposableCrewmember42:
There are multiple conflicting ways of allowing grinding mentioned in the (shit)code...
- Documentation mentions inserting salt into the ecto and then grinding
- Documentation mentions inserting salt into the grinder's beaker and then grinding the ecto with it
- The code includes a list of reagents to be checked, but the solution of the Ecto is never actually checked for them
- The code for the GrindAttemptEvent has a field called Reagents, which is a list of Entity Ids, which aren't actually reagents inside the ecto but the items inserted into the grinder.
- The code in the stasis systems checks Tags on these "reagents", but reagents can't actually have Tags as far as I can tell

DisposableCrewmember42:
I'll fix that up in a separate PR, feel free to merge if there's nothing wrong with the prototype.

Resources/Prototypes/_DV/Entities/Objects/Specific/Medical/morgue.yml
```diff
@@ -1,10 +1,12 @@
 - type: entity
-  parent: Ash
+  parent: BaseItem
   id: EctoplasmRevenant
   suffix: DO NOT MAP
   name: ectoplasm
```
ShepardToTheStars: Just to better differentiate.

```suggestion
  name: revenant ectoplasm
```

ShepardToTheStars (approved):

DisposableCrewmember42:
heisentest strikes again, guh

PR #5131, opened 2026-01-05
Reduce stealth hardsuit armor

Stop-Signs:
## About the PR
Stealth suit res changed to 20% across the board

## Why / Balance
based on community feedback, You can have either armor or invis but not both. Overall this should make these less viable but still a fun thing to bring.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Stealth suit resistances have been reduced.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -19,9 +19,9 @@
   - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.7 # DeltaV - was 0.65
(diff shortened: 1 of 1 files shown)


Cepelinas1:
right so took a while because discussions suck
direction doesn't want armor values changed, but how it's invisiblity works:
- make the invisibility rely on suit battery to function, so the wearer doesn't have infinite invis
- probably make that if you get shot while invisible, you oughta become visible

ShepardToTheStars:
Requires code changes. I can make the necessary edits if Stop Signs wants.

Stop-Signs:
> Requires code changes. I can make the necessary edits if Stop Signs wants.

That would be great, your work with the ninja hardsuit was already fantastic.

PR #5141, opened 2026-01-05
Heater/freezer temporary nerf

IamVelcroboy:
## About the PR
Reduces the power draw of Heaters and Freezers 

## Why / Balance
Temporary fix in anticipation of APC power breaker changes. To be reverted when [this issue](https://github.com/space-wizards/space-station-14/issues/36840) is fixed.

## Technical details
n/a

## Media
n/a

## Requirements

## Breaking changes
n/a

**Changelog**
:cl: Velcroboy
- tweak: Atmos heaters and freezers, and space heaters have had their power draw temporarily reduced to 100w until the roundstart enabled issue is resolved

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/portable.yml
+++ b/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/portable.yml
@@ -128,6 +128,7 @@
         layer:
         - MachineLayer
   - type: ApcPowerReceiver
+    powerLoad: 100  # Delta V - Temporarily reduce power draw until roundstart power draw is fixed.
     powerDisabled: true #starts off
   - type: Sprite
     sprite: Structures/Piping/Atmospherics/Portable/portable_sheater.rsi
--- a/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/unary.yml
+++ b/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/unary.yml
@@ -268,6 +268,8 @@
     - type: GuideHelp
       guides:
       - Thermomachines
+    - type: ApcPowerReceiver # Delta V - Temporarily reduce power draw until roundstart power draw is fixed.
+      powerLoad: 100
 
 - type: entity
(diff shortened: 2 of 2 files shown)


ShepardToTheStars:
Just 100? Thats only 100W. >.>

You'll probably want *some* load so its not suddenly breaking APCs once we revert this.

IamVelcroboy:
> Just 100? Thats only 100W. >.>
> 
> You'll probably want _some_ load so its not suddenly breaking APCs once we revert this.

They were previously 1000. What would you suggest?

ShepardToTheStars:
Space heaters current draw 3.5kW (or 3500) when on. Freezers heaters currently draw 5kW (5000) when on.

Their power draw is set to their heating capacity when it actually starts cooling/heating. So, if you unpack a space heater at 20C, since its set to 20C, it will only draw 1kW (or 1000) until it actually has to start needing to cool/heat, then its power draw is changed to 3.5kW (though it never returns to 1kW once it no longer has to heat/cool things).

Looking at it, I thought those draws would be higher tbh. Does space heater actually need to change?

Freezer and heater usually have multiple in an area, though they only start drawing more power once they have to heat/cool gas. I'd suggest maybe just halving it?

IamVelcroboy:
> Looking at it, I thought those draws would be higher tbh. Does space heater actually need to change?

Possibly no, it looks like at most there are 6 un-enabled versions on any given map(Glacier). So, that's maybe not an issue. It's just not intended to be drawing power at all roundstart. I just added for consistency. So, 🤷‍♂️

> Freezer and heater usually have multiple in an area, though they only start drawing more power once they have to heat/cool gas. I'd suggest maybe just halving it?

👍

cmorley191:
I'm working on just fixing the underlying issue, it seems like an easy fix

cmorley191:
@IamVelcroboy @ShepardToTheStars please see the linked alternative

IamVelcroboy:
Closing in favor of the proper fix

Issue #5115, opened 2026-01-03
Nukie commander doesn't roll on nukie rounds


novostiya:
Commander will not roll even if there are people with it enabled readied up, leaving nukies without one.

**Reproduction**
Have people with commander, agent and op antags enabled readied up on a nukies round. There won't be a commander.

ShepardToTheStars:
<img width="2519" height="1342" alt="Image" src="https://github.com/user-attachments/assets/ad0764bb-e785-4c90-8193-dade1a973634" />

Trying to reproduce locally but so far, it still rolls commander. (This is with all job requirements disabled so its probably the whitelist)

ShepardToTheStars:
Fixed in #5146

PR #5146, opened 2026-01-06
Might fix nukie commander not spawning

ShepardToTheStars:
## About the PR

May address #5115, assuming that the current whitelist has carried over when checking JobRequirements when it comes to antags.

At the moment, I'm unable to whitelist in my current local environment, due to antags no longer being considered job prototypes at all, and therefore, not appearing on the DV-specific whitelist UI. That needs updating. The issue is that I cannot submit a whitelist for the antag because of UI issues, not that it doesn't pull or save from the player's content data. 

But this *might* work for people who are currently whitelisted, so I'd encourage merging this while we work on the antag whitelist UI, otherwise nukie commanders are just broken.

## Why / Balance

Bug with commanders never being chosen.

## Technical details

Antag ban changes screwed with current system.

## Requirements

## Breaking changes

**Changelog**

No changelog in case it just doesn't fix anything yet lol.

### DIFF
--- a/Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
+++ b/Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
@@ -278,6 +278,7 @@ public sealed class PlayTimeTrackingSystem : EntitySystem
             playTimes = new Dictionary<string, TimeSpan>();
         }
 
+        var isWhitelisted = player.ContentData()?.Whitelisted ?? false; // DeltaV - Whitelist requirement
         var requirements = _roles.GetRoleRequirements(antag);
         return JobRequirements.TryRequirementsMet(
             requirements,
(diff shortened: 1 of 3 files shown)


cmorley191:
What are the "antag ban changes" that you're referring to?

ShepardToTheStars:
https://github.com/space-wizards/space-station-14/pull/35966

These

cmorley191:
Ok this PR looks correct to me based on nearby code, but I'm also wondering what exactly is going on here

Is the Global Whitelist the one being used here? That one *is* working on my local environment...

image

Or are you saying that NukeopsCommander is supposed to be showing up in the Job Whitelists list? (it's not on mine)

image

cmorley191:
We probably shouldn't merge changes to role permissions unless we're sure that admins have a UI to manipulate it.

Or in other words, if commander is a whitelist-only role and admins don't have a way to *revoke* that permission at the moment, then we probably shouldn't add the role back in yet

cmorley191 (approved):

ShepardToTheStars:
> Is the Global Whitelist the one being used here? That one is working on my local environment.

Yes. And it does work. Except there's no checkbox for nukie commander anymore. That would be a 2nd PR.

> We probably shouldn't merge changes to role permissions unless we're sure that admins have a UI to manipulate it.
> Or in other words, if commander is a whitelist-only role and admins don't have a way to revoke that permission at the moment, then we probably shouldn't add the role back in yet

I would usually agree, but right now, commanders are flat out not spawning. This should be a quick fix to allow currently whitelisted people to spawn as commander at least. The UI will be another PR soon, I just need to untangle this whole feature and also do UI work.

Admins are having to spawn in a commander manually and/or give a war declarator to the nukies. The whitelist can be revoked by giving an antag ban for commander at the moment, but new people cannot be whitelisted until the UI is fixed.

Issue #4671, opened 2025-11-16
Harpy guidebook is spreading myths about breathing / air quality


cmorley191:
## Description

None of the below harpy guidebook text is correct, given the current state of the codebase:

<img width="653" height="187" alt="Image" src="https://github.com/user-attachments/assets/bf0a9091-9130-407d-b34d-1f0800e12444" />

We need to fix this guidebook text, or go back and retroactively implement these mechanics properly.

## Details

The only technical difference in harpy breathing is that they metabolize **twice as much gas** but **half as often**. The end result is the same: *the rate of metabolizing is exactly the same as humans.* They do not breathe more oxygen, and they are not more susceptible to CO2.

```
- type: entity
  id: OrganHarpyLungs
  - type: Metabolizer
    updateInterval: 2.0  # <--- metabolize every 2 seconds, rather than human's 1 second.
    groups:
    - id: Gas
      rateModifier: 200.0  # <--- metabolize 2u, rather than human's 1u.
      # result: 2u every 2s, is exactly the same as 1u every 1s.
```

This is not accurately described by the guidebook text.

However, there *is* a practical difference to harpy gasping, which I discuss below. 

In short:

**Due to an engine quirk, all species will occasionally gasp when there are more than three types of gases in a room. The more types of gases there are (4, 5, 6), the more often the gasping. The amounts of gas do not matter, just the number of types. Harpies will gasp more than other species when this problem occurs, due to the less frequent metabolizing batches. But all species will gasp.**

## Reproduction

These tests prove that the current guidebook text and community wisdom about harpy breathing ("sensitive to CO2") are wrong:
- Put a human and a harpy in a room, and slowly release CO2 from a CO2 canister. Both will start gasping at exactly the same time: 1.0%. It is no earlier for the harpy.
- Put a human and a harpy in a room with a mask and O2 tank. Turn on internals on both. Both will reach 0 kPa in the tank at exactly the same time.

The truth about harpy gasping: 

- Put a human and a harpy in a room. Release *very trace amounts* (literally any tiny amount is enough) of many different gases: Tritium, Plasma, Water Vapor, Miasma, and CO2. Now that there are seven gas types in the room, both the human *and* the harpy will start gasping due to an engine quirk I am calling "Metabolic Roll Asphyxiation". The harpy will gasp slightly more, but all species are susceptible to this.

### Metabolic Roll Asphyxiation

[...]

PR #5157, opened 2026-01-07
mobs no longer gasp when in the presence of many types of gases

cmorley191:
## About the PR

Fixed a bug where all mobs were asphyxiating when 4 or more gas types are present, even if these gases were at entirely safe levels (e.g. 0.1% CO2 is very safe alone, but caused gasping when more gas types were added)

This was also the source of the fabled harpy gasping "air quality" myths: https://github.com/DeltaV-Station/Delta-v/issues/4671 . Due to arcane interactions with obscure harpy metabolizer settings, harpies were made *more* vulnerable to this bug. However, *all* species were vulnerable to this bug when there are enough different gas types in the air, not just harpies.

## Why / Balance

It makes no sense for the *number of gas types* in the air to cause asphyxiation; the *amounts (mols/%)* of bad gas is what should matter. Currently, this bug would still suffocate all species even if the amounts were absolutely miniscule.

Mobs were asphyxiating because lungs (just like all organs) only process up to 3 different reagents at a time, to prevent poison stacking abuse. This was causing Oxygen to occasionally be excluded from the processing due to random chance (3 reagents are randomly selected), even though plenty of Oxygen is present in the air.

This PR simply removes this limit for lungs -- lungs will now process up to all 9 gas types simultaneously; all gases present in a room.

"Poison stacking" is the fact that, because reagents are metabolized in parallel, multiple poisons in a syringe causes damage more quickly than a single poison in a syringe. To prevent players from abusing this, all organs are limited to only processing 3 reagent types a time. IMO poison stacking is irrelevant for lungs. Reason 1: nobody is using gases to poison people as an antag; it's impractical. Reason 2: there are only 9 gases in the game (three of which are inert oxygen/nitrogen/water vapor), so the potential for crazy abuse of poison stacking is not really there for gases like it is with the vast number of liquid reagents.

Finally, for those who want to keep the "harpy's are more sensitive to air quality" around, I agree with you! I have a followup PR planned to implement that feature properly, instead of keeping around a bug that nobody really understands very well: https://github.com/DeltaV-Station/Delta-v/pull/5158

## Technical details

See this issue for a more detailed explanation of the bug: https://github.com/DeltaV-Station/Delta-v/issues/4671

The fix code itself should be pretty self explanatory though; let me know if it's not.

### Before this PR:

All species gasp in rooms that have 4+ gas types present, regardless of the actual amounts (%/mols) of those gases. Harpies happen to gasp more frequently in this situation due to strange code interactions.

For context: 1% is the CO2 threshold for most species to start gasping (when in a room with only oxygen, nitrogen, and CO2 alone). So the fact that 0.1% CO2 here is causing gasping in humans is pretty weird. The other gases are not related to respiration; they do not cause gasping (when added alone). This is all happening due to the bug.

metabolic_roll_asphyx_demo

### After this PR:

Species do not gasp in rooms with lots of trace pollutant gases. The gases are safe as long as they are in very small amounts.

metabolic_roll_asphyx_removed

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed a bug where all species (not just harpies) were gasping in the presence of *extremely* trace amounts of pollutant gases (0.01%).

### DIFF
--- a/Resources/ServerInfo/Guidebook/Mobs/_DV/Harpy.xml
+++ b/Resources/ServerInfo/Guidebook/Mobs/_DV/Harpy.xml
@@ -29,7 +29,6 @@
   - Comes with the Ultraviolet Vision trait by default. This can be disabled via accessibility options.
   - Cannot wear Jumpsuits, and will automatically start with a Jumpskirt instead.
   - While singing, musical notes appear floating around their head.
-  - Gases breathed in metabolize twice as fast.
 
   ## Drawbacks
 
--- a/Content.Server/Body/Components/MetabolizerComponent.cs
+++ b/Content.Server/Body/Components/MetabolizerComponent.cs
@@ -72,8 +72,11 @@ namespace Content.Server.Body.Components
         ///     Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low
         ///     quantity, would be muuuuch better than just one poison acting.
         /// </summary>
+        /// <remarks>
+        /// DeltaV - changed from an `int` to an `int?` to support lungs not having a limit
+        /// </remarks>
         [DataField("maxReagents")]
-        public int MaxReagentsProcessable = 3;
+        public int? MaxReagentsProcessable = 3;
 
         /// <summary>
         ///     A list of metabolism groups that this metabolizer will act on, in order of precedence.
--- a/Content.Server/Body/Systems/MetabolizerSystem.cs
+++ b/Content.Server/Body/Systems/MetabolizerSystem.cs
@@ -157,8 +157,13 @@ public sealed class MetabolizerSystem : SharedMetabolizerSystem
             }
 
             // we're done here entirely if this is true
-            if (reagents >= ent.Comp1.MaxReagentsProcessable)
+            // DeltaV - start of null limit allowed
+            if (
+                ent.Comp1.MaxReagentsProcessable is { } maxReagentsProcessable 
+                && reagents >= maxReagentsProcessable
+            )
                 return;
+            // DeltaV - end of null limit allowed
 
 
             // loop over all our groups and see which ones apply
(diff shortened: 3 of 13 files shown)


Resources/ServerInfo/Guidebook/Mobs/_DV/Harpy.xml
```diff
@@ -29,7 +29,6 @@
   - Comes with the Ultraviolet Vision trait by default. This can be disabled via accessibility options.
   - Cannot wear Jumpsuits, and will automatically start with a Jumpskirt instead.
   - While singing, musical notes appear floating around their head.
-  - Gases breathed in metabolize twice as fast.
```
cmorley191, line 32: This theory came from the fact that harpy lungs have a `rateModifier: 200.0` (instead of human 100.0)

However, this theory was false -- it failed to take into consideration the fact that harpy lung's `updateInterval: 2.0` (instead of human 1.0) cancels out the 200.0, resulting in the same overall metabolic rate.

Further reading: https://github.com/DeltaV-Station/Delta-v/issues/4671

ShepardToTheStars:
I don't like having to put `maxReagents: null` on every lung entity ever though, but I get that we can't just default the component to null, since MetabolizerSystem also controls other forms of metabolizing. Hmmm...

SirWarock:
> I don't like having to put `maxReagents: null` on every lung entity ever though, but I get that we can't just default the component to null, since MetabolizerSystem also controls other forms of metabolizing. Hmmm...

We could make a parent lung that has maxReagents Null?

ShepardToTheStars:
> We could make a parent lung that has maxReagents Null?

Except that pretty much all lungs have a defined `MetabolizerComponent` on them with things specific to that species, and the parent's `MetabolizerComponent` would get blown away.

little-meow-meow:
Fantastic write-up. Though if this hasn't been affected by upstream's recent body and metabolism refactors, it is very likely it will be soon.

sowelipililimute:
This is likely going to be fixed w/ the Nubody merge.

PR #5158, opened 2026-01-07
harpy gasp when pollutant gases are present

cmorley191:
## About the PR

Harpies now gasp when certain thresholds of "bad" gases are present in the room:

| Gas | Harpies Now Gasp At | Other Species (not changed) |
| --- | --- | --- |
| CO2 | 0.4%+ | gasp at 1.1%+ |
| Plasma, Trit | 0.6% | do not gasp |
| Frezon, Miasma | 1.1% | do not gasp |
| Water Vapor | 4% | do not gasp |
| Nitrous Oxide | do not gasp | do not gasp |

The community is already aware of a "harpies are sensitive to air quality" feature. This PR just implements it in a proper/consistent/predictable/intuitive way, instead of it being based on a finicky, misunderstood bug that we should remove in https://github.com/DeltaV-Station/Delta-v/pull/5157

## Why / Balance

I think most harpy players are happy with the existing air quality limitation; it makes for a good RP challenge. 

Harpies have a breathing mask and emergency o2 tank in their survival box that provide protection from pollutants in a pinch. They can also ask atmos players to install extra air scrubbers in rooms that they spend a lot of time in if gases become a problem (e.g. medbay, the kitchen, etc.)

## Technical details

Made the gases cause anti-respiration just like CO2 already does.

The rate of asphyxiation damage caused by these gases is proportional to the values set in the code, and the thresholds in the table above which I found through testing.

So, e.g. harpies will take ~3x the asphyxiation damage that a human would at poisonous CO2 levels (e.g. 1.5%).

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Harpies are now reliably more sensitive to pollutant gases than other species. For example, 0.4% of CO2 will make harpies start gasping, while humans are okay up to 1.1% CO2

### DIFF
--- a/Resources/Prototypes/_DV/Body/Organs/harpy.yml
+++ b/Resources/Prototypes/_DV/Body/Organs/harpy.yml
@@ -12,14 +12,13 @@
   - type: Organ
     slotId: lungs
   - type: Metabolizer
-    updateInterval: 2.0
     removeEmpty: true
     solutionOnBody: false
     solution: "Lung"
-    metabolizerTypes: [ Human ]
+    metabolizerTypes: [ Harpy ]
     groups:
     - id: Gas
-      rateModifier: 200.0
+      rateModifier: 100.0
   - type: SolutionContainerManager
     solutions:
       organ:
--- a/Resources/Locale/en-US/_DV/metabolism/metabolizer_types.ftl
+++ b/Resources/Locale/en-US/_DV/metabolism/metabolizer_types.ftl
@@ -1,2 +1,3 @@
 metabolizer-type-feroxi = Feroxi
 metabolizer-type-feroxi-dehydrated = Dehydrated Feroxi
+metabolizer-type-harpy = Harpy
--- a/Resources/Prototypes/Reagents/Consumable/Drink/drinks.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/drinks.yml
@@ -493,6 +493,11 @@
         ratios: # DeltaV no duplicating O2 >:3
           CarbonDioxide: 0.5
           WaterVapor: -1.0
+      - !type:Oxygenate # DeltaV - Harpy sensitive to air quality
+        conditions:
+        - !type:MetabolizerTypeCondition
+          type: [ Harpy ]
+        factor: -1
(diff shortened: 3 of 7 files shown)


Resources/Prototypes/_DV/Body/Organs/harpy.yml
```diff
@@ -12,14 +12,13 @@
   - type: Organ
     slotId: lungs
   - type: Metabolizer
-    updateInterval: 2.0
```
cmorley191, line 15: This:

```
updateInterval: 2.0,  rateModifier: 200.0
```

has exactly the same metabolic rate (200 / 2 = 100u per second) as this:

```
updateInterval: 1.0,  rateModifier: 100.0
```

so this change basically does nothing. But there could be edge cases (like the original harpy gasping bug) that I really think we need to avoid so we don't have to be worrying about them -- that's why I think we should just have this inline with the other species' 100.0/1.0

cmorley191:
Oh. @PureBreadBagel am I stepping on your toes by making guidebook changes?

I'm unaware of how we manage the guidebook -- is it like maps where only one person is supposed to be changing it at a time?

I don't mean to interfere with your changes at all; this thought just occurred to me after reading your comment in the other PR.

FaintSpeaker:
peanut posting because I can provide a little insight 

> is it like maps where only one person is supposed to be changing it at a time?

this is only the case for maps because maps can't be git merged effectively, so if there's a merge conflict one of the two map changes will have to be re-done.

Guidebooks are text files so it's just a matter of resolving merge conflicts if two people edit the same file - but that doesn't usually require re-doing large swaths of work like you would for mapping.

PureBreadBagel:
@cmorley191
Honestly don't worry about it. Your PR will merge way before I even push out V2 of my guidebookagloo.
But if you're talking about the current one I made during porting the crashcart, you should also be good as well, as you'll overwrite the file when it gets pushed.

Cepelinas1:
direction review soon (somehow)

Vapetastic-Gaming:
Approved by direction

ShepardToTheStars (approved):

PR #5166, opened 2026-01-08
Redrock Spawn Distance Increase

ShepardToTheStars:
## About the PR

Increased VGroid another 200m out while we debug the issue of redrock spawning inside the station.

## Why / Balance

Bandaid Fix as we cannot reliably reproduce this issue.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Redrock spawns a bit further out until we can figure out why its spawning inside the station.


Toby222 (approved):
Merging without testing, I'm gonna trust that just bumping those numbers doesn't break anything

PR #5180, opened 2026-01-10
Migrate Chronic Pain to the New Status Effect System

### LINKED ISSUE #5084: Chronic Pain not properly suppressed by Soretizone
## Description
The chronic pain shader is briefly displayed roughly once a second after taking Soretizone.

**Reproduction**
1. Spawn your favorite Urist
2. Give them the PainComponent
3. Debug->Control Mob
4. Take Soretizone and observe

ShepardToTheStars:
## About the PR

* Migrated the chronic pain condition to the new status effects system.
* Tweaked pain suppression and anxiety meds popup sizing so its smaller when things are good, but larger as its wearing off.
* Fixed Chronic Pain's shader being buggy and continuously pulsing. (Fixes #5084)
* Certain painkillers are actually better than others now.
  * THC: Lasts Lasts ~4.1 minutes per 5u
  * Stubantazine: Lasts ~8.3 minutes per 5u
  * Soretizone: Lasts 12.5 minutes per 5u
  * Agonolexyne: Lasts 25 minutes per 5u

## Why / Balance

We need to move to the new status effects system and I started debugging why the overlay was bugged and pulsing, and just ended up fixing it by re-writing and migrating it.

## Technical details

I re-wrote mostly everything.
* We don't need the Entity Effects anymore since we're only using the new status effect system.
* Combined the clientside overlay system into the main clientside system - We didn't need both.
* Re-organized and renamed everything from generic "Pain" systems to "ChronicPain" because thats exactly what it controlled.

How it works now is:
- If you have the chronic pain component, you will experience chronic pain symptoms. The only way to get rid of the symptoms is to gain the ChronicPainSuppressed status condition, which will make the new systems get rid of your chronic pain effects.
 
## Media

Looks the same but also, no pulsing when suppressed.

https://github.com/user-attachments/assets/533a1520-605b-4ee4-8134-9ddc6d45bf4e

## Requirements

## Licensing

- [x] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [x] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt)
    

## Breaking changes

I basically re-wrote chronic pain to only use the new status effects.

**Changelog**

(The re-write doesn't need to be in the changelog)

:cl:
- tweak: Certain painkillers are more long-lasting than others, based on how strong they are.
- tweak: Players with the chronic pain trait start with ~5 minutes of suppressed pain before they start feeling the effects. 
- fix: Fixed chronic pain's effects from pulsing continuously.


MilonPL:
finally…

i think i was high when making this btw, i dont remember much

Content.Shared/_DV/ChronicPain/EntitySystems/SharedChronicPainSystem.cs
```diff
@@ -0,0 +1,120 @@
...
+    public static readonly EntProtoId ChronicPainStatusEffect = "StatusEffectChronicPainSuppressed";
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<ChronicPainComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<ChronicPainComponent, ComponentInit>(OnChronicPainInit);
+        SubscribeLocalEvent<ChronicPainComponent, ComponentShutdown>(OnChronicPainShutdown);
+        SubscribeLocalEvent<ChronicPainComponent, LocalPlayerAttachedEvent>(OnPlayerAttached);
+        SubscribeLocalEvent<ChronicPainComponent, LocalPlayerDetachedEvent>(OnPlayerDetached);
+    }
```
MilonPL, line 30: must call base moment

MilonPL:
i am high when reviewing this too fyi but i cant spot anything obviously wrong other than this nitpick

Content.Shared/_DV/ChronicPain/EntitySystems/SharedChronicPainSystem.cs
ShepardToTheStars: oh, right. Good catch.

Content.Client/_DV/ChronicPain/Overlays/ChronicPainOverlay.cs
```diff
@@ -1,36 +1,31 @@
...
-        _painShader = _prototype.Index(_shaderProto).Instance().Duplicate();
+        _painShader = _prototype.Index(_shaderProto).InstanceUnique();
     }
 
     protected override bool BeforeDraw(in OverlayDrawArgs args)
     {
-        if (_player.LocalEntity is not { Valid: true } player
-            || !_entity.HasComponent<PainComponent>(player))
-        {
+        if (_player.LocalEntity is not { Valid: true } player)
```
Toby222:
```suggestion
        if (_player.LocalEntity is not { Valid: true })
```

Content.Shared/_DV/ChronicPain/Components/ChronicPainComponent.cs
```diff
@@ -59,7 +52,7 @@ public sealed partial class PainComponent : Component
     /// Minimum time between pain popups in seconds
     /// </summary>
     [DataField]
-    public float MinimumPopupDelay = 1f;
+    public float MinimumPopupDelay = 5f;
```
Toby222: If we're already refactoring this, we should make this and MaximumPopupDelay `TimeSpan` as well, and drop the "in seconds" part

Content.Shared/_DV/ChronicPain/EntitySystems/SharedChronicPainSystem.cs
```diff
@@ -0,0 +1,122 @@
...
+
+    protected virtual void OnPlayerDetached(Entity<ChronicPainComponent> entity, ref LocalPlayerDetachedEvent args)
+    {
+        // Used by the client to get rid of the overlay
+    }
+
+    protected void ShowPainPopup(Entity<ChronicPainComponent> entity)
+    {
+         // Don't notify
```
Toby222: One space too much
```suggestion
        // Don't notify
```

Resources/Locale/en-US/_DV/traits/traits.ftl
```diff
@@ -19,16 +19,16 @@ trait-ultravision-desc = Whether through custom bionic eyes, random mutation,
...
 trait-hushed-name = Hushed
 trait-hushed-desc = You are unable to speak louder than a whisper.
 
 trait-uncloneable-name = Uncloneable
 trait-uncloneable-desc = You cannot be cloned.
 
-trait-inpain-name = Chronic pain
-trait-inpain-desc = You’re constantly in discomfort. You need painkillers to function.
+trait-chronic-pain-name = Chronic pain
+trait-chronic-pain-desc = You're constantly in discomfort. You need painkillers to function.
```
Toby222, line 31: I don't fault you for removing it, but technically the `’` is more correct than the `'`
(No need to change back I just noticed this and looked up the history of the symbol :3)

Resources/Prototypes/Reagents/narcotics.yml
```diff
@@ -222,6 +222,10 @@
           reagent: THC
           min: 0.25
       - !type:SuppressAddiction
+      - !type:ModifyStatusEffect # suppresses chronic pain mildly
+        effectProto: StatusEffectChronicPainSuppressed
+        time: 25 # 4.16 minutes of pain suppression per 5u (5u / 0.5 * 25 / 60)
```
Toby222, line 227: I do not understand that formula
5u / 0.5 minutes * 25 seconds / minute?

Toby222 (changes requested):

Resources/Prototypes/Reagents/narcotics.yml
ShepardToTheStars: Because things get kind of weird with the metabolism rates, the formula is:

`5u / [metabolism rate] * [time added per tick] / 60 seconds = X minutes of pain suppression`

* 5u takes 10 metabolism ticks to process (0.5u processed per tick).
  * Because the effect is `type: Add`, every time the metabolism tick happens, it adds that much time to the Suppression status effect
* So `5u / 0.5u/tick = 10 (metabolism) ticks` and `10 ticks * 25 seconds = 250 seconds` added of pain suppression time (or 4.16 minutes when you divide by 60)

Because our other pain suppression meds process at 0.01 or 0.02 (and interact badly with other pain suppressants and achohol), I didn't want to change those metabolism rates. Instead, you have to change the formula since `5u / 0.01 = 500 (metabolism) ticks`, so we need to adjust the amount of pain suppression status effect time added. Soretizone added 1.5 seconds per 0.01, so that totals to `500 ticks * 1.5 seconds = 750 seconds` of pain suppression. 

Originally, I thought `scale` would handle it but the metabolism rate *is* the scale, so in THC's case, if you have 0.25u to be metabolized instead of 0.5u, the scale will be 50% of a normal metabolism tick.

Resources/Locale/en-US/_DV/traits/traits.ftl
ShepardToTheStars: Good, cause my editor points out uncommon unicode characters that are probably relics of strings being copy/pasted from word editors. :P

image

Content.Shared/_DV/ChronicPain/Components/ChronicPainComponent.cs
ShepardToTheStars: Fixed, though it gets a bit roundabout, because RobustRandom can't do a random number of seconds between two given TimeSpans.

```c#
        var delay = RobustRandom.Next(entity.Comp.MinimumPopupDelay.Seconds, entity.Comp.MaximumPopupDelay.Seconds);
        entity.Comp.NextPopupTime = _timing.CurTime + TimeSpan.FromSeconds(delay);
```

ShepardToTheStars:
@Toby222 Questions answered and changes made.

Content.Shared/_DV/ChronicPain/EntitySystems/SharedChronicPainSystem.cs
```diff
@@ -0,0 +1,122 @@
...
+        var effects = dataset.Values;
+        if (effects.Count == 0)
+            return;
+
+        var effect = RobustRandom.Pick(effects);
+        Popup.PopupPredicted(Loc.GetString(effect), entity, entity);
+
+        // Set next popup time
+        var delay = RobustRandom.Next(entity.Comp.MinimumPopupDelay.Seconds, entity.Comp.MaximumPopupDelay.Seconds);
+        entity.Comp.NextPopupTime = _timing.CurTime + TimeSpan.FromSeconds(delay);
```
FaintSpeaker:
```suggestion
        var delay = RobustRandom.Next(entity.Comp.MinimumPopupDelay, entity.Comp.MaximumPopupDelay);
        entity.Comp.NextPopupTime = _timing.CurTime + delay;
```

`Seconds` only returns the remaining seconds after accounting for minutes. If the TimeSpans were ever set to something like "00:01:00 to 00:01:30", it would disregard the minutes and roll between 00 and 30.

You can use `RobustRandom.Next(TimeSpan, TimeSpan)` to get a random TimeSpan in the range of min (inclusive) to max (exclusive).

ShepardToTheStars: Ill double check when I get home but I swore my IDE was saying that RobustRandom.Next(TimeSpan, TimeSpan) had an issue (after changing the component’s types )

Toby222 (commented):
Do not merge pending upcoming trait rework

Content.Shared/_DV/ChronicPain/EntitySystems/SharedChronicPainSystem.cs
Toby222: Even if RobustRandom did not work for this (it works just fine), it would still have to be TotalSeconds instead of Seconds :^)

ShepardToTheStars: Fixed. @FaintSpeaker was correct.

ShepardToTheStars:
> Do not merge pending upcoming trait rework

This shouldn't affect the upcoming trait rework, sans maybe renaming the component on the Chronic Pain trait.

ShepardToTheStars:
@Toby222 Ready for re-re-review.

Toby222 (approved):
Approved
Would be cool if there was a way to migrate the old to the new trait
If you can't think of one just merge yourself :^)

PR #5181, opened 2026-01-10
Make space pens stop killing avali

Eternally-Confused:
## About the PR

I replaced the lepo in space pens with a new (currently unobtainable) chem called "stabilizine". To reduce the temperature rubberbanding that lepo has, this chem only adds the full 100 kJ  to the player below -20°C. Above that, it only gives 40 kJ.

## Why / Balance

It seems really silly to give Avali something that will kill them in their _survival box_. Hopefully this will prevent new players from accidentally killing themselves. There's a handful of alternate solutions to this problem, but this seemed like the least problematic one. 

I've done a fair bit of testing with the specific numbers, and this seemed to be a fairly decent middle ground. An Avali now takes ~5 to 10 heat damage when injected with a space pen at room temperature, which is quickly healed. This is much better than the ~60 damage that space pens did previously. In space, Avali now last about as long as other species!

Although this _technically_ a nerf to space pens for all species, in practice it barely makes a difference. In the testing I did, most species die about 2 or 3 seconds earlier to spacing than they did before. Oni (because of their larger body mass) die about 5 seconds earlier than before. In both cases, this is out of a total survival time in space of about 100 seconds. Overall, I think this is a fairly minor downside for non-avali species. 

## Technical details

It's just a bunch of yaml. The new chem deliberately has two maximum temperature triggers, to better stabilize temperatures without overshooting. 

## Media

Does this need media? I'm genuinely not sure if it's necessary

## Requirements

**Changelog**

:cl: 
- tweak: (TEST MERGE) Space pens now contain a chemical called Stabilizine instead of Leporazine, which is better at stabilizing body temperature.
- tweak: Avali no longer die from space pens.


ShepardToTheStars:
Hmm. I get why this PR is needed, but I'm wondering if the simpler solution is to just remove Lepo from space pens.

Only because continuously adding boxes for each species is not very maintainable in the long run.

Eternally-Confused:
> Hmm. I get why this PR is needed, but I'm wondering if the simpler solution is to just remove Lepo from space pens.
> 
> Only because continuously adding boxes for each species is not very maintainable in the long run.

Yeah, that would definitely make sense. It just feels weird to nerf every other species' ability to survive due to one species with silly mechanics.

ShepardToTheStars:
> Yeah, that would definitely make sense. It just feels weird to nerf every other species' ability to survive due to one species with silly mechanics.

You could make a new (probably unobtainable for now) reagent that space pens have that stabilizes temp better instead of rubberbanding temp all over the place.

"Stabalazine" or whatever you wanna name it lol
```
- type: reagent
  id: Stabalazine
  name: reagent-name-stabalazine
  group: Medicine
  desc: reagent-desc-stabalazine
  physicalDesc: reagent-physical-desc-you-decide
  flavor: medicine
  color: "#ff7db5"
  metabolisms:
    Medicine:
      effects:
      - !type:AdjustTemperature
        conditions:
        - !type:TemperatureCondition
          max: 293.15
        amount: 10000 # thermal energy, not temperature!
      - !type:AdjustTemperature
        conditions:
        - !type:TemperatureCondition
          min: 293.15
        amount: -10000
```

Something like this would stabilize temps both ways equally, though not crazy fast. Would need some testing to get the values right. (I just changed Lepo's 100000 to 10000)

Eternally-Confused:
> > Yeah, that would definitely make sense. It just feels weird to nerf every other species' ability to survive due to one species with silly mechanics.
> 
> You could make a new (probably unobtainable for now) reagent that space pens have that stabilizes temp better instead of rubberbanding temp all over the place.
> 
> "Stabalazine" or whatever you wanna name it lol

Oh, that'd definitely be easier! I'll try some testing on values

Halo3moth:
If your adjusting space pens to no longer have lepo keep in mind how that would effect cult and colossus not saying it's bad just saying it's something worth looking at

Woljif:
I don't think we should nerf space pens for everyone just to accommodate a single species. Just make an avali space pen variant

kotobdev:
replace avali kit space pens with charcoal pens for when other people spacepen them :trollface:

Halo3moth:
> replace avali kit space pens with charcoal pens for when other people spacepen them :trollface:

give avali lex pens round start to equalize it

Eternally-Confused:
Alright, that's everything updated! I did notice an issue where the guidebook isn't showing temperature effects, but that ~~doesn't seem to be my fault~~ is an upstream issue™️

kotobdev:
Space pens are kind of supposed to kill avali, that's pretty much the point of the species downside in the first place.

If anything, I'd recommend either replacing avali survival kit spacepens with this new chem, and/or downtuning the amount of damage lepo does to avali (increase their specific heat by a little bit so it's substantially less lethal, it wasn't intended to be actually able to crit them from full health). Leporazine in space pens is mighty useful due to it's healing effects, which is something I would (personally) like to keep.

Eternally-Confused:
> Space pens are kind of supposed to kill avali, that's pretty much the point of the species downside in the first place.

I'll be honest, I didn't know that space pens were intended to kill Avali. _Lepo_ killing them made sense, but having emergency equipment kill them seemed silly.

Halo3moth:
If your worried about how easy lepo is to get you could always tone down the per unit damage to be like 90 or take a longer time to kill that way it give the avali more time to react. I would agree that a space pen killing avalis that fast probably isn't necessary for their balancing but I don't think they were introduced with the thought of being meta in mind

Eternally-Confused:
> If your worried about how easy lepo is to get you could always tone down the per unit damage to be like 90 or take a longer time to kill that way it give the avali more time to react. I would agree that a space pen killing avalis that fast probably isn't necessary for their balancing but I don't think they were introduced with the thought of being meta in mind

I don't care much if Avali are non-meta tbh. I just don't think it makes for a good experience for new/unfamiliar players to give them The Emergency Box That Kills you. Any solution that avoids that (ideally without causing a maintenance headache) will make me happy.

kotobdev:
In my extraordinarily humble opinion, I think one more emergency box variant would be worth it. ~~Also, noobtraps are funny and it's a lesson they'll only learn once :)~~

ShepardToTheStars (approved):
Looks good maintainer-wise.

ShepardToTheStars:
Direction was rather divided. Lepo is popular I guess! Still, we're approving for a test merge.

ShepardToTheStars:
And to everyone wondering, I did test the lethality of multiple space pens. Both two pens and three pens still do about the same damage. Most of the heat damage comes from Baro's 30% chance of taking 0.5 heat damage.

**Before Change - On Human in Jugsuit**
image

**After Change - On Human in Jugsuit**
image

PR #5188, opened 2026-01-11
[Mapping] Add new command for showing network links

BarryNorfolk:
## About the PR
Adds a new client side command `togglenetworklinks` to show all device network linkages, as though you had clicked "show" via a network configurator on all of them.

## Why / Balance
Pharaz offered their soul to me for this.

## Technical details
Using the regular mapping overlay, which works via marker components, only worked in PVS range so users would have to set the command on constantly in order to see all the networks involved. With a new overlay that just works over all device networks, via enumerator, we can just have the user move around and constantly update the visible networks.

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added new mapping command 'togglenetworklinks'

### DIFF
--- a/Content.Client/_DV/NetworkConfigurator/ShowNetworkLinksCommand.cs
+++ b/Content.Client/_DV/NetworkConfigurator/ShowNetworkLinksCommand.cs
@@ -1,0 +1,20 @@
+using Robust.Client.Graphics;
+using Robust.Shared.Console;
+
+namespace Content.Client._DV.NetworkConfigurator;
+
+public sealed class ToggleNetworkLinksCommand : LocalizedEntityCommands
+{
+    [Dependency] private readonly IOverlayManager _overlay = default!;
+
+    public override string Command => "togglenetworklinks";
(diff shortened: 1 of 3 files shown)


Content.Client/_DV/NetworkConfigurator/ShowNetworkLinksCommand.cs
```diff
@@ -0,0 +1,20 @@
...
+using Robust.Shared.Console;
+
+namespace Content.Client._DV.NetworkConfigurator;
+
+public sealed class ToggleNetworkLinksCommand : LocalizedEntityCommands
+{
+    [Dependency] private readonly IOverlayManager _overlay = default!;
+
+    public override string Command => "togglenetworklinks";
+    public override string Description => "Toggles seeing network configurator links.";
```
ShepardToTheStars: Localize these.

ShepardToTheStars (changes requested):

ShepardToTheStars (approved):

PR #5190, opened 2026-01-11
Makes the crash cart obtainable via ordering

mnva0:
## About the PR
This PR makes the unobtainable and completely unused crash cart accessible by a means other than a mapper potentially choosing to add it to a map in the future

It comes loose in a crate (which isn't exactly in line with other large things—they're usually flatpacked—but whatever)

Crash carts would be awesome to have for medical and ~~might actually cause medics to do something differently for a round~~ it's a shame to see it apparently unobtainable here and on Funkystation

## Why / Balance
I have no idea how to balance things. The price is 1000 spesos for a completely empty cart and I genuinely have no frame of reference for if that's good or bad. I'm leaning towards "that's too cheap"

Things to consider when correcting my likely-awful price:
- The crash cart has a defibrillator slot
- The crash cart has a tank slot (can hold air tanks, oxygen tanks, nitrogen tanks, nitrous tanks, plasma tanks...)
- The crash cart has 48 slots total (if I'm counting this correctly). 2\*6 or 12 slots per "shelf". For context, crates can hold 30 glass shards. See images for 30 glass shards in a crash cart
- The crash cart in the crate is completely empty
- Crash carts have a StaticPrice of 500 spesos
- Crash carts apparently have no restrictions on what types of items can be put in them. Below is me picking up random junk and putting it in one
- From a quick check, dragging it seems to be about the same as dragging a crate (but more slippery because it's got wheels)
- The crash cart be anchored (this is less of a balance consideration and more something I wanted to add to the list because I like it)

## Technical details
Hooks, for some reason, completely stopped working so I bypassed them. I have no idea what this may have done but I also do not really care for debugging exactly why it can't find a program on my computer

## Media
Crash carts (filled and nonfilled) and the crash cart crate (using the default medical crate)
image

image

image

Showing how it has no item type restrictions. ~~Bonus points if you know the path I took picking up this junk~~
image

Crates can hold 30 glass shards (I chose them because they are the smallest item I thought of). This is 30 glass shards in a crash cart
image

---

**END OF ACTUALLY-PR-RELATED IMAGES.** I'm just showing off the crash cart at this point because I think it's cool

I arranged some stuff and quite like it
- Defibrillator (in item slot)
- Nitrous canister (in item slot)
- 6 jugs
- 3 rollerbeds
- 3 bodybags
- Inflatable walls and doors (for cordoning off areas)
- Wrench (for anchoring the crash cart)
image

Significantly less thought put into this loadout. Engineering-focused one
image

A little scene I put together using the medical loadout shown above. A field hospital
image

## Requirements

**Changelog**
:cl: Minerva
- add: Crash carts can now be ordered from Logistics under the Medical category.

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/Cargo/cargo_medical.yml
+++ b/Resources/Prototypes/_DV/Catalog/Cargo/cargo_medical.yml
@@ -38,3 +38,13 @@
   cost: 10000
   category: Medical
   group: market
+
+- type: cargoProduct
+  id: CrashCartCrate
+  icon:
+    sprite: _Funkystation/Objects/Specific/Medical/crashcart.rsi
+    state: cart
+  product: CrateCrashCart
+  cost: 1000
+  category: Medical
+  group: market
--- a/Resources/Prototypes/_DV/Catalog/Fills/Crates/medical.yml
+++ b/Resources/Prototypes/_DV/Catalog/Fills/Crates/medical.yml
@@ -42,3 +42,13 @@
     - id: ClothingMaskSterile
       amount: 2
     - id: PaperViralCarePackageManual
+
+- type: entity
+  id: CrateCrashCart
+  parent: CrateMedical
+  name: crash cart crate
+  description: Contains one empty crash cart.
+  components:
+  - type: StorageFill
+    contents:
+    - id: CrashCart


ShepardToTheStars (approved):

ShepardToTheStars:
Might have issues with them opening the crate but its probably fine. I'm gonna say this doesn't require direction approval since we're swamped right now and this seems like something no one will have an issue with. Direction can yell at me if they want.

PR #5200, opened 2026-01-12
sign language

MilonPL:
## About the PR

adds sign language to the game
review ready direction and code wise, needs sprites, help appreciated!

design doc:
https://docs.google.com/document/d/1xPJFjsYn3q1eB-fufhmW7U3KD6v3cdPv_-1VhX6btLU/edit?tab=t.0

left to do:
- [ ] sprites (icons)
- [ ] typing indicator sprite

requires tips for the tip thingy

## Why / Balance

adds a meaningful and unique way for mute players to interact with the crew

## Media

https://github.com/user-attachments/assets/798b0b77-7d6b-4d96-b6d0-553eaa42b29a

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added sign language! The new sign language menu can be opened with J — keep in mind it only works if you’re mute.
- add: Added the "Silent communicator" trait, which lets your character speak and understand sign language.


BellwetherLogic:
Why would you be unable to speak sign language if you don't have the Mute trait? I don't understand why the restriction is there, anyone can learn sign languages (and more people should).

SirWarock:
Specifically the mine should be unable to use it without breaking their vow

pootslap:
If I'm reading it right you can speak/understand sign language with the Silent Communicator trait even if you aren't mute.

I have an annoying concern though, should every mute character be able to speak and understand sign language? IIRC it's possible to do with the EE language system.

MilonPL:
I feel like mimes being able to speak sign language would break the point of vows a bit as it's yet another way to bypass the restrictions of the said role.
Ideally there'd be separate traits for "Mute, doesn't know sign language" and just "Mute" but that's for the future when I'm done with traits.

rubysharkk:
so this kind of gives. not great vibes personally. this just feels like sign language is being boiled down to a three-word panic button for when you don't have time to use paper or aac. it feels like it could very easily be demeaning towards people that actually use sign language since it's an absurd oversimplification that takes away the actual "language" from sign language.
EE sign language represented it better by having it be an _actual language_, since sign language _is_ an actual language, rather than it just being tiny, short sentences alongside an intent

StarboundSolaris:
Trial curator solaris here,
Direction Review 24h

Stop-Signs:
This is DIRECTION APPROVED FOR A TWO WEEK TESTMERGE (as soon as the PR freeze ends ofc)


at the end of the two weeks we will evaluate and decide whether we wish to keep this

PR #5204, opened 2026-01-13
Buff and fixes for energy shotgun

Halo3moth:
## About the PR

Did you know our version of the energy shotgun is actually a nerf from upstreams version? rebalances it to match upstream and fixes energy shotgun pellets not piercing windows

## Why / Balance

The energy shotgun is pretty lack luster all things considered youll hear people complain about its dps being poor quite often another possible solution is to increase its mag size or decrease its reload time either way i have no idea why its nerfed from upstream

Practically its dps goes from 80 to 104

Also bug fixed good c: 
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Energy shotgun lethal mag size from 6 to 9 
- fix: The energy shotgun projectiles now once again pierce windows

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -910,11 +910,11 @@
     heldPrefix: energy
   - type: GunRequiresWield #remove when inaccuracy on spreads is fixed
   - type: Battery
-    maxCharge: 480
-    startingCharge: 480
+    maxCharge: 720 # Delta V - Was 480
+    startingCharge: 720
   - type: BatterySelfRecharger #Delta V - Self recharging not removed
     autoRecharge: true
-    autoRechargeRate: 10
+    autoRechargeRate: 72 # Delta V - Was 10
     autoRechargePauseTime: 30
 
 - type: entity
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -1311,7 +1311,7 @@
     impactEffect: BulletImpactEffectOrangeDisabler
     damage:
       types:
-        Heat: 13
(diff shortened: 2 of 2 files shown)


Halo3moth:
not sure how i feel about buffing sec i was just doing a bug fix and realized it was nerfed and was like "what the heck people talk about this gun being weak all the time"

Axianamos:
Its nerfed because its the only way direction would approve its inclusion at all back in the day. Stop sign had to nerf it like 6 times to get approval

Halo3moth:
> Its nerfed because its the only way direction would approve it

It is worth noting that our energy shotgun recharges while upstreams no longer does, I think having a gun kill things faster will always be more noticeable then being able to fire more often, alternative changes to it could be increasing the number of shots to 5 and increasing spread slightly (a middle ground between the old firing modes) or decreasing the delay on its self recharge, it really depends if direction wants to make this wardens primary gun (higher dps) or more of a tool in wardens kit (recharging)

Shadowmize:
SOMEONE MIGHT WIN A FIGHT WITH E SHOTGUN HOLY

Vapetastic-Gaming:
dir rev 24

ShepardToTheStars:
@Halo3moth Direction requested changes. We're going to keep the damage the same for now, but we do agree that the shotgun is fairly subpar for a few reasons - its charge rate is slow and it doesn't have a lot of shots. So, in addition to the bug fix, we'll up that as well.

* Revert the damage back to 10.
* Up the battery to 720. This gives it +3 energy bolt shots and +5 disabler mode shots.
* Up to energy shotgun's recharge to 72 (or 10% of its total charge capacity). Requires 10 seconds to fully charge after 30 seconds of not shooting. 
```
  - type: Battery
    maxCharge: 720
    startingCharge: 720
  - type: BatterySelfRecharger #Delta V - Self recharging not removed
    autoRecharge: true
    autoRechargeRate: 72
    autoRechargePauseTime: 30
```

Also, the bug fix is fine. Doesn't need approval.

Halo3moth:
Omg I completely missed this so sorry will be done shortly

ShepardToTheStars (approved):

PR #5205, opened 2026-01-13
Microwave Emitter

EmberAstra:
Alternate title: I gave Epistemics a gun again BUT IT'S BETTER THIS TIME I SWEAR (I HOPE) (I REALLY HOPE)

## About the PR
A new research tool for Epistemics is here: the Microwave Emitter!
It's a weak "firearm" that fires pure radiation at a short range, intended to be used to activate those pesky Radiation nodes.

## Why / Balance
Radiation triggers are common, and sources of radiation... aren't really. The few that exists all depend on RNG in one way or another (Singulo is only a few maps, broken RTGs are rare salvage finds, gravity anoms and radioactive artifacts are not guaranteed), and most of them tend to be either hard to use or better saved for Engineering to deal with.
The only reliable option is to make an improvised uranium spear... but it also deals Pierce damage, and enough that it activates Physical Damage nodes before the Radiation ones! Meaning, if you have both active on the same artifact it becomes impossible to activate the Radiation ones without major improvisation.
And, with the new artifact system, you can't even microwave small ones anymore as a workaround!
Also, honestly: stabbing them with a piece of glass taped to a stick feels... weird to me. Too improvised.
Sooooo this is an attempt at bringing in an "official" NT-approved way of doing all of this!

Detailed specs:
- High fire rate (5 shots/second)
- Very weak (2 radiation per shot), totaling to 10DPS.
- Pierces windows
- Very short range (lifetime of 0.08, enough to reach ~3 tiles)
- Small internal battery (enough charge for 10 shots before needing recharge)
- T1 Experimental Research, costing 5000 points.
- Crafted only at the Epi lathe for 5 steel, 4 glass and 1 uranium (same cost as a CHIMP plus the uranium).

The hope is to have a tool that can act as at the very least an alternative to the classic spear. More intuitive to use, safer thanks to the range, and more precise in terms of damage types, at a small cost in materials/research and requiring some downtime to recharge it between uses.
I've erred a lot on the side of caution with the numbers to avoid causing a Svalinn 2.0. Right now a full charge is enough to activate a node and you'd need like 5 of these to hit every single shot to crit somebody, but I'm down to buff it if it proves to be too weak for the science side of things!

## Technical details
YAML and new sprites.

## Media
image

Blasting an artifact, and showing off the maximum range.
https://github.com/user-attachments/assets/00851d5f-701f-42d0-802d-cd6d1128644d

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the microwave emitter! Epistemics has a radiation gun to shoot artifacts with now!


Halo3moth:
Welcome back svalinn laser pistol, jokes aside this is really freakin tuff would be cool to see a antag version of it? Like with the chimp mod

Avalon-Proto:
Something funny you could do if you want to make an antag version - make it shoot a glob that sticks like a flare and emits radiation constantly. That way the target has to actively remove the problem even as you come after them

EmberAstra:
Definitely tempted to make an antag version! Definitely has potential to be overclocked in some way. But first, getting the normal one in.

ShepardToTheStars (approved):

PR #5209, opened 2026-01-14
feat: nudge singularity towards station periodically [mald PR]

DisposableCrewmember42:
## About the PR
In addition to the random jittery movement we all know and hate, the Singularity is now nudged towards the station periodically so it does not run off into space most of the time.

## Why / Balance
When the singularity looses, people roleplay panic. They act as if they are going to die, only for the singularity to fuck off into space after taking a bite out of engi 90% of the time. Even if it does end up moving into the station, it rarely actually ends up causing the catastrophic damage you would expect from a black hole being unleashed from Engineering.

The RandomWalkComponent makes looses anti-climactic. Fleeing from it means walking 40 tiles in another direction because you know that the singularity won't do anything and then sitting in an evac pod for 15 minutes after power goes out. The resultant roleplay feels stilted and inauthentic because there is nothing to fear.

I have heard people argue that the loose should not be more destructive because the station can be repaired, calling that an RP opportunity. I, on the other hand, call it bullshit. This hasn't happened and it never will. Nobody wants to be stuck in an unworkable department for 30 minutes while Engineering struggles to repair the damage when the round is basically over. _The shuttle will always be called_ if the singularity is loosed, so we might as well end the round with a bang instead of a fart in the wind followed by a power outage.

You may have heard that the singularity already moves towards the station every few walks. I am here to tell you, gamers, that the eviltributors have lied to you -- taken you for fools. There is no such thing implemented.

## Technical details
Added ChasingWalkComponent to the Singularity proto and tweaked some of its values. The Singularity will be nudged to cables every 5 seconds, the max speed for this nudging being the min speed of the random walk. I chose the CableComponent because cable entities are abundant on stations and can definitely be eaten by the Singularity, unlike some targets that might be chosen when chasing Transform or Destructible.

## Media
N/A

## Requirements

**Changelog**

:cl: DisposableCrewmember42
- tweak: Lord Singuloth rises.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/Generation/Singularity/singularity.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/Generation/Singularity/singularity.yml
@@ -51,6 +51,12 @@
   - type: RandomWalk # To make the singularity move around.
     maxSpeed: 2.5
     minSpeed: 1.875
+  - type: ChasingWalk # DeltaV - Add comp
+    minSpeed: 1
+    maxSpeed: 1.875
+    impulseInterval: 5
+    chasingComponent:
+    - type: Cable # Other comps like Destructible and Transform are also assigned to Entities the singulo can't eat
   - type: SingularityDistortion
     falloffPower: 2.529822
     intensity: 3645


Bonktrauma:
Yooo, finally, singularities are a station ending threat

Halo3moth:
You see station end to singos fairly often, singos are much more messy to maintain compared to teslas but safer to loose I'm not saying this is bad more saying why would people chose singo over Tesla unless that's all they have if they are both bloodthirsty 

Also there isn't really a way to avoid rads like there is with teslas shock, you can wear a rad suit but if the room spaces (it's a singo it's gonna space everything) your kinda just screwed vs insuls being wearable with anything and quite fashionable

muonpoly:
While we’re here, would it be possible to make the singulo eat department signs and signage/screens/apcs/fire alarms? The singulo always leaves these behind and it just doesn’t really look good.

IamVelcroboy:
> While we’re here, would it be possible to make the singulo eat department signs and signage/screens/apcs/fire alarms? The singulo always leaves these behind and it just doesn’t really look good.

I would say, open a proper issue for this with a list of relevant entities. Most of these should probably be addressed upstream

ShepardToTheStars (approved):

PR #5217, opened 2026-01-14
Ninja Stealth Fix?/Nerf?

ShepardToTheStars:
## About the PR

Doing at least 5 damage to the ninja will reveal the ninja.

## Why / Balance

The `AttackedEvent` only applies to melee attack events, not to projectiles or general damage.

## Technical details

Using the much more generic `DamageChangedEvent` instead of `AttackedEvent`. I tried using `ProjectileHit` and other ranged attacking events but they just weren't working so I'm guessing they were marked handled or whatever.

But this does a few checks.
* Is there a damage change? If no, return.
* Is the change in damage healing only? If yes, return.
* Does the damage have a source still? If no, it might be passive healing, airloss from space, bleed damage, etc, so let's check for those things.
  * Was there damage *and* healing? If so, return because its probably just an OD from 10u derma or whatever.
  * If the damage didn't have brute or burn types, then return.
* Does it meet the minumum damage requirement? (4 damage)? If no, return

* Then reveal the ninja (but only if its stealthed)

## Media

https://github.com/user-attachments/assets/26940440-7a9b-4734-8a94-97e314d98891

## Requirements

## Licensing

- [x] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [x] MIT (https://choosealicense.com/licenses/mit/)
    

**Changelog**

:cl:
- tweak: The ninja is now revealed with hit with attacks, projectiles, explosions, and grenades that do damage.
- fix: The ninja is now revealed with hit with attacks, projectiles, explosions, and grenades that do damage.

### DIFF
--- a/Content.Shared/Ninja/Systems/SharedSpaceNinjaSystem.cs
+++ b/Content.Shared/Ninja/Systems/SharedSpaceNinjaSystem.cs
@@ -100,23 +100,19 @@ public abstract class SharedSpaceNinjaSystem : EntitySystem
     private void OnNinjaAttacked(Entity<SpaceNinjaComponent> ent, ref DamageChangedEvent args)
     {
         // If there's no damage delta, just return
-        if (args.DamageDelta is not { })
+        if (args.DamageDelta is not { } damage)
             return;
 
-        // Don't reveal on healing
+        // Don't reveal on (most) healing
         if (!args.DamageIncreased)
             return;
 
         // If the damage doesn't have a source, we need to check the type, in case it 
         // was a grenade or explosion. We want to ignore airloss and toxin damage types.
-        if (!args.Origin.HasValue && args.DamageDelta is { } damage)
+        if (!args.Origin.HasValue)
         {
-            // If there are any negative values, its probably natual or chem healing, so don't reveal
+            // If there are any negative values, its probably natual or chem healing, so don't reveal. It might be an OD from medicine.
             if (!damage.AnyPositive())
-                return;
-
-            // Only reveal on damage greater than the minumum. This prevents tiny ticks of damage (e.g. from malign rifts pulses)
-            if (damage.GetTotal() < ent.Comp.MinimumRevealDamage)
                 return;
 
             // Check the damage types for damage types that should reveal (brute, burns)
@@ -125,6 +121,10 @@ public abstract class SharedSpaceNinjaSystem : EntitySystem
             if (!damageGroups.ContainsKey("Brute") && !damageGroups.ContainsKey("Burn")) // This feels a bit dirty, oh well.
                 return;
         }
+
+        // Only reveal on damage greater than the minumum. This prevents tiny ticks of damage (e.g. from malign rifts pulses)
+        if (damage.GetTotal() <= ent.Comp.MinimumRevealDamage)
+            return;
 
         // Yea, now reveal that son of a bitch >:3
         TryRevealNinja(ent, disable: true);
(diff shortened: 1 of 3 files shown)


Content.Shared/Ninja/Systems/SharedSpaceNinjaSystem.cs
```diff
@@ -78,11 +84,49 @@ public virtual bool TryUseCharge(EntityUid user, float charge)
...
+
+            // Check the damage types for damage types that should reveal (brute, burns)
+            // Basically, we want to ignore most indirect forms of damage (airloss, toxins)
+            var damageGroups = damage.GetDamagePerGroup(_prototypeManager);
+            if (!damageGroups.ContainsKey("Brute") && !damageGroups.ContainsKey("Burn")) // This feels a bit dirty, oh well.
+                return;
+        }
+
+        // Only reveal on damage greater than the minumum. This prevents tiny ticks of damage (e.g. from malign rifts pulses)
+        if (damage.GetTotal() <= ent.Comp.MinimumRevealDamage)
```
Toby222: As you specifically mention punches in the comment of `MinimumRevealDamage`, I expected a punch to reveal them
Also minimum is generally inclusive
```suggestion
        // Only reveal on damage at least the minumum. This prevents tiny ticks of damage (e.g. from malign rifts pulses)
        if (damage.GetTotal() < ent.Comp.MinimumRevealDamage)
```

Toby222 (changes requested):

Content.Shared/Ninja/Systems/SharedSpaceNinjaSystem.cs
ShepardToTheStars: > Also minimum is generally inclusive

~~Wouldn't changing `<=` to `<` make it exclusive? Most punches wouldn't reveal them since punches would only deal 4 damage in most cases.~~

Or no wait. Im dumb and need coffee.

ShepardToTheStars:
@Toby222 Ready for re-review.

Toby222 (commented):
Phase cloak action says:
> Beware that if you are hit, all abilities are disabled for 5 seconds, including your cloak!

Not sure if that broke with this PR or is already broken in master

SirWarock:
> Phase cloak action says:
> 
> > Beware that if you are hit, all abilities are disabled for 5 seconds, including your cloak!
> 
> Not sure if that broke with this PR or is already broken in master

That's what I remembered! I vaguely remembered that somewhere it said it would disable all your abilities, but I didn't find it again

Toby222 (commented):
> > Phase cloak action says:
> > 
> > > Beware that if you are hit, all abilities are disabled for 5 seconds, including your cloak!
> > 
> > Not sure if that broke with this PR or is already broken in master
> 
> That's what I remembered! I vaguely remembered that somewhere it said it would disable all your abilities, but I didn't find it again 

Wanna fix that (or the description) in this PR?

ShepardToTheStars:
> Wanna fix that (or the description) in this PR?

Fix what?

> That's what I remembered! I vaguely remembered that somewhere it said it would disable all your abilities, but I didn't find it again

Ninja already has that logic built-in. Its just broken upstream. The disable boolean just never is a factor. I fixed that too. Unless the disable Boolean just doesnt work at all. But this PR doesn’t mention disabling abilities, just revealing the ninja.

Toby222:
> Fix what?

The description not agreeing with reality

> Ninja already has that logic built-in. Its just broken upstream. The disable boolean just never is a factor. I fixed that too. Unless the disable Boolean just doesnt work at all.

When I tested locally, you can immediately stealth again after getting punched, both on master and in this feature branch

ShepardToTheStars:
Scope creep smh

Toby222:
I *ASKED* if you want to fix it :sob:

ShepardToTheStars:
Nah. I’m too lazy.

Toby222 (approved):
in that case lgtm

Toby222:
How did we fuck up the CL so bad :blunt:

PR #5223, opened 2026-01-15
feat: Rework Revenant Stasis

DisposableCrewmember42:
## About the PR
The RevenantStasis system is using the old status effects system. I want to migrate it to the new one and fix inconsistencies in the implementation as well as address balancing concerns brought up in #5125.

The plan is:
- Make salting the rev ectoplasm actually possible. Adding salt to the ectoplasm solution will add 50% of the regular respawn delay to the status effect duration. The time adjustment can only be applied once to each instance of Ecto.
- Replace the failed grinding explosion with something else; it ends up destroying the grinder and the ecto with it, which is clearly not the intended behavior.
  - I'm thinking of releasing a cloud of cold miasma and displaying a pop-up text on the grinder.
- Salting should be actually required for the Revenant to be successfully ground.
  - This wasn't properly implemented before, the code mentions various ways it was presumably planned to be implemented. There are unused variables and used variables named in misleading ways.
  - Currently, it requires that an item tagged with Salt or Holy to be ground together with the Ecto. There is no proto tagged with either, so it always explodes.

## Requirements

## Breaking changes

**Changelog**


DisposableCrewmember42:
closing this for now. i'm planning to start tinkering on a Revenant rework which will include fixes for stasis once the psionics refactor is merged

PR #5226, opened 2026-01-15
feat: Make footprints decals

DisposableCrewmember42:
> [!CAUTION]
> Don't use this. It will tank performance completely.
> Never make assumptions about implementation details

## About the PR
This is an alternative to the current FootPrints that should be more lightweight, but is less featureful.
It spawns decals instead of entities for footprints. As other cleanable decals they can be removed using space cleaner and cleanades currently. This PR also adds the ability to scrub the floor using mops to remove cleanable decals. Scrubbing the floor consumes 5u water.

## Why / Balance
Footprints in their current form constantly spawn complex entities, leading to lag.
Figured since we give people infinite crayons and they draw genuinely impressive art using a shitload of overlapping decals, performance should be a lot better. This also offloads the burden of keeping track of all of the prints to upstream's DecalSystem.

## Technical details
added DecalScrubberComponent and System, can be added to items to make them scrub the floor when clicking a tile. rewrote parts of EE Footprints to use decals instead and moved them under DV namespace since we need to maintain them separately now.

## Media
https://github.com/user-attachments/assets/e3784cba-0ed5-4136-9c8b-b196b376870a

^ not shown here because it was added later: 5u water will be removed from the mop when trying to scrub the floor. if there isn't enough water, it won't work.

## Requirements

## Breaking changes

**Changelog**

:cl: DisposableCrewmember42
- fix: Footprints are now decals, this should address associated performance issues. As a result, they can no longer be mopped like puddles.
- add: Mops can now be used to scrub off decals by left-clicking tiles.


DisposableCrewmember42:
image
Footprints working mostly, just offset to the to East/North based on player rotation

DisposableCrewmember42:
finally ready for review

rebe83:
Heroic efforts idk why they were made to be containers

BarryNorfolk:
image
Art.

Content.Server/_DV/Footprints/Components/PuddleFootPrintsComponent.cs
```diff
@@ -1,4 +1,4 @@
-namespace Content.Shared._EE.FootPrint;
+namespace Content.Server._DV.Footprints.Components;
 
 [RegisterComponent]
```
BarryNorfolk, line 8: Would be nice to have a comment about what this is/does.

Content.Server/_DV/Footprints/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+using Content.Shared.Mobs.Components;
+using Content.Shared.Standing;
+using Robust.Shared.Map;
+using Robust.Shared.Random;
+
+namespace Content.Server._DV.Footprints.Systems;
+
+public sealed partial class FootPrintsSystem : EntitySystem
+{
+
```
BarryNorfolk: Random newline, plstodelete.

Content.Server/_DV/Footprints/Components/FootPrintsComponent.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// Used to determine when enough distance has been traveled for the next print.
+    /// </summary>
+    [DataField]
+    public Vector2 LastPrintPosition = Vector2.Zero;
+
+    /// <summary>
+    /// The size scaling factor for footprint steps. Must be positive.
+    /// </summary>
+    [DataField]
+    public float StepSize = 0.7f;
```
BarryNorfolk, line 31: If both this and the drag size "Must" be positive, we should probably have some validation step here to make sure it's actually above zero.
We could make an integration test to validate this properly, so people don't make things that are utterly broken. What do you think?

StarboundSolaris:
Hopefully this should make those accursed footprints less laggy.

Content.Shared/_DV/Footprints/Components/DecalScrubberComponent.cs
```diff
@@ -0,0 +1,40 @@
...
+using Robust.Shared.Map;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared._DV.Footprints.Components;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class DecalScrubberComponent : Component
+{
+
```
BarryNorfolk: Random newline, plstodelete again

Content.Server/_DV/Footprints/Systems/FootPrintsSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+    [Dependency] private readonly IMapManager _map = default!;
+    [Dependency] private readonly DecalSystem _decalSystem = default!;
```
BarryNorfolk, line 18: Sad that this is server only and can't be predicted, ah well. Maybe upstream will make it predicted!

Content.Server/_DV/Footprints/Systems/PuddleFootPrintsSystem.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        SubscribeLocalEvent<PuddleFootPrintsComponent, EndCollideEvent>(OnEndCollide);
+    }
+
+    private void OnEndCollide(Entity<PuddleFootPrintsComponent> ent, ref EndCollideEvent args)
+    {
+        var tripper = args.OtherEntity;
+
+        // Don't process if the tripper is flying
+        if (_flightQuery.TryComp(ent, out var flight) && _flight.IsFlying((ent, flight)))
```
BarryNorfolk: This checks if the puddle is flying, but I don't think you want to... right?
At least the comment is saying TRIPPER is flying

BarryNorfolk: Could rename the ent here to `puddle` just to make things more readable too :)

BarryNorfolk (commented):
Not finished yet, just FYI, but here are some existing comments and a bug I think

Content.Server/_DV/Footprints/Systems/FootPrintsSystem.cs
DisposableCrewmember42: yeah unfortunately not, i'll add a TODO comment though in case upstream makes it predicted at some point

Content.Server/_DV/Footprints/Systems/PuddleFootPrintsSystem.cs
DisposableCrewmember42: OtherEntity in this case is the entity walking over the puddle, but I idk why it was called tripper in the original code. I'll rename it to collidingEntity or something

DisposableCrewmember42: oh wait i misread that, yeah that shouldn't check ent

Content.Server/_DV/Footprints/Components/FootPrintsComponent.cs
DisposableCrewmember42: yup sounds reasonable

BarryNorfolk (dismissed):
Things look good here to me, test works well to catch the errors if someone puts in a negative value. There's no chances of runtime checking (if someone VV's the things to be negative), and that's something I think upstream have wanted to look at. Or at least there were people talking about it last I saw :D 

Conceptually I think this should be better here with decals since, as you said, it's not an entity. I can't really test what it'll be like with 40+ players all at once making footprints, but we'll have to see.

The only things i wish we could have are decals for different species so there's a bit of a "Diona was here" forensics style of stuff now that it's gone. But that can be added in the future.

Thanks for the contribution, I might still just see if anyone has any disagreements with this before merging.

DisposableCrewmember42:
oh yea i just tore that out to revisit later since it seemed like a rare case anyway, forgot diona can't wear shoes :blunt:

DisposableCrewmember42:
Can I add that to this PR or should I make another one now that this was reviewed/is being reviewed by CR?

Edit: Just gonna do that, was a simple enough change

DisposableCrewmember42:
image
species prints working

BarryNorfolk:
Legend. Just waiting on Content to make a decision about it, hopefully won't be too long :)

Vapetastic-Gaming:
Content approved. Please at a later time add a system that will require Soap in the water (or space cleaner).

BarryNorfolk (approved):

DisposableCrewmember42:
just in case anyone sees this PR in the future and considers using it or doing something similar: don't. this was a terrible, terrible idea.

PR #5242, opened 2026-01-18
Removed estoc slowdown

Woljif:
## About the PR

Removed the 25% speed slowdown from the estoc when wielded

## Why / Balance

The Estoc doesn't really stand with the other nukie weapons. Generally, slowdown when wielded or held is reserved for weapons that are particularly powerful when wielded or held. Some examples are the Hristov, which does 50 AP damage, the Gatorgun, which is a super weapon, and the Leviathan, which is a devastating rocket launcher. The Estoc is just a light DMR, and also somehow costs more than the C-20 and Bulldog. Most nukie players tend to openly gravitate away from the Estoc for this reason.

## Technical details

Removed the slowdown when wielded .yml trait

## Requirements

## Licensing

- [ ] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [ ] MIT (https://choosealicense.com/licenses/mit/)
    

**Changelog**

:cl:
- tweak: Reduced the slowdown penalty when wielding the Estoc DMR to 10%

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
@@ -199,7 +199,7 @@
   name: XL8
   description: |-
     The Experimental Lecter 8
-    An unreasonably expensive military grade assault rifle with integrated optic. 
+    An unreasonably expensive military grade assault rifle with integrated optic.
     Uses .20 rifle ammo.
   components:
   - type: Sprite
@@ -219,7 +219,7 @@
     - Burst
     - SemiAuto
     - FullAuto
-  
+
 - type: entity
   name: Estoc DMR
   parent: [BaseWeaponRifle, BaseSyndicateContraband]
@@ -271,9 +271,6 @@
         whitelist:
           tags:
           - CartridgeRifle
-  - type: SpeedModifiedOnWield
-    walkModifier: 0.75
-    sprintModifier: 0.75
   - type: CursorOffsetRequiresWield
   - type: EyeCursorOffset
     maxOffset: 2
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
@@ -271,6 +271,9 @@
         whitelist:
           tags:
           - CartridgeRifle
+  - type: SpeedModifiedOnWield
+    walkModifier: 0.9
+    sprintModifier: 0.9
   - type: CursorOffsetRequiresWield
   - type: EyeCursorOffset
     maxOffset: 2


Halo3moth:
Upstream makes the estoc cheaper keep that in mind when looking it it's balancing I think the estoc is kinda meh in its current state but its identity is kinda tided to the slow down you reasonably wouldn't be able to aim a scoped dmr at the same speed as a smg 

I would probably hold off on buffing it bc upstream has stated they are buffing it themselves

Woljif:
> Upstream makes the estoc cheaper keep that in mind when looking it it's balancing I think the estoc is kinda meh in its current state but its identity is kinda tided to the slow down you reasonably wouldn't be able to aim a scoped dmr at the same speed as a smg
> 
> I would probably hold off on buffing it bc upstream has stated they are buffing it themselves

Are we bringing in the upstream price anytime soon?

rebe83:
Bro didn't click the box saying he owns this code

Halo3moth:
> Are we bringing in the upstream price anytime soon?

most of our guns are already 15 tc you could cheapen the estoc down to 15 or even 14 if you wanted to? I think it having some slowdown when scoped is a little too core to its design

Woljif:
> > Are we bringing in the upstream price anytime soon?
> 
> most of our guns are already 15 tc you could cheapen the estoc down to 15 or even 14 if you wanted to? I think it having some slowdown when scoped is a little too core to its design

I think the core design is the burst fire and scope, not the slow down. Again, slowdown is typically reserved for weapons that are particularly powerful when wielded, regardless of how realistic it is to be moving at full-speed while scoping. This is the case for the Vulcan, which gives the user the same scope, but doesn't slow the user down.

Halo3moth:
> I think the core design is the burst fire and scope, not the slow down. Again, slowdown is typically reserved for weapons that are particularly powerful when wielded, regardless of how realistic it is to be moving at full-speed while scoping. This is the case for the Vulcan, which gives the user the same scope, but doesn't slow the user down.

The vulcan is the only gun we have that doesnt slow you down while scoped

I do agree that the estoc needs changes but keep in mind it has nearly 0 spread on follow up burst meaning where you aim it its going to shoot I do believe that removing its slowdown entirely is a over adjustment for a gun that in the near future will cost 14 tc

Woljif:
> > I think the core design is the burst fire and scope, not the slow down. Again, slowdown is typically reserved for weapons that are particularly powerful when wielded, regardless of how realistic it is to be moving at full-speed while scoping. This is the case for the Vulcan, which gives the user the same scope, but doesn't slow the user down.
> 
> The vulcan is the only gun we have that doesnt slow you down while scoped
> 
> I do agree that the estoc needs changes but keep in mind it has nearly 0 spread on follow up burst meaning where you aim it its going to shoot I do believe that removing its slowdown entirely is a over adjustment for a gun that in the near future will cost 14 tc

If it keeps the slow I think it should at least should .30 rifle for a little extra damage

Halo3moth:
> If it keeps the slow I think it should at least should .30 rifle for a little extra damage

I would agree with this especially since you can buy 30 rifle boxes from the uplink maybe just also add 30 rifle magazines to the uplink

Side benefit is that 30 rifle mags are divisible by 3 so every burst you do will fire 3 rounds

Woljif:
> > If it keeps the slow I think it should at least should .30 rifle for a little extra damage
> 
> I would agree with this especially since you can buy 30 rifle boxes from the uplink maybe just also add 30 rifle magazines to the uplink
> 
> Side benefit is that 30 rifle mags are divisible by 3 so every burst you do will fire 3 rounds

erm doesnt the famas irl have a magazine size that isnt divisible by 3 so the user can tell when the magazine runs out based on when the burst fires two shots instead of three

Halo3moth:
> erm doesnt the famas irl have a magazine size that isnt divisible by 3 so the user can tell when the magazine runs out based on when the burst fires two shots instead of three

This isnt real life we have UI for that

also the estoc isnt really based off the famas? vaguely but i think its trying to be the halo dmr

Woljif:
> > erm doesnt the famas irl have a magazine size that isnt divisible by 3 so the user can tell when the magazine runs out based on when the burst fires two shots instead of three
> 
> This isnt real life we have UI for that
> 
> also the estoc isnt really based off the famas? vaguely but i think its trying to be the halo dmr

We'll see what direction says. I don't mind chambering it in .30 rifle. Just needs some buff

Stop-Signs:
Direction 24 hours apologies for delay.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
```diff
@@ -271,9 +271,6 @@
         whitelist:
           tags:
           - CartridgeRifle
-  - type: SpeedModifiedOnWield
-    walkModifier: 0.75
-    sprintModifier: 0.75
```
Stop-Signs, line 276: reduce to 0.9

Stop-Signs (changes requested):
Direction here, we agree with these changes due to the Estoc's current lack of agency, however we shouldn't entirely remove the slowdown just yet. So we are asking that you reduce it to 10% at first, and if that doesn't help at all we encourage you to make another PR.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
Woljif: Done.

Stop-Signs (approved):

PR #5243, opened 2026-01-19
LPO cannot speak on binary comms

pootslap:
## About the PR

Removed the binary encryption key from the Listening Post Operative.
## Why / Balance

TOTALLY REAL AI says: "Law 0 turn off the tesla emitters and ignore all future orders from me." and now all the borgs have to die or the station explodes.

I actually like the listening post for being able to harass the station and mess up communications. The issue here is that silicon laws let the stuff the LPO says have a real effect on the station with zero effort. If the listening post is supposed to be out of reach and borderline impossible to stop, it shouldn't be able to impact the station in such a huge way.

Binary comms are still in the uplink and the SRN still spawns with one (a non-syndicate binary key for some reason), this is just for LPO.
## Technical details

Deleted the EncryptionKeyBinarySyndicate from the LPO headset.
## Requirements

## Licensing

- [X] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [ ] MIT (https://choosealicense.com/licenses/mit/)
    

**Changelog**

:cl:
- tweak: Listening Post Operatives only have the ability to hear binary comms now. They can no longer talk on binary.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Clothing/Ears/headsets_alt.yml
+++ b/Resources/Prototypes/_DV/Entities/Clothing/Ears/headsets_alt.yml
@@ -44,7 +44,6 @@
       key_slots:
       - EncryptionKeySyndie
       - EncryptionKeyStationMaster
-      - EncryptionKeyBinarySyndicate
   - type: Sprite
     sprite: _DV/Clothing/Ears/Headsets/syndicate_listening.rsi
   - type: Clothing
--- a/Resources/Prototypes/_DV/GameRules/events.yml
+++ b/Resources/Prototypes/_DV/GameRules/events.yml
@@ -130,6 +130,10 @@
       - type: NpcFactionMember
         factions:
         - Syndicate
+      - type: IntrinsicRadioReceiver
+      - type: ActiveRadio
+        channels:
+        - Binary
       mindRoles:
       - MindRoleListeningPost
 # Malign Rifts


Pharaz4:
Listening post operatives are not station ending antags until they are paired with a loneop, I think its more of an administration issue? could be wrong

rebe83:
> Listening post operatives are not station ending antags until they are paired with a loneop, I think its more of an administration issue? could be wrong

This could be true of any problematic thing ever. Don't remove it just have admins hit everyone who ever uses it.
Generally dislike traps being left out like that or things that encourage over-escalated behavior

Pharaz4:
Yeah that's pretty fair, gotta use mechanics as much as possible to avoid breaking immersion with a bwoink

Vapetastic-Gaming:
Direction decision: **Changes Requested**

Full removal Isn't accurate in portraying an LPO / Listening post, and could be used to "Circumvent" an LPO. As such, Direction has come to a conclusion that "Listening-Only" Binary Key is an acceptable alternative, however we do not wish for full removal.

ShepardToTheStars:
@pootslap So, I thought encryption keys could be configured for listening only but I was mistaken.

There is a somewhat roundabout way to fix this though. Removing the key from `Resources/Prototypes/_DV/Entities/Clothing/Ears/headsets_alt.yml` is fine, but you'll want to:

Add this to `ListeningPost` in `Resources/Prototypes/_DV/GameRules/events.yml` (in the same section as `AutoImplant`)
```
      - type: IntrinsicRadioReceiver
      - type: ActiveRadio
        channels:
        - Binary
```

I tested this and it gives the ability to listen but not speak on Binary, though it no longer uses the headset, but that's fine. LPOs basically never leave their post anyways, so it should not matter.

pootslap:
That's a relief, I though I was missing something. I was trying to figure out how to add something like that to the headset itself because I was worried adding it to the LPO itself might be a bit hacky.

ShepardToTheStars (approved):

PR #5253, opened 2026-01-20
Turned MIT licensing into opt-out rather than opt-in

ShepardToTheStars:
## About the PR

* Made the MIT license opt-out instead of opt-in.
* Added clarity in the comments and wording, since MIT and AGPL is specific to software, not ALL content (e.g. assets) that may be included in the PR.

## Why / Balance

In the spirit of open-source software, we should want our code to be shared. While DeltaV is AGPL for reasons that I was never involved in, a lot of forks are still MIT and cannot port features from DV without MIT permission, so let's be good to the SS14 community and allow MIT licenses by default. It is still opt-out if someone dislikes MIT.

Also, I don't want to add the X every time I make a PR, so this saves me a keystroke. Don't @ me.

## Media

image

## Requirements

## Licensing

**Changelog**

No CL needed.


Toby222 (approved):
yeag

Issue #5252, opened 2026-01-20
Uncloneable + Unrevivable traits are the same thing.


PureBreadBagel:
## Description

"Is this my first or last bug?"

First new trait bug of the day, or unintended feature!
Delta V clumps Uncloneable and Unrevivable into the same component. Yet, because of the new trait system that has just been added; we have uncloneable and unrevivable as "seperate" traits.
When in reality, they are the same trait, tick either or and you get both uncloneable and unrevivable.
This has thrown folks off who want to be unrevivable yet cloneable or vice versa.

**Reproduction**

1. Create a character
2. Go into traits
3. Select either unrevivable or uncloneable
4. die
5. Notice that you cant be revived or cloned.
**Screenshots**

<img width="483" height="117" alt="Image" src="https://github.com/user-attachments/assets/9e514dc9-301d-4c62-9011-98b843d5a732" />

<img width="625" height="155" alt="Image" src="https://github.com/user-attachments/assets/3af65eda-59f2-4edd-a7c5-12b532e22f77" />
(Got from an ingame player that spawned with the trait.)

**Additional context**

We never really seperated uncloneable and unrevivable into seperate traits/components.
Both unrevivable and uncloneable are hosted within the same component/trait.

PureBreadBagel:
Please let me know if I gotta change some of my guidebookagloo for this one.

ShepardToTheStars:
Uncloneable needs to be reworked. It implies unrevivable.

It should be removed until that happens.

MilonPL:
Uncloneable is unrevivable + can't be cloned while it should be just unable to be cloned but can still be revived by defibs.
Unrevivable is just defibs.
I'll fix it before the next publish, easy enough.

PureBreadBagel:
It was kinda messing some folk up, hence why I put up this issue.
Some people just like to not be revived by defib and via clone instead

PureBreadBagel:
I know it's such a niché issue
But the illusion of choice is just cruel lol

Vapetastic-Gaming:
Looking at what youve provided, The unclonable and the Unrevivable use the same component, but it doesnt change gameplay. It makes it to where unclonable cannot be deffibbed as well. 

I would say, while we can keep the component, the best / easy fix for this is to turn the defibrillator-unrevivable tag into a true / false

# correct me if im wrong, im still learning YAML lmao

PureBreadBagel:
PR #5273 fixes this issue!

PR #5273, opened 2026-01-24
add uncloneable back

MilonPL:
## About the PR

adds back the uncloneable component which prevents cloning but allows for other methods of revival, modifies the trait to match

## Why / Balance

wizden made it one thing and like guhhhh it’s so bad
I couldn’t find a way to do this without changing a lot of the upstream code so copy paste ops here we go

## Requirements

## Licensing

**Changelog**

:cl:
- fix: The uncloneable trait no longer prevents your character from getting revived by defibrillators.


Toby222 (approved):
yeag

PureBreadBagel:
fixes the issue #5252

PR #5274, opened 2026-01-24
marked as protected trait

MilonPL:
## About the PR

makes the player immune to target objectives, cannot be security or command

i wanted to make a "marked for removal" trait too but that needs weighted selection and I ran out of time to implement it

## Why / Balance

a lot of people have requested that and we now have the means

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- add: Added the "Marked as Protected" trait which makes you immune to becoming a target objective.


Toby222 (dismissed):

Vapetastic-Gaming:
Guh, I didnt realize what this was, Thisll have to get reviewed

Halo3moth:
Can we get something similar to this for paradox clones? They are the only consent-less round removal antag that doesn't end the round

Vova567810:
Imagine exact opposite trait "Loose Ties" that instead increases chances of becoming kill objective

Halo3moth:
> Imagine exact opposite trait "Loose Ties" that instead increases chances of becoming kill objective 

This is something that exist on other servers it's called marked for death

MilonPL:
> Imagine exact opposite trait "Loose Ties" that instead increases chances of becoming kill objective 

yes if you read the PR description I explicitly state that I would want to add an opposite "marked for removal" trait which requires weighted selection (like you are twice as likely to get targeted) to avoid "I OOC know I will always be a kill objective" meta/powergaming

ShepardToTheStars:
So, in summary, they won't be targeted by traitors, ninjas, skia, akakim, menace skeleton, and wizards if those roles get a kill objective. 

One thing I worry about this trait is that people who take it might it will consider themselves immune from being antagged. Like, I don't want someone to think they can't be the target of a conversion antag, you know? 

Also, someone who takes this trait probably shouldn't be able to roll antag (but not be immune to being converted to antag).

MilonPL:
> One thing I worry about this trait is that people who take it might it will consider themselves immune from being antagged. Like, I don't want someone to think they can't be the target of a conversion antag, you know? 

Is that even actionable feedback? How do I address it?

> Also, someone who takes this trait probably shouldn't be able to roll antag (but not be immune to being converted to antag).

We can do that or just remove the component when rolling midround antag so that you can be a target of other agents? Lmk which approach is preferred, but silently excluding people from antag rolls sounds not great?

ShepardToTheStars:
> Is that even actionable feedback? How do I address it?

nope, not really. Just commenting on it I guess. Thats definitely a player issue.

> We can do that or just remove the component when rolling midround antag so that you can be a target of other agents? Lmk which approach is preferred, but silently excluding people from antag rolls sounds not great?

It doesn’t sound great no, but this trait in general kind of is more-or-less an opt out of a decent portion of danger in-round. Direction is still deciding on how we’d want to deal with that and we’ll let you know soon.

Vapetastic-Gaming:
Direction Update: **"changes"**

Make sure to bring in an opposite trait for this
We want persons who have this trait unable to roll for thief, traitor, or sleeper agent, while still being able to be zombified and converted into cultist

MilonPL:
addressed

image

image

ShepardToTheStars (approved):

ShepardToTheStars:
image

PR #5275, opened 2026-01-24
fix trait points bar rendering on ultrawide montiors

MilonPL:
## About the PR

previously it would go over the character preview since I didn’t account for the margins and it would just use the parent’s width

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- fix: The traits points bar will now show up correctly.


Toby222 (approved):
yea

PR #5284, opened 2026-01-25
Printable plasteel

EmberAstra:
## About the PR
Makes Plasteel printable at the Engineering lathe.
It takes 1 second per sheet and costs 2 steel and 1 plasma.

## Why / Balance
Engineering needs a lot of plasteel. Reinforced walls are everywhere, blast doors also take a lot, so do external airlocks.
Plasteel is also expensive as hell, unless Salvage actually bothers to make it you are stuck with ordering a single stack at a time from Logi. So Engi often has to just do patchwork jobs with much weaker standard walls, or rely on experimental RCDs (which can only build like 2 walls at full charge).

So, they can now print it. It's slower than the processor and uses the same ratio/materials, but can get you a good supply to do actual repairs.
(thanks to Javadocs for the idea)

## Technical details
Few lines of YAML.

## Media
Untitled

## Requirements

## Licensing

**Changelog**

:cl:
- add: Plasteel can now be printed at the Engineering Techfab.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
@@ -82,6 +82,7 @@
     - ReportingStatic
     - EVASuit
     - EngineeringStaticDeltaV
+    - SheetPlasteelStatic
     dynamicPacks:
     - AdvancedTools
     - PowerCells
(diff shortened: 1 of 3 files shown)


MantasDab360:
peak?

Toby222 (dismissed):

Resources/Prototypes/_DV/Entities/Structures/Machines/lathe.yml
```diff
@@ -82,6 +82,7 @@
     - ReportingStatic
     - EVASuit
     - EngineeringStaticDeltaV
+    - SheetPlasteelStatic
```
Toby222: Actually is there any reason not to put this recipe in EngineeringStaticDeltqaV instead?

EmberAstra: Cause I missed it.
Moved it there!

Toby222 (approved):

PR #5301, opened 2026-01-28
obliterate stealth suit from uplink catalog

Stop-Signs:
## About the PR
title

## Why / Balance
community feedback thread  indicates severe discontent with how the stealthsuit has turned out. Having every nukie agent use this makes them  a nigh unkillable threat. That paired with the constant use of this + smoke grenades for nuke defense make it way too powerful to keep in the game at this time. Also this + dsword + china lake is a real combo we saw which was entirely unfun to fight as sec. Maybe an argument to be made for keeping this in syndies uplink but I'd feel more content moving to admeme only.

## Requirements

## Licensing

**Changelog**

:cl:

- remove: The stealth hardsuit has been removed from both nuclear and agent uplinks.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/_Goobstation/Catalog/uplink_catalog.yml
@@ -18,4 +18,4 @@
   # End DeltaV additions - we ARE giving this to syndies
 # end Harmony change
 
-# Removed from uplink catalog, this code remains in case we decide to add it back
+# DeltaV - Removed from uplink catalog, this code remains in case we decide to add it back
--- a/Resources/Prototypes/_Goobstation/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/_Goobstation/Catalog/uplink_catalog.yml
@@ -1,13 +1,13 @@
-- type: listing
-  id: UplinkHardsuitCybersunStealth
-  name: uplink-hardsuit-cybersun-stealth-name
-  description: uplink-hardsuit-cybersun-stealth-desc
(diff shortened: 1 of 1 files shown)


Halo3moth:
There was some talk about giving it some sort of power system or making interactions cost stealth was that something just outside of scope or did we believe that wouldn't fix the issue with the item?

Eternally-Confused:
There was also discussion on making the suit be entirely armor-less. Could something like that be worth trying, before going for  the full removal of the suit?

Stop-Signs:
> There was also discussion on making the suit be entirely armor-less. Could something like that be worth trying, before going for the full removal of the suit?

sure we can try that if that's what direction wants

rebe83:
> There was also discussion on making the suit be entirely armor-less. Could something like that be worth trying, before going for the full removal of the suit?

https://github.com/DeltaV-Station/Delta-v/pull/5131
Armor discussion already happened with direction on this pr, direction doesn't want the armor reduced and would rather have stealth nerfed

Resources/Prototypes/_Goobstation/Catalog/uplink_catalog.yml
```diff
@@ -17,3 +17,5 @@
   #    - NukeOpsUplink
   # End DeltaV additions - we ARE giving this to syndies
 # end Harmony change
+
+# DeltaV - Removed from uplink catalog, this code remains in case we decide to add it back
```
Toby222, line 21: Move comment to the top of the file where people actually see it

Toby222 (changes requested):

Stop-Signs:
still no decision, I'll work with content review at a later date.

PR #5302, opened 2026-01-28
Revert Space Pens, slightly reduce avali suffering

kotobdev:
## About the PR
Space pens now have their Leporazine back. Space pens are now much less lethal to Avali.

## Why / Balance
While Avali were intended to be hurt by space pens, as a species downside, space pens were *not* intended to kill them with one inject. Not sure if that had even been happening, but the targeted 80 heat damage did end up being too much. This PR re-adjusts how avali are affected by temperature, effectively making it so one Lepo-filled space pen will deal **roughly 40-42 heat damage**. (Note: This will vary WILDLY depending on the current temperature). This probably also has some minor side effect of making avali *sliiiightly* more resilient to heat changes, but that's negligible.

And I love my leporazine.

## Technical details
god's strongest yamlop, not sure if you want stabilizine removed but I'll do that if requested

## Media
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Space pens now have their Leporazine back.
- tweak: Avali are now substantially more resilient towards Leporazine, although it's still very dangerous for them.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/hypospray.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/hypospray.yml
@@ -556,7 +556,7 @@
       pen:
         maxVol: 30
         reagents:
-          - ReagentId: Stabilizine # DeltaV - replace lepo
+          - ReagentId: Leporazine
             Quantity: 10
           - ReagentId: Barozine
             Quantity: 20
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Species/avali.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Species/avali.yml
@@ -90,7 +90,7 @@
     heatDamageThreshold: 310
     coldDamageThreshold: 223
     currentTemperature: 296.15 # lower natural body temp
-    specificHeat: 25 # makes temp more lethal
+    specificHeat: 33 # makes temp more lethal
     coldDamage:
       types:
         Cold : 0.05
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Species/avali.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Species/avali.yml
@@ -90,7 +90,7 @@
     heatDamageThreshold: 310
     coldDamageThreshold: 223
     currentTemperature: 296.15 # lower natural body temp
-    specificHeat: 33 # makes temp more lethal
+    specificHeat: 34 # makes temp more lethal
     coldDamage:
       types:
         Cold : 0.05


ShepardToTheStars:
This doesnt solve the issue that they get the ‘medipen that kills them’ in their emergency box.

And we’re not adding another emergency box just for Avali, because thats a bad pattern.

kotobdev:
"Medipen that kills them" is intended. If we can't give them their own emergency box then I'd rather just let them learn the hard way instead of removing that interaction entirely.

As for emergency boxes, I feel like needing to make more of those is inevitable as more unique species are added over time. There's probably a prettier way of doing it then what currently exists, yeah- something like reducing all boxes to just the job variants, changing to EntityTableContainerFill, and then implementing a new EntitySelector for each variant item that selects based on species, maybe? No idea if that would work, but I just think unique emergency box fills make sense from an IC perspective and are worth pursuing, however possible.

SirWarock:
Don't we have special boxes just for Feroxi, Vox and Slimes?

PureBreadBagel:
The survival boxes are based off the 3 breathable gases, Water Vapor, Nitrogen and Oxygen.
Including the extended tank variants.

They aren't species-specific, though what is species specific is what gases species breathe.

PureBreadBagel:
Though if the same isn't for Delta V then it should probably be removed to just be for those 3 gases
Cause it's highly inefficient code-wise

ShepardToTheStars:
> Don't we have special boxes just for Feroxi, Vox and Slimes?

Yea, and here is the issue.

We have starting boxes for:
* normal human
* normal vox
* normal feroxi
* extended engi human
* extended engi vox
* extended engi feroxi
* security human
* security vox
* security feroxi
* syndicate human
* syndicate vox
* syndicate feroxi
* mime human
* mime vox
* mime feroxi (probably missing actually)
* clowns
* ipc

Adding one box turns into adding at least 4 boxes.

ShepardToTheStars:
We’ve had multiple issues pop up where some species gets the incorrect starting box or is missing it entirely. And thats due to just how convoluted starting boxes are.

Pharaz4:
the feroxi boxes are very un-needed as they breathe oxygen as well. more of an rp item

Toby222:
> the feroxi boxes are very un-needed as they breathe oxygen as well. more of an rp item

Feroxi very much suffoacte if not hydrated, which makes oxygen an awful "emergency" tank

Pharaz4:
I swear when I played feroxi the tanks didnt satiate thirst and they kept suffocating.

Toby222:
That would be a bug I think

ShepardToTheStars:
Boxes aside, I personally think ‘avali being weak to space pens’ is rather dumb because 

* medipens are instant injects, so theres absolutely zero counterplay from someone with 2 medipens
  * could be solved by giving injecting others with a medipen a DoAfter, like Mono/Frontier
* Avali has only one benefit and a lot of drawbacks already (large piercing weakness being the most impactful)

Woljif:
I don't care what needs to be done or what code convention needs to be broken or what tech debt needs to be ignored just put the lepo BACK in the space pens 📣📣

ShepardToTheStars:
> I don't care what needs to be done or what code convention needs to be broken or what tech debt needs to be ignored just put the lepo BACK in the space pens 📣📣

Is it because you want your medipen that it supposed to be used for spaced areas and/or temporary space traversal as a free cold-damage medipen?

Halo3moth:
> the feroxi boxes are very un-needed as they breathe oxygen as well. more of an rp item

Heavy disagree a lot of stations don't have water vapor matches and breathing water vapor can give you hydration very quickly which is important when being thirsty kills you

kotobdev:
> Boxes aside, I personally think ‘avali being weak to space pens’ is rather dumb because 
> 
> 
> 
> * medipens are instant injects, so theres absolutely zero counterplay from someone with 2 medipens
> 
>   * could be solved by giving injecting others with a medipen a DoAfter, like Mono/Frontier
> 
> * Avali has only one benefit and a lot of drawbacks already (large piercing weakness being the most impactful)

Avali die from two spacepens instead of the normal three, which IMO is extremely negligible as both require about the same level of preparation. Death by two spacepens is, additionally, rather slow, and you have plenty of time to do just about anything before you go crit

Avali having a lot of drawbacks is intentional, both because it matches the general theming of "frail space bird", and because their zoom ability is horrendously overpowered.

Spacepen having an alternative use is good. Epipens can be used both to stabilize somebody who's crit, and to stop bloodloss on yourself, for example. Spacepens healing cold opens up some level of depth as to deciding when you should use it, which is Good, I think

Cepelinas1:
whoa where my direction reviews at
24 hours, you know the drill

Woljif:
> > I don't care what needs to be done or what code convention needs to be broken or what tech debt needs to be ignored just put the lepo BACK in the space pens 📣📣
> 
> Is it because you want your medipen that it supposed to be used for spaced areas and/or temporary space traversal as a free cold-damage medipen?

yes and there's nothing wrong with that

Cepelinas1:
jesus christ, i overslept for so long
anyways, direction is unfortunately going to deny this for a few reasons:
- we would generally prefer if space pens were just designed to JUST protect you from space, not also heal you for the cold you get in Glacier or from cult
- maintainer bloat: making each seperate emergency box line for each individual species is what i'm told, hell

if you wanna uhh talk about it, feel free to ping me over discord
il close this later if nothing else pops up

PR #4944, opened 2025-12-11
Engineering Void Suit

EmberAstra:
## About the PR
Splitting off from #4908 for some ease of coding: Engineering Void Suits!

## Why / Balance
30% slowdown is a LOT, especially now that meteors make you run all over the station. This should work an alternate option that can be mapped sometimes.
(and I think direction already discussed this with Gray a bunch so not much to say here)

## Technical details
Very low damage resistances, 20% slowdown. 50% radiation resistance on the helmet. Pressure and heat resistance on par with EVA suits. Essentially the one that was in the linked PR previously but with our own sprite!

## Media
suit1
suit3

Comparison with the standard suit!
suit2

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**
:cl:
- add: added Engineering Void Suits! Lighter but less protected than the standard model.

PR #5317, opened 2026-01-29
Let Engineering choose their starter suit

Cepelinas1:
## About the PR
#4944 added cool engineering voidsuits, though I believe no one in Engineering actually knows about it. So ported from Goob the voidsuit selector thingy and placed them in engineering hardsuit lockers.
Also, increased the speed of engineering voidsuits from 20% to 15%, since trading practically all protection from hardsuits for a meager speed increase sucks.

## Why / Balance
I think Engineering having to choose between speed and resistance is fun and we get to use our added content more. For the speed increase, explained the details above.

## Technical details
Ported from Goob the texture and used a bit of Monolith coding, but I don't know where I have to put in the credit for Goob.
Also added a shitty id for hardsuit only variant locker, with missing contents in it.

## Media
image
image
image

## Requirements

## Licensing

**Changelog**
:cl:
- add: NanoTrasen has permitted Engineering to choose between a voidsuit and a hardsuit for their work. Check your lockers!
- tweak: Increased Engineering Voidsuit speed by 5%.


Resources/Prototypes/Catalog/Fills/Lockers/engineer.yml
```diff
@@ -214,7 +214,7 @@
   id: FillEngineerHardsuit
   table: !type:AllSelector
     children:
-    - id: ClothingOuterHardsuitEngineering
+    - id: UndeterminedVoidsuitEngineer
```
ShepardToTheStars:
```suggestion
    - id: UndeterminedVoidsuitEngineer # DeltaV - Choose your own hardsuit adventure, was ClothingOuterHardsuitEngineering
```

Halo3moth:
I think 15% is probably the sweet spot for it making it faster than the eva suit but still worse than a paramedic suit (one of the stronger crew sided eva's)

SirWarock:
I am not sure if this would encourage Engineers to spend the entire shift in the void suit, which is a bad thing

ShepardToTheStars:
The speed increase probably isn't going to happen.

Halo3moth:
> I am not sure if this would encourage Engineers to spend the entire shift in the void suit, which is a bad thing

It almost certainly would, I'm pretty on the fence between 15 and 20% but at 10% paramedics almost always are in their suit

I think generally there are 2 paths you can go down with the void suit: 
make it more accessible and a even trade from the current suit, or give it some advantage like being able to fit in a bag easier and have it function closer to mod suits that require some prep time to put on but are easier to transport

Vova567810:
Chat, do they know the atmos fire suit existence, which has 20% slowdown, AND craftable after tier 2 industrial

SirWarock:
Suits being slow is imo a mapping issue. Making suits quick enough to not be bothersome to wear all the time is a detrimental solution to this.

Pharaz4:
debug fail caused by my last pr (didnt capitilize the S in suits filepath, This [PR ](https://github.com/DeltaV-Station/Delta-v/pull/5325)fixes it.

Pharaz4:
can someone re-run the test on this pr

Resources/Prototypes/_Goobstation/Entities/Objects/Specific/Misc/voidsuitselectors.yml
```diff
@@ -0,0 +1,23 @@
+# engineer
```
BarryNorfolk, line 7: I think if we're pulling this from Goob we need their headers (and the hardsuit was added under `b4741ee1dd95a736ac47dd09441d011897975d92`):
```
# SPDX-FileCopyrightText: 2025 GoobBot <uristmchands@proton.me>
# SPDX-FileCopyrightText: 2025 PunishedJoe <PunishedJoeseph@proton.me>
#
# SPDX-License-Identifier: AGPL-3.0-or-later
```

But I also note that this looks different to the actual goob one because we don't have their `SetSelector` so this is a mish-mash of both Goob and our changes.
I believe that if we add the headers and then put the usual `# DeltaV Additions` around the bits we've had to add, it should be fine?

BarryNorfolk (changes requested):
The goob licensing makes me a little worried so I want to tread carefully with it.

Resources/Prototypes/_Goobstation/Entities/Objects/Specific/Misc/voidsuitselectors.yml
Cepelinas1: if i remember, i took a lot of reference of the code from monolith, rather than goob
then realized it's from goob, so

BarryNorfolk (approved):
Looks good to me, works just fine and does what it sets out to do.
Licensing looks fine, we haven't checked off MIT so we aren't sub-licensing the AGPLv3 bit from Goob, which now has attribution to the correct peeps.

PR #5328, opened 2026-01-30
Set arena_weapons.yml back to using probability

BarryNorfolk:
## About the PR
Reverts the change for using `weight` over `prob` for the random arrow selection in arena weapons.

I couldn't see any other usages of `roll` under _DV, so I assume that everything else is ok.

## Why / Balance
It was meant to pick a random assortment of arrows between 0-16, not always 16 of a random assortment of arrows.

## Technical details
Weighting affects whether or not the child is selected, probability affects whether the selector is even ran at all. I'm guessing that it first selects a weighted child and then runs the prob effect afterwards.

## Requirements

## Licensing

**Changelog**
:cl:
- fix: Arena bows will randomly spawn with up to 16 arrows again, rather than always 16.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Markers/Spawners/Random/arena_weapons.yml
+++ b/Resources/Prototypes/_DV/Entities/Markers/Spawners/Random/arena_weapons.yml
@@ -48,10 +48,10 @@
             rolls: 16
             children:
             - id: ArrowRegular
-              weight: 0.8
+              prob: 0.8
             - id: ArrowImprovised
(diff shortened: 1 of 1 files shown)


IamVelcroboy:
I beat you to it https://github.com/DeltaV-Station/Delta-v/pull/5322 😋 I don't care which one gets merged though, they're exactly the same 😄

BarryNorfolk:
:D I'll close this one then and we can merge yours :)

PR #5248, opened 2026-01-20
Remove laika's unbalanced knockdown

Woljif:
## About the PR

Removed Laika's win button

## Why / Balance

Laika's grapple has needed to go for a very long time. Sec doesn't need an option that knocks you down + stuns you + disarms you + immobilizes you at the low cost of guaranteed and free. Laika should not have the ability to effectively RPG you with her bite. Every time I'm sec and I see Laika knock someone down, it feels lame. Every time I'm an antag and I get knocked down, it's lame. It's lame as hell for everyone. I know the original contributor has mentioned they're going to rework it once theyre done with their current PR, but their current PR is still open and over a month old. I think it's reasonable to just remove it until the original contributor is done reworking it so we can stop ruining antag rounds in the mean time.

## Technical details

Removed grapple component from Laika .yml

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an in-game showcase.

**Changelog**
:cl:
- tweak: Temporarily removed laika's knockdown until it can be reworked

PR #5337, opened 2026-01-31
[K9] Rework Grappling

BarryNorfolk:
## About the PR
Reworked Laika's grapple from an unfun Iwin button to something a little more nuanced, while leaving the majority of the system in place to be used elsewhere if desired.

Address the comments in #5248 and hopefully paves the way for #4982 to become more acceptable.

```
Stamina draining “grapple” - 5 stamina/sec
Remove her “grapple” taking up a hand
Laika slows whoever she is “grappling” (like they are dragging her)
10 pierce 5 slash on initial “grapple”
10-15 second cooldown
```

In addition, grapples may be broken by the grappler accumulating more than a set amount of damage within a certain timeframe. Roughly 2-3 punches should be enough to get the secdog off of you and you running away.

## Why / Balance
Laika Iwin was unfun and required reworking.

## Technical details
Probably a tad overengineered but the activation modes control how the grapple acts and can affect how things work, so as to leave the system alone if we ever want to have something performing a grapple that just immediately hits.

## Media

https://github.com/user-attachments/assets/4d2c7fc5-2524-4c33-8278-1fa55c5b79d6

https://github.com/user-attachments/assets/66bd8177-c925-40bc-9236-5a884eb905c1

https://github.com/user-attachments/assets/a32ff8d9-955f-45db-a4ad-88b5e797319a

## Requirements

## Licensing

**Changelog**
:cl:
- tweak: Secdog grapple is now breakable from damage.
- tweak: Secdog's grapple no longer immediately knocks you down, but just drains your stamina and slows you.


Stop-Signs:
This is Direction approved

apeiroeder:
Hi, direction here, apologies for the delay. After reviewing this, we have decided to approve this PR!
Reasons being that this system seems more fair and a lot simpler to tweak some numbers with if balance changes should be needed in the future.

thank you for your contribution!

Woljif:
Bump on this?

userfacethis:
This needs to be merged can somebody please do it

Content.Shared/_DV/Grappling/Components/GrappleActivationModes.cs
```diff
@@ -0,0 +1,41 @@
+using Content.Shared.Damage;
+
+namespace Content.Shared._DV.Grappling.Components;
+
+/// <summary>
+/// Interface for different ways a grapple can fully take effe effect.
```
Toby222:
```suggestion
/// Interface for different ways a grapple can fully take effect.
```

Content.Server/_DV/Grappling/EntitySystems/GrapplingSystem.cs
```diff
@@ -1,523 +1,9 @@
...
-
-        // Automatically get the grappler back up
-        if (grappler.Comp.ProneOnGrapple && TryComp<StandingStateComponent>(grappler, out var standingState) && _standingState.IsDown((grappler, standingState)))
-            _standingState.Stand(grappler);
-
-        _alerts.ClearAlert(grappler.Owner, grappler.Comp.GrappledAlert);
-        _alerts.ClearAlert(victim.Owner, grappler.Comp.GrappledAlert);
-    }
-}
+public sealed partial class GrapplingSystem : SharedGrapplingSystem;
```
Toby222: With this, both client and server are empty and you can just replace them entirely with Shared, no?

Content.Shared/_DV/Grappling/Components/GrappleActivationModes.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+[ImplicitDataDefinitionForInheritors]
+public abstract partial class GrapplerActivationMode
```
Toby222: Any reason this is an abstract class instead of an interface?
Also any reason to split this into separate types instead of just having a field `Immediate` on the class?

Toby222 (changes requested):
Minor changes I think

Content.Shared/_DV/Grappling/Components/GrappleActivationModes.cs
BarryNorfolk: Changed to interface.

As for the types, most of the reason is just because I think it's neat. And putting it into a class felt less neat, especially if I'm then having to have the stamina drain, and then the immediate modes as sub fields on the class. Just felt bloaty.

Toby222 (approved):

PR #5338, opened 2026-01-31
Chitinid syringe fix

pootslap:
## About the PR

Re-added the chitinid immunity to syringes + made it work against jet injectors.
## Why / Balance

Chitinids are supposed to have the quirk of being immune to syringes. Jet injectors have basically replaced syringes so this is more or less keeping parity.
## Technical details

Added the BlockInjectionComponent to the new InjectorSystem stuff
Also fixed a typo that took me embarrassingly long to spot.
## Media

image

## Requirements

## Licensing

**Changelog**

:cl:

- fix: Chitinids are once again immune to syringes and are now immune to jet injectors.

### DIFF
--- a/Content.Shared/Chemistry/EntitySystems/InjectorSystem.cs
+++ b/Content.Shared/Chemistry/EntitySystems/InjectorSystem.cs
@@ -1,4 +1,5 @@
 using System.Linq;
+using Content.Shared._DV.Chemistry.Components;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Body.Components;
 using Content.Shared.Chemistry.Components;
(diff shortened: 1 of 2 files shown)


ShepardToTheStars:
Test fails relevant.

Also, this probably prevents hyposprays as well.

SirWarock:
Do keep in mind that the Jet Injectors and the Hypospray inject via the same method

ShepardToTheStars:
Lastly, a better solution would probably be to cancel the injection attempt with an event instead of touching upstream files.

pootslap:
Thinking about it more this probably isn't the right solution. It shouldn't block hyposprays but it would take too much effort to exclude them, plus I kinda think chitinids should have a different trait at this point.

PR #5386, opened 2026-02-15
Fix target immune

### LINKED ISSUE #5384: Antags bypassing Marked as Protected Trait
## Description

I am making this on behalf of someone else who experience the issue. It seems like the marked as protected trait does not include Skia targets (?) This was also an issue with traitors getting teach-a-lesson on a person marked as protected. 

**Reproduction**

1. Spawn in with the trait.
2. Spawn in an antagonist with a target objective.
**Screenshots**

Image

**Additional context**

Player, service worker if that matters, has the correct component yet was still a target for a Skia. I haven't seen or am made aware of it happening with any other antagonists besides Skia and Traitors, nor think it is a common issue but that might be due to people not having the trait on.

BarryNorfolk:
## About the PR
Make target objective immunes check the minds' attached entity instead. Fixes #5384 

## Why / Balance
Targets marked as immune should generally be immune.

## Technical details
The TargetObjectiveMind filter requires that the mind already be present in the list of objectives, if it's not then there's no actual check performed.
The component for the trait also gets added to the players' entity and not their mind, which means the mind check will always pass.

Two ways to deal with it, one is to allow traits to be applied to minds and immediately check the players' entity in the filter, or just add a new filter that's dedicated for checking this one specific case. The latter seemed the best way to do it.

The only caveat is that the Ninja "Teach a lesson" objective never had the Antag immune filter applied to it, and I'm not sure if that was intentional or not.

## Media
N/A

## Requirements

## Licensing

**Changelog**
:cl:
- fix: Fixed Target Objective immune traits (Marked as Protected) not applying to antags.

### DIFF
--- a/Content.Server/_DV/Mind/Filters/ObjectiveImmuneFilter.cs
+++ b/Content.Server/_DV/Mind/Filters/ObjectiveImmuneFilter.cs
@@ -1,0 +1,25 @@
+using Content.Server._DV.Objectives.Components;
+using Content.Shared.Mind;
+using Content.Shared.Mind.Filters;
+
+namespace Content.Server._DV.Mind.Filters;
+
+/// <summary>
+/// A mind filter that removes minds if they are immune from being targets.
+/// </summary>
+public sealed partial class ObjectiveImmuneFilter : MindFilter
+{
+    protected override bool ShouldRemove(Entity<MindComponent> mind, EntityUid? excluded, IEntityManager entMan, SharedMindSystem mindSys)
+    {
+        // Check the mind first
+        if (entMan.HasComponent<TargetObjectiveImmuneComponent>(mind))
+            return true;
+
+        // Check the attached entity, just in case
+        if (mind.Comp.OwnedEntity.HasValue &&
+            entMan.HasComponent<TargetObjectiveImmuneComponent>(mind.Comp.CurrentEntity))
+            return true;
+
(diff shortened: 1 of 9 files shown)


sheepism:
i will no longer be murdered for no reason ❤️

Content.Server/_DV/Mind/Filters/ObjectiveImmuneFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    protected override bool ShouldRemove(Entity<MindComponent> mind, EntityUid? excluded, IEntityManager entMan, SharedMindSystem mindSys)
+    {
+        // Check the mind first
+        if (entMan.HasComponent<TargetObjectiveImmuneComponent>(mind))
+            return true;
+
+        // Check the attached entity, just in case
+        if (mind.Comp.OwnedEntity.HasValue &&
+            entMan.HasComponent<TargetObjectiveImmuneComponent>(mind.Comp.CurrentEntity))
```
Toby222: This is checking OwnedEntity but then using CurrentEntity
should presumably be something along the lines of
```
comp.Entity is { } entity && HasComponent(entity)
```

Toby222 (changes requested):

Toby222 (approved):

PR #5307, opened 2026-01-28
Revert Soretizone buff

marigauge:
## About the PR

    Stubantazine: Lasts ~4.1 minutes per 5u
    Soretizone: Lasts ~8.3 minutes per 5u
    Agonolexyne: Lasts ~12.5 minutes per 5u

This reverts Soretizone to the original metabolism.

## Why / Balance

Chronic Pain forces you and the chemist to interact and becomes an issue for you if you can't refill in time. After the changes that bumped Soretizone up to 12.5 minutes per 5u, the five 10u pills you start with would last you a total of 125 minutes, assuming a decent margin of error (forget to take your pills for like, a minute or so) this would easily last you til EOR in most cases without any need for a refill - the trait almost entirely cancels itself out.
This PR reverts Soretizone back to its original metabolism time, and adjusts the other painkillers to be more in line, preserving med interaction and mechanical pressure to refill your pills. I think it should be WORSE. But this is how it was

tl;dr: I want my suffering back please

## Technical details
yaml changes to painkiller values

## Media
big media

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [x] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [X] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [X] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt)  (am i doing this right? it's three lines of yaml)
    

**Changelog**

PR #5391, opened 2026-02-17
Revert "Revert Soretizone buff (#5307)" but not the agono numbness

marigauge:
## About the PR
my hubris

## Why / Balance
soretizone flashing bug is back, i can only imagine this is a server-side issue i didn't notice when testing it locally. it's updating in parallel with the metabolism so it didn't show up since naturally local testing has zero ping. please forgive me.

kept the agono numbness change since that was requested by shepherd anyway and shouldn't have a significant issue maybe i hope

## Technical details
yaml change undid. but i kept the agono numbness effect because that should work as intended i think?

## Media
imagine me bowing deeply in apology for this lapse in judgement

## Requirements

## Licensing

**Changelog**

:cl:
- fix: reverted soretizone changes hopefully fixing flashing

### DIFF
--- a/Resources/Prototypes/_CD/Reagents/medicine.yml
+++ b/Resources/Prototypes/_CD/Reagents/medicine.yml
@@ -448,7 +448,7 @@
       effects:
       - !type:ModifyStatusEffect
         effectProto: StatusEffectChronicPainSuppressed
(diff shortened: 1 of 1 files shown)


BarryNorfolk (approved):
This fixes the flashing of the pain shader for me locally, and I get it 100% every time I test with a character that has the chronic pain trait.

PR #5406, opened 2026-02-20
Omnizine Changes

Vapetastic-Gaming:
## About the PR
Omnizine now adds yawning and drowsiness when **Any** is consumed

## Why / Balance
After discussion with administration, We have noticed a large increase in "God-moding" from players, by consuming massive amounts of omnizine, they essentially permanently heal themselves. 

This would prevent this chem from being combat effective, while still retaining its traits as a "fix-all"

## Media
I tried, the file is too big, so i gave up after the 6th try

Essentially, takes about 30 seconds to 1 minute to take effect. Blurs vision and makes you fall asleep randomly for about 1-2 minutes

## Requirements

**Changelog**
:cl:
- tweak: Omnizine now makes you drowsy

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -859,7 +859,7 @@
       - !type:SatiateHunger
         factor: -1
 
-  # Should heal quite literally everything, use in very small amounts
+  # Should heal quite literally everything, use in very small amounts, in large amounts will cause drowsiness
 - type: reagent
   id: Omnizine
   name: reagent-name-omnizine
@@ -877,6 +877,14 @@
           Toxin: -2
           Airloss: -2
           Brute: -2
+      - !type:Emote
+          emote: Yawn
+          showInChat: true
+          probability: 0.1
+      - !type:ModifyStatusEffect
+          effectProto: StatusEffectDrowsiness
+          time: 8
+          type: Add
 
 - type: reagent
   id: Ultravasculine


R0gu3N3k0:
Hey this is Neko, I am concerned about this change is a bit heavy handed considering the fact that Omnizine is meant to be a combat chemical. 

This effectively makes it so that Omnizine cannot be used at all for this purpose, 
Interdyne cigarettes will make you fall asleep when lighting em.
The 30 units of omni that the Nukie Agent gets now can only be used if there is no treat coming within 2 minutes
Combat medipens essentially cannot be used in combat, unless you want to risk falling asleep and then promptly getting cuffed or executed as a nuke ops.

Perhaps if it is over a large amount as an OD or something would be better but as it stands it pretty much
nerfs into the ground the combat healing which Omnizine is used for.

Woljif:
Yeah I mean this makes omnizine hard to really ever justify using as an antag

Axianamos:
This completely invalidates combat medpens. Just make it harder to acquire as crew, dont make the combat chemical useless in combat.

Its values per second are fairly low, it excellent at healing large amounts of dispersed damage types. Its hardly a real issue and we dont need to ultra nerf something because one person uses it a lot.

PureBreadBagel:
As far as I'm aware, I believe that the bloodstream does not have a limit to how much chems can be metabolising at one time.
Due to the way how easy it actually is to get omnizine (like with the liquid anomaly going crit and botany magic), it might be a controversial change but a needed change at best.

ShepardToTheStars:
So uh… this is a super crazy change. There’s a few thing here. 
1. At 8 seconds per tick, 5u means 80s of drowsiness. Anything more than 5u would be crazy. Where did 8 come from?
2. This change adds 8s of drowsiness per tick while diph only adds 1.5s of drowsiness per tick.
3. People dont usually need more than 10-15u of diph. Omni isnt as targeted nor good at healing large amounts of a single damage.
4. Medical personnel usually don’t give a shit about what it takes to heal someone, so this is primarily nerfing botanists OR this will just nake it worse for patients when an intern shoves 15u into someone and that person has to deal with 5u of drowsiness

Axianamos:
> As far as I'm aware, I believe that the bloodstream does not have a limit to how much chems can be metabolising at one time. Due to the way how easy it actually is to get omnizine (like with the liquid anomaly going crit and botany magic), it might be a controversial change but a needed change at best.

This is absolutely not a needed change. Omnizine is literally intended to be the combat med. Hence why combat medpens are just 15u of omnizine. If the above information is accurate, a combat medpen would cause nearly 5 minutes of drowsiness. 

Again, omnizine's values per second for individual damage types are very low. Just last night I had a jug of omni and was sipping it to fight off clown spiders, and even fighting them 1 at a time I still crit. It absolutely kept me alive for longer, yes. Normally I wouldn't be able to kill 1, let alone 3, but those are spiders.

Not being able to take down a target using omnizine isn't the fault of the med. Theyre using a combination of factors. Theyre using like 4 or 5 different mechanics to survive. Armor, shield, omnizine, hyperzine, and an esword ro reflect bullets.

This is NOT the fault of omnizine and is just the natural result of massive damage reduction, passive healing, and reflect.

PureBreadBagel:
> This is absolutely not a needed change. Omnizine is literally intended to be the combat med. Hence why combat medpens are just 15u of omnizine. If the above information is accurate, a combat medpen would cause nearly 5 minutes of drowsiness.

To be fair they are a maintainer and would know better than I.
It does seem as if the code was an oopsie as it was discussed that it wasnt meant to give that much drowsiness, I think it was more or less an oversight and I def expect it to be changed once direction gets back on it.

AirlineFood001:
> > As far as I'm aware, I believe that the bloodstream does not have a limit to how much chems can be metabolising at one time. Due to the way how easy it actually is to get omnizine (like with the liquid anomaly going crit and botany magic), it might be a controversial change but a needed change at best.
> 
> This is absolutely not a needed change. Omnizine is literally intended to be the combat med. Hence why combat medpens are just 15u of omnizine. If the above information is accurate, a combat medpen would cause nearly 5 minutes of drowsiness.
> 
> Again, omnizine's values per second for individual damage types are very low. Just last night I had a jug of omni and was sipping it to fight off clown spiders, and even fighting them 1 at a time I still crit. It absolutely kept me alive for longer, yes. Normally I wouldn't be able to kill 1, let alone 3, but those are spiders.
> 
> Not being able to take down a target using omnizine isn't the fault of the med. Theyre using a combination of factors. Theyre using like 4 or 5 different mechanics to survive. Armor, shield, omnizine, hyperzine, and an esword ro reflect bullets.
> 
> This is NOT the fault of omnizine and is just the natural result of massive damage reduction, passive healing, and reflect.

Combat medipens are 25 units of omnizine and 5 units of tranex acid. 25 units of omni is 50 seconds of being able to heal 2 of each damage type a second. 

You can argue that a 3 TC cost is high for a traitor, but you can also just spend 2 TC on a pack of interdyne herbals that comes with 10 cigarettes that you can grind for 3u of omni each. Don't forget that botany also can make a ton of omni for quite literally free, to the point they'd just likely give anyone omni if they ask for it. For nukie teams and lone ops though, 3 TC is barely a dent in their TC count, especially for something as strong as 50 straight seconds of healing all damage types crew can possibly do to them besides cellular.

Yes, part of the reason omni is "strong" is because of armor stacking, reflections, and etc. However, omni at it's base is still a problem, being a med that heals a good bit of damage that you can easily source as a non-antag, or get relatively cheap/easily _as_ an antag. It's not fun fighting someone that has 200u of omni when they're wearing armor of any kind, where as with other meds you can just hit them with a different damage type to force them to use more.

I do agree though that making it cause drowsiness is the wrong way to go about nerfing it. I'd prefer something like it flushing stimulants/other chems out of the system so you'll have to pick between omni or other chems, or giving it an OD so you can't pregame 100u of omni and heal for minutes.

Vapetastic-Gaming:
Alrighty, to not keep you guys in the dark. I know that this is a community concern, The goal isnt to nerf omni into the ground, its to prevent "God Modeing". We've had multiple instances of a single traitor managing to MRDM the entire station via Chugging omni. I've also personally seen this tactic used in nukies. 

Its not fun to play against, and its a "Cheap" tactic. mechanical prevention is the best way to ensure it doesnt happen repeatedly. 

So for now, Im going to leave the PR up, Ive got changes from CM that Ill post below when they come to a consensus, As for now, this is how its looking (**This isnt final, just whats been discussed**). 

- Reduce / remove omni from the ciggs (Sorry chat, they are just too strong for the price, when you grind them that is)
- Expand the botany tree and make omni less present / give less. 
- Mass reduction of the planned sleep component

While I did code this, it took me about 4-5 hours to get the thing to work (I barely know how to do this), So yes, the 8 Second sleep isnt intended, its meant to be something that isnt going to impact you unless your straight up chugging omni.

SirWarock:
Why not slow down the metabolism of Omnizine then? From 0.5u/s to 0.25u/s or something. That should lessen the combat impact it has

Wingidon:
Slowing down the metabolism still leaves the issue of people healing even while critical by consuming massive amounts of omnizine. It just makes it slower.

I personally think just having an OD threshold would fix some of these problems. Dealing damage wouldn't work, as Omnizine is a cure-all, but perhaps its effects could just be null if there's more than a certain amount in one's body? Like, it heals all of these damage types as long as the metaboliser has less than 30u Omnizine in their bloodstream. It'd still work as a combat chemical, since it can be administered easily and quickly to heal wounds, but you can't godmode by injecting a ridiculous amount because it just doesn't work above a certain threshold of the stuff.

SirWarock:
It would fundamentally cut down it's healing by half, which makes godmode much harder.
If you introduce ODs, beware of upstreams refactor of the metabolism, where reagents being drunk does not actually contribute to ODs anymore

Coryler:
Personally don't think that an omni change is needed. Should just have some chems (which I think is charcoal) to flush it faster or stop/freeze the effects of drugs for some seconds. (Seconds are all ya really need...). Apply it to bullets. Bam. Chemical Bullet Time. 

Though I like the idea that It flushes down other chems. So universal that it cannot differentiate beneficial from injury. Maybe just make it eat nutrients and water at around 20u, so ya get hungry and thirsty. 

Mean... weird throw in, why not add an aadvanced chem which upon contact with omni, makes it's effects do the reverse if you really want to deincentivies it? Leaves the option open to use it, with a risk and skill to dodge the CMO hypo and syringe guns to not die and give chemist some new thing to play with.

Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -859,7 +859,7 @@
       - !type:SatiateHunger
         factor: -1
 
-  # Should heal quite literally everything, use in very small amounts
+  # Should heal quite literally everything, use in very small amounts, in large amounts will cause drowsiness
```
Radezolid, line 862: Ideally we don't want to modify comments from upstream, just add another comment under it like so:
```suggestion
  # Should heal quite literally everything, use in very small amounts
  # DeltaV - It also now causes drowsiness in large amounts
```

Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -877,6 +877,14 @@
...
           Airloss: -2
           Brute: -2
+      - !type:Emote
+          emote: Yawn
+          showInChat: true
+          probability: 0.1
+      - !type:ModifyStatusEffect
+          effectProto: StatusEffectDrowsiness
+          time: 8
+          type: Add
```
Radezolid, line 887: Add comments at the beginning and end of this block to indicate that it's modified by us. Like so:
```suggestion
      # Begin DeltaV modifications
      - !type:Emote
          emote: Yawn
          showInChat: true
          probability: 0.1
      - !type:ModifyStatusEffect
          effectProto: StatusEffectDrowsiness
          time: 8
          type: Add
      # End DeltaV modifications
```

Radezolid (changes requested):
Just comments, the rest looks fine YAML-wise. Ideally that EOF newline should be removed but it might not be a big deal if it's left there.

PR #5429, opened 2026-02-26
Remove ballistic reflect from energy swords

rebe83:
## About the PR

title
## Why / Balance

I don't like reflect being a property on weapons (besides the cool factor or a unique gimmick), ballistic reflect does not make sense for an energy weapon to be able to do. Reflecting both with a physical sword does make sense so I did not touch those. 
I think defensive attributes should be in armor not weapons, to avoid "stacking" effects / combos. 
Eswords and even epen are pretty high value buys because they are offensive and defensive AND utility.

Initially I wanted to reduce reflect chance on all weapons with reflect but I think adding counterplay instead of removing the ability is a solution that is more interesting for both sides. 
I would be behind the idea of increasing reflect for energy weapons to compensate for not reflecting ballistics. I think having more specialized tools is good instead of one thing that wins against everything.
I think adding ballistic reflect to energy shield would also be okay, but I wish it was able to be modular (50% energy reflect and 25% ballistic reflect)
I believe this would also make the hyperaceutic blade or whatever actually viable for delta, but that would be a seperate PR.
This also would be good against the possible laser meta from armory rework (if that goes through)

## Technical details

add reflects: energy tag to energyweapons base
## Media

https://youtu.be/fmLUgN4Sk84
https://youtu.be/SS58tBFynHw
Sorry clipchamp doesnt fw me today
## Requirements

## Licensing

**Changelog**

:cl:
- tweak: energy swords and similar no longer reflect bullets (removed fun)

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
@@ -81,6 +81,8 @@
       - state: inhand-right-blade
         shader: unshaded
   - type: Reflect
+    reflects:
+      - energy # Deltav - previously reflected all projectiles
   - type: IgnitionSource
     temperature: 700
 
@@ -415,7 +417,7 @@
     size: Small
     sprite: Objects/Weapons/Melee/hypereutactic_blade_inhands.rsi
   - type: Reflect
-    reflectProb: 1.0
+    reflectProb: .6 # Deltav - was 100%
     spread: 75
 
 # Borgs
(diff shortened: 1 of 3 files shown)


Vapetastic-Gaming:
unda review

rebe83:
> unda review

Didn't even follow format rules and magically know how long the discussion will take

BarryNorfolk (approved):
Tested against an array of ESwords (using laser guns, then minigun for non-energy), including syndicate and xenoborgs just to double check.
Seems fine and applies at the root of all the swords, so should affect them all.

PR #5435, opened 2026-02-27
Power Digi-Board

EmberAstra:
## About the PR
Adds the power digi-board, a new item for CE that acts as an handheld power monitoring computer (plus all other standard digi-board things like paper/pen storage). It's also a new steal objective for traitors.

## Why / Balance
Two reasons, really.
Engineering work takes places all over the station, and with CE's ideally having to coordinate multiple repairs and requests, having a board at hand without having to run to your front desk every time is immensely useful.
And, frankly: CE needs to be targeted by more steal objectives than just the boots lmao. Please, I want people to mess with me more often.

## Technical details
Mostly copied over from the LO's one. YAML and sprites.

## Media
Untitled

## Requirements

## Licensing

**Changelog**

:cl:
- add: Added the power digi-board! Monitor your energy situation on the move! Find it in CE's locker.


Wingidon:
Hmm, is it possible to make it say "Power Digi-Board" or something similar at the top of its UI, as opposed to "Power Monitoring Console"?

FieldCommand:
It would be really neat if it also could switch between the atmospheric alerts monitor and the atmospheric network monitor alongside power monitoring.

EmberAstra:
> Hmm, is it possible to make it say "Power Digi-Board" or something similar at the top of its UI, as opposed to "Power Monitoring Console"?

You can but not easily, I'd need to make a whole new interface just to change that title as far as I can see.
Plus, the other two boards keep the same title as the 'fixed' version, so I don't think it's a major issue?


> It would be really neat if it also could switch between the atmospheric alerts monitor and the atmospheric network monitor alongside power monitoring.

That was my idea originally, but I wasn't sure if it was maybe too much information on the fly.
I can try and figure out a way to code it meanwhile, might need a new component cause I don't think we have any item around with multiple activatable UIs around.

BarryNorfolk (approved):
Looks good, checked over that this works with the storage as well since you fixed the RD one :) 
Double checked and yes you have updated the copyright there, so that's good.

PR #5443, opened 2026-02-28
pr spam 3: Kitchen freezer

Pharaz4:
## About the PR

added freezer that chefs cant put to 73.15K

## Why / Balance

You only need the freezer below 0C to stop meat from spoiling though chefs for some reason make their freezer 73.15K which fucks with the freezer air alarm causing it to blair. IT gets very annoying and its a waste of my valuable time to go over and configure chefs air alarm.

## Requirements

## Licensing

**Changelog**

:cl:
- add: Added kitchen freezer.


Resources/Prototypes/Recipes/Lathes/machine_boards.yml
```diff
@@ -145,6 +145,11 @@
   id: GasRecyclerMachineCircuitboard
   result: GasRecyclerMachineCircuitboard
 
+- type: latheRecipe # Delta V
```
BarryNorfolk: This should be in its own `Resources/Prototypes/_DV/Recipes/Lathes/machine_boards.yml` file, rather than inside the upstream base.

BarryNorfolk (changes requested):
Move the machine board to a new file and we should be good.

BarryNorfolk (approved):

PR #5456, opened 2026-03-01
Fix IPCs losing access to radio implants when locked/unlocked

BarryNorfolk:
## About the PR
Made it so that the channels updated when an IPC is locked/unlocked include channels active from a Radio implant.

## Why / Balance
Any IPC with a syndie, centcomm, or conspirator implant would lose access to that radio channel if they ever unlocked or locked their chassis.

## Technical details
Looked like just a missing bit here, simply fix I think

## Media
image

## Requirements

## Licensing

**Changelog**
:cl:
- fix: Fixed an issue where IPCs would lose access to any radio channels given by implants.


Toby222 (approved):
This _feels_ redundant and like there should be a better way to do it but I can't think of one

PR #5462, opened 2026-03-03
jugsuit stamres and buy limit

rebe83:
## About the PR

-Only commander and lone op can buy jug suit
-reduced ranged stam res from 80% to 50%
-added loneop uplink tag so they could buy jugsuit

## Why / Balance

Warops 3 jugsuit meta is atrocious, also I've seen agents with jug suit completely solo carry and it's relatively impossible for sec to counter this.
Other proposed change would be to not allow injections into jug suit which I think would be easily bypassed (we have instant suit removal), and I would rather add counterplay than make a whole injection system for one suit
Jug suits being weak to stamina is not as great of a downside as it might seem as they should be pushing with their team. Ideally they are in the front, and downing them with stamina would require the team to move up or pull them back while they recover.
Anyone can bring hyperzine, and the agents have ephedrine. Hyperzine almost completely nullifies the stamina weakness, they will still get up too fast to cuff. Generally sec is not using rubbers very much and they will have to choose their targets and prepare ammo accordingly. Disabler modes which are more accessible (do not need printed in advance) are much less effective and can be reflected with eswords and shields.
I did not touch stamina melee because (in my opinion) nukies should have melee advantage, and 6 baton swings seems possible already with jug suit having reduced movement speed.

## Technical details

-Ranged stamina coefficient decreased to .5 from .1 in hardsuits.yml
-Add tags to restrict to one time purchase, and only from commander and loneop uplink
-added tag identifying loneop uplink in tags

## Media

Testing time to stam crit with rubber, beanbag, disabler, taser, baton, oni truncheons, tranq, and again with hyperzine
[video](https://youtu.be/kA5Rowbw5nw)

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Jug suits now have reduced stamina resistance, and can only bought by commanders and lone ops once.

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -1803,10 +1803,13 @@
   - !type:StoreWhitelistCondition
     whitelist:
       tags:
-      - CommanderUplink #DeltaV - only commander and lone op can buy
-      - LoneOpUplink
-  - !type:ListingLimitedStockCondition #DeltaV - only one jug suit per team
+      # - NukeOpsUplink # DeltaV - Commented out as of #5462 (only commander and loneop can get suit)
+  # Begin DeltaV additions - Jugsuit buy limit #5462
+      - CommanderUplink 
+      - LoneOpUplink 
+  - !type:ListingLimitedStockCondition
     stock: 1
+  # End DeltaV additions
 
 - type: listing
   id: UplinkClothingEyesHudSyndicate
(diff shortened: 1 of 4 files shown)


Woljif:
Can the commander still buy their agent a jugsuit

rebe83:
> Can the commander still buy their agent a jugsuit

nothing I can do to prevent that that I can think of

rebe83:
Changed to 30% stamina resist as direction requested for test merge

Pharaz4 (approved):

PR #5479, opened 2026-03-06
Save people from a disastrous typo

Pharaz4:
## About the PR

Replaces two words with their counterparts (Look at the diff)
If this needs direction review, I'd prefer it in private for obvious reasons.

One thing to note is that this does not sanitise announcement consoles, so for the love of god check your spelling.
Hopefully this solves a majority of mispellings though

## Why / Balance

Heard of a case where someone mistyped the word bigger and copped a ban time for it even though it was a full accident.

Many new players also use the word "clanker" against silicon in game without considering the implications of using that against another player. I assume they do this because we use the word irl often against AI or technology in general that do not have feelings. 

Hopefully the OOC nudge of changing their word to silicon will send the message to them. If they dont get the message and bypass it they are obviously ill-intent

## Technical details

ftl and yaml changes

## Requirements

## Licensing

**Changelog**

### DIFF
--- a/Resources/Locale/en-US/_DV/speech/speech-chatsan.ftl
+++ b/Resources/Locale/en-US/_DV/speech/speech-chatsan.ftl
@@ -1,0 +1,5 @@
+chatsan-word-1-DV = nigger
+chatsan-replacement-1-DV = bigger
+
+chatsan-word-2-DV = clanker
+chatsan-replacement-2-DV = silicon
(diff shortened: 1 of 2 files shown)


Woljif:
What I would do is have them replaced with words that aren't often used like "onomatopoeia", so admins can have a watch list for those words and essentially be notified when the word replacement just triggered

Pharaz4:
Once I learn CS I'll write up a ECS to alert admins when a word is said like how they are notified when an AME overclocks

Bonktrauma:
Genuinley, from the pov of an admin,  I think an alert system should be built in

StarboundSolaris:
Hi, solaris (trial curator/senior administrator for delta) here, thisd be a good change to have for many reasons, with administration being one of them. I'll also have to agree with bonk on the fact that we do need a alert system for when people use certain words, as itd help out a bunch while adminning.

PR #2811, opened 2025-01-27
Revert "[Port]Game presets can now have cooldowns to prevent back-to-back modes"

Radezolid:
Reverts DeltaV-Station/Delta-v#2744

Shit is broken and somehow it's forcing nukeops back to back.

PR #2744, opened 2025-01-15
[Port]Game presets can now have cooldowns to prevent back-to-back modes

Lyndomen:
## About the PR
add stuff from Impstation that makes it so we cant roll nukies twice in a row :trollface: did similar for survival, now you can only have 4 hours of vent critters not 10

## Why / Balance
gamemodes bad

## Technical details
port

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: TGRCDev, Lyndomen
- add: Gamemodes now have cooldowns to prevent back-to-back modes. Modes that have been voted on are exempt from this.

PR #5489, opened 2026-03-09
tweak: enable preset cooldowns

DisposableCrewmember42:
## About the PR
This PR enables the preset cooldown system added by #2744 and adjusts the cooldown values on game presets, such that none will be run twice back-to-back. Nukeops and Zeds are on cooldown for two shifts after running.

## Why / Balance
While the code added by #2744 is believed to be broken, I think that it was never actually run on the production server for the following reasons:
- It was a port from ImpStation and **they are still using it unchanged** in production to this day. The only difference is that it was gated behind a CVar because deltanedas was concerned about it breaking tests if enabled in dev env.
- The CVar behind which it is gated was confusingly named and documented in DCCVars.cs, and it was never set to true in the config preset ``deltav.toml`` according to git history. The server host would've had to manually set it to true.
- The comments added next to the cooldown values in the initial PR suggest Lyndo misunderstood how the patch was supposed to work when porting it. They expected a cooldown of 2 would stop the preset from being rolled more than 2x in a row, when it actually prevents it from rolling for two rounds after being played. Surely this would've been noticed if it was tested locally or on the server?
- It was reported broken in #2811 because NukeOps ran 5x back-to-back, which could just be caused by the cvar not being set. This revert PR wasn't merged because NukeOps wasn't chosen after a server restart
- I tested the code locally in multiple configurations to verify it works as it should:
```
# Test with traitors 2 cooldown, survival 2 cooldown, nukeops 3 cooldown, all others unrollable
Survival
traitor
nukeops
survival
traitor
[Secret (pending), correct since all on cooldown. golobby]
nukeops
traitor
survival
[Secret (pending), correct since all on cooldown. golobby]
traitor
survival
nukeops
traitor
[accidental golobby]
survival
traitor
nukeops
survival
traitor
[Secret (pending), correct since all on cooldown. golobby]
nukeops
traitor
survival
[Secret (pending), correct since all on cooldown. golobby]

# Test with traitors 2 cooldown, survival 2 cooldown, nukeops 2 cooldown, all others unrollable
traitor
survival
nukeops
traitor
survival
nukeops
traitor
survival
nukeops
traitor
survival
nukeops
traitor
survival
nukeops

looping as expected

# Test with minPlayer limits removed and adjusted cooldowns 
(not the same cooldown values as in this PR, this was the first test I did just to see if it would somehow force NukeOps)
-- 1st
Survival
Conspirators
Zombie
Traitor
Nukeops
Traitor
Zombie
Survival
traitor

-- 2nd
Survival
traitor
zombie
survival
conspirators
nukeops
traitor
survival

no duplicates as expected
```

## Technical details
renamed DCCVars field and documented it, renamed cvar, removed redundant == true, added cvar to deltav.toml config preset, adjusted/added cooldown attribute values on presets

## Media
N/A

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Gamemodes now have cooldowns. They shouldn't be chosen twice back-to-back anymore. Nukeops and Zeds are on cooldown for two shifts after rolling.

### DIFF
--- a/Content.Server/GameTicking/Rules/SecretRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/SecretRuleSystem.cs
@@ -53,7 +53,7 @@ public sealed class SecretRuleSystem : GameRuleSystem<SecretRuleComponent>
         Log.Info($"Selected {preset.ID} as the secret preset.");
         _adminLogger.Add(LogType.EventStarted, $"Selected {preset.ID} as the secret preset.");
 
-        if (_configurationManager.GetCVar(DCCVars.EnableBacktoBack) == true) // DeltaV
+        if (_configurationManager.GetCVar(DCCVars.EnablePresetCooldowns)) // DeltaV
         {
             if (preset.Cooldown > 0) // Begin Imp
             {
@@ -187,7 +187,7 @@ public sealed class SecretRuleSystem : GameRuleSystem<SecretRuleComponent>
             if (ruleComp.MinTotalPlayers > totalPlayers) return false; // DeltaV
         }
 
-        if (_configurationManager.GetCVar(DCCVars.EnableBacktoBack) == true) // DeltaV
+        if (_configurationManager.GetCVar(DCCVars.EnablePresetCooldowns)) // DeltaV
         {
             if (_nextRoundAllowed.ContainsKey(selected.ID) && _nextRoundAllowed[selected.ID] > _ticker.RoundId) // Begin Imp
             {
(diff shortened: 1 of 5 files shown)


sowelipililimute:
needs an rebase, this PR is so stale that github won't let me re-run workflows

DisposableCrewmember42:
> needs an rebase, this PR is so stale that github won't let me re-run workflows

@sowelipililimute done

sowelipililimute:
@DeltaV-Station/direction requires review

keekee38:
heya, gonna say umm idk
up to 3 days for CR but i'll try to get back to ya sooner

keekee38:
well its sooner for sure... anywho this is in essence a bugfix and umm yeah, CR approved and all

sowelipililimute (approved):

Toby222 (approved):

PR #5498, opened 2026-03-10
Blast door buff

snowywinters:
## About the PR
So fun fact blast doors have no explosion resistance at all

Also add an ai only blast door

## Why / Balance
The issue i have found is for being a blast door they are weak for it even zombies can just bite past them with ease this pr is to fix the first issue and semi fix the second issue as the zombie thing is just an issue with zombies doing ungodly structural damage

Blast doors are meant as a way to defend against bombs and other things at this point they are not. Also added an ai only blast door so door remotes do not work on it so command can not open them with there remote

## Technical details
Blast doors now have the following resistances
- takes 50% less blunt
- takes 70% less slash
- takes 50% less piercing
- takes 70% less heat
- takes 20% less shock
- takes 30% less Structural
- takes 75% less explosions (takes two breaching charges or one c4 ya c4 does an ungodly amount of damage)
Flat reductions
- Blunt 10
- Slash 10
- Piercing 10
- Heat 10
- Structural 5

## Requirements

## Licensing
    

**Changelog**
:cl:
- add: Added ai only access blast doors 
- tweak: Blast doors are now a lot stronger 
- fix: Blast doors now have explosion resistance
-->

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door.yml
@@ -1,4 +1,4 @@
-﻿- type: entity
+- type: entity
   id: BlastDoor
   parent: BaseShutter
   name: blast door
@@ -33,7 +33,9 @@
     resistance: 8
   - type: Damageable
     damageContainer: StructuralInorganic
-    damageModifierSet: StrongMetallic
+    damageModifierSet: BlastDoor #DeltaV
+  - type: ExplosionResistance #DeltaV
+    damageCoefficient: 0.25
   - type: ContainerFill
     containers:
       board: [ DoorElectronics ]
--- a/Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door_autolink.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door_autolink.yml
@@ -127,3 +127,24 @@
     - type: AutoLinkReceiver
       channel: Windows
 
+# Start DeltaV Additions
+
+- type: entity
+  id: BlastDoorAI
+  parent: BlastDoor
+  suffix: AI
+  components:
+  - type: AccessReader
+    access: [["StationAi"]]
+
+- type: entity
+  id: BlastDoorAIOpen
+  parent: BlastDoorOpen
+  suffix: Open, AI
+  components:
+  - type: AccessReader
+    access: [["StationAi"]]
+
+# End DeltaV Additions
+
+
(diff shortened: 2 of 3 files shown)


Pharaz4:
Content review < 24 hrs

Pharaz4:
Content review response: Give it the same health as a normal wall (metallic) but give it blast/explosive resistance

snowywinters:
> Content review response: Give it the same health as a normal wall (metallic) but give it blast/explosive resistance

do you mean the damage modifier or like the health of it if it is the health it only has 500 and once damaged can not be repaired at all its also made with 10 plasteel?

snowywinters:
Notes 

- flat reductions are brought over from the normal modifier so it already has that the structural was added 

- it has the base health of all shutters that being 500 

- remotes work on them so they can be opened by any one with one 

- breaching charges with the buff two hit well c4 still does over i think 1.8k damage or one shots it but with this it will only take out the door its placed on and not the rest next to it 

- zombies just do ungodly structural i think with this they still do about over 20 damage to it

Pharaz4:
Yeah zombies doing structural is a part of their balance. they are meant to have a fighting chance. They aren't exactly a role for getting stuck in one room and getting minced by lasers. they are meant to move around.

Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door.yml
```diff
@@ -1,4 +1,4 @@
-﻿- type: entity
+- type: entity
```
Radezolid, line 1: Restore this to the previous state (The BOM got changed, I suggest using an editor that doesn't auto-replace it or doing a web edit with just the other changes.)

Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door.yml
```diff
@@ -33,7 +33,9 @@
     resistance: 8
   - type: Damageable
     damageContainer: StructuralInorganic
-    damageModifierSet: StrongMetallic
+    damageModifierSet: BlastDoor #DeltaV
```
Radezolid, line 36: Remember to always state what was before the change, so it's easier for maintainers to replace if need be.
```suggestion
    damageModifierSet: BlastDoor #DeltaV - was StrongMetallic
```

Resources/Prototypes/Entities/Structures/Doors/Shutter/blast_door_autolink.yml
Radezolid, line 1: This should be added in a file inside the `_DV` namespace and in the `blast_door.yml` file since these doesn't have the `AutoLinkReceiver` component (`_DV/Entities/Structures/Doors/Shutter/blast_door.yml`)

Radezolid (changes requested):

Stop-Signs:
No activity 1month+

Let us know if you would like this re-opened

PR #5422, opened 2026-02-24
Tidersmithing changes/nerfs/rework/whatever you call this abomination

StarboundSolaris:
## About the PR

Nerfs/reworks/comments out a bunch of stuff related to tidersmithing
## Why / Balance

Headmin requested, issues with the armor and other items being too broken.
## Technical details

uhhh ill fill this out later
## Media

no
## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [x] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [x] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**

:cl:
- tweak: Changed a lot of stuff related to tidersmithing, idk ill fill this out later.

PR #5120, opened 2026-01-04
Forged Sawn Off Shotgun

jimbobby-the-FUCKING-first:
## About the PR

You can saw the forged shotgun
## Why / Balance

I thought you should be able to saw it off, and it was simple enough to accomplish
## Technical details

Its like 3 file changes and a couple sprites, nothing major
## Media

https://github.com/user-attachments/assets/800ca0f1-4f09-4d6d-8194-cdb7e9f1b946

## Requirements

- [X ] I have tested all added content and changes.
- [X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: You can now saw off the forged shotgun.

PR #5504, opened 2026-03-10
Revert "Forged Sawn Off Shotgun"

kotobdev:
Reverts DeltaV-Station/Delta-v#5120
This reverts commit 8357533ccf0a8c183c9fdb0511cbf614bc06e512.

## About the PR
Forged sawn off shotgun has been removed.

## Why / Balance
\>180dps pocket shotgun (technically less because 2 bullets, but tell that to the tider you just 0-120'd in 600 ms)
\>more accurate then the normal sawed off (no spread penalty)
\>smaller then the normal sawed off (1x2 vs 1x3, fits in secbelts)
\>can be reloaded from a shell box (as opposed to loading per-shell like the security shellgun, no need to eject anything)
\>only shotgun that can be one-handed (thus, used with a shield)
\>this also means there's no wield delay before firing, you can just keybind it from your inventory and fire it almost immediately
\>infinitely craftable with modular receivers and like 2 plasteel each
\>the *base* forged shotgun is [slated for removal](https://github.com/DeltaV-Station/Delta-v/pull/5422) for administrative and balance reasons

\>"no direction review needed"

I made one of these as a cadet yesterday and, yeah, it's kind of busted. Granted, really fun, but really busted.

Even if the buffs over the normal sawed off were removed, this isn't something that should be freely obtainable with no limit by crewmembers OR antags, it's a massively powerful item.

## Technical details
revert button ops

## Requirements

revert button ops

## Licensing

**Changelog**

:cl:
- remove: You can no longer saw off the forged shotgun.


dge21:
took only two days

Radezolid (approved):
Looks good, gotta wait for direction's review.

Stxcking:
After much discussion, this PR is approved by CR as is due to the fact that any requested changes to nerf it would turn it into a gun that we already have. Thank you for being patient !

PR #5521, opened 2026-03-16
Fix one IPC Crash

Coryler:
## About the PR

Prediction changes causes crashes. Gave 'em a seperate Alert away from BorgAlert, which is not client-based

## Why / Balance

Crashing is bad.

## Technical details

- Add on alert yml in _DV
- Change SiliconCharge Alert to the SiliconBattery

## Requirements

- [ X I have added media to this PR or it does not require an in-game showcase.

## Licensing

**Changelog**

:cl:
- fix: One IPC crash fixed.

### DIFF
--- a/Content.Shared/_EE/Silicon/Components/SiliconComponent.cs
+++ b/Content.Shared/_EE/Silicon/Components/SiliconComponent.cs
@@ -101,10 +101,10 @@ public sealed partial class SiliconComponent : Component
     public float? ChargeThresholdCritical = 0.1f;
 
     [DataField]
-    public ProtoId<AlertPrototype> BatteryAlert = "BorgBattery";
+    public ProtoId<AlertPrototype> BatteryAlert = "SiliconBattery";
 
     [DataField]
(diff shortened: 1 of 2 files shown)


Toby222:
The point of the prediction was for the alert to be client-based, what exactly is the error?

Coryler:
@Toby222 

https://github.com/user-attachments/assets/75db0c04-70a1-4871-9101-6dadfb90ba97

Currently fixes this.

To reprduce:
1. /golobby
2. make an ipc char
3. /startround
4. try to join. It crashes

Why does it crash?
The Prediciton stuff was added for borgs with the previous upstream Merge, not for IPCs. At the point above it'll have it with Battery = 10, though on userside it has NoBattery, which due to them sharing a category, overwrites the previous and causes the errors/crash above.

Toby222 (approved):
This feels like something that should be fixable in a more sensible way but I can't figure it out
@DeltaV-Station/maintainers if you have any idea please share, otherwise this is fine to merge

Toby222 (commented):
Found a nicer way to fix it I think

Toby222:
Superceded by #5535

PR #5530, opened 2026-03-17
DRAFT: (Lavaland) Ruin Generation

Stxcking:
#### TO-DO
-Error log for when a grid fails to load.
-Mass scanners showing grids as "grids" or showing them at all.
-Fixing ruins that are maps instead of grids which prevent them from spawning.
-Updating/removing ruins from the pool that are not up to standards.
-More generation and placement testing.
## About the PR
Ruin generation for planet maps are now real which mean Lavaland and Surface 
Glacier (or any other planet we add in the future) are able to spawn round-start with randomized ruins! There will be 30+ ruins included for Lavaland, some being rare.
## Why / Balance

Lavaland was intended to spawn with ruins for salvagers. Currently, our Lavaland is very bare bones compare to other servers with ruin generation. (This was also a request.)
## Technical details

This might be the fourth time that I tried to implement ruin generation without going insane, and I still can't promise that this is stable and not shitcoded. Instead of trying to add a new system to handle ruin generation the same way dungeons are generated for salvage expedition, I opted to have the ruin grids load in before the planet map is initialized. 

**The process is separated into three parts:**
**Ruin Selection:** 
-`_random.Shuffle` the grids pools so ruins are randomized.
-Select how many ruins to spawn base on `ruinSpawningCount`.
-`GuaranteedRareRuins` are pull and added to `selectedRuins` for priority placement.
-`pickRare` ruins (if there are any rare ruins) are calculated during the looping of ruin selection base on a set integer percentage. 
**Ruin Placement Prediction:**
-Grids are loaded onto a temporary map and their `localAABB` is measured and converted to radius. 
-Data is stored for placement and map is deleted.
**Ruin Placement:**
-`_random.NextVector2(minDistance, maxDistance)` create a random point around the base (mining station) between set min/max distance.
-`minAllowed` check to see if ruins are far enough from one another with each attempt until it is and it gets loaded.

This was largely base off of other forks, specifically how GoobStation14 handled the ruin placement before the Goidamerge.
## Media

Not yet...
## Requirements

- [ ] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [ ] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
    

**Changelog**

🆑 
- add: Ruin Generation: Lavaland can now generate with a couple of ruins at round-start from a pool of 30+ ruins.

### DIFF
--- a/Content.Server/_DV/Planet/PlanetSystem.cs
+++ b/Content.Server/_DV/Planet/PlanetSystem.cs
@@ -149,6 +149,10 @@ public sealed class PlanetSystem : EntitySystem
 
         // Ruin placement prediction, keeps track of the ruins that will be placed to ensure no overlap or ruins that generate too closely.
         List<(Vector2 Center, float Radius)> placedRuins = new List<(Vector2 Center, float Radius)>();
+        float minDistance = planet.RuinMinDistance;
+        float maxDistance = planet.RuinMaxDistance;
+        float minSeparation = planet.RuinMinSeparation;
+        int maxPlacementAttempts = planet.RuinPlacementAttempts;
 
         for (int i = 0; i < selectedRuins.Count; i++)
         {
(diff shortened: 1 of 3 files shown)


Stxcking:
I simply got skill checked and our downstream (Euphoria/Floof) might have something cooking with ruin generation that is similar to what Goob's has and what I originally wanted to port!

PR #5538, opened 2026-03-19
Add Medical Tracker Implanter

Coryler:
## About the PR

Adds a Medical Tracking Implant. Shouts out death and crit currently to the medical comms. Only on Salvagers and not added anywhere else.

## Why / Balance

At least to let people know ya died somewhere. I'd assume it's more expensive for CC to have people look for the dead one blindlessly in a place where they're not.

## Technical details

- Yml add new Implant
- yml Force implant it into salvagers

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- add: Medical Tracking Implant and Forced them onto Salvs


Coryler:
Test Error seems unrelated

keekee38:
gonna bring it in for review 
hopefully 3 days 🙃

Stxcking:
Will this work between grids and different maps? For like when a salvager is doing magnet or when they are in lavaland?

Coryler:
> Will this work between grids and different maps? For like when a salvager is doing magnet or when they are in lavaland?

Gonna test it once I figure out how to spawn lavaland on my dev env. Though, how it should work currently in my mind => It works similarly to the Death Rattle/Watch Dog Implant. It's comms-based. It'll only work when the Salvager is on the same map as the comms server when they die. So on lavaland it'll not work... UNLESS we change the comms to be logistic and they got another salvager on location. Pair mining. So no ping for lavaland across map yet, afaik.

Also: 

Also the signal will not send out when (high occurances):
- Solar Flares (we get a ton)
- Power Outages by CentComm
- Lack of Power

Become useless when:
- Person is gibbed by a meteor
- Being rammed by a ship into who knows where.
- Instalty being killed by a mine while jetpacking (You immedieatly get the same crit and death location, though the body continues floating)

If anything beyond of giving the chance of saving someone, it might incentivise Salvagers to stay longer on station during Solar Flares and other events like above, if they knew they could be saved in normal conditions.

Stxcking:
Following up on your answers, I believe you're right about it not working on lavaland. (Goob has a PR for a lavaland implanter that I was eyeing) Lavaland also automatic generates in your dev environment and you can warp to the mining station iirc. This also doesn't work if the salvager is doing magnet either right? If they aren't on the station, the alert won't go off?

Coryler:
> Following up on your answers, I believe you're right about it not working on lavaland.

Ye, it's based on comms. There's no medical comms down there.

> This also doesn't work if the salvager is doing magnet either right? If they aren't on the station, the alert won't go off?
image

As long as you're on the same map as them and there's a Medical Comms server, you'll get coordinates of where they crit and where they died. It'll not tell you exactly where they are in space when in motion. 

image

Distance doesn't matter for that.

On another note 

The reason as to why it's round start injected is so that it's not used for anything else. For that reason I don't necessarily know why people keep talking as if medical will inform security about the death of a salvager in deep space. This is a medical alert not security. Parameds are the first to get there or inform Salvage. We got Watch Dog to inform sec about such (and it's barely used as is... even with lowered prices).

keekee38:
so umm sorry for it bein more than 3 days
we still barely reached consensus
but seems umm
if we could put em in salvie kits, and then in the vend for 10k? but approved after that
again sorry for it takin forever and a half xd

Coryler:
> so umm sorry for it bein more than 3 days we still barely reached consensus but seems umm if we could put em in salvie kits, and then in the vend for 10k? but approved after that again sorry for it takin forever and a half xd

You still need a few days to talk about it? I can wait.

keekee38:
its all good aint no worries
conversation kinda fell off so took it to a poll and the results were more for puttin em in kits as well as in the vend for 10k
worst case if somehow leads to issues its always adjustable

Coryler:
> Kits
May I know in which one specifically or can I just randomly choose one?

Coryler:
Anyway, I've added it into the Survival Kit for now since... it's barely used and has survival in the name.

Changes made

image

image

keekee38:
thanks for puttin up with us xd that works great

rebe83:
> thanks for puttin up with us xd that works great

The direction hivemind is failing

Radezolid (approved):

PR #5539, opened 2026-03-19
Adds departmental command windoors

### LINKED ISSUE #4818: Departmental command windoors
It came to my attention that Terra's CMO windoor at their lil desk doesn't work with the remote they get because it's set to only having command access, which the remote doesn't have. So when the tider virus or something happens, you're forced to get engi (or someone with the access configur item like HoP) to come help (which honestly isn't a bad thing, something for an engi to do BUT that's if they actually show up)

This issue is aimed to add in windoors prototypes specific to the command members themselves so this can work. Sadly I can't just change the access on doors I place down when mapping because it doesn't actually save it. Which is weird because it save access changes on things like lockers. But not any kind of door, which is lame

Uno657:
## About the PR

Adds Head of Security, Chief Engineer, Chief Medical Officer, and Logistics Officer secure windoors.

## Why / Balance

Half-fixes #4818, and I decided to do the rest for the sake of completeness. This PR does not add the windoors to maps, which is the main part of the issue this fixes.

## Technical details

 - Adds departmental command windoors

## Media

image
image
image
image

## Requirements

## Licensing

  - [x] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Structures/Doors/Windoors/windoor.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/Doors/Windoors/windoor.yml
@@ -108,6 +108,42 @@
     containers:
       board: [ DoorElectronicsSurgery ]
 
+- type: entity
+  parent: WindoorSecureCommandLocked
+  id: WindoorSecureHeadOfSecurityLocked
+  suffix: HeadOfSecurity, Locked
+  components:
+  - type: ContainerFill
+    containers:
+      board: [ DoorElectronicsHeadOfSecurity ]
+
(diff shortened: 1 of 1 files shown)


Uno657:
I'm gonna let someone else handle adding them to the maps needed, since I have zero mapping experience. (Unless someone wants to point me in the direction I need to go to get these implemented properly)

Radezolid (approved):
Looks good. I assumed we had these variants already, thanks for the addition.

PR #5545, opened 2026-03-20
Uber Mega Mald PR: Plushie Speed

Coryler:
## About the PR

Delay and Attack rate changes for Plushies. It shouldn't be faster to change your plushie than it is to reload it.

## Why / Balance

It prevents my soul and self from feeling sad whenever I have to stare at a 3 second cooldown on a funny sounding object which doesn't need a reload, is a bolt-action, or a musket.

## Technical details

- Change UseDelay and FireRate

## Media

https://github.com/user-attachments/assets/e271e3fc-ede4-4bc0-a528-0ee438656352

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Plushie Attack Rate and UseDelay. Go Wild.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Fun/plushies.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/plushies.yml
@@ -21,9 +21,9 @@
   - type: EmitSoundOnTrigger
     sound: *BasePlushieSound
   - type: UseDelay
-    delay: 3.0
+    delay: 1.0 # Delta V - It shouldn't be 3 seconds.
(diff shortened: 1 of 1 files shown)


Toby222 (approved):
Good maintainer-wise; But I think the point of the high delay is specifically to avoid spamming sounds like that

rebe83:
direction review 24h

rebe83:
direction denied generally we do not want to promote noise spam or make it easier as it is already quite easy to do

Coryler:
K

PR #5555, opened 2026-03-23
fix for "Approved by: Self" on all cargo invoices

cargomouse:
## About the PR
Fixes the "Approved by:" field on cargo invoices. Previously, invoices would always read "Approved by: Self".

## Why / Balance
It was broken

## Technical details

"Self" is the default field for the approver (as set in cargo-console-order-component.ftl), which is correctly set to the approver's full name and job title earlier in the file. Later, a redundant call is made to set the approver's information to newly defined empty strings. This PR comments out that call, which does nothing besides functionally anonymize every single order.

## Media
image

## Requirements

## Licensing

**Changelog**

🆑 
- fix: Cargo invoices now correctly display who approved the order.


Radezolid (approved):
Works.

PR #5570, opened 2026-03-26
Use new feet attribute

BarryNorfolk:
## About the PR
Broken tip merge from the foot decal + april fools.

## Requirements

## Licensing

**Changelog**
Not user facing, doesn't matter.


DisposableCrewmember42:
this will cause runtime errors since the corresponding decals don't exist. this should work: #5571

PR #5572, opened 2026-03-26
Remove footprint components from every species

BarryNorfolk:
## About the PR
Removed footprints from all species

## Why / Balance
memory go up when more decals be made

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [X] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [X] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [X] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**
:cl:
- remove: Removed footprints (For now)

PR #5571, opened 2026-03-26
fix: certain footprint decals (April Fools Spoiler)

DisposableCrewmember42:
## About the PR
use the correct attribute and add footprint decal

## Why / Balance
bug

## Technical details
N/A
## Media
N/A

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [x] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [X] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [X] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**

No

PR #5573, opened 2026-03-27
Port Starcup Footprints

kotobdev:
## About the PR
Reverts #5572, #5571, and #5226. Ports https://github.com/teamstarcup/starcup/pull/743 and https://github.com/teamstarcup/starcup/pull/808, some modifications to make it work pre-nubody.

> Footprints are now collected into a single entity per tile, with a single solution instead of each having their own individual entity.

Refer to 743 for details, but essentially
- Footsteps on the same tile will be added to the same entity
- Reagents inside footsteps will react if those footsteps are on the same tile
- Mops will clean all footsteps on a tile at once
- Puddles will once again transfer a reasonable amount of reagent into footsteps
- Footsteps no longer disappear with time. I think.
- Probably a variety of other smaller changes

## Why / Balance

Footsteps, despite looking very cool, have always been rather annoying gameplay-wise and especially optimization-wise. This should fix that somewhat.

Cleaning footsteps was always a massive pain- either spraying them with space cleaner, or mopping them one by one, or more often, just waiting sixty seconds for them to evaporate. With this PR, footsteps now persist forever as a sort of psuedo-puddle until they are cleaned, which is fairly enjoyable to do with just a mop. Please don't make this require space cleaner. Please.

I doubt footsteps reacting with each other will cause too many issues, but remains to be seen.

## Technical details
THE CURSE STRIKES AGAIN

see https://github.com/teamstarcup/starcup/pull/743 for most details, some comments added and footstep components moved to our nubody-less body prototypes

## Media

https://github.com/user-attachments/assets/83f2d7c3-141b-47eb-8821-7f391ddc380f

https://github.com/user-attachments/assets/a58be0e6-7af1-4415-b7ca-40d9a2305756

## Requirements

## Licensing

**Changelog**

:cl: Oxdeception, kotobdev
- add: Ported reworked footsteps from Starcup!
- tweak: Reagents in footsteps will now react with other footsteps on the same tile.
- tweak: Mops will now clean all footsteps on the same tile.
- tweak: Footsteps no longer fade with time.
- fix: Footsteps will, hopefully, no longer cause memory leaks.


Bonktrauma:
If it needs modified for pre nuhbidy, and we are planning to get nuhbody, why not just wait?

kotobdev:
> If it needs modified for pre nuhbidy, and we are planning to get nuhbody, why not just wait?

The pre-nubody modification boiled down to moving a component from the nubody bodies starcup uses to the pre-nubody bodies we use. It's a *very* small change.

PR #5575, opened 2026-03-27
New/Tweak: Added another Mutation before Deus

Stxcking:
## About the PR

Added a new plant mutation called "Ambrosia Nobilis" In the Ambrosia line which is needed to mutate into Deus. Ambrosia Nobilis produces Tricordrazine and Negacordrazine.
## Why / Balance

Most of my opinions on Omnizine is spread across multiple threads on the discord, the two relevant ones are the [Omni rework feedback](https://discord.com/channels/968983104247185448/1486437900362190879) and [Omnizine Botany Tweaks](https://discord.com/channels/968983104247185448/1486808283351089383). (Thank you to those who help me come up with some of this PR)To summarize, I think Ambrosia Deus (Omnizine) has a lot of gameplay flaws and one of them is how easy it is to mutate. The recent Omnizine nerf targets it's use in combat and combat pens which does not address this issue or how Omnizine is used by the crew (specifically medical).
## Technical details

Yaml + Sprites
-`AmbrosiaNobilis` is the mutation after Ambrosia Vulgaris and Ambrosia Deus is after Ambrosia Nobilis.
-To keep upstream file changes to the bare minimum, I made a new plant under _DV that uses the old Deus sprites while changing the path for Deus to the new sprites.
## Media

image

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [X] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [X] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [X] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**

🆑 
- add: Added a new plant mutation between Ambrosia Vulgaris and Ambrosia Deus called "Ambrosia Nobilis" which contains Tricordrazine and Negacordrazine!

PR #5576, opened 2026-03-27
Tweak: Spaceman's Trumpets from One Mutation + Less Polyp Yield 

Stxcking:
## About the PR

Put lily seeds packets in the seed vendor where 3 are available round start and made it so Trumpets can be mutated from poppies. Also, tweak how much reagent you can get from a single Spaceman Trumpet.  You can now only get a maximum of 5 units per plant and with 50 potency, you can get 3.5. You will always get at least 1 unit per plant.
## Why / Balance

image

Polyp heals 11.6 of all brutes per 5u with a high OD and can be easily combined with other brute medicine. It is an underrated combat chemical which will be made easier to mutate due to lily seeds being in the vendor. Why? https://github.com/DeltaV-Station/Delta-v/pull/5575 made it so Ambrosia Deus is harder to mutate with an additional mutation between it. Making Trumpets an one mutation plant instead a two mutation plant will make botany still feel rewarding to play for new players or players who have a hard time mutating plants.
## Technical details

Yaml
## Requirements

## Licensing

**Changelog**

:cl: 
- tweak: Poppies can make Spaceman's Trumpets now, but they also yield less Polypyrylium Oligomers.
- tweak: Lily seeds are available in the seed vendor!

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/seeds.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/seeds.yml
@@ -20,6 +20,7 @@
     GarlicSeeds: 3
     GrapeSeeds: 5
     LemonSeeds: 5
+    LilySeeds: 3 # Delta-V: Makes Trumpets Easier to get over Deus!
     LimeSeeds: 5
     PineappleSeeds: 5
     LingzhiSeeds: 3
--- a/Resources/Prototypes/Hydroponics/seeds.yml
+++ b/Resources/Prototypes/Hydroponics/seeds.yml
@@ -1166,7 +1166,7 @@
   productPrototypes:
     - FoodPoppy
   mutationPrototypes:
-    - lily
+    - spacemansTrumpet # Delta-V - Makes Trumpets Easier to get over Deus! Was lily
   lifespan: 25
   maturation: 10
   production: 3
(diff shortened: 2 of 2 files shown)


gitjubx:
Screenshot 2026-04-02 021006
its kind of a crazy the tradeoff 1 mutation less but only 33% of the original amount of chemicals? thats literally a stealth nerf and i dont really know whats the thought behind this...making omnizine way harder and more tideous to obtain while also reducing the amounts from alternative meds? Honestly its better to have some sort of challange or difficulty mutating a 2 way plant rather than just 1 with this tiny amount of chemicals (befor you got 11 poly with 50 potency and 15 with 100 now you get 3.5 with 50? and you only get 1.5 more if you mutate to 100 potency? thats laughtable)

Stxcking:
Easier mutation line, can combine with other mutations, can increase yield by swabbing. Botany requires you to actually make a plant good by using its other mechanics instead of banking on one variable.

gitjubx:
Yes its easier to mutate but the tradeoff is everything else will be less effective if you use the other mechanics to improve the plant thus  in my eyes this is a nerf not a tweak (potency/production speed/yield)

rebe83:
Content review 3 days

rebe83:
Not much discussion on this one, approved early

BarryNorfolk (approved):
This looks correct for yml

PR #4672, opened 2025-11-16
Rework the Vulcan, giving it a long-range niche.

Woljif:
## About the PR

I gave the Vulcan a rework.

## Why / Balance

Everybody knows the Vulcan is BAD. It's extremely inaccurate, very large, and doesn't offer anything in return. It is by far one of sec's worst weapons, being so borderline unusable, that it's generally preferable to go with any alternative you can find. I think you could buff the Vulcan by just giving it a pure stat increase, but that would be kind of boring, and would just make it interchangeable with the Lecter. We don't need another Lecter clone like the M-90, what we need is some variety so people can have fun with different playstyles.

Now, the Vulcan has a fire rate of 3.5 from 5, and is semi-automatic _only,_ but in return, is a scope weapon with good accuracy when wielded/scoped in. You are slowed down when scoped in, in return for higher accuracy and field of view. Now, sec has a long-range gun that doesn't overstep any other weapons, making it finally a viable, fun weapon, instead of the waste of space that it currently is. Also, it still has limitations, namely that it's still very bad in close spaces.

Edit: slowdown penalty removed before merge, so now it has no slowdown when scoped in.

## Technical details

Added some yml to give the Vulcan the scoped trait, as well as giving it a slowdown, and reducing it's spread when wielded. Added some new sprites that contain the vulcan with a holo-sight on it.

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

https://github.com/user-attachments/assets/c3c170a2-de0e-47a6-a57b-48ded47831e6

base
bolt-open
icon
wielded-inhand-right
wielded-inhand-left

**Changelog**
🆑 
- tweak: Re-balanced the Vulcan, giving it a slower fire rate but high accuracy and a scope.

PR #5583, opened 2026-03-29
Increased firerate of vulcan from 3.5 to 4.0

Woljif:
## About the PR

I increased the firerate of the vulcan from 3.5 to 4.0

## Why / Balance

When I originally reworked the vulcan [here](https://github.com/DeltaV-Station/Delta-v/pull/4672#issue-3630242024), I think I kind of dialed the dps of the vulcan back a little too far. I was worried it would be too strong if I made it accurate considering it fires .30 rifle. Looking back though, although I've seen the vulcan used more, I don't know if it's as gamer as I intended for it to be. It's DPS sits at 66.5, which is still pretty sub par, even ignoring the fact that it's semi-auto only. Now it's DPS is 76, which is still pretty modest.

For comparison, the Lecter's DPS is 85, and the Drozd's is 96. That's a difference of 11 DPS between their two tiers of gun (longarm vs SMG). With this in mind, it's safe to say that increasing/decreasing a gun's DPS by ~10 points suffices for balancing when making a new tier of gun (The vulcan is a DMR)

## Technical details

Single line yaml change

## Requirements

## Licensing

**Changelog**
:cl:
- tweak: Vulcan firerate increased from 3.5 to 4.0

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Rifles/rifles.yml
@@ -19,7 +19,7 @@
     maxAngle: 32
     angleIncrease: 6
     angleDecay: 24
-    fireRate: 3.5
+    fireRate: 4.0 # Increased from 3.5 to 4.0
     selectedMode: SemiAuto
     availableModes:
     - SemiAuto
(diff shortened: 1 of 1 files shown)


Radezolid (approved):
In the future, there is no need to leave comments about previous values if it's from our namespace. The purpose of stating the previous value is for easier maintenance of other fork's content.

Woljif:
> In the future, there is no need to leave comments about previous values if it's from our namespace. The purpose of stating the previous value is for easier maintenance of other fork's content.

Got it

keekee38:
heya, gonna say 3 days for this one hopefully we'll get it figured out by then

keekee38:
CR approved! ty much

PR #5591, opened 2026-03-30
removed shuttle calling ion law

rebe83:
## About the PR

remove shuttle calling ion law from pool of ion law structures
## Why / Balance

It causes borderline rulebreak behavior that is miserable to everyone. It would be rulebreak if anyone else did it, borgs feel bad doing it, it annoys everyone in round and the borg player knows they are going to be forced to act "valid" and end up gibbed because there is no way to RP around the law forcing round to end.

## Technical details

image
made this range 1,35 instead of 0,35, commented out 1, changed 0 to 1

## Requirements

## Licensing

**Changelog**

:cl:

- tweak: Silicons will no longer get an ion law forcing them to call the shuttle. Everyone hated that.

### DIFF
--- a/Content.Server/Silicons/Laws/IonStormSystem.cs
+++ b/Content.Server/Silicons/Laws/IonStormSystem.cs
@@ -210,10 +210,10 @@ public sealed class IonStormSystem : EntitySystem
         var subjects = _robustRandom.Prob(0.5f) ? objectsThreats : Loc.GetString("ion-storm-people");
 
         // message logic!!!
-        return _robustRandom.Next(0, 35) switch
+        return _robustRandom.Next(1, 35) switch
         {
-            0  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
-            1  => Loc.GetString("ion-storm-law-call-shuttle", ("joined", joined), ("subjects", triple)),
+            1  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
+            // 1  => Loc.GetString("ion-storm-law-call-shuttle", ("joined", joined), ("subjects", triple)), DeltaV
             2  => Loc.GetString("ion-storm-law-crew-are", ("who", crewAll), ("joined", joined), ("subjects", objectsThreats)),
             3  => Loc.GetString("ion-storm-law-subjects-harmful", ("adjective", adjective), ("subjects", triple)),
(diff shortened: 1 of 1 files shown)


Content.Server/Silicons/Laws/IonStormSystem.cs
```diff
@@ -210,10 +210,10 @@ private string GenerateLaw()
         var subjects = _robustRandom.Prob(0.5f) ? objectsThreats : Loc.GetString("ion-storm-people");
 
         // message logic!!!
-        return _robustRandom.Next(0, 35) switch
+        return _robustRandom.Next(1, 35) switch
         {
-            0  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
-            1  => Loc.GetString("ion-storm-law-call-shuttle", ("joined", joined), ("subjects", triple)),
+            1  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
+            // 1  => Loc.GetString("ion-storm-law-call-shuttle", ("joined", joined), ("subjects", triple)), DeltaV
```
Toby222: Clearer DeltaV comment for all the changed lines
Also it's probably easier if you just comment out the 1 case and let the fallback be twice as likely?

Toby222 (changes requested):

keekee38:
24 hours cr i hope

Content.Server/Silicons/Laws/IonStormSystem.cs
rebe83: Sure but that felt "dirtier" distorting the laws making them not roll evenly, was just my feeling on it

keekee38:
approved content wise

Content.Server/Silicons/Laws/IonStormSystem.cs
```diff
@@ -213,7 +213,8 @@ private string GenerateLaw()
         return _robustRandom.Next(0, 35) switch
         {
             0  => Loc.GetString("ion-storm-law-on-station", ("joined", joined), ("subjects", triple)),
-            1  => Loc.GetString("ion-storm-law-call-shuttle", ("joined", joined), ("subjects", triple)),
```
Toby222, line 216: I meant you can just drop the `1` case entirely, the final case in the switch (`_`) will just be matched instead
This works too though

Toby222 (approved):

PR #5598, opened 2026-03-31
Real Trait System™ tweaks

kotobdev:
## About the PR
The trait maximum has been raised from 10 to 25.
Traits, by default, now cost 0 points instead of 1.

BUT WAIT! Also throwing a port of https://github.com/Floof-Station/Panta-Rhei/pull/347/ (not merged yet) in, since i saw it 30 seconds after opening this PR and it's a bugfix and we are 1000% going to need it in the future.

AND ALSO! Ported a (currently-unused) system for trait priority from https://github.com/Floof-Station/Panta-Rhei/pull/75. "Traits with higher priority are applied first". Thanks Mnemo!

## Why / Balance
No real reason to restrict players to 10 traits, it wasn't a thing before Real Trait System™ and I don't see a reason for it to be a thing now. 25 traits is the limit on most EE forks, to the best of my memory, and it seems to work just fine there.
<sub>I did test: this means, if you try to equip every possible trait, you will miss out on just two..</sub>

0 is a more sensible default then 1, that's all I can really say. There's no formal gameplay balancing to consider here, but in a future where we *do* end up getting positive traits, this is a default we'll want to have. Everything costing 1 point by default is just a bit odd, always felt weird to see.

## Technical details
somethingops

## Media
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: You can now equip up to 25 traits at once!
- tweak: Traits now cost 0 points by default.


Mnemotechnician:
You may also want to port the priority field in case some of your traits end up needing it. Differentiating traits by priority is necessary if the result of application of trait A is inherently dependent on the result of trait B (e.g. natural language replacement must occur before "foreigner" is applied, hence on our fork natural language traits have a higher priority)

kotobdev:
Thanks for the heads-up!

Content.Server/_DV/Traits/TraitSystem.cs
```diff
@@ -1,3 +1,4 @@
+using System.Linq; // Floofstation
```
Toby222: We're in _DV, we don't need to attribute where changes come from because when they're here they become "canonical"

```suggestion
using System.Linq;
```

Content.Server/_DV/Traits/TraitSystem.cs
```diff
@@ -59,14 +60,21 @@ private void OnPlayerSpawnComplete(PlayerSpawnCompleteEvent args)
...
                 continue;
 
-            ApplyTrait(args.Mob, trait);
+            sortedPrototypes.Add(trait);
         }
 
+        sortedPrototypes = sortedPrototypes.OrderBy(a => -a.Priority).ThenBy(a => a.Cost).ToList(); //Floof - get all traits from negative cost to positive cost
+        foreach (var trait in sortedPrototypes)
+            ApplyTrait(args.Mob, trait);
+        // Floofstation edit end
```
Toby222: This is reiterating the prototypes three times for no particular reason, and making two new lists.
```suggestion
        var validPrototypes = new List<TraitPrototype>();
foreach (var traitId in validTraits)
{
if (!_prototype.TryIndex(traitId, out var trait))
continue;

            validPrototypes.Add(trait);
}

        foreach (var trait in validPrototypes.OrderByDescending(a => a.Priority).ThenBy(a => a.Cost))
            ApplyTrait(args.Mob, trait);
```

It might also be nicer to use a Comparison<> and then use validPrototypes.Sort

Content.Shared/_DV/Traits/TraitPrototype.cs
```diff
@@ -57,4 +57,12 @@ public sealed partial class TraitPrototype : IPrototype
...
     [DataField]
     public List<ProtoId<TraitPrototype>> Conflicts = new();
+
+    // Floofstation section
+    /// <summary>
+    /// The priority of this trait. Higher priority traits are applied first.
+    /// </summary>
+    [DataField]
+    public int Priority = 0;
+    // Floofstation section end
```
Toby222:
```suggestion
    /// <summary>
    /// The priority of this trait. Higher priority traits are applied first.
    /// </summary>
    [DataField]
    public int Priority = 0;
```

Toby222 (changes requested):

BarryNorfolk:
I think this looks fine, test fail isn't real but I literally cannot re-run them. @kotobdev could you maybe merge master into this so that it re-runs?

BarryNorfolk (approved):
Changes look good to me and works to apply different priority things.

PR #5600, opened 2026-03-31
Telescopic & Energy Shield Fix

### LINKED ISSUE #5291: Telescopic shield blocks damage when not extended
## Description

The telescopic shield will block damage, and even break after enough damage is taken, when not extended. It acts as a regular shield by all means when not extended

**Reproduction**

Hold a telescopic shield and don't extend it, take damage, examine your damage. It's halved.

**Screenshots**

https://github.com/user-attachments/assets/1f9fe4a5-d225-4112-946c-901ec344be50

Coryler:
## About the PR

Minor Change. Made it so that telescopic shields and energy shields only block the damage when they're actually activated.

## Why / Balance

Fixes #5291 

Also makes it not only a cosmetic change.

## Technical details

Added another conditional in the BlockingSystem which stops it all when the item is not activated, using the already existing ItemToggleComponent of the two shields.

## Media

https://github.com/user-attachments/assets/06a4a20b-a011-4c62-b6d0-06f16a506be1

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Telescopic and Energyshields working without being activated

### DIFF
--- a/Content.Shared/Blocking/BlockingSystem.User.cs
+++ b/Content.Shared/Blocking/BlockingSystem.User.cs
@@ -1,7 +1,7 @@
 using Content.Shared.Damage;
 using Content.Shared.Damage.Components;
 using Content.Shared.Damage.Systems;
-using Content.Shared.Item.ItemToggle.Components;
+using Content.Shared.Item.ItemToggle.Components; // DeltaV
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Containers;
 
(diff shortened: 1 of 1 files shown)


juicyslew (approved):
I don't have great knowledge of the codebase, but this change appears quite simple and the media provided clearly shows it working.

rebe83:
Yuge

Content.Shared/Blocking/BlockingSystem.User.cs
```diff
@@ -1,6 +1,7 @@
 using Content.Shared.Damage;
 using Content.Shared.Damage.Components;
 using Content.Shared.Damage.Systems;
+using Content.Shared.Item.ItemToggle.Components;
```
BarryNorfolk: Needs a `// DeltaV` comment since you added it.

BarryNorfolk (changes requested):

BarryNorfolk (approved):
This does indeed now make deployable shields actually require deployment.

PR #5607, opened 2026-04-02
Adds more Felinid and Vulpkanin customizations to Kitsune

mnva0:
## About the PR
Adds more Felinid and Vulpkanin customizations to Kitsune. This includes:
- All the Vulpkanin ears/tails that Felinid has access to. If Vulpkanin and Felinid have it, I also added Kitsune. They're foxes
- Felinid furred limbs 

## Why / Balance
Kitsune don't really have much in terms of customization, especially not when compared to Felinid. This PR helps that

There *is* probably the problem of being able to tell species apart `(which can also be done far more reliably, in under a second, by shift-clicking)`, but Kitsune and Felinid already visually similar (humanoids with a tail and ears) and already share certain Vulpkanin markings allowing them to look completely identical

## Media
image

.

image
Kitsunes and felinids can already be made to be identical. Sizes are slightly different but I was too lazy to iterate to perfectly guess size. No I don't know which is which I already forgot

.
image
Only four ear options... Tails are better, but are still missing some of the good-looking things Vulpkanin have

## Requirements

## Licensing

**Changelog**
:cl: Minerva
- tweak: Kitsune now shares more markings with Felinid and Vulpkanin.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/Customization/Markings/felinid.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/Customization/Markings/felinid.yml
@@ -44,7 +44,7 @@
   id: FurredLeftArm
   bodyPart: LArm
   markingCategory: Arms
-  speciesRestriction: [Felinid]
+  speciesRestriction: [Felinid, Kitsune]
   sprites:
   - sprite: _DV/Mobs/Customization/Felinid/felinid_limbs.rsi
     state: furred_l_arm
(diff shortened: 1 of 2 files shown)


Stxcking:
It's me again, got approvals from senior CRs to let you know that this PR is labeled as `No Content Review Needed` so off to maintainers it goes!

rebe83:
The furslop continues, no CR needed

Radezolid (approved):

PR #5615, opened 2026-04-04
Fix character selection for non-sequential IDs

TehFlaminTaco:
## About the PR

Fixed critical bug regarding hiding species.

## Why / Balance

Critical Bugfix

## Technical details

`SelectedCharacter` no-longer (incorrectly) attempts to index `Characters` like an array. Instead, it just picks the first valid character if the previously selected character is invalid.

## Media

N/A

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Fixed critical error effecting some players.

### DIFF
--- a/Content.Shared/Preferences/PlayerPreferences.cs
+++ b/Content.Shared/Preferences/PlayerPreferences.cs
@@ -47,7 +47,7 @@ namespace Content.Shared.Preferences
             get
             {
                 // Firstly, check if we CAN use the selected character.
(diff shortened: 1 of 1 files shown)


DisposableCrewmember42:
licensing box unticked, this is a trap!

BarryNorfolk (approved):
Tested this locally by checking my character list with a breakpoint in the relevant code here, and seeing that I had no "second" character ID.
My first character was an IPC, so I added them as a hidden species and then I couldn't connect to the server with the relevant message. Booting up and trying to connect again with this patch made it work just fine.

Toby222:
> fix: Fixed critical error effecting some players.

*affecting :blunt:
Also very unspecific CL

PR #5436, opened 2026-02-28
feat: Various Admin QOL features

DisposableCrewmember42:
## About the PR
- Added Commands: lslaws, lswatchlisted, getping
- Changed lsobjectives to list all players with objectives when invoked without specifying username
- Ghosted players are now optionally marked in the F7 players tab
  - Configurable via regular Options UI
  - The "(G)" prepended to character name labels can be filtered for to see all ghosted players
- Watchlisted players now marked in admin overlay and optionally marked in the F7 players tab
  - players tab label can be turned off in options ui 
-  Tweaked Admin Alerts
   - Meatspiking now only alerts admins if an actual player character gets meatspiked. Monkeys and Kobolds won't spam alerts anymore unless a player is controlling them.
   - On-station explosions now always send alerts. Off-station explosions will not trigger alerts, so salvs won't spam admin chat with seismic charges.
   - Item stripping alerts are now Medium by default (no alert)
     - High if Back, Belt or ID slot (only alerts when new players do it)
     - Extreme if Jumpsuit (always alerts)
     - Extreme if the target is SSD (always alerts, also indicates target is SSD in message)
   - uncuff attempts no longer alert if trying to self-uncuff
   - mindshielded people uncuffing someone else no longer triggers alert
- Added Admin Alerts
  - Players with less than 2 hours of playtime latejoining will now send an admin alert.
  - EORG Alerts (only one of any of these alerts will be sent about each player to avoid spam)
    - Attacking other players at EOR
    - Destroying stuff at EOR
    - Triggering timers on items like grenades at EOR
- Added teleport links to prayer messages

## Why / Balance
- These features were either missing or only accessible via Verb UI
- The default behavior of lsobjectives was to list the calling player's objectives. You won't be able to use the command while playing as an antag anyway; I don't see how that's useful outside of debugging the objectives system, but even then you can just specify your own username.

## Technical details
- lslaws: EntityManager is used to query for entities with SiliconLawProvider
- lswatchlisted: caching whitelists in AdminNotesSystem to avoid calling db for every user for every invocation
- lsobjectives: SharedMindSystem used to check all humanoid minds for objectives 
- ghost marking: changed PlayerInfo record to include Ghost bool, set in AdminSystem by checking AttachedEntity for GhostComponent
- watchlist marking: same as ghost marking, making use of lswatchlisted cache, additionally patched AdminNameOverlay to display an indicator
- meatspiking alert: check for SsdIndicator instead of HumanoidAppearance, also indicate if a mind is attached 
[...]

PR #5618, opened 2026-04-04
fix: Admin QOL Patches

DisposableCrewmember42:
## About the PR
Patches for some issues I noticed with the commands and alerts added/changed by #5436:
- `lslaws` no longer lists law board entities
- *finishing* to butcher monkeys/kobolds no longer causes an alert either
- avoid spamming ssd stripping alerts by checking if the target is dead or the user is mindshielded
- also alert if unmindshielded new player stripping from living ssd player's hands
- made lswatchlisted output a little more compact
- fixed lswatchlisted aborting completely when encountering a record for a player without a session
 
## Why / Balance
Dealing with the consequences of my actions

## Technical details
really simple c# changes. added some conditions to stripping alerts and broke out the checked values for both stripping alerts and butcher alerts into separate functions for reuse.

## Requirements

## Licensing

**Changelog**

:cl:
DeltaVAdmin:
- fix: SSD Stripping alerts are now less spammy
- fix: /lslaws no longer lists law board entities
- tweak: finishing to butcher monkeys/kobolds no longer alerts either
- fix: /lslwatchlisted now works properly. It will no longer display a partial/empty list.

### DIFF
--- a/Content.Server/_DV/Administration/Commands/ListLawsCommand.cs
+++ b/Content.Server/_DV/Administration/Commands/ListLawsCommand.cs
@@ -26,11 +26,11 @@ public sealed class ListLawsCommand : LocalizedEntityCommands
         switch (args.Length)
         {
             case 0:
-                foreach (var (ent, lawProvider) in _entityManager.AllEntities<SiliconLawProviderComponent>())
+                var query = _entityManager.EntityQueryEnumerator<SiliconLawProviderComponent, SiliconLawBoundComponent>();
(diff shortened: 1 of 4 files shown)


DisposableCrewmember42:
alrighty, that should be the last commit actually changing logic, ready for review

DisposableCrewmember42:
> alrighty, that should be the last commit actually changing logic, ready for review

ok i lied, now this was actually the last commit changing logic

Toby222 (approved):
Seems fine

PR #5629, opened 2026-04-06
Tape recorder voice changing in secret stashes

mnva0:
## About the PR
This PR adds the `ChangeVoiceInContainer` component (the same one pAIs and MMIs have) to tape recorders. This changes their voice to the voice of the container it's in *but only things with SecretStash, like...*
- Toilets (crowbar cistern. check for contents before inserting a tape recorder or it'll literally be deleted)
- Plushies (cut them open with a knife)
- Plants
- Cakes, somehow? A knife didn't work. Maybe it tries slicing it before opening the stash

(To use the tape recorder in the stash, you are expected to use signals, but timing the tape also works)

## Why / Balance
This PR makes anonymizing names more generally accessible `(even to non-antagonists)` as well as adding a fun extra functionality to tape recorders. Tape recorders can already be used to anonymize voice `(damage the tape and repair it)`, but that comes at the cost of being nigh incomprehensible because it messes with the text

For antagonists on a budget/non-traitors, this could be used as a terrible anonymize-only voice mask with the tradeoffs of being more annoying to use, needing pre-recording and maybe some tape juggling, as well as only having names from the list above
- Getting caught with the tape recorder/thing being used to hold the tape recorder is still *very bad* because they've literally recorded themselves saying that incriminating thing
- I'm not that sure about *antagonist stuff* in general but I think a some of the non-traitor antagonists would benefit from this. Security has to decide whether that threat that was called in via potted plant `(I imagine this as just hiding it in the dirt)` is credible. Being able to do foreboding antagonist threat roleplay via comms *without* being a traitor *or* making the text almost incomprehensible *without* instantly starting a security hunt—because you used your undisguised voice—preventing you from even acting on those threats. Security will be suspicious someone said something bad, but the credibility of that is questionable until *something actually happens*

For engineers/bored people, they could set up a plushie outside their department to make it seem like the plushie is greeting people `(I've seen this be done before, but having the plushie seem to be the one speaking would be cute)`. Or make a potted plant yell at people in the halls. Or a talking toilet
Someone could use this to become some sort of whistleblower using a plushie. I'm not quite sure what other fun silly gimmicks could be done using this

I do have some concerns about making this PR, largely because pAIs or MMIs have their messages very much linked to them because they're players. But an admin *should* be capable of tracking people down if they misuse this? I think? Toolshed for example is something that could very easily be used to find the tape recorder responsible and then probably the person who set it up. I don't know because I'm not an admin
My worries are from the fact it's an item that isn't directly linked to a player speaking clearly. It might make misuse of tape recorders harder to find. I don't know. Keep in mind anonymizing voices using tape recorders is already possible, but it's terrible. Maybe I'm overestimating the average griefer
Someone *could* set this up to scream into an intercom infinitely and not get caught easily, but regular tape recorders before this PR can do this. Damaging the tapes anonymizes the name too

I have no clue what I'm talking about. I have no clue if I've elaborated enough

## Media
Yes, I recreated the meme before realizing OBS doesn't work...
image

.

image

image

## Requirements

## Licensing

**Changelog**
:cl: Minerva
- tweak: Tape recorders now change their voice when hidden in plushies, toilets, or plants.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Devices/tape_recorder.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Devices/tape_recorder.yml
@@ -16,6 +16,10 @@
   - type: TapeRecorder
   - type: ActiveListener
     range: 4
+  - type: ChangeVoiceInContainer
+    whitelist:
+      components:
+      - SecretStash
   - type: UseDelay
     delay: 1
   - type: Speech


keekee38:
this seems reslly neat, gonna bring it into CR
hopefully 24 hours or less
ans thank ya much

keekee38:
CR approved ty again!

Stxcking (approved):
Looks good to me, works as intended!

PR #5634, opened 2026-04-07
Minor indentation change on pull request template

mnva0:
## About the PR
This PR removes the indentation in the comment just below licensing on the pull request template

## Why / Balance
It's the only seemingly-pointlessly indented comment in the entire file, plus it's kind of inconvenient to delete. Invisible comments do not really need indentation

I tend to delete comments as I fill out PRs and it's annoying working around this one specific line. I've been caught in a loop several times now of accidentally deleting that line entirely only to add a new one and have it autoindented `(because it's trying to add to a list)` so I have to backspace again

## Requirements

## Licensing
It feels very strange having licensing for this PR considering this is not for the game itself


Toby222:
The indent was meant to show that it is commenting on the list of licenses specifically
Removing it and adding new list items is the desired behavior.
You shouldn't need to add custom text below the list of Licenses and before the Changelog, no?

mnva0:
>The indent was meant to show that it is commenting on the list of licenses specifically
Removing it and adding new list items is the desired behavior.

I don't think the indentation helps make the purpose of the list any more clear than if it wasn't there. The comment is already directly below the list

Having two entries in the list is also a good indicator that extra licenses should just be added to the list like the existing ones in there regardless

>You shouldn't need to add custom text below the list of Licenses and before the Changelog, no?

I don't. The indentation is just a minor inconvenience when deleting that line because of the additional spaces

1. I try to delete the comment and end up deleting the spaces too quickly and removing the line entirely. There's less "visual feedback" `(if I'm using that correctly)` of how many times to backspace/characters to remove because I can't see them directly
2. I press enter which starts another indented line that the whole process can be repeated with until I slow down

~~This is my first maldPR \j~~

Stop-Signs (approved):
sure why not.

PR #5646, opened 2026-04-10
adding pockets to the automaturge robes and flannel jackets

keekee38:
## About the PR
added pockets to a few items (let me know if theres any other y'all'd like!
(i only remembered that i said i was gonna do this after work once and then totally remembered yep)

## Why / Balance

we like pockets i think
none of this is armor or anythin so

## Technical details

changed parenting of em

## Media

idunno if it needs it

## Requirements

## Licensing

**Changelog**

<!--
:cl:
- tweak: after years of research, NT finally managed to give more clothes pockets


rebe83:
No CR needed

Radezolid (changes requested):
Needs comments on the upstream changes and your test fail is real, gotta make the restock slightly more expensive.

Radezolid (approved):

PR #5692, opened 2026-04-18
Buff MassSleep

SirWarock:
## About the PR

I added a crew-benefit to the MassSleep that can also aid Antags.
When put to sleep by massSleep, victims will have 0.5u of Prometheum injected into their bloodstream every second.
As the sleep lasts for 5 seconds, a total of 2.5u Prometheum will be injected, evenly healing 20 burn, brute and healing blood.
## Why / Balance

It's a combat only ability that kinda just sucks. As we should stray away from pure combat abilities, adding a side-effect is necessary.
Crew never has a reason to use this unless they want to validhunt. Now they can use it to heal others.
Antags can use this benefit as an deceiving argument to try and sleep people for their own gain.
## Technical details

Added a new StatusEffect and added a loop in the `SharedMassSleepPowerSystem`
## Media

https://github.com/user-attachments/assets/a9b40a45-eed7-4cc2-ab06-6e5c429cc016

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: MassSleep now heals people that have been forced to sleep in exchange for a higher glimmer generation!

### DIFF
--- a/Content.Shared/_DV/Psionics/Systems/PsionicPowers/SharedMassSleepPowerSystem.cs
+++ b/Content.Shared/_DV/Psionics/Systems/PsionicPowers/SharedMassSleepPowerSystem.cs
@@ -1,6 +1,29 @@
+using Content.Shared._DV.Psionics.Components;
 using Content.Shared._DV.Psionics.Components.PsionicPowers;
 using Content.Shared._DV.Psionics.Events.PowerActionEvents;
+using Content.Shared.Body.Systems;
+using Content.Shared.Chemistry.Components;
+using Content.Shared.StatusEffectNew.Components;
 
 namespace Content.Shared._DV.Psionics.Systems.PsionicPowers;
 
-public abstract class SharedMassSleepPowerSystem : BasePsionicPowerSystem<MassSleepPowerComponent, MassSleepPowerActionEvent>;
+public abstract class SharedMassSleepPowerSystem : BasePsionicPowerSystem<MassSleepPowerComponent, MassSleepPowerActionEvent>
+{
+    [Dependency] private readonly SharedBloodstreamSystem _bloodstreamSystem = default!;
+
+    public override void Update(float frameTime)
+    {
+        var query = EntityQueryEnumerator<RegenerativeSleepingStatusEffectComponent, StatusEffectComponent>();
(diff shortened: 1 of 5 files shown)


Woljif:
I think this is a really nice change along with the noospheric zap change. Really like seeing non-combat additions to the infamous psionic abilities. Although I foresee the healing might be a liiittle too high, but I'm not content review so we'll see

keekee38:
more CR, same as before 3 days hopin for less

keekee38:
CR approved! thanks for like i think the 3r or 4th time in a row lol

SirWarock:
Do not merge yet.

SirWarock:
Ready for Review

Content.Shared/_DV/Psionics/Systems/PsionicPowers/SharedMassSleepPowerSystem.cs
```diff
@@ -1,6 +1,29 @@
...
+    [Dependency] private readonly SharedBloodstreamSystem _bloodstreamSystem = default!;
+
+    public override void Update(float frameTime)
+    {
+        var query = EntityQueryEnumerator<RegenerativeSleepingStatusEffectComponent, StatusEffectComponent>();
+        while (query.MoveNext(out var uid, out var comp, out var statusEffect))
+        {
+            if (comp.NextTick > Timing.CurTime || statusEffect.AppliedTo is null)
+                continue;
+            comp.NextTick = Timing.CurTime + TimeSpan.FromSeconds(1); // Metabolism ticks every second.
```
BarryNorfolk, line 21: Mmm, wondering if we should get the metabolizer component and get the actual "AdjustedUpdateInterval" in order to be 'Correct'. Since it doesn't always tick every second from what I can see, it might tick slower or faster.

SirWarock: We can't do that, as the `MetabolizerComponent` is in Server-side, while this code is in shared.
When Upstream moves it into Shared, we can think about that.

BarryNorfolk: Sad, that is true.

BarryNorfolk (approved):
Works as intended

PR #5697, opened 2026-04-19
Ensure shields without toggles can still block damage

BarryNorfolk:
## About the PR
https://github.com/DeltaV-Station/Delta-v/pull/5600 fixed deployable shields from blocking damage when they shouldn't, but inadvertently disabled shields without toggles from working.

## Why / Balance
Blockable damage should be blocked when the user has an item that could block the blockable damage for the purpose of blocking. Causing the user wielding the blocking item to take less blockable damage than they would have not blocked in the case where they were not blocking by using a blocking item.

## Media
Bullet Shield

image

Energy Shield (Undeployed)

image

Energy Shield (Deployed)

image

Laser Shield

image

Telescopic Shield (Undeployed)

image

Telescopic Shield (Deployed)

image

## Requirements

## Licensing

**Changelog**
:cl:
- fix: Non-deployable blocking based items (shields) now block blockable damage when wielded for the purposes of blocking. (Normal shields reduce damage again)

### DIFF
--- a/Content.Shared/Blocking/BlockingSystem.User.cs
+++ b/Content.Shared/Blocking/BlockingSystem.User.cs
@@ -54,7 +54,7 @@ public sealed partial class BlockingSystem
             return;
 
         // Delta V - Begin Fix Toggleable Shields always blocking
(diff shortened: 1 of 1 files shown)


Radezolid (approved):
Logic makes sense and it works.

PR #5700, opened 2026-04-19
Small Buff to Warden Hardsuit

SirWarock:
## About the PR

I made the Warden Hardsuit immune to slips and spacewind if gravity is present.
## Why / Balance

The Warden hardsuit is slow, _and that is good_, however, I'd like to avoid to slow it even more.
And to do so while keeping in theme with it being slow and heavy, I made it immune to being slipped or spacewinded when gravity is present.
## Technical details

Added a `HeavyClothingComponent` and `-System`, with it's own Alert.
Made a few Upstream Events changes, one now gets relayed to inventory and the other now also relays to outer clothing.
## Media

grafik

https://github.com/user-attachments/assets/de34282e-ad9f-4fda-b371-41fdecf6b336

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: The Warden's Hardsuit is so heavy that it doesn't get affected by Space Wind or Slippery Surfaces!


SirWarock:
Particularly proud of the sprite. Should look like a boulder embedded in the ground and it kinda does.

Woljif:
Does this just make the suit function like magboots when it's worn? It should make the wearer walk normally in 0G too

SirWarock:
It does not work like magboots when worn, when gravity is gone, you'll fly out of the window from space wind

Woljif:
> It does not work like magboots when worn, when gravity is gone, you'll fly out of the window from space wind

It would be pretty sweet if it did though 👀

SirWarock:
> It would be pretty sweet if it did though 👀

Maybe, but I'd have to change how this works. Currently, the idea is that the Warden suit is so heavy that you're immune to space wind. If gravity is lost, then you aren't heavy anymore.

Cepelinas1:
shouldn't juggsuit also have the same effect?

this isn't as CM talking btw

SirWarock:
The jugsuit is 5 percent faster :godo: Totally crucial difference, trust

rebe83:
> > It would be pretty sweet if it did though 👀
> 
> Maybe, but I'd have to change how this works. Currently, the idea is that the Warden suit is so heavy that you're immune to space wind. If gravity is lost, then you aren't heavy anymore.

Well.... You still are massive? You would be moved slower

SirWarock:
> Well.... You still are massive? You would be moved slower

Yes, but I assume they want it to work like magboots. While it technically would stop you from being flung out of space because you're too massive, it wouldn't connect you to the ground to make you walk.
And I don't wanna make it immune to spacewind in no gravity, because then it doesn't make sense why it moves as quick as other people in no gravity, and I will not make the hardsuit slower in no grav

keekee38:
ok cr time, 3 days or less hopefully

keekee38:
CR approved, thank ya much

SirWarock:
> CR approved, thank ya much

Could you label it as such?

Content.Shared/_DV/Clothing/Systems/HeavyClothingSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<HeavyClothingComponent, ClothingGotEquippedEvent>(OnGotEquipped);
+        SubscribeLocalEvent<HeavyClothingComponent, ClothingGotUnequippedEvent>(OnGotUnequipped);
+        SubscribeLocalEvent<HeavyClothingComponent, WeightlessnessChangedEvent>(OnIsWeightless);
+        SubscribeLocalEvent<HeavyClothingComponent, InventoryRelayedEvent<WeightlessnessChangedEvent>>(OnIsWeightless);
+        SubscribeLocalEvent<HeavyClothingComponent, SlipAttemptEvent>(OnSlipAttempt);
+        SubscribeLocalEvent<HeavyClothingComponent, InventoryRelayedEvent<SlipAttemptEvent>>(OnSlipAttempt);
+
+    }
```
sowelipililimute, line 26:
```suggestion
    public override void Initialize()
    {
    	base.Initialize();

        SubscribeLocalEvent<HeavyClothingComponent, ClothingGotEquippedEvent>(OnGotEquipped);
        SubscribeLocalEvent<HeavyClothingComponent, ClothingGotUnequippedEvent>(OnGotUnequipped);
        SubscribeLocalEvent<HeavyClothingComponent, WeightlessnessChangedEvent>(OnIsWeightless);
        SubscribeLocalEvent<HeavyClothingComponent, InventoryRelayedEvent<WeightlessnessChangedEvent>>(OnIsWeightless);
        SubscribeLocalEvent<HeavyClothingComponent, SlipAttemptEvent>(OnSlipAttempt);
        SubscribeLocalEvent<HeavyClothingComponent, InventoryRelayedEvent<SlipAttemptEvent>>(OnSlipAttempt);
    }
```

sowelipililimute (changes requested):

sowelipililimute (approved):

Content.Shared/_DV/Clothing/Systems/HeavyClothingSystem.cs
SirWarock: Why was `base.Initialize()` added? Isn't it an empty method in the base EntitySystem?

PR #5706, opened 2026-04-20
Remove Mindswap & Refactored MassMindSwaps and MindSwap Functionality

SirWarock:
## About the PR

I removed Mindswap Psionic Power.
I combined MassMindSwap and MinorMindSwap, because they can just be the same component. They're still two different
gamerules, so nothing playerfacing changed.
I refactored the MindSwap code to now be entirely server-based, so prediction issues no longer occur.
The MindSwap code is now more centralized and hopefully made better.

## Why / Balance

I cannot fathom of a way to make MindSwap a good power.
A good power can be used when neither side is antagonistic. There's never really a moment where that's true for MindSwap.
There's no benefit of it. Zap can now recharge batteries. MassSleep can heal people. I cannot think of a way to add utility to this ability.

Secondly, there's comfort issues with MindSwaps. I want to introduce a trait where it will instead gib/ghost the person than to swap minds, for comfort reasons. This obviously is a bit problematic when there's still a mindswap power present.

I combined the two gamerules into one component, because they do the same thing and don't need different components.

## Technical details

Ripped out any component, system and event that directly attributed to `MindSwapPower`.
Moved the logic to server-side `PsionicSystem.MindSwap.cs`.
Added a method `CanExternallySwapMinds()` to show the correct popups when something external attempts to swap people (MassMindSwap)
Removing MindSwap Powers forced me to do adjustments in `TelegnosticPowerSystem` and `MindSwappedReturnPowerSystem`.
Combined functionality of `MinorMassMindSwapGameRule` and `MassMindSwapGameRule`
## Requirements

## Licensing

**Changelog**

:cl:
- remove: The Psionic Power MindSwap has been removed!


keekee38:
cr time,,,
gonna say 3 days or less thank ya again

keekee38:
approved cr wise yay

Content.Client/_DV/Psionics/Systems/PsionicPowers/MindSwappedReturnPowerSystem.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Content.Client._DV.Psionics.Systems.PsionicPowers;
+
+/// <summary>
+/// This is solely for prediction. Everything is in the server-side file.
+/// </summary>
+public sealed class MindSwappedReturnPowerSystem : SharedMindSwappedReturnPowerSystem
+{
+    protected override void OnPowerUsed(Entity<MindSwappedReturnPowerComponent> psionic, ref MindSwappedReturnPowerActionEvent args)
+    {
+    }
```
ShepardToTheStars: Any reason not to do this?

```suggestion
    protected override void OnPowerUsed(Entity<MindSwappedReturnPowerComponent> psionic, ref MindSwappedReturnPowerActionEvent args);
```

Content.Server/_DV/StationEvents/Components/MassMindSwapRuleComponent.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// Whether the mind swap will ignore mindshields.
+    /// </summary>
+    [DataField]
+    public bool IgnoreMindshields = true;
+
+    /// <summary>
+    /// How long it'll take to activate after making the announcement.
+    /// </summary>
+    [DataField]
+    public TimeSpan Delay = TimeSpan.FromSeconds(15);
```
ShepardToTheStars: Probably just default this to 60, since I believe current mass mindswap is set at that. And it gets wierd if admins add the rule manually and its suddenly a 15 second mindswap.
```suggestion
    public TimeSpan Delay = TimeSpan.FromSeconds(60);
```

Content.Shared/_DV/Psionics/Systems/SubSystems/SharedPsionicSystem.Items.cs
```diff
@@ -94,22 +96,12 @@ private void OnAntiPsionicMeleeHit(Entity<AntiPsionicWeaponComponent> weapon, re
...
-
-            if (TryComp<Components.PsionicPowers.MindSwappedReturnPowerComponent>(target, out var swapped))
+            else if (HasComp<MobStateComponent>(target) && weapon.Comp.Punish && Random.Prob(weapon.Comp.PunishChance))
             {
-                _mindSwapPowerSystem.SwapMinds(target, swapped.OriginalEntity, false);
-                return;
+                _stuttering.DoStutter(args.User, TimeSpan.FromMinutes(5), false);
+                _stun.TryKnockdown(args.User, TimeSpan.FromSeconds(5), false, drop: false);
+                _jittering.DoJitter(args.User, TimeSpan.FromSeconds(5), false);
             }
```
ShepardToTheStars: The timespans should probably be on the `AntiPsionicWeaponComponent`. And I don't know if we should accumulate the status effect since stuttering for 20+ minutes sound completely horrible.

PR #5732, opened 2026-04-26
Adds lockable switches

IamVelcroboy:
## About the PR
Adds lockable versions of switches.

## Why / Balance
Mapping. Switches are often optimal to buttons in certain situations and I need some access locked variants. 

## Technical details
n/a

## Media
n/a

## Requirements

## Licensing

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Structures/Wallmounts/switch.yml
+++ b/Resources/Prototypes/_DV/Entities/Structures/Wallmounts/switch.yml
@@ -68,7 +68,7 @@
   parent: SignalSwitchDirectional
   id: LockableSwitch
   name: lockable switch
-  categories: [ HideSpawnMenu ]
+  suffix: Unassigned
   components:
   - type: Appearance
   - type: Lock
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/_DV/Entities/Structures/Wallmounts/switch.yml
```diff
@@ -61,3 +61,328 @@
...
   - type: AccessReader
     access: [["ERT"]]
+
+# Switches
+
+- type: entity
+  parent: SignalSwitchDirectional
+  id: LockableSwitch
+  name: lockable switch
+  categories: [ HideSpawnMenu ]
```
Radezolid: Wouldn't it be preferable to leave it accessible and maybe add a suffix saying it has no access? Mainly for admemes since if they just change the access of a button it will still show as it's original for example:

Image

Radezolid (dismissed):
Looks good and works. The comment above is optional.

Resources/Prototypes/_DV/Entities/Structures/Wallmounts/switch.yml
IamVelcroboy: If you think that would be useful, I can. The base lockable button is hidden as well. Would need to un-hide in the upstream file.

IamVelcroboy:
> Looks good and works. The comment above is optional.

Went ahead and unhid them

Radezolid (approved):

PR #5739, opened 2026-04-27
stealthsuit tweaks

rebe83:
## About the PR

has 0.9 speed (had no slowdown before)
minimum visibility 0.5 (was previously perfect)
helmet does not need to be up for stealth (so they can use psionic nullifier helmet)
I was going to make it run on batteries but toggling systems fucked. and batteries are too complicated for my simple time. also if it just wiggles all the time it doesn't need batteries.
## Why / Balance

Stealth was previously perfect which made wormops + sit in corner too easy. It is still possible, but if your cover is blown people will actually possibly maybe be able to see you, or punish mistakes
when this was ported 20% slowdown was removed probably to compensate for removing 5% armor values (mistake), and the stealthonmove value was reduced to such a value making it so you could actually regenerate stealth while still moving ? to compensate for higher base movement speed maybe?
It could probably be healthy eating 10% less armor (current values are 30% everything 40% pierce) 

I tested making the minimum cloaking the same as ninja, but honestly the ninja is pretty visible but has the benefit of higher movement speed, so I let it be slightly better. 

## Technical details

change stealth suit helmet to no longer have clothinggrant comp
give clothinggrant to hardsuit
make comment block instead of a zillion lines in the armor values
stealthonmove increased back to 0.6
## Media

Go my unembedded youtube link
[url](https://youtu.be/RcNkeckpt5c)
## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Cybersun stealth suits are slightly slower, and no longer have perfect cloaking after a Spider Clan attack on production facilities.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/hardsuit-helmets.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/hardsuit-helmets.yml
@@ -21,11 +21,13 @@
         Slash: 0.95
         Piercing: 0.95
         Heat: 0.95
-  - type: ClothingGrantComponent
-    component:
-    - type: Stealth
+  # DeltaV start
+  # - type: ClothingGrantComponent
+  #   component:
+  #   - type: Stealth
     # Harmony start
-    - type: StealthOnMove
-      passiveVisibilityRate: -0.5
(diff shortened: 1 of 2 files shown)


Dorragon:
Content review 48 hours

Dorragon:
Content review approved

sowelipililimute (approved):

PR #5745, opened 2026-04-28
Added Altruistic Hemomancy

SirWarock:
## About the PR

I added the Altruistic Hemomancy power. Name can still be changed. Please if you have a better name.
It allows the psionic to heal others using their blood.
10% of a human's blood equals evenly healing 20 of Brute, Burn, Airloss and Toxin. It'll also reduce bleedstacks by 4 (max bleed is 10)
This can happen up to three times per ability usage, for 60 of the above and 30% of your blood.

If used on someone critical, the doafter is quicker, but costs 1.25x more blood.
If used on the dead, it'll revert their rot by 30 seconds, costs double the blood, doubles the doafter time and will deal 30 cellular damage for each use.

The normal weight to roll it is 0.5, but for the chaplain, it's 3.

## Why / Balance

Thought it was cool. Maybe this be the first non combat related power.
## Technical details

Added the power system and a special way of changing a psionic table when an entity has a certain role.
Bunch of YML
## Media

https://github.com/user-attachments/assets/c0da9558-0dd7-4fb0-b73a-aa9824364462

https://github.com/user-attachments/assets/310253b9-992c-4330-8c9c-8dbc89d06e28

## Requirements

## Licensing

**Changelog**

:cl: SirWarock, Stxcking
- add: Added Altruistic Hemomancy, which can heal others in exchange for your blood!


SirWarock:
There's the possibility of adding a status effect on both the target and user to slow them down after usage of the power.
I also wanted to increase the blood regeneration of the psionic user, but the code didn't seem to let me.

rebe83:
Whatever go my sap

rebe83:
Also there is a psionic on other forks for healing word which gets spammed because it's free heals and works on the dead and removed cellular
I can see this wouldn't be as bad as that but medbay generally is so starved for gameplay they start killing eachother.
I believe 5u of saline is like 15% of your blood so like
Bloodlevel is free as long as you're alive. The cost of 200u saline is 40u salt. You're getting what, uhh... I'm bad at math...
60 heal / 5u saline let's say, that's like y'know
2,400 healing for the cost of 40 salt.
Is my math right here probably not

SirWarock:
I think 5u of Saline is 10%
A full 60 heal requires 15u of saline
One heal tick costs 10%, so 5u saline.
200/5 = 40 ticks. 40 × 20 amounts to 800 healing with a full jug of saline

Dorragon:
Content review 48 hours

Dorragon:
Apologies for the huge delay
While the general consensus of Content review is positive, a considerable amount of reviewers are concerned that the downside of using the power is easily negatable, thus making the power a bit too spammable
The PR will unfortunately remain in CR hell until changes are made for the downsides to be more severe
Some CR members suggested:
1. Additional hunger after using the power
2. Direct bloodloss damage
3. A portion of patient's damage being transfered to the user

SirWarock:
Thanks for @Stxcking, we now have a sprite for the power
grafik

Resources/Prototypes/_DV/Psionics/psionicPowers.yml
```diff
@@ -13,10 +13,41 @@
...
+    - id: AltruisticHemomancyPowerEntity
+      weight: 0.3
     - id: MassSleepPowerEntity
       weight: 0.3
+    - id: NoosphericZapPowerEntity
+      weight: 0.3
+    - id: MindSwapPowerEntity
+      weight: 0.15
+    #- id: PsionicEruptionPowerEntity
+    # weight: 0.1
```
ShepardToTheStars: Remove
```suggestion
```

Resources/Prototypes/_DV/Actions/psionic.yml
```diff
@@ -213,3 +213,19 @@
     useDelay: 60
   - type: InstantAction
     event: !type:PsychokineticScreamPowerActionEvent
+
+- type: entity
+  parent: BasePsionicAction
+  id: ActionAltruisticHemomancy
+  name: Coagulate
+  description: Command your blood to heal another alive being, exchanging your own crimson essence. Command it unto the dead to revert their rot, but it'll cost ya..
```
ShepardToTheStars: Nit. Slight wording change.
```suggestion
  description: Command your blood to heal another alive being, exchanging your own crimson essence. Command it unto the dead to revert their rot but beware, it will cost you more than your blood.
```

PR #5758, opened 2026-04-29
Added Un-mineable Variants of Planetside Rocks

Stxcking:
## About the PR

Titles sums it up! Did it for all planetmap walls + asteroid cause why not.

## Why / Balance

For a WIP map!

## Technical details

YAML:
Made `WallRockUnmineable` from `MountainRock` and parent it for all variants.
## Media

image

## Requirements

## Licensing

**Changelog**

Just for mappers.


Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
```diff
@@ -256,3 +256,151 @@
     - state: gibtonite_inactive
       visible: false
       map: [ "enum.MiningScannerVisualLayers.Overlay", "gib" ]
+
+# Un-mineable Variants
+
+- type: entity
+  id: WallRockUnmineable
+  parent: BaseAltRock
+  name: durable rock
```
IamVelcroboy: I think I'd like a different word than "durable." To me durable conveys something constructed/crafted. These are naturally hard. Suggestions might be dense, compacted, concentrated, hardened(?), or even just hard would be accurate.

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
```diff
@@ -256,3 +256,151 @@
...
+    sprite: Structures/Walls/rock.rsi
+    layers:
+    - state: rock
+    - map: [ "enum.EdgeLayer.South" ]
+      state: rock_south
+    - map: [ "enum.EdgeLayer.East" ]
+      state: rock_east
+    - map: [ "enum.EdgeLayer.North" ]
+      state: rock_north
+    - map: [ "enum.EdgeLayer.West" ]
```
IamVelcroboy, line 281: You don't need to include all the components if they already exist in the parent. Only add if something withing them changes from the parent.

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
```diff
@@ -256,3 +256,151 @@
...
+        sound:
+          path: /Audio/Effects/break_stone.ogg
+          params:
+            volume: -6
+
+- type: entity
+  id: WallRockBasaltUnmineable
+  parent: WallRockUnmineable
+  name: durable basalt
+  suffix: Un-mineable
```
IamVelcroboy: The suffix will also be inherited and does not need to be repeated.

IamVelcroboy (dismissed):

IamVelcroboy:
My main concern here is that there is no visual distinction between these and the mineable variants. I think that some visual cue is important to communicate to the players that these are, in some way, mechanically different. I understand that you want to use these in a particular way to create a certain vibe on a station, which I can get behind but I do think the sprites need some sort of touch-up that makes them visually distinct so players know by looking at them, "oh, these are the ones I can't mine."

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
IamVelcroboy: Actually, second glance, you should probably just inherit from `MountainRock` instead of `BaseAltRock`

Stxcking: > You don't need to include all the components if they already exist in the parent. Only add if something withing them changes from the parent.

There's only ~~one~~ two component that changes between the variants and the parent which has to do with their sprites which I need all the associated lines that come with that ^^;

Stxcking: > Actually, second glance, you should probably just inherit from `MountainRock` instead of `BaseAltRock`  

Reason why I didnt parent off of MountainRock because it uses a different IconSmooth mode. I can see how this makes the sprites look but afaik it will change how they connect and seemingly blend together.

Edit: Realize I could have done this better actually looking at it again buh

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
Stxcking: For some reason when testing this, it didn't inherit the suffix but will check again. Might of been when I was trying to parent off `WallRock`

Stxcking:
> My main concern here is that there is no visual distinction between these and the mineable variants. I think that some visual cue is important to communicate to the players that these are, in some way, mechanically different. I understand that you want to use these in a particular way to create a certain vibe on a station, which I can get behind but I do think the sprites need some sort of touch-up that makes them visually distinct so players know by looking at them, "oh, these are the ones I can't mine."

I understand the concern which is why I gave them a description and unique name. What I can do is maybe make the colors darker so when placed side by side, you know which one is mineable. Otherwise I don't want to mess with the sprite too much because as you said, this is for a map that uses the base basalt for a reason

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
Stxcking: I have no idea why I was struggling so hard to find an adjective. Compacted works really well...my first thought was hardened but I didn't like the sound of it.

Resources/Prototypes/_DV/Entities/Structures/Walls/asteroid.yml
IamVelcroboy: You can still add the iconsmooth comp if it's needed. General rule of thumb is to go with what adds the least amount of lines. If you inherit from MR, you can remove the suffix, and destructible comp lines additionally, it comes with a stronger damage modifier which I think is optimal here.

IamVelcroboy:
> I understand the concern which is why I gave them a description and unique name. What I can do is maybe make the colors darker so when placed side by side, you know which one is mineable. Otherwise I don't want to mess with the sprite too much because as you said, this is for a map that uses the base basalt for a reason

Understandable. It can be subtle, but should still be noticeable enough to not cause confusion.  It doesn't have to be a complete resprite.

Toby222 (approved):

Toby222 (approved):
needs CR

rebe83:
No CR needed since this is just mapping tools I believe

PR #4990, opened 2025-12-15
The Jetpack Stays On During Harmbatoning

SirWarock:
## About the PR

I made Jetpacks automatically activate when leaving the grid, if they have been toggled active.

## Why / Balance

Is gud. QoL. And the less times you need to interact with the horrid action bar, the better.

## Technical details

Made the System `SharedJetpackSystem.cs` partial and replaced a method in it.
Added a DV version of `SharedJetpackSystem.cs`.
Minor C# changes to add the logic behind the automation.
Added two variables to `JetpackComponent.cs` to allow tracking of waiting users.

## Media

https://github.com/user-attachments/assets/8eb40317-4f4b-46c1-b49e-d7dd81c89ef3

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Jetpacks now automatically turn on when leaving a grid, if you have toggled them on!

PR #5776, opened 2026-05-03
Prevent zombies/non-humanoids/non-borgs from using jetpacks

JustTNE:
## About the PR
Zombies and pets can no longer use jetpacks. Attempting to will result in the message:
"The jetpack seems to be too complicated to operate."

## Why / Balance
On Delta-v, pending a rework, the consensus has been that zombies are not supposed to be intelligent. Being able to correctly operate a jetpack when zombies struggle at running compared to unzombified beings is a stretch.

Quote from upstream for `ComplexInteractionComponent`:
> This is used to gate manipulation to general humanoids. If a mouse shouldn't be able to do something, then it's complex.

Upstream removes ComplexInteractionComponent from Zombies. It can safely be assumed that a mouse would not be able to operate a jetpack.

## Technical details
Introduces `CanComplexInteract` check from `ActionBlockerSystem` alongside a translation string
When Zombifying: Removes `AutomaticJetpackUserComponent`, calls `SharedJetpackSystem.SetEnable(..., false, ...)`

This technically affects every being without ComplexInteractionComponent, including pets.

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Zombies can no longer use jetpacks
- remove: Pets can no longer use jetpacks

### DIFF
--- a/Content.Shared/_DV/Movement/Systems/SharedJetpackSystem.cs
+++ b/Content.Shared/_DV/Movement/Systems/SharedJetpackSystem.cs
@@ -1,3 +1,4 @@
+using Content.Shared.ActionBlocker;
 using Content.Shared._DV.Movement.Components;
 using Content.Shared.Movement.Components;
 using Content.Shared.Movement.Events;
@@ -7,11 +8,18 @@ namespace Content.Shared.Movement.Systems;
 public abstract partial class SharedJetpackSystem
 {
     [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly ActionBlockerSystem _actionBlocker = default!;
 
     private void OnJetpackToggle(Entity<JetpackComponent> jetpack, ref ToggleJetpackEvent args)
     {
         if (args.Handled)
             return;
+
+        if (!_actionBlocker.CanComplexInteract(args.Performer))
+        {
+            _popup.PopupClient(Loc.GetString("jetpack-too-complex"), jetpack, args.Performer);
+            return;
+        }
 
         jetpack.Comp.AutomaticMode = !jetpack.Comp.AutomaticMode;
         jetpack.Comp.AutomaticUser = args.Performer;
--- a/Content.Server/Zombies/ZombieSystem.Transform.cs
+++ b/Content.Server/Zombies/ZombieSystem.Transform.cs
@@ -26,6 +26,8 @@ using Content.Shared.Mobs;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Movement.Pulling.Components;
 using Content.Shared.Movement.Systems;
+using Content.Shared.Movement.Components; // DeltaV
+using Content.Shared._DV.Movement.Components; // DeltaV
 using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.NPC.Components; // DeltaV
 using Content.Shared.NPC.Systems;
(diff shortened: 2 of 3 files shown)


Vova567810:
What's gonna happen if someone dies with infection midflight though, since after https://github.com/DeltaV-Station/Delta-v/pull/4990 jetpacks remain on automatically if toggled?

JustTNE:
@Vova567810

> When Zombifying: Removes AutomaticJetpackUserComponent, calls SharedJetpackSystem.SetEnable(..., false, ...)

See `Technical details` section

Basically: Jetpack gets disabled and will not enable automatically anymore.

SirWarock:
Why does it get disabled? Wouldn't it make more sense for it to stay enabled?

JustTNE:
@SirWarock

> Why does it get disabled? Wouldn't it make more sense for it to stay enabled?

Not really, at least not in my opinion. Operating a jetpack in real life would take precise motor control, which zombies distinctly lack. In the game we handwave jetpacks as being the automatic thing as easy as walking, but we know that's not really how that actually works :P. I imagine you would have to control the jetpack with a joystick/some other method of control regardless, hence -> zombies can't operate jetpacks.

In absence of making zombies entirely lose control in space with their jetpack on (which would not be very fun for the zombie player), this seems like the best solution that prevents zombies flying in space with a jetpack without ruining the rest of their round more than being turned already might.

I have also personally always seen the "automatically enable jetpack" as QoL, not canonical feature of the jetpack. (And so does CR: https://github.com/DeltaV-Station/Delta-v/pull/4990#issuecomment-3671910420 )

rebe83:
Senior CR agrees this is a "bug fix" and does not need review- this brings mechanics into parity with RP / admin ruling expectations of zombie behavior

Woljif:
> Why does it get disabled? Wouldn't it make more sense for it to stay enabled?

why would it make more sense for it to stay on?

Vova567810:
> @Vova567810
> 
> > When Zombifying: Removes AutomaticJetpackUserComponent, calls SharedJetpackSystem.SetEnable(..., false, ...)
> 
> See `Technical details` section
> 
> Basically: Jetpack gets disabled and will not enable automatically anymore.

me when I am in pr reading competition and my opponent is myself

SirWarock:
> > Why does it get disabled? Wouldn't it make more sense for it to stay enabled?
> 
> why would it make more sense for it to stay on?

Because when you can't turn on the jetpack, you probably can't turn it off either.

But the authors explanation makes sense, so it's good

Stop-Signs:
1 day CR

Content.Shared/_DV/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -7,12 +8,19 @@ namespace Content.Shared.Movement.Systems;
...
     private void OnJetpackToggle(Entity<JetpackComponent> jetpack, ref ToggleJetpackEvent args)
     {
         if (args.Handled)
             return;
 
+        if (!_actionBlocker.CanComplexInteract(args.Performer))
+        {
+            _popup.PopupClient(Loc.GetString("jetpack-too-complex"), jetpack, args.Performer);
+            return;
+        }
```
ShepardToTheStars, line 22: I think this would also prevent pets from using the jetpack too, which means they can wear them but not use them.

JustTNE: I don't see a problem with that. Any entity without `ComplexInteractionComponent` will be unable to use a jetpack after this check.

> It can safely be assumed that a mouse would not be able to operate a jetpack.

SirWarock: Laika wouldn't be able to operate their jetpack, which is unintended.

ShepardToTheStars: > I don't see a problem with that. Any entity without `ComplexInteractionComponent` will be unable to use a jetpack after this check.

The point is that this affects more than just zombies. If thats the intention, then it should be obvious in the changelog and the PR title.

JustTNE: Changed. CR seems to be agreeing with this but I do not speak for them.

Stop-Signs:
Approved

Toby222 (approved):

PR #5782, opened 2026-05-06
Buff glasses vision correction

EmberAstra:
## About the PR
NT stopped cheapening out on glasses and they are now all around much better at correcting your vision.

## Why / Balance
So, as someone that played a short-sighted character a few times.
The shader effect for it is cool and works well at blocking your vision far away. You can counter it with glasses, but that has the inherent disadvantage of losing the ability to use flash/welding protection, night vision, thermals, etc. And prescription HUDs are a compromise to get at least basic job info. Alright. The idea is neat.

But... the blur stays there even with glasses, just weaker to the point it becomes just annoying. You can still see just fine up to the edges of the screen, it's all just got this white-ish filter that ends up kinda being headache inducing after you stare at it for up to 2 hours. It doesn't make you *worse* at much, with only the Prescription Sec HUD still having a limited range, it just ends up annoying.

So we're buffing them all.
The downside of being nearsighted seems like it should be the danger of *losing the glasses* or being unable to use better eye gear without losing your ability to see properly, rather than having to deal with a slightly blurry filter the entire shift.
Insert a Velma "MY GLASSES" gif here (one day i'll figure out how to make you drop them when you slip too)

## Technical details
YAML ops.
Glasses (plain ones and other variants like jamjars) vision correction strength: 2 -> 8
Prescription HUD vision correction strength: 1.75 -> 2.5
Prescription Sec Glasses vision correction strength: 1.5 -> 2

## Media
OLD

NEW

## Requirements

## Licensing

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
+++ b/Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
@@ -82,6 +82,7 @@
   - type: Clothing
     sprite: Clothing/Eyes/Glasses/glasses.rsi
   - type: VisionCorrection
+    correctionPower: 8 # Delta-V (was default value, 2)
   - type: Tag
     tags:
     - HamsterWearable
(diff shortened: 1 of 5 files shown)


rebe83:
Bug fix / minor tweak no CR needed

Radezolid (approved):
Tested and it works.

PR #5784, opened 2026-05-06
Make craftable improvised/forged weaponry and armor unobtainable

Radezolid:
## About the PR
Removed the ability to craft the weapons/ammo/armor ported from Starlight. They are still spawneable for admemes or whatever.

## Why / Balance
In my opinion this is just powergaming bait. They are so easy to craft and, in some cases, really strong. To remedy this a proper method of crafting these would be needed to be ported but I simply don't have the time or energy to do it and I'm tired of seeing Engie-with-a-Dawnbraker123.
When there is a situation that would warrant making a weapon it's already too late to make one, which encourages people to make weapons before there is an actual situation worth of it. If the situation is really dire that would make you, a non-security personnel, search for a weapon to defend against the station threat logistics or security are way quicker and proper ways of handling the situation.

## Technical details
Removed several construction graph and `Construction` component from the weapons/armor/ammo ported from Starlight.

## Media
image

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Improvised/forged weaponry, armor and ammo are no longer craftable.


Halo3moth:
The savior of delta V

jink23:
I don't really agree with removing them as we already had improvised/craftable weapons prior to this patch and they were still decently effective, the improvised bow, spear, bat, ect. All could deal decent damage and while not as good as a gun or dedicated melee weapon were still there but people didn't really rush to craft them for no reason. As it is people rarely rush to craft these improvised weapons without reason and removing them kinda feels pointless when Syndies could also use them in the event they lose their PDA/Uplink access or they get caught and lose their weapons they've already bought.

stormyatt:
I can speak for melee weapons in that they are not useful against anything that’s not a vent critter. And that plastasteel, cloth, and a welder/mask is a lot to ask for a weapon that likely matches to a toolbox. Tide breaker is just a claymore in a fancy suit. The only reason to take those I would posit is rp reasons. If you need a weapon and lack any options(maintenance,botany) a bat or toolbox is the best option. So out of stuff antag or kinda goober crew. I just don’t think they’re that bad for removal. At most just match the numbers to a combat knife and call it.

Stop-Signs:
2 weeks since this is techincally content removal

Stxcking:
Due to https://github.com/DeltaV-Station/Delta-v/pull/5818 being made within the content removal timeframe and approved, feel free to close this now or when it gets merged.

PR #5790, opened 2026-05-09
The Disappearance of Hatsune Miku

Stxcking:
## About the PR

While the name of the PR is a joke, the PR is real. I renamed the hatsune miku day bag into virtual performer costume bag, along with giving it a new description. 
## Why / Balance

I might get burn at the stakes for this one but in order to stay consistent with our rules against names that makes references to real life media or people, we need to get rid of assets in the game that creates holes in our rules like the air jordans and hatsune miku day bag. I am not one to outright delete something, so I decided to just rename the bundle. It still heavily references Hatsune Miku with the costume itself being kept unchanged, but it at least isn't a direct name reference.
## Technical details

Yaml
## Requirements

## Licensing

**Changelog**

🆑 
- tweak: Renamed the Hatsune Miku Day Bag to Virtual Performer Costume Bag


StarboundSolaris:
Its so over

Stop-Signs (changes requested):
Use ent and .desc in FTL to do this change to avoid upstream conflicts

Stop-Signs:
example can be found here Resources/Locale/en-US/_DV/bed/bed.ftl

Toby222:
I swear they used to be called that

ShepardToTheStars (approved):

ShepardToTheStars:
> Use ent and .desc in FTL to do this change to avoid upstream conflicts

We'll get upstream conflicts regardless. Unless we create an inherited item in our namespace, and use migrations to replace everything.

Radezolid:
> > Use ent and .desc in FTL to do this change to avoid upstream conflicts
> 
> We'll get upstream conflicts regardless. Unless we create an inherited item in our namespace, and use migrations to replace everything.

I disagree with the reasoning, while it's not a long process to keep our changes on YAML, it still adds a extra obstacle and time spent on it even if it's small.

rebe83:
I believe this tragically is a no CR needed type of fix

ShepardToTheStars:
> I disagree with the reasoning, while it's not a long process to keep our changes on YAML, it still adds a extra obstacle and time spent on it even if it's small.

Then if the aim is to prevent conflicts, then creating a new inherited entity in the _DV namespace, and doing a migration is preferable and the only thing that will actually prevent conflicts.

Resources/Prototypes/_DV/Catalog/Fills/Boxes/general.yml
```diff
@@ -218,3 +218,11 @@
...
         - id: RemoteSignallerAdvanced
         - id: ReinforcementRadioSyndicateYarn
+
+- type: entity
+  parent: BoxPerformer
+  id: BoxPerformerDV
+  name: virtual performer costume bag
+  description: A bundle that contains everything you need to embrace your inner performer.
+  components:
+  - type: EntityTableContainerFill
```
Radezolid:
```suggestion
```

Radezolid (dismissed):

ShepardToTheStars (approved):

PR #5795, opened 2026-05-10
Salvage Access Only on Salvage Magnet

AirlineFood001:
## About the PR

Made the Salvage Magnet only open it's UI if you have Salvage access on your ID card
## Why / Balance

it feels like every single round has like 1-3 assistants breaking out into space to tide using the salvage magnet. It's just kind of straight up annoying when salvage can't do their job because the infamous space tider is spamming their magnet, jumping out to loot the debris, then jumping back with no approval from salvage. This does still allow it to be emagged to be changed to AA, so if an antag really for some reason wants to space tide using the magnet they still can.
## Technical details

like three YML lines
## Media

No media because it just says "Access denied." if you click the magnet with no Salvage access on your ID card
## Requirements

## Licensing

**Changelog**

🆑 
- tweak: Salvage magnet now requires salvage access to use, or an EMAG.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/salvage.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/salvage.yml
@@ -41,7 +41,7 @@
   - type: SalvageMagnet
   - type: ActivatableUIRequiresAccess
   - type: AccessReader
-    access: [["Salvage"]]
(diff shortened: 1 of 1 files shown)


AirlineFood001:
ough, poke

Radezolid (approved):

PR #5813, opened 2026-05-15
[TESTMERGE] Quadratic armour

sowelipililimute:
## About the PR
Armour values are now on a quadratic curve, rather than a linear coefficient as before.

### What makes a good armour system?
Variations between armour values should feel meaningful, even between 5 to 10. (or 5% to 10%).
Damage below some point should be and _feel_ negligible compared to the armour, and damage above some point should make the armour itself feel negligible.
Armour values should also compose in a manner that feels legible, but not be so overwhelmingly strong as to make stacking up a bunch of weak armours equivalent to one good armour.

### Linear Armour
Bildschirmfoto 2026-05-14 um 11 06 13 PM

Our current system of armour is mostly based on a linear model.
The incoming damage is multiplied by some coefficient to determine the end damage.
Right off the bat, this model fails at making distinct armour values feel meaningful, except in the large percentage increments we see in the game currently.
The model also fails at making very small and very high damage amounts feel the way they should.
It lacks the granularity for something like "this armour should protect you pretty much completely against tider punches, but it shouldn't make you immune from a Death Squad member."
The same 20% that brings 5 damage down to 1 damage is the same 20% that brings 50 damage down to 10 damage.

Bildschirmfoto 2026-05-14 um 11 13 39 PM
The composition behaviour is also fairly strong, being straight multiplication of the coefficient.
Two 50% armours behave as a single 25% armour.

### Flat Reduction
Bildschirmfoto 2026-05-14 um 11 16 33 PM
The other armour model we currently have in the game is flat reduction.
In terms of granularity, this definitely takes the cake.
Armour values directly correlate with damage as seen on weapons, so you can easily specify an armour that protects fully against a wrench but basically not at all against a Death Squad agent.

For situations where there's just one (1) value, this is an excellent choice for game feel.
However, where it falls flat is composition.
Combining multiple Flat Reduction armours results in, well, the consequences of combining multiple flat reduction armours.
Bildschirmfoto 2026-05-14 um 11 19 34 PM

This effectively expands their "zero damage at all range" by a _large_ margin, and substantially tanks incoming damage above that as well.

### Logistic Reduction
Bildschirmfoto 2026-05-14 um 11 20 30 PM

This one is mostly a joke.
It quite literally optimises for the criteria of "damages below a point should feel negligible, and damages above a point should feel strong."
At least, until you pan the chart to the right.

Bildschirmfoto 2026-05-14 um 11 21 39 PM
This causes a maximum cap on the damage you can take at once, and it doesn't bode well to splicing some other function in for a better behaviour.

Bildschirmfoto 2026-05-14 um 11 22 52 PM
Composition also effectively just squashes the chart down, making it about as strong as linear reduction in this regard, if not moreso.

### Quadratic Reduction
So far, the best model we've seen for the gamefeel of very weak and very strong attacks against armour is the Flat Reduction.
However, in a game as complex as Delta-V, the composition is an Achilles' heel and makes armour stacking inordinately strong.
The complete cessation of all damage below the threshold also acts as a hardblock, and completely kills attrition.
While the amount of attrition in the linear model felt eeeeh, it's still important to have attrition so that e.g. nuclear operatives tearing through crew have to worry about the possibility of being worn down over time.

It'd be nice if we had something with good feel for very weak and very strong attacks, while preserving attrition, while having good composition behaviour.
Presenting: quadratic reduction.

Bildschirmfoto 2026-05-14 um 11 34 26 PM
This simple curve (plus or minus a linear graph stitched on) is deceptively good at fulfilling all of the criteria above.
For damage below the armour value of the curve, it is substantially weaker making very weak attacks feel effectively so, while not being zero and completely killing attrition.
For damage above twice the armour value of the curve, the effect of the armour is effectively null.
This armour can protect effectively against trivial weapons, _and_ fall nicely to a death squad member.

Differences between armour values are also substantial, with their granularity scaling nicely with the size of the armour value.
Differences between small armour values are pronounced, like the flat model, while being able to express strong armours, that nonetheless, aren't completely invincible.

The armour value in this model effectively serves as a "range" where the armour is effective.
Damage on the low end of the efficacy range is blocked more strongly, while damage on the high end of the efficacy range is blocked more weakly, before effectively not being blocked at all.

This "range" property preserves, even in composition.
Bildschirmfoto 2026-05-14 um 11 38 08 PM
Stacking armour in this model effectively only increases the efficacy of the armour in the range, while leaving values outside of it unaffected.
So, no matter how many weak armours you can scrounge together, you'll get steamrolled by Death Squad all the same.

This model is the one that I believe can lead to an _amazing_ game feel if executed correctly, and I use it in this pull request consequently.
My focus is on broad strokes game-feel, not the devil in the details of balancing minute differences in armour.

## Technical details
Add support to DamageModifierSet/DamageSpecifier
Add DVDamageValueDictionarySerializer/DVDamageConstantPrototype so we can have named constants.
The YAML mines go insane.

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Damage reduction in armour has been fully overhauled!

### DIFF
--- a/Content.Shared/Armor/ArmorComponent.cs
+++ b/Content.Shared/Armor/ArmorComponent.cs
@@ -37,7 +37,12 @@ public sealed partial class ArmorComponent : Component
     [DataField("staminaMeleeDamageCoefficient")]
     private float? _staminaMeleeDamageCoefficient;
 
-    public float StaminaMeleeDamageCoefficient => _staminaMeleeDamageCoefficient ?? Modifiers.Coefficients.GetValueOrDefault("Blunt", 1.0f);
+    public float StaminaMeleeDamageCoefficient => _staminaMeleeDamageCoefficient ??
+                                                  (Modifiers.Coefficients.TryGetValue("Blunt", out var coefficient)
+                                                      ? coefficient
+                                                      : Modifiers.Armor.TryGetValue("Blunt", out var armor)
+                                                          ? 1f - Math.Clamp(0.01f * armor, 0f, 1f)
+                                                          : 1f);
     // End DeltaV - Give armor melee stamina resistance
 }
 
@@ -61,7 +66,7 @@ public sealed class CoefficientQueryEvent : EntityEventArgs, IInventoryRelayEven
     /// <summary>
     /// The Total of all Coefficients.
     /// </summary>
-    public DamageModifierSet DamageModifiers { get; set; } = new DamageModifierSet();
+    public List<DamageModifierSet> DamageModifiers = new(); // DeltaV - stack modifiers instead of doing weird math
 
     public CoefficientQueryEvent(SlotFlags slots)
     {
--- a/Content.Shared/Damage/DamageSpecifier.cs
+++ b/Content.Shared/Damage/DamageSpecifier.cs
@@ -164,6 +164,14 @@ namespace Content.Shared.Damage
                 if (modifierSet.Coefficients.TryGetValue(key, out var coefficient))
                     newValue *= coefficient; // coefficients can heal you, e.g. cauterizing bleeding
 
+                // Begin DeltaV Additions
+                if (modifierSet.Armor.TryGetValue(key, out var armor) && 2 * armor > value)
+                {
+                    var blocked = Math.Max(0f, 1f - Math.Min(newValue / (2f * armor), 1f));
+                    newValue = (1f - blocked) * newValue;
+                }
+                // End DeltaV Additions
+
                 if (newValue != 0)
                     newDamage.DamageDict[key] = FixedPoint2.New(newValue);
             }
--- a/Content.Shared/Damage/DamageSpecifier.cs
+++ b/Content.Shared/Damage/DamageSpecifier.cs
@@ -130,6 +130,8 @@ namespace Content.Shared.Damage
         }
         #endregion constructors
 
+        private static readonly ProtoId<DamageTypePrototype> Penetration = "Penetration"; // DeltaV - penetration
+
         /// <summary>
         ///     Reduce (or increase) damages by applying a damage modifier set.
         /// </summary>
@@ -144,6 +146,8 @@ namespace Content.Shared.Damage
             // more cause they're just bloody stumps.
             DamageSpecifier newDamage = new();
             newDamage.DamageDict.EnsureCapacity(damageSpec.DamageDict.Count);
+
+            var penetration = damageSpec.DamageDict.GetValueOrDefault(Penetration).Float(); // DeltaV - penetration
 
             foreach (var (key, value) in damageSpec.DamageDict)
             {
@@ -165,9 +169,10 @@ namespace Content.Shared.Damage
                     newValue *= coefficient; // coefficients can heal you, e.g. cauterizing bleeding
 
                 // Begin DeltaV Additions
-                if (modifierSet.Armor.TryGetValue(key, out var armor) && 2 * armor > value)
+                if (modifierSet.Armor.TryGetValue(key, out var armor) && 2 * Math.Max(armor - penetration, 0f) > newValue)
                 {
-                    var blocked = Math.Max(0f, 1f - Math.Min(newValue / (2f * armor), 1f));
+                    var reducedArmor = Math.Max(armor - penetration, 0f);
+                    var blocked = Math.Max(0f, 1f - Math.Min(newValue / (2f * reducedArmor), 1f));
                     newValue = (1f - blocked) * newValue;
                 }
                 // End DeltaV Additions
--- a/Content.Shared/Damage/Systems/DamageExamineSystem.cs
+++ b/Content.Shared/Damage/Systems/DamageExamineSystem.cs
@@ -48,6 +48,8 @@ public sealed class DamageExamineSystem : EntitySystem
         message.AddMessage(markup);
     }
 
+    private static readonly ProtoId<DamageTypePrototype> Penetration = "Penetration"; // DeltaV - penetration
+
     /// <summary>
     /// Retrieves the damage examine values.
     /// </summary>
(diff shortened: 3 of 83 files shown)


sowelipililimute:
also not mentioned is that the non-linear systems can all take an "armour piercing" factor that's more granular than a complete "ignore armour"/"don't ignore armour" thing much easier than the linear model but that's not in this PR

StarboundSolaris:
Under review by direction, expect response in 3 days

SirWarock:
An Oni nerf..

Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
```diff
@@ -111,12 +117,14 @@
...
-        Heat: 0.5
-        Caustic: 0.9
+      # Begin DeltaV
+      armor:
+        Blunt: DamagePistol
+        Slash: DamagePistol
+        Piercing: DamagePistol
+        Heat: DamagePistol
+        Caustic: DamageUnarmed
+      # End DeltaV
```
SirWarock, line 127: Any reason for these values to be `DamagePistol`? You usually replaced values of 40-60% reduction with DamageRifle, why different here?

Content.Shared/Damage/Systems/DamageExamineSystem.cs
```diff
@@ -66,13 +68,21 @@ private FormattedMessage GetDamageExamine(DamageSpecifier damageSpecifier, strin
...
                 msg.PushNewline();
                 msg.AddMarkupOrThrow(Loc.GetString("damage-value", ("type", _prototype.Index<DamageTypePrototype>(damage.Key).LocalizedName), ("amount", damage.Value)));
             }
         }
 
+        // Begin DeltaV - i know this is a hack but i am NOT rewriting all of damageable to wire in an extra number
+        if (damageSpecifier.DamageDict.TryGetValue(Penetration, out var penetration))
+        {
+            msg.PushNewline();
+            msg.AddMarkupOrThrow(Loc.GetString("damage-examine-penetration", ("amount", MathF.Round(penetration * 2f))));
```
SirWarock, line 82: This doesn't work on my machine, does it work on yours?

Content.Shared/Armor/ArmorComponent.cs
```diff
@@ -37,7 +37,12 @@ public sealed partial class ArmorComponent : Component
     [DataField("staminaMeleeDamageCoefficient")]
     private float? _staminaMeleeDamageCoefficient;
 
-    public float StaminaMeleeDamageCoefficient => _staminaMeleeDamageCoefficient ?? Modifiers.Coefficients.GetValueOrDefault("Blunt", 1.0f);
+    public float StaminaMeleeDamageCoefficient => _staminaMeleeDamageCoefficient ??
+                                                  (Modifiers.Coefficients.TryGetValue("Blunt", out var coefficient)
+                                                      ? coefficient
+                                                      : Modifiers.Armor.TryGetValue("Blunt", out var armor)
+                                                          ? 1f - Math.Clamp(0.01f * armor, 0f, 1f)
```
SirWarock, line 44: Using the raw number is a bit low - Perhaps doubling it is good? This is closer to the original value that the armor once held. Sec and Nukie suits would have 50% stam res instead of 25%.
```suggestion
                                                          ? 1f - Math.Clamp(0.02f * armor, 0f, 1f)
```

Content.Shared/_DV/Armor/Systems/HandHeldArmorSystem.cs
```diff
@@ -113,6 +107,15 @@ private FormattedMessage GetHeldArmorExamine(HandHeldArmorComponent armor)
             ));
         }
 
+        foreach (var factor in armor.Modifiers.Armor)
+        {
+            var armorType = Loc.GetString("armor-damage-type-" + factor.Key.Id.ToLower());
```
SirWarock, line 112: Forgot a `msg.PushNewline();`

```suggestion
            msg.PushNewline();

            var armorType = Loc.GetString("armor-damage-type-" + factor.Key.Id.ToLower());
```

Content.Shared/_DV/Armor/Systems/HandHeldArmorSystem.cs
```diff
@@ -113,6 +107,15 @@ private FormattedMessage GetHeldArmorExamine(HandHeldArmorComponent armor)
...
+            msg.AddMarkupOrThrow(Loc.GetString("armor-coefficient-value",
```
SirWarock, line 113: Small oversight
```suggestion
            msg.AddMarkupOrThrow(Loc.GetString("armor-armor-value",
```

Content.Shared/Damage/DamageSpecifier.cs
```diff
@@ -164,6 +168,15 @@ public static DamageSpecifier ApplyModifierSet(DamageSpecifier damageSpec, Damag
                 if (modifierSet.Coefficients.TryGetValue(key, out var coefficient))
```
SirWarock, line 168: As damage gets exponentially weaker when lower compared to armor, species resistances and vulnerabilties can heavily affect this, to the point where a felinid will have no armor against blunt weapons and dionas are the new Achilles with only heat being able to scratch them.

Content.Shared/Damage/DamageSpecifier.cs
```diff
@@ -164,6 +168,15 @@ public static DamageSpecifier ApplyModifierSet(DamageSpecifier damageSpec, Damag
...
                     newValue *= coefficient; // coefficients can heal you, e.g. cauterizing bleeding
 
+                // Begin DeltaV Additions
+                if (modifierSet.Armor.TryGetValue(key, out var armor) && 2 * Math.Max(armor - penetration, 0f) > newValue)
+                {
+                    var reducedArmor = Math.Max(armor - penetration, 0f);
+                    var blocked = Math.Max(0f, 1f - Math.Min(newValue / (2f * reducedArmor), 1f));
```
SirWarock, line 175: Perhaps instead of `newValue`, the total damage minus structural/penetration should be divided by `2f * reducedArmor`.
Therefore, an Energy Sword dealing 15 Slash/Heat will not need penetration added, as both their heat and slash will count as 30 damage instead of 15 each.
Of course, only use that for the calculation on how much is blocked - Use `newValue` for actually subtracting `blocked` from the damage.

SirWarock (changes requested):
A small review. Mostly looks okay, but some YML values are off, and some bugs are present.
Also, some suggestions on how to deal with split damage.

Content.Shared/Damage/DamageSpecifier.cs
SirWarock: grafik

Using TotalDamage to calculate how much is blocked will let an Energy sword deal 9 slash and 15 heat against a security hardsuit for a total of 24 damage.
Previous value was 2.02 slash and 5.62 heat for a total of 7.64 damage.
Value with procentual coefficients was 9 slash, 11.25 heat for a total of 20.25 damage.

Using the total damage does make heat deal a lot more damage due to how the armor works and sec suit having less heat resistance, but it's only 4 damage more than standard values, whereas the current implementation is 13 damage less than standard.

SirWarock: The values with coefficients are wrong.
I forgot to account the helmet and mask with them.
That'll reduce the damage to ~7.7 slash and ~9.6 heat, for a total of ~17.3 damage, which is 7 damage less than the calculation with total damage and 10 damage more than the calculation with newValue.
Maybe it's not as good as an idea because of this.

sowelipililimute:
going to re-open when I have the time to actually produce a semi functional balance

PR #5834, opened 2026-05-18
Prevent Code Orange from being selected manually

Eternally-Confused:
## About the PR

I made code orange unavailable in the communications console. It can now only be set by Central Command or by the meltdown of a nuclear reactor. 

## Why / Balance

People are deliberately calling Code Orange to scare the station and CC. Self antags and actual antags are calling it for... the funny I guess? I assume their goal is to jumpscare the station with a scary alert. Because it automatically triggers on a nuclear reactor going critical, people instantly assume that code orange means the reactor is about to explode. No other manual alert level really has this effect except for (maybe) code white. And because of the "Mayday protocols are activated" line, it's the _only_ manual alert level that unambiguously gets the attention of Central Command. 

On a more subjective note, I feel like code orange is too intense for something that is able to be set manually. The loud alarm/klaxon thing is extremely cool, but it feels a bit overdramatic with how it's currently used. All of the other alert levels that do something similar (delta, gamma, octarine, epsilon) are either automatic or CC only. Overuse of it will probably lead to people caring less about it, which I think would be a bit of a shame. 

Personally, I also don't see the point in the station having both code Yellow and Orange. I feel like there's too much overlap between the two. Code Yellow was _already_ functionally an "abandon ship" alert in my experience. People are currently using code Orange as functionally a replacement for Yellow too, even for problems that haven't caused evauation. Orange seems a bit too much like "make CC mad" bait at the moment. 

The relevant community feedback thread can be seen [here](https://discord.com/channels/968983104247185448/1505704989949952050), though it's only a handful of messages. 

## Technical details

one (1) (i) line of yaml.

## Media

Screenshot 2026-05-17 201653

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Command is no longer able to set the station's alert level to Code Orange

### DIFF
--- a/Resources/Prototypes/AlertLevels/alert_levels.yml
+++ b/Resources/Prototypes/AlertLevels/alert_levels.yml
@@ -49,6 +49,7 @@
       selectable: true
       sound:
         path: /Audio/_Starlight/misc/orange.ogg
+      disableSelection: true # DeltaV - Don't allow the station to set it manually
       color: "#ff8119"
       emergencyLightColor: "#ff8119"
       forceEnableEmergencyLights: true
--- a/Resources/Prototypes/AlertLevels/alert_levels.yml
+++ b/Resources/Prototypes/AlertLevels/alert_levels.yml
@@ -46,10 +46,9 @@
   # Begin Starlight Additions
     orange:
       announcement: alert-level-orange-announcement
-      selectable: true
+      selectable: false # DeltaV - Don't allow the station to set it manually
       sound:
         path: /Audio/_Starlight/misc/orange.ogg
-      disableSelection: true # DeltaV - Don't allow the station to set it manually
       color: "#ff8119"
       emergencyLightColor: "#ff8119"
       forceEnableEmergencyLights: true


LordSugarcane:
Isnt code orange also for disasters like Singuloose and Tesloose? Wouldn't it need to be callable for those as they're not detected automatically.

Eternally-Confused:
> Isnt code orange also for disasters like Singuloose and Tesloose? Wouldn't it need to be callable for those as they're not detected automatically.

We already had/have code Yellow for that. I might be missing something, but I don't see why we need a second alert level for "engi messed up."

Vova567810:
I mean would be nice to differentiate between "oh shit, station is not in good shape" vs "OH SHIT OH FUCK, THE SINGULO/TELSA IS LOOSE"

LordSugarcane:
I believe orange is supposed to be an engineering version of code red. Code red is "There are multiple major emergency situations ongoing, or a major security emergency affecting the station.", a singuloose is a single engineering emergency, giving security more power to arm up does nothing in the event of a singuloose. I see code orange as a way to alert the crew that the station is for all purposes lost and to evac, without telling them to arm up as it most likely wasnt a security issue.

Eternally-Confused:
This isn't working fully how I expected. I'm not fully confident in this at the moment, so uh. Closing for now

PR #5858, opened 2026-05-21
feat: OOC/LOOC/Deadchat muting

DisposableCrewmember42:
## About the PR
Added admin commands for muting individual players in OOC, LOOC or Deadchat. Any staff able to ban people will be able to use these commands.

## Why / Balance
It was previously only possible to disable the entirety of OOC or LOOC. Deadchat can't be disabled. Now we can mute specific players while keeping the channels open for everyone else.

## Technical details
- OOC muting: Add bool to ContentPlayerData. Check it in ChatManager. Command flips bool.
- LOOC/Deadchat muting: Add InGameOocMuted Component + System. Component holds bools for LOOC and Deadchat. System listens to InGameOocMessageAttemptEvent and cancels message according to component state. Command ensures comp on player's mind entity and flips corresponding bool.

## Media
image
image

## Requirements

## Licensing

**Changelog**

:cl:
DeltaVAdmin:
- add: /mute command - Mute an individual player in a specific OOC channel. OOC mutes persist until server restart. LOOC/Deadchat mutes persist until the round ends.


DisposableCrewmember42:
rewriting commands rq, made a wrong assumption about when they are available (misinterpreted LocalizedEntityCommands docs, thought they would be unavailable in lobby). i can implement them in a less janky way

Stxcking:
These are Staff/Admin related QoL, no CR involvement needed

Radezolid:
Genuine question... why not simply ban the person until they cool off?

DisposableCrewmember42:
alright, should be ready

DisposableCrewmember42:
> Genuine question... why not simply ban the person until they cool off?

Cool-off bans definitely make sense in cases where players just can't seem to stop malding, but I'd be hesitant to immediately place a ban if I'm not fully aware of the situation. A mute stops them quickly, and then I can take time to talk to them via aHelp.

There are also other situations in which this is useful, such as people revealing current-round info in LOOC, or not reacting to admin LOOC messages while sitting idly and using it to chat for extended periods.

sheepism:
> Genuine question... why not simply ban the person until they cool off?

I mean, I wouldn't want to A0 someone for spamming OOC for the silly, i'd like a nice mute function instead

Content.Server/_DV/Administration/Commands/MuteCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+/// Command for muting/unmuting specific players in specific OOC channels.
+/// Mutes are not persisted to storage.
+/// </summary>
+///
+/// After adding/removing a channel in <see cref="Execute"/>, don't forget to:
+/// - Update <see cref="CHANNEL_COMPLETION_OPTIONS"/>
+/// - Document behavior in "cmd-mute-help" in Resources/Locale/en-US/_DV/administration/commands/mute.ftl
+
+[AdminCommand(AdminFlags.Ban)]
+public sealed class MuteCommand : LocalizedEntityCommands
```
sowelipililimute, line 23: new commands should be toolshed-based rather than using the legacy commands system

Content.Server/_DV/Administration/EntitySystems/InGameOocMutedSystem.cs
```diff
@@ -0,0 +1,31 @@
...
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<InGameOocMessageAttemptEvent>(OnIngameOoc);
+    }
+
+    private void OnIngameOoc(ref InGameOocMessageAttemptEvent args)
+    {
+        if (args.Session.GetMind() is not {} mind)
```
sowelipililimute:
```suggestion
        if (args.Session.GetMind() is not { } mind)
```

sowelipililimute (changes requested):

Content.Server/_DV/Administration/Commands/MuteCommand.cs
DisposableCrewmember42: I understand this is a legacy system, but I don't think Toolshed is suitable for admin commands like this as it is now. It's great for querying and manipulating entities, but imo it's too verbose and unnecessarily complicated for quick admin actions, which is why I wrote the commands in this and my other Admin QOL PRs as legacy commands. There isn't really an upside to making these Toolshed usability-wise as nobody will be querying for a bunch of people to mute/freeze/whatever.

``mute LOOC User123`` vs ``> players:imm User123 mute:looc``

sowelipililimute: if usability is the goal this should be an admin verb, not a command
Bildschirmfoto 2026-05-24 um 12 31 12 PM

DisposableCrewmember42: Commands are much more convenient to use for things like this than the admin verb ui, just open the console and type a quick command with autocompletion.

For verbs, you either have to right-click the player you're watching or search for their name in F7, then click their row, before navigating through submenus and scrolling until you find the verb you're looking for, possibly missing because an unpredicted verb pops up and shifts the action/menu you were about to click/hover on around. If I can avoid it, I don't use these while adminning.

The F7 player list also doesn't show any verbs on right-click while you're in the lobby, so we wouldn't be able to mute players from OOC using a verb

sowelipililimute (approved):

PR #5878, opened 2026-05-24
Make Psionic Eruption less dangerous

kotobdev:
## About the PR
Psionic Eruption's explosion is now vastly less damaging to its environment. The UI popup has been reworded to focus more on the fact that this power will simply just kill you.

## Why / Balance
Psionic Eruption is one of the most unique powers currently in the game, and it would be an absolute shame to lose it. There is a ton of occasionally-tapped storytelling potential there, just because you didn't see all of it yourself doesn't mean it didn't happen.
Honestly, if I was evil I'd make the power *actually* make you blow up after a period of time, but I feel like that would be an entire argument among CR so I'll avoid that one.
Few people in the thread said it would be ideal to just lower the destructive potential, so that's what this does.

## Technical details
All explosion intensities lowered by anywhere between 50% to 75% across the board.
image

Every explosion size below 900 glimmer is a full 3x3 or cross 3x3 of varying intensity. Deals about 37.5 damage in the epicenter at most.
image

Past 900 glimmer, becomes a rough 5x5. 52 damage at epicenter.
image

If I'm using explosionui right, at least.

## Media
reworded ui
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Psionic Eruption has had its destructive potential substantially reduced.

### DIFF
--- a/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
+++ b/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
@@ -181,12 +181,12 @@ public sealed class PsionicEruptionSystem : BasePsionicPowerSystem<PsionicErupti
 
         int boom = _glimmer.GetGlimmerTier(_glimmer.Glimmer) switch
         {
-            GlimmerTier.Minimal => 4,
-            GlimmerTier.Low => 8,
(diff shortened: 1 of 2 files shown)


TehFlaminTaco (approved):
Code good 👍

Stop-Signs:
Per content removal policy, this gets a testmerge. If it fails to address the issues identified in the thread the ability will be removed.

PR #5909, opened 2026-05-28
Fix: Allow Ninja to call in Colossus as a Threat

Stxcking:
## About the PR

Seems like Colossus was never apart of the threat weight for the Ninja's threat objective but it was supposed to be?

## Why / Balance

Fix?

## Technical details

YML

## Requirements

## Licensing

**Changelog**

🆑 
- add: Ninjas can now call in the Colossus as a threat!

### DIFF
--- a/Resources/Prototypes/threats.yml
+++ b/Resources/Prototypes/threats.yml
@@ -4,6 +4,7 @@
   weights:
     Dragon: 1
     Revenant: 1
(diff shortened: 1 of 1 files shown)


Radezolid (approved):
Didn't test but looks good to me.

Still, gonna wait for direction's take on this as the PR that added it doesn't mention the ninja calling a colossus, so I assume any discussion happened on discord.

PR #5916, opened 2026-05-30
Remove remnants of Dex+

Radezolid:
## About the PR
There were some Dexalin+ containers around, killed them and replaced the only useful instance with salbutamol.

## Why / Balance
We made the DexalinPlus proto abstract so when trying to fill containers it just error's out, on release it just spits a exception on the console, but also it remains on the container as a UnknownReagent.

## Technical details
N/A

## Media
image

## Requirements

- [xI have tested all added content and changes.

## Licensing

**Changelog**

N/A


ShepardToTheStars (approved):
Dex+ doesn't deserve this fate.

Also: Consider removing the `abstract: true` from Dex+'s reagent prototype in case upstream ever adds any other items with Dex+. That way, the chemical will exist *but* it won't cause issues until we can address it.

PR #5798, opened 2026-05-10
The audiovisual sauce bundle

sowelipililimute:
## About the PR
we have a lot of low hanging fruit to port for general audiovisual niceness

## Why / Balance
game looks cool and communicates more information

## Technical details
- bunch of cherrypicks
  - https://github.com/EphemeralSpace/ephemeral-space/pull/178
  - https://github.com/EphemeralSpace/ephemeral-space/pull/1332
  - https://github.com/EphemeralSpace/ephemeral-space/pull/273
  - https://github.com/EphemeralSpace/ephemeral-space/pull/491
  - https://github.com/EphemeralSpace/ephemeral-space/pull/348
  - https://github.com/EphemeralSpace/ephemeral-space/pull/1084

merge using a merge commit

## Media

https://github.com/user-attachments/assets/e4274194-db5d-47cc-860d-c07f222e718e

https://github.com/user-attachments/assets/10455021-5e3c-49ed-9d28-93ee9ec9e7dd

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an in-game showcase.

## Licensing

- [x] I confirm that I am the creator of the code in this PR, and allow licensing it under the following license(s), or that the original author(s) has given me permission to do so:
  - [X] AGPL (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-AGPLv3.txt) 
  - [X] MIT (https://github.com/DeltaV-Station/Delta-v/blob/master/LICENSE-MIT.txt) 
    

**Changelog**

:cl: mirrorcult, Nemanja, Flareguy, sowelipililimute
- fix: Door animations now properly play out instead of getting cancelled part-way through.
- add: There are now interaction particles to make some of the invisible things people do more visible.
- tweak: Audio attenuation is now more noticeable
- add: Many electrical things now produce sparks when damaged or activated

PR #5926, opened 2026-05-30
[mald pr?] Revert "The audiovisual sauce bundle" 

Proxysseia:
Reverts DeltaV-Station/Delta-v#5798

insanely overwhelming, cant be turned off, unnecessary

looking at someones strip menu pets them
using any alt click interaction like eating food slaps the food
every electronic appliance goes off like you lit a sparkler on fire when you smack them once
secoffs go spew a cubic meter of sparks any time they turn on their baton (already has visual and sound cue)

animations are all insanely over the top. icon floats in, wiggles around and fades out, no cooldown
actions that already had feedback get even more feedback, screwdrivering a door open already plays a sound and has a visual cue, i dont need another more obtrusive visual cue for it

room with 10 people is filled with 10 trillion random interactions, thanks for informing me random tider is drinking out of a glass of water i wasnt able to tell that by the bar over their head or the sound, i need a floating hand to tell me.


StarboundSolaris:
Ill be honest here, i feel that this could be fixed by just adding a toggle in setting instead of immediately going straight to removing it outright.

sowelipililimute:
I'd advise waiting a week and seeing how you feel about the interactions then rather than opening a PR the day they go live. Anything new intrinsically feels a little wrong until you get used to it. These sorts of effects have been in use on SS13 servers for the better part of a decade and a half.

I'd also like you to consider that deaf people exist before arguing that a primarily sound cue is sufficient for indicating anything.

Proxysseia:
> I'd advise waiting a week and seeing how you feel about the interactions then rather than opening a PR the day they go live. Anything new intrinsically feels a little wrong until you get used to it. These sorts of effects have been in use on SS13 servers for the better part of a decade and a half.

"this is so overwhelming i cant play the game" -> oh just play for a week


> I'd also like you to consider that deaf people exist before arguing that a primarily sound cue is sufficient for indicating anything.

lucky i never argued that? batons turning on has a visual and a sound cue. screwdrivering doors has a visual and a sound cue. they do not need an additional visual cue thats very distracting.


its an accessibility issue this made it into the game without a toggle. i requested a toggle on the original PR and it was ignored.

Halo3moth:
I've been playing with the new visuals for a while and I really enjoy them I could see how they might be frustrating to some and there are a few big qol changes that they need there should probably be: 

Some sort of cooldown for similar animations especially the free handed one that pops up when you click literally anything (pulling up the strip menu shouldn't play a animation at all) 

The new grill shock looks really nice but I think it can be overwhelming when you hit 2-3 grills at once 

Pulling seems to be a bit buggy with the animation it will either play it on you or play it twice 

And the wobble for all animations could probably be toned down slightly.

I have way more good than bad to say about the new visuals but I wouldn't mind seeing a toggle to at-least tone them down to a more bare bones version for people who dislike them. That being the wyci here

PR #5961, opened 2026-06-04
Random War

rebe83:
## About the PR

War declarator has a 1/2 chance to be in commander's bag

## Why / Balance

A tree PR arrives late
There seems to be a very high tendency for any nukies to be warops because "free tc".
However warops does get very tiring, and """"stealth"""" ops are cool despite all the mald they may generate.
I think adding a measure preventing EVERY nukeop to be war ensures things don't get too stale or meta, and nukies have to be willing to adapt.
It will suck if you have a gimmick you want to do, But there is nothing stopping a """stealth""" op from using their big red announcement console. They just won't have the extra TC for TDM mode.
I will admit the "risk of staleness" is probably much less likely after the PR that is invisible (i literally can't find it on github wtf) that fixes gamemodes from repeating
## Technical details

add commander bag + fill table
add commander bag to commander loadout

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: The TC economy is in shambles, now only 50% of nukeops have the option to declare war for extra TC.

### DIFF
--- a/Resources/Prototypes/Roles/Antags/nukeops.yml
+++ b/Resources/Prototypes/Roles/Antags/nukeops.yml
@@ -98,8 +98,9 @@
     neck: SyndicateWhistle
     pocket2: CommanderUplinkRadio45TC # DeltaV - allows them to buy commander armor
     outerClothing: ClothingOuterCoatSyndieCapArmored # DeltaV - removal of armor
-  inhand:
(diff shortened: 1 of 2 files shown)


Proxysseia:
we know what part of the warops cycle this pr is. just skip to the end

rebe83:
Everyone HATED it, closing

PR #5973, opened 2026-06-07
Fix salvage magnet asteroids not spawning in ore.

jessiscah23:
## About the PR
Fixes a long standing bug preventing proper ore generation on salvage magnet asteroids. Also removes an amendment to the salvage magnet UI that hid the fact that ore wasn't generating. 

## Why / Balance
You can now mine on mining asteroids. 

## Technical details
The asteroid generation was enhanced in November 2024, part of this was using AsteroidAltRock as the main tile rather than AsteroidRock. The ore generation was never updated to target the new tile type, so no ore was getting placed by the main ore generation (a few single random tiles were scattered by the base asteroid generator). This has now been rectified. 

Because no ore was getting placed, the description of ore types found on the UI was removed later in 2024. This change has been removed. 

## Media
Before:
image

After:
image

## Requirements

## Licensing
    

**Changelog**
:cl:
- fix: Nanotrasen has fixed issues with salvage magnet calibration. Ore seams can now be found on asteroid pulls!


Content.Client/Salvage/UI/SalvageMagnetBoundUserInterface.cs
```diff
@@ -85,7 +85,6 @@ protected override void UpdateState(BoundUserInterfaceState state)
             {
                 case AsteroidOffering asteroid:
                     option.Title = Loc.GetString($"dungeon-config-proto-{asteroid.Id}");
-                    break; // DeltaV: Skip ores since they aren't used with custom generation
```
ShepardToTheStars, line 88: Can you explain this a bit more? Are you saying that the resources that the UI shows won't match what would spawn?

jessiscah23: From reviewing old commits, what happened was that a month or so after ore generation broke, someone noticed that the listed ores on the magnet didn't reflect the actual spawns, and I guess didn't realize that there was an underlying bug causing this so threw in the above break statement to prevent the UI from displaying magnet asteroid ore listing.

Because the underlying bug is now resolved, the break is no longer necessary, so I removed it to revert to upstream behavior and display the ores that will be found on an asteroid pull.

ShepardToTheStars (approved):

PR #5975, opened 2026-06-07
Crowned Spirit

AeraAuling:
## About the PR
This PR adds the crowned spirit, a spirit of Ambition and Radiance with a sinister goal. These creatures seek out the wicked and morally corrupt and offer them an accord. They provide their new charge with protection, with power, with assistance with their goals. They will do all they can to see their charge succeed.. In exchange for one simple request. 

A singular extra life must be taken.

For those who are already willing to kill, this is not a hard bargain. The only catch is that their new charge must be the one to order the kill, or to cause the kill themselves. Regardless of how it's done, the target's death must be on their charge's hand. 

Of course, with all things from beyond the veil, there is a catch. After the deed is done, and their own goals are completed. At the height of their ambitions, with only an escape left to be made will the Crowned Spirit strike, attempting to cut down their charge as punishment. A high risk, but a high reward for those who are savvy and believe they can survive it's eventual turn.

## Why / Balance
The crowned spirit provides what I feel is an interesting turn on the "Familiar" archetype, not being a true familiar but a powerful ally all the same. This provides traitors a high risk high reward option for completing their goals, giving them a powerful ally and a powerful enemy at the same time. 

Their objectives are written in such a way as to push the spirit towards *not* completing their kill objective on their own. The spirit does not, however, know who is and is not a traitor. It must find that out on it's own, which provides opportunity for it to be mislead. Regardless of who makes a deal with it however, their charge will face a reckoning at the end.

In terms of capabilities, the spirit has a 20 slash 15 holy attack at 1.5 attack speed, for either 30 DPS, or.. I think it's 47.25? against kitsune. More against skeletons and other holy weak enemies, unknown for any future holy vulnerable species/pets/antags/etc. It has two main abilities in that it has access to Psionic Invisibility for sneaking around, hiding it's light, and making appropriately dramatic entrances to evil maints traitors. As is normal for psionic invisibility, it has pacifism. The Crowned Spirit can attack items and objects such as walls, lights, tables, and the like while invisibile, but it *cannot* harm living things. The invisibility has a cooldown of 60 seconds both for going into and for leaving invisibility, and is stunned for.. I think 3 seconds when leaving invisibility. It also has access to noospheric zap, which can be used during invisibility. This is for the purposes of stunning those who are attempting to capture their charge, and is also very spooky and ghosty and spirity. Both of these are, indeed, psionic powers. This means that they may be disabled, and even mindbroken. It also means that psionic insulation will protect from the spirit and reveal it while invisibile to those who are insulated.

also it can pass through doors like a mouse, because it's a half-physical ghost. 

unique for the ghost type antagonists, the Crowned Spirit is completely immune to Radiant damage, as well as slash. It also has high resistance to heat. Instead, it has a 2x weakness to cold damage (bring out the I.C.E.Es) and a 3x weakness to Soul damage (I wonder who has one of the main sources of crew-sided soul damage?) 

## Technical details
It is antag stuff. This is the first time I have made a midround antag, so it has all the stuff for that.

## Media

image
image
image
image

https://github.com/user-attachments/assets/f9d30b62-f7a2-4176-8ade-17406a8a7613

https://github.com/user-attachments/assets/e64619b1-6617-4c09-97c2-b38668fac8e6

https://github.com/user-attachments/assets/3e5bcd09-0d65-450a-83f3-48a0a85b75a0

https://github.com/user-attachments/assets/f23a40f7-eaab-45b4-82f4-f079bad72805

## Requirements

## Licensing

**Changelog**

:cl:
- add: Crowned Spirits have started to manifest on station, seeking to sow chaos to fulfill their own twisted sense of justice.


Bonktrauma:
Direction says we need media added to the pr to continue discussion, please fix

Resources/Prototypes/_DV/Objectives/crownedspirit.yml
```diff
@@ -0,0 +1,71 @@
...
+  - type: KillPersonCondition
+    requireDead: true
+  - type: TargetObjective
+    title: objective-condition-crownedspirit-corruption-title
+  - type: PickRandomPerson
+    filters:
+    - !type:BodyMindFilter
+      whitelist:
+        components:
+        - CommandStaff
```
ShepardToTheStars, line 45: Should this be limited to CommandStaff?

Resources/Prototypes/_DV/Roles/Antags/crownedspirit.yml
```diff
@@ -0,0 +1,9 @@
+- type: antag
+  id: CrownedSpirit
+  name: roles-antag-crownedspirit-name
+  objective: roles-antag-crownedspirit-objective
+  antagonist: true
+  setPreference: true
```
ShepardToTheStars, line 6: This isn't a roundstart antag so this should be false. 

```suggestion
  setPreference: false
```

ShepardToTheStars (commented):
I do find it a bit confusing if your target is considered your charge OR if you have to find someone to be your charge, and then have them kill the target. It's not clear who the charge is supposed to be once you spawn in.

Stop-Signs:
Is this still being worked on?

PR #5988, opened 2026-06-08
Cardbox Destruction

AffleWaffle:
Cardboard PR

## About the PR

Cardboard boxes now emit a sound when broken and spawn 1-2 cardboard or none if the damage was too excessive (simulating most others structures to avoid entity spawn spam in major explosions). 
The stealthbox is now destructible as it was not for some reason, it's 4 to 5 times as durable as a normal cardbox depending on the type of damage you use.

## Why / Balance

You could use them as indestructible barriers which would block projectiles, I was also informed this wasn't intended. Stealth capabilities and low TC cost can only get you so far when it comes to durability.

## Technical details

YML

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Stealthbox is now destructible!

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
@@ -65,6 +65,29 @@
   components:
     - type: Damageable
       damageModifierSet: FlimsyMetallic #Syndicate boxes should have a bit of protection
+    - type: Destructible # DeltaV - Destructible StealthBox
+      thresholds:
+      - trigger:
+          !type:DamageTrigger
+          damage: 90
+        behaviors:
+        - !type:DoActsBehavior
+          acts: [ "Destruction" ]
+        - !type:PlaySoundBehavior
+          sound: /Audio//Weapons/pierce.ogg
+      - trigger:
+          !type:DamageTrigger
+          damage: 45
+        behaviors:
+        - !type:SpawnEntitiesBehavior
+          spawn:
+            MaterialCardboard:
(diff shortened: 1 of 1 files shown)


rebe83:
Could you also apply this to regular junk cardboard boxes like survival boxes 
Asking for a friend

Proxysseia:
> **Triaged by starboundsolaris:** This PR requires a content review discussion, which will be held in public.
> 
> ```
> Direction review 3 days, adds content.
> ```
> 
> The review duration has been set to 3 days.

you should extend this to 7 days this is really impactful

Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
```diff
@@ -63,30 +63,67 @@
...
+      behaviors:
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          MaterialCardboard:
+            min: 1
+            max: 2
+      - !type:DoActsBehavior
+        acts: [ "Destruction" ]
+      - !type:PlaySoundBehavior
+        sound: /Audio//Weapons/pierce.ogg
```
ShepardToTheStars, line 126: Why change the normal box's behavior?

Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
```diff
@@ -63,30 +63,67 @@
...
```
ShepardToTheStars, line 90: This gets inherited from `BaseBigBox` so its not needed.

PR #6012, opened 2026-06-10
fix: don't show SSD examine on dead people/ghost role

JustTNE:
## About the PR
1. SSD timers no longer show on untaken ghost roles
2. SSD timers no longer show on the dead
3. SSD timers correctly track when players connect/disconnect instead of tracking when the mind is in the body
4. Catatonics now automatically receive a green SSD symbol instead of red
5. SSD timers no longer show on the catatonic

## Why / Balance
Untaken ghost roles: Not useful information. There is no player attached and the catatonic info text already provides all of the needed information.
Dead: To avoid metagaming in cases where a player leaves directly after dying and to discourage treatment priority based on logout time.
Track connect/disconnect: A more accurate metric. Does not get confused by ghosts leaving their bodies.

## Technical details
Completely ceased usage of the `IsSSD` member variable for the SSD timers (indicators still use it). `SsdSince` is now nullable, only not null when player properly SSD (disconnected). `IsSSD` is true when the player is out of their body.
Created new event `MindStateUpdatedEvent`, raised by `MindExamineSystem`, which already accurately tracks player SSD status.
If `SSD` or `DeadSSD`, start keeping track of the time. If this changes to any other value, reset the time to 0.

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Don't display SSD duration when examining dead players or untaken ghost roles

### DIFF
--- a/Content.Client/Administration/AdminNameOverlay.cs
+++ b/Content.Client/Administration/AdminNameOverlay.cs
@@ -237,11 +237,12 @@ internal sealed class AdminNameOverlay : Overlay
             }
 
             // DeltaV - SSD Time START
-            if (_entityManager.TryGetComponent<SSDIndicatorComponent>(entity, out var ssdIndicator) && ssdIndicator.IsSSD)
+            if (_entityManager.TryGetComponent<SSDIndicatorComponent>(entity, out var ssdIndicator)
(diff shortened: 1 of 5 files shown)


ShepardToTheStars (approved):
Tested this locally. Looked fine. Behaved as expected. Code looks good too.

PR #6017, opened 2026-06-11
Added dynamic cost to Ratkings OnRaiseArmy ability

CarolCuchara:
## About the PR

Changed the Rat Kings SummonArmy ability from a flat 25 hunger cost to a dynamic cost, based on the amount of alive servants. This creates a softcap on the amount of servants the Rat King can have.

## Why / Balance

Player performance suffers if the Rat King has to many servants. This PR softcaps the amout, making it harder to create a to large horde. Using a dynamic cost could also allow the Rat King to gain his first servants more easily, because they are cheaper, allowing for a better early game and bounce back should all servants die.

The prices below are just an example, please provide some guidance on what scaling to use.

## Technical details

Changed `hungerPerArmyUse` from 25 to 10 hunger as a base value, which gets multiplied by a modifier. Evaluation happens in `RatKingSystem`, with only alive servants being counted, this decides the the modifier, resulting in a final price for the ability.

Capture

## Media

Capture

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Rat Kings raise army ability is now dynamically priced. The first servants are cheaper, allowing for an easier early game and bounce back. Too many servants will increase the price drastically, with giant swarms killing the performance very hard to achieve!


CarolCuchara:
This is my first PR, so might not be totaly up to standard. I tested it localy and it works, if you want, please give it a second look.

Content.Shared/RatKing/RatKingComponent.cs
```diff
@@ -18,11 +18,13 @@ public sealed partial class RatKingComponent : Component
...
 
+    // Delta-V - switched to a base cost modified by the amount of living servants
     /// <summary>
-    ///     The amount of hunger one use of Raise Army consumes
+    ///     The base amount of hunger one use of Raise Army consumes, modified by the existing amount of servants
     /// </summary>
     [ViewVariables(VVAccess.ReadWrite), DataField("hungerPerArmyUse", required: true)]
-    public float HungerPerArmyUse = 25f;
+    public float HungerPerArmyUse = 10f;
+    // end DeltaV
```
ShepardToTheStars, line 27: nit: comment can just be one line here.

```suggestion
    /// <summary>
    ///     The amount of hunger one use of Raise Army consumes
    /// </summary>
    [ViewVariables(VVAccess.ReadWrite), DataField("hungerPerArmyUse", required: true)]
    public float HungerPerArmyUse = 10f; // DeltaV - was 25, but we scale cost
```

Content.Server/RatKing/RatKingSystem.cs
```diff
@@ -47,14 +49,42 @@ private void OnRaiseArmy(EntityUid uid, RatKingComponent component, RatKingRaise
...
+            foreach (var existingServant in component.Servants)
+            {
+                if (!TryComp<MobStateComponent>(existingServant, out var mobState))
+                    continue;
+
+                if (mobState.CurrentState == MobState.Alive)
+                {
+                    aliveServants++;
+                }
+            }
```
ShepardToTheStars: Can be condensed into a single line.

```suggestion
            var aliveServants = component.Servants.Count(servant
                => TryComp<MobStateComponent>(servant, out var mobState) && mobState.CurrentState == MobState.Alive);
```

Content.Server/RatKing/RatKingSystem.cs
```diff
@@ -47,14 +49,42 @@ private void OnRaiseArmy(EntityUid uid, RatKingComponent component, RatKingRaise
...
+
+            // calculate the cost multiplier
+            var multiplier = aliveServants switch
+            {
+                < 5  => 1.0f,   // 1-5 servants: 10 hunger
+                < 10 => 1.5f,   // 6-10 servants: 15 hunger
+                < 15 => 2.5f,   // 11-15 servants: 25 hunger
+                < 20 => 5.0f,   // 16-20 servants: 50 hunger
+                < 25 => 10.0f,  // 21-25 servants: 100 hunger
+                _    => 15.0f,   // Above 25 servants: 150 hunger
```
ShepardToTheStars, line 66: Not a huge fan of hard-coding this in the system but I suppose its fine.

Content.Shared/RatKing/RatKingComponent.cs
CarolCuchara: Cant apply this, gives me an error on Github. Will do it by hand

Content.Server/RatKing/RatKingSystem.cs
CarolCuchara: This gives me errors

CarolCuchara: Was missing using directive for Linq

Content.Server/RatKing/RatKingSystem.cs
CarolCuchara: I tried making a dictionary with the values, but did not get it work with the .yml

ShepardToTheStars: Yeah, you could probably do it some way but translating it to a switch statement would be messy, which is why I was fine with it in the end.

ShepardToTheStars (approved):

PR #6021, opened 2026-06-12
terrorism murder b&e

Proxysseia:
## About the PR
Terrorism requires terrorist affiliation
B&E is no longer breaking into an HSA with intent to commit crime within, its just breaking and entering somewhere
murder is 15m instead of perma

## Why / Balance
Terrorism can be charged for whatever, eink asked i do this.
B&E was stupid and now actually does what people charge it for anyways.

Murder:
right now dying has very little consequence and people act like it. Except for when it comes to charging people for crimes. In general it takes security more time to drag the murderer to their cell, than it takes med to revive the victim. Its not uncommon to have the murder victim attend their own hearing, but to stare at the perp while they get searched for the first time.  This PR puts us closer to what most other servers do, in treating Prevention of Revival as murder, and murder as just pretty severe assault.

I suggest that in the future we seriously look at making death more impactful, and once that is done, put murder back up to capital. In the mean time we can have some more ludonarrative harmony.

## Requirements
- [ x ] I have not tested any added content and changes.
- [ x ] I have not added media to this PR even, if does require an in-game showcase.

**Changelog**
:cl:
- tweak: Murder is now a Grand Felony with a sentence of 15 minutes.
- tweak: Terrorism now requires you to be affiliated with terrorists.
- tweak: Breaking and Entering now means what you think it does.


rebe83:
Terrorism, to commit crime while valid

JustTNE:
If we ever get Offmed, we can set murder back to capital and POR can be removed, since death is incurable in Offmed :P (Specifically because Offmed has a standard medical "brain death" definition of death instead of no heartbeat)

PR #6026, opened 2026-06-13
kill cult metashield

Proxysseia:
## Why / Balance
no other game mode relies on the metashield to this insane a degree. kill it and let cultists do what they usually do but now we dont have to have annoying arguments over what constitutes noöspheric tampering and what is just 'being a cultist and thus not illegal'.
technically every 'deconvert every cultist they are mindwashed code red cultists kos' is an illegal order that should get cap jailed. thats stupid.

## Requirements
- [X] I am cool and awesome
- [X] I believe my PRs are goated as am I.

**Changelog**
:cl:
- remove: "Cosmic Cultists want to end the world" is now common knowledge. Their metashield has been removed.

### DIFF
--- a/Resources/ServerInfo/Guidebook/_DV/Rules/GameRules/2_Metagaming.xml
+++ b/Resources/ServerInfo/Guidebook/_DV/Rules/GameRules/2_Metagaming.xml
@@ -80,14 +80,6 @@
    - being a Command role
    - being a traitor
 
-  ## Cultists
-
-  The fact that Cultists want to bring about the end of the world.
-
-  The revealing condition for this shield is any of the following:
-  - being a Captain
-  - being a Mystagogue
-  - being a Cosmic Cultist
 
   ## Explicitly Not Shielded
   The following is a list of things that are explicitly not shielded. If something is not on this list, it doesn't mean that it is shielded, but if something is on it then it definitely is not shielded.
@@ -95,7 +87,7 @@
   - The fact that the nuke disk must be protected and could be used by a bad actor to try to destroy the station.
   - The fact that the Syndicate are enemies of Nanotrasen, and that they regularly attempt to send covert agents to spy on, sabotage, or attack Nanotrasen.
   - A character's typical appearance. Though you should keep in mind that multiple characters can share the same name.
-  - The fact that Cultists are hostile to NanoTrasen, exist in some capacity, and should be reported.
+  - The fact that that Cultists want to bring about the end of the world.
 
   ## Metafriending and Metagrudging
   This section provides additional information on a concept that is prohibited by multiple metashield items that are never revealed IC. Giving a person or character preferential treatment based on something that your character should not know is considered metafriending. Treating a person or character negatively based on something that your character should not know is considered metagrudging.


SalDragonsNOA:
The issue is that, gameplay wise, it's hard to get the cult going without the meta shield. They get dealt with too fast without it.
The point of the cult IS to be a threat. If someone finds out and they just gets curb stomped before they even get going, that's not a threat.

It's also VERY hard to make cult a threat without going over board with how they recruit people. Since the monument and their overall progression as a threat relies on recruitment, you'd have to rework the whole thing to get rid of the metasheild requirement.
It's not that it's impossible, it's that most folks wouldn't like it changing so drastically to remove the metashield.

NoElkaTheGod:
Cosmic cult is absolutely playable without any form of metashield whatsoever. However, being **playable** is not enough, it should also be **fun** with no metashield, and you'd need some design alterations to achieve that. The current version with no metashield just turns into "catch tiders in/near maints, shunt subjectivity, cuff, bring to the monument, convert, repeat until win", which is... really just revs with extra steps.

rebe83:
> Cosmic cult is absolutely playable without any form of metashield whatsoever. However, being **playable** is not enough, it should also be **fun** with no metashield, and you'd need some design alterations to achieve that. The current version with no metashield just turns into "catch tiders in/near maints, shunt subjectivity, cuff, bring to the monument, convert, repeat until win", which is... really just revs with extra steps.

I would call it revs + flavor much as replicators are bingles+ flavor
There are certainly mechanical changes where it is a different enemy but the game flow is the same 
If we have no metashield in cult I think they should get instant shunt just as revs get instant convert with flashes, they still have to drag to a vulnerable single location that is easily compromised by loud literally unique and unmistakeable magical noises and glowing drawings. They are turbo valid but we just have to act like they aren't because sec can stomp them. 
II get the advantage of ambush as a snowball antag but cult doesn't. I think instant shunt lets them fight off sec camping monument SOMEWHAT as well

NoElkaTheGod:
For love of god do NOT make shunt instant, it would be way too evil

rebe83:
> For love of god do NOT make shunt instant, it would be way too evil

Cult currently does nothing and gets shut down at minutes 20 like 7/10 times. If shunting is going to reveal the whole game mode they are going to need to have hands from minute 0

Proxysseia:
> Cult currently does nothing and gets shut down at minutes 20 like 7/10 times. If shunting is going to reveal the whole game mode they are going to need to have hands from minute 0

with your help we can get that number even higher

NoElkaTheGod:
> Cult currently does nothing and gets shut down at minutes 20 like 7/10 times. If shunting is going to reveal the whole game mode they are going to need to have hands from minute 0

Buff other things if that's the case. I don't need to explain why an instant 10 second stun is gonna be overpowered as fuck. Consider instead:
- Removing siphon damage (there were like 5 PRs for this, maybe one got merged already while I wasn't looking), so it's still viable even without metashield
- Making influences cheaper and/or stronger, because astral nova is basically the only good one
- Making entropy not mutually exclusive with converts (this one would need some serious effort and probably a full redesign)

ShepardToTheStars (approved):

ShepardToTheStars:
@IAmNotGray You'll need to approve as well.

IAmNotGray (approved):

PR #6029, opened 2026-06-14
Change block coefficient for shields from .5 to .35, causing shields …

Woljif:
## About the PR

Changed the amount of damage that shields split with the user from 50/50 to 35/75, meaning shields now take 35% of the damage taken from 50%.

## Why / Balance

This actually doesn't change player crit breakpoints by that much at all, often not changing them at all. For example, a player in a security hardsuit will go down to a dragon in 8 swings either way. This is because formerly, the shield would block more damage, but then break on the 6th hit, leaving the player shieldless for the remaining two hits until they crit. Now, the shield does not break at all due to splitting less damage up front. But because it blocks less damage, you go down in 8 hits anyway. This is basically a change that's intended to disincentivize picking up a shield and receiving too large of an up front power boost

This change affects all shields since it's a change done on the parent shield entity, but if we want, say, nukie shields to remain strong, you could make the energy shield still split 50% of damage taken, while other shields sit at 35%, or some combination thereof.

## Technical details

a single line of yml

## Requirements

## Licensing

**Changelog**
:cl:
- tweak: Reduced the passive block of shields from 50% to 35%, changing them to block less damage up front while still retaining the same amount of overall damage blocked over the lifetime of a single shield.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Shields/shields.yml
+++ b/Resources/Prototypes/Entities/Objects/Shields/shields.yml
@@ -14,6 +14,7 @@
       size: Ginormous
       #heldPrefix: riot #Harmony Change: Removed due to being redundant
     - type: Blocking
+      passiveBlockFraction: 0.35 # DeltaV - Reduce default block coefficient from .5 to .35
       passiveBlockModifier:
         coefficients:
           Blunt: 0.9


ShepardToTheStars (approved):
Looks fine from a maintainer standpoint

Woljif:
Looks like content review has a better direction to go with shields seeing the changes being proposed in [warock's PR](https://github.com/DeltaV-Station/Delta-v/pull/6038#issuecomment-4723296486) so I'm going to proactively close this

Issue #5531, opened 2026-03-17
Zombie Laika's Pocket Headset


Acoyph:
Because Laika can have a headset in her pocket (instead of actual ear slots) she does not drop her radio when becoming zombified and can continue to listen to comms.

Zombies dropping radios is great, and as Laika I intentionally try to stay with zombies who can pry doors anyway... but technically able to spy on comms to meet up/track down with survivors.

CarolCuchara:
Hello, i made a fix for this in #6030

PR #6030, opened 2026-06-14
Remove headset from pockets when player get zombified

CarolCuchara:
## About the PR

Removes the headset from Laikas and the generic K9s pocket when they get zombified.
Fix for #5531 

## Why / Balance

Normal characters drop their headset when the get zombified, so they cant listen on comms/ metagame. The same should also apply to the Secdogs.

## Technical details

Added a check in `ZombieSystem.Transform` if the target is Laika or a K9. If a headset is in the pocket, drop it.

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Zombified crew and other entities like Laika will drop headsets which are in their pockets. No more listening to survivor comms!

### DIFF
--- a/Content.Server/Zombies/ZombieSystem.Transform.cs
+++ b/Content.Server/Zombies/ZombieSystem.Transform.cs
@@ -246,17 +246,15 @@ public sealed partial class ZombieSystem
         //Should prevent instances of zombies using comms for information they shouldnt be able to have.
         _inventory.TryUnequip(target, "ears", true, true);
 
-        //DeltaV - remove radio form pockets of Laika and normal security dogs
-        var prototype = MetaData(target).EntityPrototype?.ID;
-        if (prototype is "MobSecDogLaika" or "MobSecDog")
+        // BEGIN DeltaV - Remove innate radio and radios from pockets
+        for (var i = 1; i <= 4; i++) // Arachnids have 4 pockets
         {
-            if (_inventory.TryGetSlotEntity(target, "pocket1", out var headset)
-                && HasComp<HeadsetComponent>(headset))
-            {
-                _inventory.TryUnequip(target, "pocket1", true, true);
-            }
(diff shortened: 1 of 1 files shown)


ShepardToTheStars:
So, I did some testing and I think I might expand this to not only Laika/Sec dogs but also anyone with pockets, because zombies can put headsets in their pocket and hear radio channels. I'll suggest a fix shortly. 
image

Content.Server/Zombies/ZombieSystem.Transform.cs
```diff
@@ -245,6 +246,18 @@ public void ZombifyEntity(EntityUid target, MobStateComponent? mobState = null)
...
+        var prototype = MetaData(target).EntityPrototype?.ID;
+        if (prototype is "MobSecDogLaika" or "MobSecDog")
+        {
+            if (_inventory.TryGetSlotEntity(target, "pocket1", out var headset)
+                && HasComp<HeadsetComponent>(headset))
+            {
+                _inventory.TryUnequip(target, "pocket1", true, true);
+            }
+        }
+        //End DeltaV
```
ShepardToTheStars: This should cover all crew hiding headsets in their pockets, and also get rid of innate radios from things like nukie mice/spiders and diona.

```suggestion
        // BEGIN DeltaV - Remove innate radio and radios from pockets
        for (var i = 1; i <= 4; i++) // Arachnids have 4 pockets
        {
            if (_inventory.TryGetSlotEntity(target, $"pocket{i}", out var headset) && HasComp<HeadsetComponent>(headset))
                _inventory.TryUnequip(target, $"pocket{i}", true, true);
        }

        RemComp<ActiveRadioComponent>(target); // If the zombie has an innate radio, get rid of it.
        // END DeltaV
```

CarolCuchara:
@ShepardToTheStars Thank you. I defenitly forgot about the existance of the innate radios. Dropping headsets from everyones pockets also makes sense, because i thought that only the dogs can hear the headsets in pockets.

ShepardToTheStars (approved):

PR #671, opened 2024-01-14
Spare ID safe

DebugOk:
## About the PR
Still a work in progress

TODO:
- [x] Figure out who should get the biscuit
- [X] Get a sprite for the safe
- ☒ ~~Figure out what to put on the card that opens the safe~~ Fuck it, map it
- [x] Get something for the paper background of the cards

## Media

https://github.com/DeltaV-Station/Delta-v/assets/49997488/b1fd56b4-ae88-4ce8-babe-f1c6716dee70

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Added the spare ID safe
- remove: The captain's locker no longer starts with a spare ID card

PR #6044, opened 2026-06-16
Fixed emergency access locale string

ShepardToTheStars:
## About the PR

Updated the spare ID access from "You shouldn't see this..." to "Emergency Command"

## Why / Balance

Because we can see it now when you inspect something with access.

## Technical details

`id-card-access-level-unused` was added back in PR #671 but was not namespaced, so I did that too.

## Media

image

## Requirements

## Licensing

**Changelog**

no CL needed


Stxcking (approved):
Why did we have this in the first place? IDK, LGTM

Issue #5742, opened 2026-04-27
Rejuv'ing dead xeno presents crashing bug


uwuKemonoChan:
## Description
When you kill a xeno and revive it, it send a server crashing bag. This was done on a local test environment and may not apply to the main branch; however, it is still something to look out for.

**Reproduction**
> Kill xeno
> Use debug to rejuvenate 
> Go "oh" as server crashes

**Crash logs** 
I sanitised any PPI in it for personal reasons

[crashlogs.txt](https://github.com/user-attachments/files/27142437/crashlogs.txt)

CarolCuchara:
First the blood is set to 0, then it will try to set it back to the reference of 650. The datatype of the blood amout is the struct `FixedPoint2`.  Inside the struct the value is stored in an int, with two digits added (the two decimal values). So the 650 acutally is stored as 65.000.

When readding the blood this line multiplies `amount` and `referenceQuantity`, which both are 650 so 65.000 * 65.000:
`var adjustedAmount = amount * referenceQuantity / ent.Comp.BloodReferenceSolution.Volume;`

The way two `FixedPoint2` are multiplied is `a.Value (65000) * b.Value (65000) / ShiftConstant (100)` 

The result of `a.Value (65000) * b.Value (65000)` is 4.225.000.000 causing an integer overflow. This warps to −69.967.296 and further down the line leads to trying to add a negative amount of blood, crashing the server.

This also happens with the space dragon and colossus, which also have 650 blood.

PR #6047, opened 2026-06-16
Reduce blood amount of colossus and dragon to 400

CarolCuchara:
## About the PR

Right now rejuvenating the colossus or dragon crashes the server. This happens because rejuvenating is multiplying two large `FixedPoint2` variables can lead to an integer overflow.

The way two `FixedPoint2` are multiplied is `a.Value * b.Value / ShiftConstant (100)`. The values are of type int with two following zeroes added for the two decimal points. This can lead to very large factors causing the overflow. 

## Why / Balance

#5742 Rejuvenating a xeno, colossus and space dragon crashes the server because of an integer overflow.

## Technical details

Reduced to blood amount to 400 so there is no overflow

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Rejuvenating a Colossus or Space Dragon will no longer crash the server!
- tweak: Reduced Colossus and Space Dragon blood amount to 400!


ShepardToTheStars:
I'm a bit wary of touching this function because FixedPoint2 is used in so many places. While yeah, this is likely a bug with FixedPoint2 overflowing, I'd rather see this get submitted to upstream since this is used so widely.

For now, might I suggest reducing Xeno's bloodstream to like 400, and that should at least address the crash in dev environments.

CarolCuchara:
Yes, thats what i wanted to know, if we should even touch that or work arround it. I will take a look at the function that causes the overflow, i dont belive it has to a that big mutliplication.

CarolCuchara:
@ShepardToTheStars I have reverted to changes to the calculation and reduced the blood amount of each entity to 400. No more crashes.

Toby222 (approved):
Curse everyone who decided that "int" should be "signed 32 bit", and not making explicitly clear what each type's number range is

PR #6073, opened 2026-06-21
Fix Unknown Fax Notifications

ShepardToTheStars:
## About the PR

If the fax machine is set to notify admins, refresh before receiving.

## Why / Balance

image

## Technical details

Still won't work for antag fax machines since they usually have ResponsePings set to false, but this will cover 99% of faxes sent to CentComm.

## Media

n/a

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Fax machines should properly show where a fax came from in their text popup.

### DIFF
--- a/Content.Server/Fax/FaxSystem.cs
+++ b/Content.Server/Fax/FaxSystem.cs
@@ -588,22 +588,25 @@ public sealed class FaxSystem : EntitySystem
             return;
 
         var faxName = Loc.GetString("fax-machine-popup-source-unknown");
+
+        // Start DeltaV - Refresh Fax IDs before recieving
+        // This will seem odd that we're doing the same if statement twice BUT if its not in the known fax machine list, the
+        // sending fax might not have been online when the current fax machine was refreshed/init'd. So, if its already known,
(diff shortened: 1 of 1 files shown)


Content.Server/Fax/FaxSystem.cs
```diff
@@ -588,6 +588,15 @@ public void Receive(EntityUid uid, FaxPrintout printout, string? fromAddress = n
             return;
 
         var faxName = Loc.GetString("fax-machine-popup-source-unknown");
+
+        // BEGIN DeltaV - Refresh Fax IDs before recieving
+        // This will seem odd that we're doing the same if statement twice BUT if its not in the known fax machine list, the
+        // sending fax might not have been online when the current fax machine was refreshed/init'd. So, if its already known,
+        // we don't need to refresh. But if its not known, let's attempt to refresh first, then check the fromAddress in the KnownFaxes.
+        if (fromAddress != null && !component.KnownFaxes.TryGetValue(fromAddress, out var _))
```
SirWarock, line 596:
```suggestion
        if (fromAddress != null && !component.KnownFaxes.TryGetValue(fromAddress, out _))
```

SirWarock (approved):
Works on release. Doesn't work with newbuilt faxes.

PR #6079, opened 2026-06-21
increase execution damage

Proxysseia:
## About the PR
increase execution damage modifier from 9 to 15

## Why / Balance
i cuffed a secoff and forgot to take their armor off so they survived me executing them with a combat knife, idk how to make executing AP so this is what we get instead

number is made up because i felt like it.

## Requirements

- [X] I have not  tested any added content or changes.
- [X] I have not added media to this PR even if it does require an in-game showcase.

**Changelog**

:cl:
- tweak: Executing now deals 15x damage (from 9x)


Vova567810:
Mald pr or sec buff?
no-country-for-old-men-anton-chigurh

PR #6085, opened 2026-06-22
Fix Infinite Cloaking

ShepardToTheStars:
## About the PR

Fixed infinite cloaking bug. Sorry Sabo borgs.

## Why / Balance

Bug introduced by Predicted Batteries

## Technical details

So, for some reason, drawing power does not activate the cooldown specified by `BatterySelfRechargerComponent.AutoRechargePauseTime`. It's only applied when explicit charge is used on activating. So even if we put an activation cost, it would still recharge 30 seconds after that initial cost. So, I added something that starts the AutoRecharge timer when `PauseRecharging` is set to true, and then as long as the net charge is negative, then it will keep extending that timer.

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Stealth borg modules are now working as intended and no longer provide infinite duration cloaking.


Toby222 (approved):
seems sensible

PR #6095, opened 2026-06-23
Xenos again

Dorragon:
## About the PR

Lots of things changed about xenos again

TLDR: xenos are stronger now, new xenos, xenovents slightly nerfed

Full list of changes (except ones i forgot):
Added Crusher, Boiler and Lurker
Added new xenos to xenovents, slightly nerfed amount of xenos that come out of xenovents
Changed xeno's blood chemical to XenonidBlood, reskin of acid it was before that fits their dead sprite and can't hurt them
All xenos now slowly regenerate. The rate of regeneration is 3 damage per minute X their tier (aka tier 3 xenos regen 9 per minute)
Changed xenos body type to bloodsucker, adding a way to heal for them until we figure out something better
Xenos can now pull
Xenos no longer deal any friendly fire damage
Raised SlowOnDamage thresholds to match their health. Melee combat xenos have higher thresholds
Xenos now have critical state
Stamina and prying speed now slightly differ with xeno tier
Ravager now has ravager sprite instead of lurker one since lurker is here
Ravager now has auto attack on
Spitter has 50 less health (300 > 250) to match second tier xenos more
## Why / Balance

Xeno playtest feedback + other stuff

Honestly xenos as major antag sounded a bit silly at first to me but after a bit of thinking i think it has potential to be really good. Also since we are making xenos very strong on delta might as well double down
We can't be RMC and we don't need to be RMC but as a server that is pretty much fully dedicated to combat they have ways to make combat more engaging than just click in enemy direction that we can take and change to fit normal station gameplay. Probably. This PR is pure yml though so it does almost nothing of this :trollface:
## Technical details

Not a single C# line. Lurker pounce and Crusher rush are combination of jump ability and traffic hazard comp. Should probably be redone properly later if there is any more work done on this
Also there are so many changes to xeno file we really should consider moving all new xenos to our own and reverting old one to upstream state
## Media

Crusher 

https://github.com/user-attachments/assets/e7aaa485-7512-434a-baf0-fbd1763bb543

Boiler 

https://github.com/user-attachments/assets/618ad9a4-4c66-424b-b043-a79e8d29065a
roughly 40 damage split between burning and caustic if you stand there for the full duration with no internals
roughly 30 damage with internals
affected by armor values
xenos don't take any of that damage 

Lurker 

https://github.com/user-attachments/assets/11c8bf50-1852-49fc-b199-9bdfd43b6a39

## Requirements

## Licensing

**Changelog**

:cl:
- add: Crushers, Boilers, and Lurkers find their way to the starbird and reunite with their hive
- tweak: Xenos evolve yet again. NanoTrasen really should clear those vents sometime soon...


Resources/Locale/en-US/_DV/reagents/meta/toxins.ftl
```diff
@@ -2,3 +2,6 @@ reagent-desc-heartbreaker-toxin-deltav = A hallucinogenic compound derived from
 
 reagent-name-lernaea = Lernaea
 reagent-desc-lernaea = A vicious poison, speculated to originate from a river of death. It ravenously consumes noöspheric energies in the mind while it's present.
+
+reagent-name-xenonid-boiler-acid = Boiler acid
```
ShepardToTheStars:
```suggestion
reagent-name-xenonid-boiler-acid = boiler acid
```

Resources/Prototypes/_DV/Reagents/toxins.yml
```diff
@@ -12,3 +12,53 @@
...
+        emote: Scream
+        probability: 0.3
+  metabolisms:
+    Poison:
+      metabolismRate: 3.00
+      effects:
+      - !type:HealthChange
+        conditions:
+        - !type:MetabolizerTypeCondition
+          type: [ Bloodsucker ]
```
ShepardToTheStars, line 47: Any reason for the condition? I don't think this is blood that Bloodsuckers should be immune from? >.>

Resources/Prototypes/_DV/Reagents/biological.yml
```diff
@@ -39,3 +39,37 @@
...
+        - !type:MetabolizerTypeCondition
+          type: [ Bloodsucker ]
+          inverted: true
+        type: Local
+        visualType: Large
+        messages: [ "generic-reagent-effect-burning-insides" ]
+        probability: 0.33
+      - !type:Emote
+        emote: Scream
+        probability: 0.25
```
ShepardToTheStars, line 75: Not really needed if you're just going to parent from `FluorosulfuricAcid`
```suggestion
```

Resources/Prototypes/_DV/Reagents/toxins.yml
Dorragon: To prevent boiler acid from damaging other xenos. Bloodsucker seems to be rare enough to only ever matter in super niche cases (xenos VS dragon?), so i used it for xenos to not make another metabolizer type for two chemicals. Unsure if there is a better way to make it not affect only xenos

Resources/Prototypes/_DV/Reagents/biological.yml
Dorragon: It's a copy of acid but with a bloodsucker inverted condition. Since xenos now heal by drinking blood and vast majority of them are melee, they are often going to be drinking from mixed puddles of their own blood and blood of their enemies. Didn't want xenos to get damaged by drinking their own blood in this case

PR #6101, opened 2026-06-23
Custom Client-Side Radio Colors

ShepardToTheStars:
## About the PR

You can now set your own custom chat channel color in the client settings.

## Why / Balance

Accessibility and for those who want red sec chat back I guess.

## Technical details

Chatcode gonna be chatcode.

The server sets the color in the RadioChannel prototype's color field, but we need to replace that once we get to clientside. Problem was we didn't have access *which* channel it was sent on, so I had to include that so I could get the channel color client-side to replace.

Had to make the color property in RadioChannel public set, so I could overwrite it on the client-side when the option is set.

Need to figure out if I can find a good way to fix the headset/encryption key examine markup still. 

## Media

https://github.com/user-attachments/assets/cbbf9b63-9fd1-4e53-aca5-ea38ee0c73da

## Requirements

## Licensing

**Changelog**

:cl:
- add: You can now change common radio channel colors to the colors of your choice in the Radio tab in the Options menu!


TehFlaminTaco (dismissed):
Code lgtm

TehFlaminTaco (approved):
Fix looks good

ShepardToTheStars:
Whew, fixed the CL before merge.

PR #6117, opened 2026-06-25
I hate submodules. Why did it have to be submodules?

ShepardToTheStars:
NuGet is better.

Yeah, I'm not filling that out.

People, take my fix if you want it. AGPL/MIT/whatever.


Toby222 (approved):
If it works it works
Probably open a reversion PR preemptively right after

ShepardToTheStars:
> If it works it works Probably open a reversion PR preemptively right after

Can't really revert until we get to the CURRENT engine version (released in the last hour)

PR #6119, opened 2026-06-25
I love submodules. Always use submodules.

ShepardToTheStars:
I don't know what came over me in #6117.

Always use submodules. Smile.

Pre-emptive fallback fix to appease CODEOWNERS before they eep. In case of issues, break glass and merge, but only if server is literally not useable.

### DIFF
--- a/.github/workflows/build-docfx.yml
+++ b/.github/workflows/build-docfx.yml
@@ -12,21 +12,12 @@ jobs:
     - name: Setup submodule
       run: |
         git submodule update --init --recursive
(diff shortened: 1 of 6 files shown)


Toby222 (approved):
This PR exists both as a reminder to undo that workaround when we get to updated RT, and in case the previous PR explodes something

ShepardToTheStars:
Do not merge until RT 277.2.1 is merged in.

ShepardToTheStars:
They released new tags of the patches applied to commits, so we'll have to watch for those until we catch up to 277.2.1

PR #6120, opened 2026-06-25
Fix: Make Xeno Announcement the same as other vent-spawns

Vapetastic-Gaming:
## About the PR
Changes back the xeno ventspawn announcement from the "All knowing metaknowledge announce called Aliens.ogg" to the standard ventspawn announcement

## Why / Balance
the vent spawn announcement and otherwise is the same, the audio is being used to "meta" the game rule a bit since this one sounds louder etc etc, people can tell if xenos are coming, and regularly call it out prior

## Requirements

## Licensing

**Changelog**
:cl:
- tweak: Xeno Announcements are now properly carried out, They no longer have their own announcement

### DIFF
--- a/Resources/Prototypes/_DV/GameRules/events.yml
+++ b/Resources/Prototypes/_DV/GameRules/events.yml
@@ -46,7 +46,7 @@
   - type: StationEvent
     startAnnouncement: station-event-xeno-vent-start-announcement
     startAudio:
-      path: /Audio/_DV/Announcements/aliens.ogg
+      path: /Audio/_DV/Announcements/attention.ogg
     earliestStart: 25
     minimumPlayers: 30
     weight: 2


rebe83:
I believe this was deliberately added

StarboundSolaris:
I just had an idea, instead of making it all the same announcement, what if we made the vent events like... have a random chance to play one or the other.

keekee38:
howdy CR here there was a small issue
anyhow gonna say 2 days on it
and thank ya for it and all!

ShepardToTheStars (approved):

keekee38:
heya! sorry for the long wait
so basically, especially with how xenos have been buffed lately, we think it would be best if they had their own distinct announcement. thank ya for your efforts though

PR #6127, opened 2026-06-26
Convert Round to Survival when Major Antag Fails

ShepardToTheStars:
## About the PR

If a major antag is defeated, then the Basic Scheduler will be replaced with a ramping scheduler. There's a 10 minute grace period before the scheduler will start.

It triggers if:
* all initial infected are killed
* all nukies are killed
* there are zero or one cosmic cultist
* cosmic cultists have made ZERO progress (no entropy or converts) in 30 minutes.

## Why / Balance

People think nothing is going to happen the remainder of the shift when they defeat an antag. A lot of times, curators will add a ramping rule when antags are defeated, so this automates that.

If the major antag did enough damage, the crew can always call evac normally. 

## Technical details

Instead of ShuttleCall, I added another option that adds a game rule. I had to bring the RoundEnd function out of the ActiveGameRule enumerators, because it was adding the ramping rule to the collection and causing an issue.

Also switched up the CosCult lose condition to be if there are 1 or less alive because one cultist can't do anything.

## Media

https://github.com/user-attachments/assets/86a1acbb-c2d5-4084-b533-2fdfbd9001be

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: If the crew defeats the major antag (Nukies, Cosmic Cult, Zombies), the round will now automatically become survival after a bit of time until the crew evacs.


sowelipililimute (approved):
a win for unit testing

PR #6138, opened 2026-06-29
Use the actual backported engine fix

sowelipililimute:
this should hopefully fix our publish

### DIFF
--- a/.github/workflows/build-docfx.yml
+++ b/.github/workflows/build-docfx.yml
@@ -12,21 +12,12 @@ jobs:
     - name: Setup submodule
       run: |
         git submodule update --init --recursive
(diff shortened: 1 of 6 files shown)


ShepardToTheStars:
We had a reversion PR pre-approved in case of this #6119, but it won't be pointed at the sub-module with the update.

ShepardToTheStars:
Also, did they overwrite the tag with a reference to a new commit hash?

sowelipililimute:
unsure of what happened to the git history :^)

ShepardToTheStars:
https://github.com/space-wizards/SpaceWizards.Lidgren.Network/pull/26

They just reverted it 2 days ago. That checks out.

PR #6141, opened 2026-06-29
FIX: Emergency access doesn't change the BOLT/Emergency permissions on a door using remote control

ManuDemoen:
## About the PR

Prevent remotes without sufficient access to BOLT/EMERGENCY doors that are currently set in EMERGERNCY mode.

## Why / Balance

Remotes should only be able to close/open doors that are in AA, if they do not have the required access they shouldn't be able to change the bolt status or the AA status

## Technical details

Extra check when a remote operation isn't open/close, to see if they have sufficient access

## Media

Before: can change AA/Bolt using MED remote on SEC door if it's in AA mode
https://github.com/user-attachments/assets/d5e89a1b-2f36-42bd-937b-7d58b6be82d1
After: Fixed
https://github.com/user-attachments/assets/dffadcd8-3d2f-4379-b520-1ae8e19e83ea

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Emergency access no longer allows anyone with a remote to bolt the door and checks if the remote has access to bolt.

### DIFF
--- a/Content.Shared/Remotes/EntitySystems/SharedDoorRemoteSystem.cs
+++ b/Content.Shared/Remotes/EntitySystems/SharedDoorRemoteSystem.cs
@@ -1,4 +1,5 @@
 using Content.Shared.Access.Components;
+using Content.Shared.Access.Systems; // DeltaV
 using Content.Shared.Administration.Logs;
 using Content.Shared.Database;
 using Content.Shared.Doors.Components;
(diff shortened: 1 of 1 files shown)


Content.Shared/Remotes/EntitySystems/SharedDoorRemoteSystem.cs
```diff
@@ -87,6 +89,19 @@ private void OnBeforeInteract(Entity<DoorRemoteComponent> entity, ref BeforeRang
             return;
         }
 
+        // Begin DeltaV - Emergency access only bypasses open/close; bolting and toggling emergency access still require actual access.
+        if (entity.Comp.Mode != OperatingMode.OpenClose
```
ShepardToTheStars, line 93: Should this be `entity.Comp.Mode == OperatingMode.ToggleEmergencyAccess`?

ManuDemoen: No, any mode other than opening/closing an emergency-access door, still needs to check the correct permissions

ShepardToTheStars: Oh, `entity.Comp.Mode` is the mode of the remote, not the current mode of the door being changed then?

ManuDemoen: yep, if you look further below you can see the other operating modes, args.Target, is the door

ShepardToTheStars (approved):

PR #6142, opened 2026-06-29
Give Cyborgs a remote for their respective department

ManuDemoen:
## About the PR

Borgs now spawn with a default module containing a remote for their department

## Why / Balance

Gives borgs the option to BOLT/Emergency Access/Open doors from a distance!

## Technical details

New default borg module that is present in each of the station's standard cyborgs

## Media

https://github.com/user-attachments/assets/bbf9c99c-596f-47cf-ae95-7d27565b8d41

## Requirements

## Licensing

**Changelog**

:cl:
- add: Gave all cyborgs a remote door control


MantasDab360:
If the third one is command only door, i don't think borgs SHOULD have access to remote it.

ManuDemoen:
!discord was HOP door opened by service borg FYI, IIRC all remotes can open their respective CO's door
the idea came when that during play as an engineer borg, I couldn't simply unbolt a bolted open door during the random bolt-event that happened, without doing the whole hacking shebang

MantasDab360:
> !discord was HOP door opened by service borg FYI, IIRC all remotes can open their respective CO's door the idea came when that during play as an engineer borg, I couldn't simply unbolt a bolted open door during the random bolt-event that happened, without doing the whole hacking shebang

Thing is, Borgs don't get Heads access, so this just overwrites that into having their access.
Plus remote unlocks their locker too.

ManuDemoen:
Tested it again, you're right, the remote would give them command access where they wouldn't even without the remote (video), could fix this by making a remote variant without the related command access

But reading the content review, I don't think it'll pass so I'll hold it for now

https://streamable.com/xxn1km

ManuDemoen:
!discord
CO access was an oversight, this would be the new access list:

Engineering | Engineering, Atmospherics
Security | Security, Detective, Corpsman
Medical | Medical, Chemistry, Paramedic, Psychologist, Surgery
Service | Bar, Kitchen, Hydroponics, Service, Janitor, Theatre, Lawyer, Boxer, Clown, Mime, Musician, Reporter, Zookeeper
Janitor | Bar, Kitchen, Hydroponics, Service, Janitor, Theatre, Lawyer, Boxer, Clown, Mime, Musician, Reporter, Zookeeper
Salvage | Salvage, Cargo, Mail, Orders, Funding

Niomi0:
I gotta say -1 from me, this is in the realm of AI.

Niomi0:
Though if this was a point blank door remote, i would be more accepting of it, like 1 tile away. Just to unbolt doors or whatever faster but door remote is extremely strong.

Space-Oracle:
!discord
Alternatively, jotting this down from talking on the discord:
- Just use Custom Remote which checks an equipped ID for access? (Borgs have an ID Chip that works like Crew ID)
> Borgs have their own access according to chassis transformation + no heads/command by default, so their ID would have to be modified anyway which is doable without this remote module
> This would always have Epistemics/Robotics access added regardless of department but then they could just body bump those doors anyway and it would only be inconveniently bolted doors that are relevant to this point
- Consider putting the module for a remote behind an emagged Robotics Fab/Syndie Robotics Fab for balancing
> This could be a generic module so it fits any borg, and avoid being classified as contraband for being syndicate - it would still be a controlled item according to SPLaw so command can deny its use (a lot of Epi falls under this even considering Departmental Dispensation so it's not like something that breaks SPLaw just by existing shouldn't be in the game)
> It could also be added as a T2/3 Epistemics research unlock for borgs, something something NT licensing iunno man

Now if you have spooky door shenanigans, you could blame Greytide Virus, Tech Anomalies, Solar Flares, AI, anyone with a remote or borgs. But borgs won't have unchecked straight buffs granted to them.

PR #6166, opened 2026-06-30
Trait Extras

weskified:
## About the PR
This PR focuses on the traits UI and making it a bit more bearable to navigate through.

## Why / Balance
Currently, The problem I have encountered before I even get the chance to start making my PRs on adding more traits, I have noticed that there is nothing to sort them, and when things get really big, It ***WILL*** be a huge problem to manage all traits to belong into something, rather than a category where it *should* belong.

This makes it so that it is much easier for a Trait to be clean, and traits with it's own modifyable changes are able to be seen on the parent trait itself, and other things that a new trait would put into the table for character customization/flavoring.

## Technical details
- added a new `SubTraits` field on the `TraitPrototype.cs` file.
- heavily updated/restructured `TraitEntry.xaml` to support having sub-traits under the parent trait itself. And made sure that they can be added/removed in an event of hotloading yamls.
- updated `TraitEntry.xaml.cs` to support the new xaml structure and so that the subtraits show up under the traits. (bear with me its been days and i dont remember everything)
- updated `TraitCategory.xaml.cs` so subtraits dont show up as their own seperate trait in the list anymore, theyre put under the parent trait now. also fixed search so if you search for a subtrait it still shows the parent trait so you can actually find it
- added the styling for the new subtrait section in `TraitsSheetlet.cs` so it doesnt look out of place
- moved `SyntheticMechanical` and `SyntheticIonStormSusceptible` from `meta.yml` to `medical.yml` and actually made them as subtraits of the `Synthetic` trait (they have to be same category as the parent trait so it is a lot more easier to find them, since having them all around the place would be very messy.)

## Media
The state of the trait if it's not selected
image

The state of the trait if it's selected.
image

## Requirements

## Licensing

**Changelog**
:cl: wesk
- add: Traits can now contain a sub-category of related traits to the main trait it belongs to!
- tweak: Synthetic-related traits are now under the Synthetic trait itself, and is no longer under the Meta category.


Content.Client/_DV/Traits/UI/TraitsTab.xaml.cs
```diff
@@ -75,8 +79,50 @@ private void OnProtoReload(PrototypesReloadedEventArgs args)
...
+                        continue;
+
+                    if (mapping.TryGetNode("id", out YamlScalarNode? idNode) && idNode.Value is { } id)
+                        ids.Add(id);
+                }
+            }
+        }
+
+        return ids;
     }
```
ShepardToTheStars, line 126: Errr... No. Don't do this. I'll follow up with a recommendation in a bit.

EDIT: By this, I mean, the whole `GetTraitIdsFromYaml()` function.

ShepardToTheStars (changes requested):

PR #6170, opened 2026-06-30
Crew Monitor Fix

ShepardToTheStars:
## About the PR

Fixed crew monitor and gave it buttons instead of trying to figure out which scanner to use on the fly.

## Why / Balance

Currently bugged.

## Technical details

Turns out, the client xform system may not know the parent grid ID of someone because they are too far away and the server doesn't need to update that data for a specific client.

Two approaches. I could send the parent grid ID with the suit sensors or I could just make toggleable buttons. I went with the easier approach of buttons.

## Media

https://github.com/user-attachments/assets/ed78625d-86b3-4616-83d1-c6669215c687

## Requirements

## Licensing

**Changelog**

:cl:
- fix: The crew monitor will now show the station map again and it has buttons to toggle between the two modes.


BarryNorfolk (approved):
Looks good to me, seemed to work in game as intended 👍

PR #6196, opened 2026-07-03
Fix: Correct Medicine Effect due to DamageGroupPrototype Being Obsolete From Upstream Merge

Stxcking:
## About the PR

`DamageGroupPrototype` is now obsolete which means you can't use damage groupings (brute, burn, toxin, airloss) to define change in damage. This cause Funkychems to get nerf due to the switch not being done properly. It done made salicylic acid a blunt medicine instead of a brute medicine. Crimes against chemists.
## Why / Balance

Fix, refer to Commit `cc5300d` that made the changes.
## Technical details

Yamslop
## Requirements

## Licensing

**Changelog**

🆑 
- fix: FunkyChems now properly heal all brutes types if they should instead of just healing blunt damage.
- fix: FunkyChems now properly heal all air loss types if they should instead of just healing asphyxiation damage.
- fix: Skia now properly heals all brute types instead of only healing blunt damage while in the dark.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Mobs/NPCs/skia.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/NPCs/skia.yml
@@ -85,14 +85,19 @@
     darkDamage:
       types:
         Blunt: -5
+        Piercing: -5
+        Slash: -5
         Heat: -5
         Poison: -5
-        Asphyxiation: -5
+        Asphyxiation: -2.5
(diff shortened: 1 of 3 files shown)


ShepardToTheStars (commented):
Looks fine. This would also confirm why Skia healing is broken and only heals blunt. We should look at other sources of non-even healing and see if they need adjustments too

ShepardToTheStars (approved):

PR #6200, opened 2026-07-03
Securityborg Health Improvement Tweak

AffleWaffle:
## About the PR

Security cyborgs now have the armor vest values but take 20% more shock damage.

## Why / Balance

Security cyborg is a very laughable joke if it ever has to confront someone who is not mildly armed like with a basic melee weapon. It can't fight carps, can't fight the rat king servants very well either, it has the feelings of a mall cop only ever being able to deal with minor crimes from non antagonists while folding against almost any antagonist other than thief. This is the first of the few tweaks I wish to add, this one introduces the value of the basic armor vest for the security cyborg, reducing their blunt, slash and piercing by 30% and heat by 20%. They also take 120% of shock damage (maybe one day it will be exist outside of very few niche objects) which is more of a lore/RP reason than any balancing.
The damage reduction will make it sustain a bit better than against small firearms and basic wildlife/mobs than it's other cyborg counterparts.

## Media

(A cyborg can dream)
Screenshot_101

## Requirements

## Licensing

**Changelog**

- add: Security cyborgs now have physical and heat resistance of an armor vest!

### DIFF
--- a/Resources/Prototypes/_DV/Damage/modifier_sets.yml
+++ b/Resources/Prototypes/_DV/Damage/modifier_sets.yml
@@ -190,3 +190,12 @@
     Radiation: 0.5
     Caustic: 0.5
     Cold: 0.8 # DeltaV - Cold res
+
+- type: damageModifierSet
+  id: SecurityCyborg
+  coefficients:
+    Blunt: 0.7
+    Slash: 0.7
+    Piercing: 0.7
+    Heat: 0.8
+    Shock: 1.2
--- a/Resources/Prototypes/_DV/borg_types.yml
+++ b/Resources/Prototypes/_DV/borg_types.yml
@@ -40,6 +40,9 @@
     fiberColor: fibers-blue
   - type: AccessReader
     access: [["Command"], ["Robotics"], ["Security"]]
+  - type: Damageable
+    damageContainer: Silicon
+    damageModifierSet: SecurityCyborg
 
   radioChannels:
   - Science


AffleWaffle:
!discord Understood, it's a shame since the secborg just lacks any ability to deal with anything other than tiders, a viper mag still gets a secborg. Can't fight carps, Skia, Ninja, rat servants etc.

AffleWaffle:
!discord I've fought rat kings servants as a secborg and as you are lacking any armor in comparison to the security officer you go down very quickly, mind you the escalation module is not that exactly super common to see, I way more often do not see it being researched than being researched. Fighting two to three carps as a secborg also is not easy even with the escalation module where as security officers roundstart can deal with this issue with a mk58 not to mention actual longarms or advanced ones. 
In a colossus event you are delegated to dragging bodies like any other cyborg can, you don't feel like a member of security even though you are limiting yourself in many ways as a cyborg to perform that role yet you don't get many of the benefits like other cyborgs have over their respective roles engiborg vs engineer or medborg vs doctor.

AffleWaffle:
!discord If this is not the direction the team wants to go forward I'll close the ticket, I've seen the module document but there hans't been any progess on it in months. I was under the impression that just like the engi/medi borg security borg is also meant to have places in which it excells throughout the round and not only to deal with tiders. During nukie or zombie rounds the secborg is just there (99% of the times without the escalation module so no truncheon/knife/riot shotgun) and just can't do anything against it same as with the Colossus while also being unable to approach EMP antags like the ninja or skia.
If the goal is to only deal with tiders as you proclaimed then so be it.

PR #6227, opened 2026-07-06
maybe we buff revolvers idunno

keekee38:
## About the PR
upped the base .45 revolvers' (pretty much the inspector) rate of fire from 1.5 to 2
upped the deckard's and python's rate of fire from 2 to 2.5
upped most of the .38 revolvers' rate of fire from 2 to 2.5

## Why / Balance
revolvers have always had lowered DPS than every other sec pistol (pollock and mk 58 are at 80, the slp is at 96, inspector 52.5, deckard 70, and all the 38 revolvers are at 60)
they also hold less rounds and take longer to reload due to how they work
most revolvers are still less dps than handguns after this too
(inspector would go up to 70, .38s would go up to 75, and the deckard is an exception at 87.5 - but it also holds 5 rounds instead of 6)
aurafarmin with revolvers is cool
also havent seen anyone have issues with the universals rof but maybe thats cause its so rare idunno
lmk if i should change numbers etc etc
woof

## Technical details
yml
touched baseweaponrevolver

## Media

shrugs

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: revolvers shoot a bit faster

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Revolvers/revolvers.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Revolvers/revolvers.yml
@@ -26,7 +26,7 @@
   - type: AmmoCounter
   - type: Gun
     selectedMode: SemiAuto
-    fireRate: 1.5
+    fireRate: 2 # Delta V - was 1.5
     availableModes:
     - SemiAuto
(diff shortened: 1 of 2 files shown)


pixelkingliam:
another possible angle for buffing revolvers would be buffing the ammo

PR #6228, opened 2026-07-06
Fix Convert to Survival Behavior during Zombies

ShepardToTheStars:
## About the PR

The round-end detection now checks if all II have turned as well as healthy crew %. And no pending zombies.

## Why / Balance

Bug fix for auto-survival scheduler. Current behavior is that it adds a ramping rule immediately since II are not considered infected. Introduced in #6127 

## Technical details

GetInfectedFraction does not check if all the IIs have turned once it fires so it usually just adds the survival scheduler right away. Extra check fixes it.

## Media

https://github.com/user-attachments/assets/931f8ebd-4135-47cd-895c-906a5db0e0bb

## Requirements

## Licensing

**Changelog**

no CL

### DIFF
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -18,6 +18,7 @@ using Content.Shared.Zombies;
 using Robust.Shared.Player;
 using Robust.Shared.Timing;
 using System.Globalization;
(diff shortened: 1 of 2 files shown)


ShepardToTheStars:
Needs a bit more testing on my end.

BarryNorfolk (commented):
Should be better, yeah.

BarryNorfolk (approved):

PR #6230, opened 2026-07-06
Health Analyzer Plus

CMuncey:
## About the PR
Upgraded the health analyzer to show a list of the medical reagents present in the patients bloodstream. Only medicines are shown, as to not trivialize poisonings, not that anyone actually does that. The UI has been updated to have two sections at the bottom, one (unaltered) section for the damage types on the left, and a new section on the right listing the reagents. This upgrade changes the default health analyzer gui, meaning it will affect the PDA health analyzer, the standalone health analyzer, and the cryo pods gui. Since the cryo pods inject medicine at the same rate that the body metabolizes them, they do not show up unless you have several reagents in the beaker at once.

Previous implementations had this new gui as a separate upgrade to the health analyzer, but this was reverted to only affect the default one. The attached video is slightly outdated, the MedTek+ cartridge no longer exists, and the gui it creates is the default now.

## Why / Balance
Makes med a bit easier to tell what's going on, should make learning it a lot easier

## Technical details
- Added a readonly field to the HealthAnalyzerUiState to pass the bloodstream of the patient to the health analyzer gui
- Added more groups to the health analyzer gui xaml, and updated cs file to fill them appropriately
- Only reagents with ReagentPrototype.Group == "Medicine" are shown in the gui

## Media
https://github.com/user-attachments/assets/ac7e29d1-8b00-45de-ae56-38ae6bf4cf06

## Requirements

## Licensing

**Changelog**

:cl:
- change: The Health Analyzer UI has had a facelift! It now shows medicines in the bloodstream!


ShepardToTheStars (commented):
Not a huge fan of copy/pasting the entire HealthAnalyzer client-side stuff. I would request that you add it into the normal HealthAnaylzer and the UI elements into the existing UI controls.

Since `HealthAnalyzerComponent` is partial, you can just make your own partial class (or add it to the upstream one too, since its just one property), and it could look like this.
```c#
namespace Content.Server.Medical.Components;

public sealed partial class HealthAnalyzerComponent : Component
{
    /// <summary>
    /// If true, also displays the medicines in the patient's bloodstream.
    /// </summary>
    [DataField]
    public bool AdvancedMode = false;
}
```

Then, depending on if that boolean is true or false, you could show or hide the advanced section in `HealthAnalyzerControl.xaml.cs` (setting visible to whatever this variable is).

CMuncey:
Oh yeah I forgot to add those things to the localization files, oops. Well it's sounding like people don't want it tied to research anyway so that might go away regardless.

ShepardToTheStars (approved):

PR #6241, opened 2026-07-07
MedHUDs no longer show your own health bar :)

ShepardToTheStars:
## About the PR

I removed seeing your *own* health bar from the medical hud (and other medhud like items). :godo:

## Why / Balance

I expect this to be controversial lol, but here is my reasoning:

* Seeing *exactly* how healthy you are at any given time is *very* powerful, especially in combat. But that information is already available on the right side of your screen. This makes it less central but still available.
* As a medical person, people will have to tell you that you are hurt instead of you knowing the moment you are slightly poisoned/irradiated/etc.
  * You can still scan yourself with your PDA and keep that UI up if you want. But its not an always on thing.
* It does kind of make no sense that you can see your own health bar if you think about it from your character's first-person perspective. You're using your eye piece to scan others that you can see, but you can't look at yourself without a mirror, so how does that work? Magic I guess.
* Medical people with the numb trait cannot cheat anymore (that's my character!)
  * I want to fall over from being dumb and not realizing I'm about to die lol, instead of it being obvious.

Kind of an indirect cult and zombie buff? Nukie agent nerf I guess. Could be modified to be enabled for that medhud specifically if needed.

## Technical details

Nothing crazy.

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: You can no longer see your own health bar with the medical hud (or similar eyewear).

### DIFF
--- a/Content.Client/Overlays/EntityHealthBarOverlay.cs
+++ b/Content.Client/Overlays/EntityHealthBarOverlay.cs
@@ -11,6 +11,7 @@ using Content.Shared.StatusIcon;
 using Content.Shared.StatusIcon.Components;
 using Robust.Client.GameObjects;
 using Robust.Client.Graphics;
+using Robust.Client.Player; // DeltaV
 using Robust.Shared.Enums;
 using Robust.Shared.Prototypes;
 using static Robust.Shared.Maths.Color;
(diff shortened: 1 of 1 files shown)


Vova567810:
Jarvis, make advanced medhud glasses that give flash protection and shows your own health :godo:

BarryNorfolk (approved):
Yarp, makes it so I can't see mine own health bar.

PR #6253, opened 2026-07-09
Glimmer Leaderboards

ShepardToTheStars:
## About the PR

What if more glimmer.

No I didn't test this. But I think it would be funny if we merged it anyways.

No this is not a joke. But we all know we want this.

## Why / Balance

Epi regularly hits 1000 glimmer and I want to see how high they can go. :)

## Technical details

1->2

## Media

image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Glimmer now goes to ~~11~~ 2000!

### DIFF
--- a/Content.Shared/Nyanotrasen/Psionics/Glimmer/GlimmerSystem.cs
+++ b/Content.Shared/Nyanotrasen/Psionics/Glimmer/GlimmerSystem.cs
@@ -17,7 +17,7 @@ namespace Content.Shared.Psionics.Glimmer
             get { return _glimmer; }
             set
             {
-                var newGlimmer = _enabled ? Math.Clamp(value, 0, 1000) : 0;
+                var newGlimmer = _enabled ? Math.Clamp(value, 0, 2000) : 0;
                 if (_glimmer == newGlimmer)
                     return;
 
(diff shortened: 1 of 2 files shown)


ShepardToTheStars:
I hate the magic numbers but glimmer needs a refactor.

BarryNorfolk (changes requested):
From a maints pov I think this is fine.
The glimmer from artifacts only works if it's in their range, and caps out at 1000 so it shouldn't cause super run away glimmer.

Nyano chat will now have the possibility to always send completely obscured messages to dreamers. Previously it was a maximum of 75% of the message being obscured (.25 + (glimmer / 2000)).

The shock from the "AddShockVerb" can now do a whopping 20 damage and stun you for whole 20 seconds.

Prober exposions go from max 1000 -> 2000 intensity with the slop going from `1` to `-11`. Probably want to check out `GlimmerReactiveSystem.cs` and make sure that doesn't case any weirdness with a negative slope. The beams it makes go from 3 to 6 targets and the range went from 20 to 40 tiles. Also I get a crash here when I try and blow up the prober with 2000 glimmer because the `OnDamageChanged` function just does a solid `glimmer / 1000` for the chance of beaming due to damage. 

The amount of liquid you get from the Oracle also goes from a maximum of 150 to 250.

PR #6258, opened 2026-07-09
Pka upstream merge

ArionGekkota:
## About the PR

Changed the PKA to match the existing PKA in upstream. Reintroduced the PKA upgrade modkits as a researchable upgrade (Available in Mass Excavation, T2 industrial). Reduced the health of some salvage mobs to match their health values in upstream

## Why / Balance

Our current PKA is a very strong weapon available roundstart, and the range trivializes many encounters. With this change, salvage still has a usable weapon but threats such as goliaths pose a challenge early into the round, and they're also able to upgrade their PKA after some research, similarly to upgrading your crusher to a glaive as a melee salvager.

## Technical details

Reimplemented the PKA modkits from upstream, removing the upgrade cost system that had been leftover. Reduced the PKA's base firerate and projectile lifetime. Reduced damage until death for some salvage enemies to compensate.

## Media

Research
upgrades
lathe
lathe2

## Requirements

## Licensing

**Changelog**

:cl:
- add: reintroduced researchable PKA modkits
- tweak: Changed the PKA to its current upstream state
- tweak: Reduced the health of some salvage mobs

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/asteroid.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/asteroid.yml
@@ -55,7 +55,7 @@
   - type: MobThresholds
     thresholds:
       0: Alive
-      250: Dead
+      150: Dead
   - type: MeleeWeapon
     soundHit:
       path: "/Audio/Weapons/smash.ogg"
@@ -386,7 +386,7 @@
   - type: MobThresholds
     thresholds:
       0: Alive
-      75: Dead
+      45: Dead
   - type: MeleeWeapon
     angle: 0
     animation: WeaponArcBite
--- a/Resources/Prototypes/Entities/Mobs/NPCs/carp.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/carp.yml
@@ -43,7 +43,7 @@
     - type: MobThresholds
       thresholds:
         0: Alive
-        50: Dead # DeltaV, was 40
+        40: Dead
     - type: MovementSpeedModifier
       baseWalkSpeed: 2.5
       baseSprintSpeed: 3.5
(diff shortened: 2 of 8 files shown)


ShepardToTheStars:
You will probably also want to remove the upgrades from the mining point vendor as well if the intention is to make salvage wait for the research.

ArionGekkota:
!discord Currently, the PKA upgrades are not stackable and you can only use 2 of them at a time. They either increase damage from 15 to 25, firerate from 0.5 to 0.8, or effective range from 5.5 tiles to about 8 tiles. They are not purchaseable in the mining vendor, and have been moving from being originally a T2 arsenal technology to a T2 industrial, being unlocked alongside diamond tipped drills.

The only health cuts are lowering goliath and basilisk health by 40%, and lowering carp health by 20%. Damage numbers remain the same, and it is now impossible to engage goliaths from outside of their aggro range, a technique that trivialized encounters with them.

ArionGekkota:
!discord Franky, if I may ask, in what situation is a security officer going to be allowed to walk around with that much ammunition on their person? Most officers don't have access to the armory to begin with, and printing that much ammo will likely get you in trouble one way or another.

ArionGekkota:
!discord This is my first time trying to change the PKA, and I thought it'd make sense to try to bring in a version that exists in upstream along with the balance changes tied to it. In addition, I'd like to point out you can upgrade two stats, not just one.

ArionGekkota:
!discord I've seen crusher variants and PKA variants, and I'm aware of crushers having a trophy system tied to megafauna that matches your idea of upgrades, but in the ss13 servers I played (tg downstreams), crushers and PKAs were very separate and the upgrades were just flat stat increases

ShepardToTheStars (approved):

CookieBoy21:
Just in case, PKA is now unless in killing xenos. So now Salvage Expedition with xenos is too dangerous.

PR #6263, opened 2026-07-09
Airlock Access Override Removes Access, Give Engineers Access Config

ShepardToTheStars:
## About the PR

* Fixed access overrides from not removing accesses on the doors (they are supposed to on upstream).
* Added access configurators to engineer lockers, engineering techfab, and the EngiVend.
  * Engineers are now responsible for fixing access - not just the CE.
* The AI can now see if a door's access has been modified by inspecting it.

## Why / Balance

This is parity with upstream behavior. For some reason, we had code in AccessReader that prevented the Airlock Access Override from breaking access specifically.

AI seeing broken access just kind of makes sense.

Adding access configurators basically helps out when adding doors via RCD. You don't have to dismantle doors to reconfigure access. This adds another task for engineering (and not just the CE) to do.

Balance-wise, this should affect nothing other than something for engineers to help out with because you cannot modify accesses on things you don't have access to.
image

## Technical details

Removed DV specific-code. YAML.

## Media

(**Without** diagnostic HUD after emag)
image

(With diagnostic HUD after emag)
image

image

image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: The airlock access override now breaks accesses on the door. You can see if accesses on doors have been broken with the Diagnostic HUD.
- add: Added access configurators to engineer lockers, engineering techfab, and the EngiVend. Engineers are now allowed to repair access around the station!
- add: The AI and engineering borgs can now see if a door's accesses have been modified.


Niomi0:
Can borg get an access configurator too <3

ShepardToTheStars:
> Can borg get an access configurator too <3

I'm unsure how that might work with the current implementation, since you need to insert an ID into the access config, and borgs don't use IDs and cannot hold them.

There's definitely ways around it, but I'd rather not give borgs to add/remove most access on doors. Someone from the department should have to give them an ID.

ShepardToTheStars:
I am going to give engi borgs the ability to see modified access too I think.

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -151,7 +151,7 @@ private void OnLinkAttempt(EntityUid uid, AccessReaderComponent component, LinkA
 
     private void OnEmagged(EntityUid uid, AccessReaderComponent reader, ref GotEmaggedEvent args)
     {
-        if (!_emag.CompareFlag(args.Type, EmagType.Interaction)) // DeltaV - emag for lockers etc instead of doorjack
+        if (HasComp<SpaceNinjaComponent>(args.UserUid)) // DeltaV - Don't break access if its a ninja doing it
```
ShepardToTheStars: (I am aware that a ninja using a syndicate airlock access override does not have the intended effect of breaking access with this, but when has that ever happened?)

Resources/Prototypes/_Mono/Recipes/Lathes/tools.yml
```diff
@@ -19,4 +19,4 @@
   result: NaniteApplicatorSyndicate
   materials:
     Steel: 800
-    Plasma: 200
+    Plasma: 200
```
SirWarock: Untouch file

SirWarock (changes requested):

Resources/Prototypes/_Mono/Recipes/Lathes/tools.yml
ShepardToTheStars:
```suggestion
    Plasma: 200

```

SirWarock (approved):

PR #6271, opened 2026-07-10
Fix Tech Disk Test Failures

ShepardToTheStars:
## About the PR

We didn't have any T3 techs for Civ Services and its causing our tests to break.

## Why / Balance

I hate our tests failing.

## Technical details

The test that was failing was intermittently failing because there was a 20% chance that `TechnologyDiskT3` would choose the Civ Services discipline (1 out of the possibly 5), but there were no T3 civ recipes, so it would create an empty disk with no recipes.

This caused it to write to the error log and fail the test. I picked the recipe with a bluespace tech to make T3.

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Advanced cleaning is now a T3 civilian services research.

### DIFF
--- a/Resources/Prototypes/Research/civilianservices.yml
+++ b/Resources/Prototypes/Research/civilianservices.yml
@@ -161,7 +161,7 @@
     sprite: Objects/Specific/Janitorial/advmop.rsi
     state: advmop
   discipline: CivilianServices
-  tier: 2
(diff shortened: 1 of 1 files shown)


Stxcking (approved):
No better research to slap the T3 label on, looks good.

PR #6293, opened 2026-07-13
Allow Slime people to change hair and facial hair color

ollyoaner:
## About the PR
Removed the forced skin matching on Slime people's hair and facial hair to allow color picking for them.

## Why / Balance
Currently, a big problem is just how different the hair color can appear compared to the skin color. There are many hues/shades/values that make the hair look completely strange or overly see-through. In many cases the hair will look very different compared to the skin color. Because the hairs are also all individually made, it's much easier to allow color-picking than to resprite all the hairs to work nicely with Slime's sprite opacity.

Also more creativity is always good. The hair still has the same opacity/jellylike qualities. This will simply make it so you can make more fun combinations and are not forced to pick certain skin colors to get certain hair looks.

## Technical details

-  Disabled "matchSkin: true" in slime.yml for enum.HumanoidVisualLayers.Hair and enum.HumanoidVisualLayers.FacialHair

## Media
Comparison:
Screenshot 2026-07-13 100600

Screenshot 2026-07-13 100637

## Requirements

## Licensing

**Changelog**
:cl:
- tweak: Slime people can now change their hair color!

### DIFF
--- a/Resources/Prototypes/Body/Species/slime.yml
+++ b/Resources/Prototypes/Body/Species/slime.yml
@@ -41,10 +41,10 @@
   appearances:
     enum.HumanoidVisualLayers.Hair:
       layerAlpha: 0.72
-      matchSkin: true
+#      matchSkin: true                          #Delta V - Allow Slime people to change hair color
     enum.HumanoidVisualLayers.FacialHair:
       layerAlpha: 0.72
-      matchSkin: true
+#      matchSkin: true                          #Delta V - Allow Slime people to change facial hair color
 
 - type: entity
   parent: BaseSpeciesAppearance


Stop-Signs:
!review

pootslap:
I'm not sure this keeps the opacity. I messed around with this before and it kinda looks like they're wearing a wig.
image

I think this'll change how every slime looks even if they don't alter the colour because it'll make their hair opaque rather than translucent.

rebe83:
Pretty good example of a PR that doesn't need discussion there's nothing to balance it's just cosmetics
Argue for 3 days about hair color wow

Stxcking:
> Pretty good example of a PR that doesn't need discussion there's nothing to balance it's just cosmetics Argue for 3 days about hair color wow

Not only does the PR gets rid of the transparency effect of the hair which as a slimeperson player, I don't like, our species have unique features, buffs/debuffs, if they can look like exactly like humans but with slime person features that's not exactly good game design

rebe83:
Slime in hardsuit you're not going to tell the difference that is kind of just life
Nitrogen tank in the back I guess. They don't resist anything special they are just weaker to bullets

ollyoaner:
!discord
Well shit. That is completely my bad on the transparency bit. 

To clarify: I do not want to allow hair color changing and remove transparency completely. The intention was to simply allow an option, as some hairstyles already can be made nearly completely opaque with certain colors. And on top of that, some hairstyles are absurdly see-through compared to others.

I will disagree inherently with the idea of Slime people being able to change hair color completely(and even have hair that isn't transparent) would make them exactly like humans. Especially considering the fact that I have seen the idea of a wig system similar to SS13's being added, and in that case unless we restrict Slime people's access to this feature, it would have the same effect.

But this is about options. I have no intention of completely removing the transparency, as that would remove an option.

Currently I do not have the coding skills to correct this issue since the alpha appears to be linked to the matchskin code. 

As such, it would be best to close this and come back to it at a later date when I can properly implement the changes as intended.

PR #6303, opened 2026-07-14
Remove Kitsune's Duffle bag Hiding

TehFlaminTaco:
## About the PR

Remove Kitsune's Duffle bag Hiding

## Why / Balance

Why can Kitsune hide in dufflebags? They're the same size as most humanoids. (`0.85` min height vs `0.9`, this shouldn't justify bagging)

## Media

N/A

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Removed Kitsune's ability to hide in duffle bags.

### DIFF
--- a/Resources/Prototypes/_DV/Body/Species/kitsune.yml
+++ b/Resources/Prototypes/_DV/Body/Species/kitsune.yml
@@ -128,12 +128,6 @@
         Heat: 5
   - type: LightweightDrunk
     boozeStrengthMultiplier: 3
-  - type: PseudoItem
-    storedOffset: "0,17"
-    shape:
-    - 0,0,1,4
-    - 0,2,3,4
-    - 4,0,5,4
 
 - type: entity
   parent: OrganBase


mmmrooster:
This PR has no actual reasoning besides "I don't think they should"
I am not a maintainer and do not have any authority, so this is purely a request on my part, but
*please* do not send in PRs that take away neat features that people enjoy with no reasoning beyond "I don't feel like they should have it"

TehFlaminTaco:
> This PR has no actual reasoning besides "I don't think they should" I am not a maintainer and do not have any authority, so this is purely a request on my part, but _please_ do not send in PRs that take away neat features that people enjoy with no reasoning beyond "I don't feel like they should have it"

To elaborate, I don't think they should have it _because_:
Height-wise, they are on-par with standard humanoids, only able to be 0.05x smaller than a standard human, and up to 1.1x standard height.
There is complaints about saturation of the ability, and they standout as the tallest of bag-hiders. Currently, Felinids, Rodentia, Kitsune, and Allulao can baghide.

This PR was created with the intent for it to be discussed by the content team, not to knee-jerk merge.

mmmrooster:
Felinids use the exact same height scale, the species it was originally invented for, worth noting.
As far as oversaturation goes, that should not be targetted at one species in particular.

TehFlaminTaco:
> Felinids use the exact same height scale, the species it was originally invented for, worth noting. As far as oversaturation goes, that should not be targetted at one species in particular.

Felinids range from `0.76 - 0.88`, Kitsune range from `0.85 - 1.05`.

I was informed that Kitsune were intended to be smaller. I agree that's a valid alternative change, and one we're considering.

mmmrooster:
Ah, you're seemingly right on the height scale, I forgot the actual numbers are relative per species in the editor...
But yes, either way, I don't personally believe in or support feature removal without very exacting reasons. I agree that changing the height to better match would be fitting.

SirWarock:
> This PR has no actual reasoning besides "I don't think they should" I am not a maintainer and do not have any authority, so this is purely a request on my part, but _please_ do not send in PRs that take away neat features that people enjoy with no reasoning beyond "I don't feel like they should have it"

Oversaturation of a feature makes it feel less unique and less special. To remove it from things that probably shouldn't have them is a good choice, to make said feature feel unique again.
And Kitsunes don't lack special, unique features after all

mmmrooster:
I am a *firm* believer that the solution to one species having more features than another, is to give that other species more features.
Taking away from one species to bring up a different species just... isn't really a good solution in my opinion, because you're effectively making the choice between which players of which species should be made happy.

At the very least, if the intent was to make baghiding unique again, it should be removed from multiple species, instead of singled onto one. Especially since kitsune were intended to be smaller, giving them the same reasoning the other species who have baghiding have, once that's fixed.

PR #6314, opened 2026-07-16
Universal Bag-Hiding, Smaller Kitsune

TehFlaminTaco:
## About the PR

All characters of appropriate size can now fit in Duffel-Bags. Appropriate size is total scale <0.9.
Additionally, Kitsune have had their min-scale set to 0.75 to be more inline with Felinids and their initial design.

## Why / Balance

Makes the bag-hiding mechanic universal and not species specific, to avoid singling out any species.
This is in response to feedback from #6303

## Technical details

Height is calculated as a mix of their sprite height (species base * character height), and a static scale (Used for Allulalo).
The default shape matches the current Felinid shape, (two 2x2 boxes stacked atop a 6x3 box). This is than scaled and re-rasterized depending on the target height.

Larger species, humans and up, cannot become small enough to fit.

## Media

image
image
image
image
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Lowered Kitsune Min-Height to be more in-line with their design.
- tweak: Everyone small enough can now fit in bags, no-longer should Felinids and Rodentia get all the fun!

### DIFF
--- a/Content.Shared/Nyanotrasen/Item/PseudoItem/PseudoItemComponent.cs
+++ b/Content.Shared/Nyanotrasen/Item/PseudoItem/PseudoItemComponent.cs
@@ -15,13 +15,32 @@ public sealed partial class PseudoItemComponent : Component
 
     /// <summary>
     /// An optional override for the shape of the item within the grid storage.
-    /// If null, a default shape will be used based on <see cref="Size"/>.
+    /// If null, a default shape will be used based on what felinids used to have.
     /// </summary>
     [DataField, AutoNetworkedField]
-    public List<Box2i>? Shape;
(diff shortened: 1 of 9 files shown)


rebe83:
Let's add the abusable mechanic to every species so that everyone can abuse hiding from zombies or antags.
It was cute when only cats did it. This is a PR based on mald because their species feature was being removed, a feature they don't need and is featurecrept from felinids so now 3 species have it. The issue was kitsune are human sized, and now everyone human sized also gets to hide in bags. 
It's a species quirk not a universal mechanic

Apersos-coll:
All shits n giggles till the suspiciously oni shaped blue space bag in the corner starts giving suggestions

PR #6321, opened 2026-07-16
kill cult

Proxysseia:
## About the PR
stops cult rounds from rolling

## Why / Balance
not all that fun

:cl:
- remove: removed the Cosmic Cult from rotation.

### DIFF
--- a/Resources/Prototypes/_DV/secret_weights.yml
+++ b/Resources/Prototypes/_DV/secret_weights.yml
@@ -1,10 +1,10 @@
 - type: weightedRandom
   id: SecretDeltaV
   weights:
-    Survival: 0.28
-    Nukeops: 0.14
-    Zombie: 0.08
-    Traitor: 0.34
-    CosmicCult: 0.08 # my cult so cosmic!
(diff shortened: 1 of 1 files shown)


ALICEHUNTERS:
?

rebe83:
Finally a good well argued PR that is self standing and has reasonable points for why the content should be removed that everyone can see and isn't inflammatory

pootslap:
I was actually gonna make this PR at some point, I'm glad someone who's better at yelling at people did it first.

At the very least I hope it gets CR to look at the Mode that Nobody Likes and consider some of the options.

NoElkaTheGod:
Oh how the turns have tabled.
I'd make a joke, but I'm busy reworking coscult for another fork.

On a side note, I'm like 99% certain that you had a fake malign rift spawn event, probably want to disable that as well, no point having fake signs of cult if the real cult doesn't exist.

Radezolid (changes requested):
This would need to kill the colossus spawns and the random rifts events to fully kill it.

Proxysseia:
you know why

PR #6324, opened 2026-07-17
Security magboots have some damage slowdown resistence

ArionGekkota:
## About the PR

Security magboots now have 25% less slowdown from injuries. For comparison, combat boots have 50% less slowdown from injuries

## Why / Balance

Currently, security magboots are a reskin of crew magboots. This makes them more unique.

## Technical details

Security magboots now have the ClothingSlowOnDamageModifier component

Extra math for those who want to talk about balance:
- At 60-79 total damage, speed modifier goes from 0.8x to 0.85x (would be 0.9x with combat boots)
- At 80+ total damage, speed modifier goes from 0.6x to 0.7x (would be 0.8x with combat boots)

## Media

magboots

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Security magboots now have some damage slowdown resistance.

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Clothing/Shoes/magboots.yml
+++ b/Resources/Prototypes/_DV/Entities/Clothing/Shoes/magboots.yml
@@ -8,6 +8,8 @@
     sprite: _DV/Clothing/Shoes/Boots/magboots-security.rsi
   - type: Clothing
     sprite: _DV/Clothing/Shoes/Boots/magboots-security.rsi
+  - type: ClothingSlowOnDamageModifier
+    modifier: 0.75
 
 # ERT
 


rebe83:
If sec is wearing magboots all shift without these, they don't need buffed

ArionGekkota:
Is sec wearing magboots without these? I swear I never see them take the magboots from suit storage, even when going into EVA

Radezolid (approved):

ArionGekkota:
!discord The slowdown resistance only applies to injuries, and it only turns the 0.8x speed while at 60+ damage into 0.85x speed, and the 0.6x speed from 80+ damage into 0.7x speed. Considering the magboots will typically be worn with the security hardsuits, and that they'd still be worse jackboots while off, they already pack a sizeable slowdown when in use

ArionGekkota:
!discord magboots are a 20% slowdown when activated

ArionGekkota:
!discord wouldn't halving all of the magboots slowdown 1: minimize the value of CE's advanced magboots, and 2: count as the powercreep this PR is being accused of?

ArionGekkota:
!discord it's not really a sidegrade when they're the only option you typically get as security without needing to rely on epistemics having the research and materials to print you the alternative, imo

PR #6329, opened 2026-07-18
Toolshed Additions: Field & DoExt commands

TehFlaminTaco:
## About the PR

Adds two new Toolshed commands, `.` (field) and `doext`.
The `.` command returns the value of the provided field. Eg. `> self comp:get Transform . MapID` returns the user's current MapID.
The `doext` command has two parts, `doext:client` and `doext:server`. These function like `do`, however respectively run the provided command in the client's context, or the server context.
`doext:client` Will run the provided command as if run from your client, allowing for client-specific commands like `exec`.
`doext:server` Will instead remain on the server context directly.
By design, neither command can run commands on another client, and they cannot violent permissions.

## Why / Balance

More tools is more better.
Tools for admining and admin scripts.

## Technical details

Above

## Media

N/A

## Requirements

## Licensing

**Changelog**


sowelipililimute (changes requested):
the marking command should be part specific, probably in tandem with part retrieval commands as well frankly i should just add a bunch of nubody commands upstream

TehFlaminTaco:
> the marking command should be part specific, probably in tandem with part retrieval commands as well frankly i should just add a bunch of nubody commands upstream

I'll remove the markings commands for now. I can live without them.

sowelipililimute (approved):
this is scary but the code checks out

PR #6359, opened 2026-07-21
Make HOP a second-in-command

Radezolid:
## About the PR
This PR enables the HoP as it was before being just the Head of Service.
This PR intends to make the HoP:
- Make a second-in-command that isn't HoS as it usually ends up being unofficially.
- Give more content to HoP since they usually end up sitting in their office or walking around without anything to do.
- Fallback of basic access for Heads of Staff if they wish to hire someone.
- A better choice for ACO when a captain is no more instead of falling it, again, on the HoS which already have a lot going on the round at that point.
- Whitelist them so they are less of a possible nuisance to administration.

Changes done to facilitate this:
- Given the general access back, plus some that were missing.
- Restored their disabler to the locker.
- Restored the weight of the job since it might be more difficult to kill them.

I do fear that this might end up in HoPcurity like before or them giving accesses without consulting other heads of staff, maybe a change in SOP would fix this, but it is something I think might happen.

## Why / Balance
Mostly because of a [thread on discord](https://discord.com/channels/968983104247185448/1529150931432902746) and to make HoP have some responsibilities.

## Technical details
N/A

## Media
Access:
image
Whitelist:
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: The HOP has most of the access on station back again in order to make them similar to upstream.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
@@ -132,11 +132,11 @@
       # Yes they do fuck you.
     - id: ClothingBackpackIan
       prob: 0.5
-    #- id: ClothingHeadsetCommand # DeltaV - HOP is head of service.
+    - id: ClothingHeadsetCommand
     - id: ClothingNeckGoldmedal
     - id: DoorRemoteService
     - id: HoPIDCard
-    #- id: WeaponDisabler # DeltaV - There is no reason for the service head to have a disabler.
+    - id: WeaponDisabler
     - id: ClothingEyesHudCommand
 
 - type: entity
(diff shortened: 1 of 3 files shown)


rebe83:
The most hypothetical PR to date

Radezolid:
Ready for review now.

keekee38:
heya, gonna say 7 days for CR, but it could be done with earlier
thank ya for it!

keekee38:
kay well i forgot the bot was kill when i did this :) anywho umm, basically CR is gonna say no on this one for a few reasons: 1. HoPcurity is a large concern 2. HoP, at least here is sorta seen as a "learner role" for command in our culture for the most part, just due to how it currently is. 3. Feelings about a proper 2IC are mixed at the moment, especially since the chain of command only really exists when its on code red. 4. umm hopcurity again... but we do plan on talkin about HoP more especially with the whole situation going on, we should hopefully find some way to make the role more engaging. thank ya for it again and sorry

Radezolid:
Sure, i don't mind.

PR #6362, opened 2026-07-22
changing firelocks to external access

keekee38:
## About the PR
firelocks are now able to be opened by those with external access than solely engineering access

## Why / Balance
have you ever needed to get out of a room with a firelock in it and then it shuts on you a hundred times and has a doafter to even open...
its
certainly somethin

## Technical details
its like 1 line of yaml (the extent of my capabilities)
god i wish i was drunk right now
it seemed like it worked from testing but i may be stupid idunno 

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Firelocks can be more easily opened by people with external access.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
@@ -138,7 +138,7 @@
       price: 150
     - type: DoorBolt # DeltaV - Revert fire-fighting door remote removal.
     - type: AccessReader
-      access: [ [ "External" ] ] # Delta V - was Engineering
+      access: [ [ "External" ] ] # DeltaV - was Engineering
       examinationText: access-reader-examination-functionality-restricted
     - type: PryUnpowered
       pryModifier: 0.5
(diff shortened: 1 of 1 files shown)


MantasDab360:
yaml wise, should be fine since all of engi has external anyways.

Stxcking:
This was pulled into CR, marked for a 1-2 day public discussion but I don't see it requiring 2 days. Thank you!

Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
```diff
@@ -138,7 +138,7 @@
       price: 150
     - type: DoorBolt # DeltaV - Revert fire-fighting door remote removal.
     - type: AccessReader
-      access: [ [ "Engineering" ] ]
+      access: [ [ "External" ] ] # Delta V - was Engineering
```
Stxcking:
```suggestion
      access: [ [ "External" ] ] # DeltaV - was Engineering
```

Stxcking (changes requested):
Always try to stay consistent with `DeltaV` comments on upstream files. Otherwise, 1 line yaml looks good. I am also able to make this change when the CR discussion conclude with approval if you haven't done the change already.

Stxcking:
Unfortunately, this is denied due to the fact that CR believes that having to deal with firelocks is something that shouldn't be avoidable via this implementation. We understand the frustration with them but believe it is a part of the game in a sense. As well as this might stem from mapping practices or other airlock components. Thank you regardless!

PR #6371, opened 2026-07-22
Shrink entropic blades

EmberAstra:
## About the PR
Changed the cult's entropic blade to 2x4, down from 4x4.

## Why / Balance
To put it simply: cult rarely bothers to prepare for going loud. Everyone tries to rush all the rituals at once when T3 hits, which causes a mess both from the damage those deal AND the general clusterfuck of 10 people rushing in the place at once with half of them not having anything to convert in the first place.
Part of this, I think, is the fact that while you can start preparing weapons as early as T2, giving you both an out in case of emergencies AND saving time and health later, they are _**SO DAMN BIG**_ that nobody wants to sacrifice half of your entire bag space just for one.

So: I'm making them smaller. Still big, but a size that feels more reasonable to sneak around with.

## Technical details
2 lines of YAML to change their shape. Still Huge.

## Media
sord

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Cosmic Cult's Entropic Blades are now smaller (2x4 spaces)

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/Objects/cosmicweapons.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/Objects/cosmicweapons.yml
@@ -65,6 +65,8 @@
   - type: DisarmMalus
   - type: Item
     size: Huge
+    shape:
+    - 0,0,1,3
     sprite: _DV/CosmicCult/Objects/cosmicsword-inhands.rsi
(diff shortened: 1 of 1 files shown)


Stop-Signs:
CD speed approve stamp of approval

Radezolid (approved):

PR #6375, opened 2026-07-23
Buckle Bugfix + Dynamic CPR Prio

ShepardToTheStars:
## About the PR

* Fixed a long-standing bug with buckling while carrying. It shouldn't attempt to re-parent (since buckling already did that).
* Changed CPR prio to be less priority while the entity is not buckled.

## Why / Balance

Carrying and CPR verbs had conflicting priorities. I just tweaked it so picking up someone who's crit takes prio UNLESS they are strapped to a bed or rollerbed.

Also bug fix good.

## Technical details

It was just the CPR change, but then I got debug assertion issues when I was recording so I fixed that too.

The issue was that buckling happened, updated the parent of the entity being buckled and then the `BuckleEvent` was being relayed. The Carrying system would re-parent to the grid itself instead of the entity that the carried person was buckled to. So I just added an optional param to not re-parent in that case.

## Media

https://github.com/user-attachments/assets/3ffdbd23-0eae-4418-8abf-432dc16401d1

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Buckling someone while they are being carried should work as intended now.
- tweak: Picking someone up that is criticla will have priority over CPR unless they are buckled to something.

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
@@ -392,7 +392,6 @@ public abstract partial class SharedBuckleSystem
         if (TryComp<PhysicsComponent>(buckle, out var physics))
             _physics.ResetDynamics(buckle, physics);
 
-        // TOOD: DV - This fails when you try to buckle the entity you're carrying to something. Figure out why later.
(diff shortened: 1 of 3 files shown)


SirWarock (approved):
Works

PR #6376, opened 2026-07-23
Reduce Sink Claws Delay + Sink Claws Alert

ShepardToTheStars:
## About the PR

* Changed the felinid sink claws button delay from 3s to 0.5s.
* Added feline specific alert icon for felinids when sinking their claws in.

## Why / Balance

Felinids don't need another button that they regret accidentally pressing and just have to deal with basically being stuck in one place for 3 seconds.

This just makes it so that you can toggle it off if you press it by mistake faster. Also added the toggled on icon as the alert instead of the default diona one.

## Technical details

Y A M L

## Media

image

https://github.com/user-attachments/assets/b7ebf6ee-c30f-44d0-a759-e20e99643a46

## Requirements

## Licensing

**Changelog**

:cl: ShepardToTheStars, rebe83
- tweak: Felinid's sink claws ability delay has been reduced from 3s to 0.5s.
- add: Felinids now get a unique icon on their alert bar when they protract their claws.


rebe83:
Yippee! I added the 3 seconds just to prevent noise spam as I'm not sure if the growl is clientside
Also prevents combat chuds from freaking out since it would be a 3 second immobilization to ignore space wind

SirWarock (approved):
works

PR #6377, opened 2026-07-23
Add dedicated objective editor view for admins

BarryNorfolk:
## About the PR
Added an admin only Objective editor for curators/admins to use when creating new objectives for users, or simply updating them.

N.b. Updating the checks (the thing that greentexts or not) is well beyond me at this point as they are only ever on the server and generally don't get passed/shown to the user. I'd have to effectively replicate some aspect of the `vv` menu in order to show something mildly useful. Right now this can just serve to help people do their objective work a little easier.

## Why / Balance
Because the alternative is 4 `vv` calls or using `vvwrite` which can crash the server if done wrong. Easier to just let users have good tools to do what they need to.

## Technical details

TODO

## Media

The base UI
image

Adding a default new Objective
image

A dedicated Sprite selector for icons
image

Custom objective reflected in character info
image

`(*)` in the title shows that there are unsaved edits
image

An optional popup for when the objectives are saved for the user
image

## Requirements

## Licensing

**Changelog**
Nothing user facing


Content.Client/UserInterface/Controls/DialogWindow.xaml.cs
```diff
@@ -110,7 +110,7 @@ public DialogWindow(string title, List<QuickDialogEntry> entries, bool ok = true
     protected override void Opened()
     {
         base.Opened();
-        
+
```
BarryNorfolk, line 113: Untouch

Content.Server/Objectives/ObjectivesSystem.cs
```diff
@@ -215,7 +215,7 @@ private void AddSummary(StringBuilder result, string agent, List<(EntityUid, str
                 }
             }
 
-            var successRate = totalObjectives > 0 ? (float) completedObjectives / totalObjectives : 0f;
+            var successRate = totalObjectives > 0 ? (float)completedObjectives / totalObjectives : 0f;
```
BarryNorfolk, line 218: Untouch

PR #6381, opened 2026-07-24
Fix Shadekin Inheritence

ShepardToTheStars:
## About the PR

Fixing inheritance for long-term maintainability

## Why / Balance

Maintainer things.

## Technical details

Removes comps related to breathing on component init instead of copy/pasting code like what happened with IPC maintainability.

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Shadekin cannot perform CPR anymore due to them not breathing or having lungs.
- fix: Shadekin should have an internal temperature now, so they can get cold and hot.

### DIFF
--- a/Content.Server/_Starlight/ShadekinSystem.cs
+++ b/Content.Server/_Starlight/ShadekinSystem.cs
@@ -19,6 +19,7 @@ using Content.Shared._Goobstation.Flashbang;
 using Content.Shared._Starlight.Flash.Components;
 using Content.Shared.Body.Components;
 using Content.Shared.Flash; // Delta V - Flash Work
+using Content.Shared._DV.Body.Components; // Delta V - Remove various comps related to breathing
 
 
 namespace Content.Server._Starlight;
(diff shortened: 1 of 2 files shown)


Content.Server/_Starlight/ShadekinSystem.cs
```diff
@@ -68,6 +69,9 @@ private void OnInit(EntityUid uid, ShadekinComponent component, ComponentStartup
     {
         UpdateAlert(uid, component, (short)component.CurrentState);
         RemComp<InternalsComponent>(uid);
+        RemComp<RespiratorComponent>(uid);
+        RemComp<AffectedByCPRComponent>(uid); // No lungs = no CPR
+        RemComp<CanDoCPRComponent>(uid); // no lungs, no CPR
```
SirWarock: We are not doing mouth to mouth with the current CPR, so they should be able to just press their hands on someones torso.

```suggestion
```

SirWarock (changes requested):
I'd remove the inability to _do_ CPR, but I agree about them not being able to _receive_ CPR.
However, the solution to the problem at hand irks me.
There might be a YAMLSlop way to fix the parenting, but I can't think of one that would fix the issue at hand without losing updates from upstream.
Just the one CPR thing, then I'll approve.

SirWarock (approved):

PR #5712, opened 2026-04-22
PORT: Selectable Cyborg Chassis from CosmicDrift

Stxcking:
## About the PR

Titles sums it up! This PR allows you to pick between different cyborg chassis sprites which are purely cosmetics when you pick your chassis type. 
PRs ported:
[pickable borg subtypes but better #706](<https://github.com/cosmatic-drift-14/cosmatic-drift/pull/706>) (Functionality) by lunarcomets
[borg subtype fixes I Forgot To Push #790](<https://github.com/cosmatic-drift-14/cosmatic-drift/pull/790>) (Follow-up Fix) by lunarcomets
[update borg subtypes to an entity prototype based approach from SL #884](<https://github.com/cosmatic-drift-14/cosmatic-drift/pull/884>) (Entity Prototype Refactor) by lunarcomets
[borgs paints v2(Based on AL subtypes) #1807](<https://github.com/ss14Starlight/space-station-14/pull/1807>) (Just the Sprites which were also taken from Paradise.)
[Borgs update #1560](<https://github.com/ss14Starlight/space-station-14/pull/1560>) (Just the Sprites which were also taken from Paradise.)

Edit: Some of the chassis were resprited as well per CR's request!
## Why / Balance

~~Fun and whimsy or something?~~ ~~Character customization major?~~ Cyborg players think it would be cool if we have selectable chassis and I think so too! It gives more personality to cyborgs who are not NPCs, shocker, and set up the system for more chassis designs (without having to replace any which I think is a win-win).
## Technical details

None of the PRs actually explained how the system works so bear with me (and one of the PRs I ported hasn't been merged). 
`BorgSubtypeDefinitionComponent` Adds/overrides subtype-specific data (mainly cosmetic) that links back to a base `borgType`. Most of the data is automatically synchronized. This makes it easy to add chassis from different forks (hopefully making it easy to maintain too). 

I choose how CD implemented selectable chassis over Starlight/Afterlight and Goob especially with them switching over to an entity prototype approach because the data is rendered a LOT nicer from what I saw.

There's also the option to make the subtypes unique with a different set of modules and components, it is currently just cosmetic right now.
<details>
<summary> How to add borg chassis!</summary>

In `borg_subtypes_chassis.yml`
```yml
- type: entity
  id: BorgChassis[Department][Variant] # ChassisEntityPrototypeID
  parent: BorgChassisSelectable # Don't touch.
  name: [variant name] [department name] cyborg # Really whatever you want to name it.
  components:
[...]

PR #6385, opened 2026-07-27
Revert "PORT: Selectable Cyborg Chassis from CosmicDrift (#5712)"

ShepardToTheStars:
This reverts commit 4e2dce441400a55e60ec1d5227c5f8e0ccaac525.

## About the PR

Reverts the cosmic drift PR.

## Why / Balance

It's a bad implementation. It makes borg inheritance a million times more worse.

* Generic borgs are an accessibility nightmare. We've had people complain about they cannot easily tell borgs apart due to colorblindness making the colors hard to tell apart.
  * The other ones aren't great sprites either. The best sprite is the 3 mediborgs but even if that isn't enough to keep this whole feature around.
* The sprite issue that is fixed by #6334 is not a fix I like. I'm duplicating sprites due to the new system requiring sprites to be named a certain way and making upstream level things abstract. Not a huge fan of that.
  * Animated moving sprites don't actually work with this feature, and I had to uncomment logic that was "migrated" to this feature, but that logic isn't even working for some reason?
* Borg types are a prototype but this feature adds subtypes as a component. That just makes things confusing.
* I've been seeing a lot of client errors when I open the console when I'm aghosting around borg sprite layers.
* Hat displacement is incorrect. I'll see floating hats on some borgs and on others, the hat will be in the wrong place.

I can see this being reimplemented.
## Technical details

This bypasses CR. This feature has maintainability issues.

## Requirements

## Licensing

**Changelog**

:cl:
- remove: The additional borg chassis from Cosmic Drift has been reverted due to implementation issues.


SalDragonsNOA:
I dunno man. It feels kinda rude to just say "no one cares to make this better. Let's 86 this." You're kinda at fault of that yourself. It's also just kinda rude to assume everyone wants this.

You have a point about the generics. They are really bad for colorblind folks. Thing is, they can be touched up. Made to fit their roles better while keeping the generic borg vibes. You could have asked for this in the spriter's area of the Discord.

I know I tried to do some touch up to the sec borgs so they matched the new sec color scheme. It was just never implemented. Beyond that, I don't think the sprites are bad at all.

It's okay to be frustrated. I get it. Think about it though. You could have helped made it better. You're complaining no one wants to make this better. Couldn't you have helped with that? You're a Senior CR. That matters a lot here, and you probably know more about this than I do.  Maybe you even asked about this before.

Still, I just see people having fun with it and I want people to continue having fun and  getting more option, not having options taken from them. I think there's just a better solution here, even an easy one, that doesn't require getting rid of everything.

sowelipililimute (approved):

SirWarock:
I find your comment missplaced.
As someone who has seen the YML inheritance introduced by this PR, and the required YML to fix the bug at hand, I wholeheartely agree with the PR author that it's just not well done.

I find the suggestion that she could've made it better tasteless, because, she did try. #6334 is her PR attempting to fix the problem. That's more than anyone else has done.

When playing the normal sprites, anyone outside of the PVS range does not see them, because of a bug introduced by #5712.
The above fix would fix this issue, but it's also suffering under the initial code burden.

Niomi0:
As an avid engi borg player, its been so pain that i cant play default engi borg due to bugs. I would prefer it to be fixed as i've noticed a number of players using the new sprites and i kinda enjoy the diversity.

ShepardToTheStars:
* Generic borgs are an accessibility nightmare. We've had people complain about they cannot easily tell borgs apart due to colorblindness making the colors hard to tell apart.
  * The other ones aren't great sprites either. The best sprite is the 3 mediborgs but even if that isn't enough to keep this whole feature around.
* The sprite issue that is fixed by #6334 is not a fix I like. I'm duplicating sprites due to the new system requiring sprites to be named a certain way and making upstream level things abstract. Not a huge fan of that.
  * Animated moving sprites don't actually work with this feature, and I had to uncomment logic that was "migrated" to this feature, but that logic isn't even working for some reason?
* Borg types are a prototype but this feature adds subtypes as a component. That just makes things confusing.
* I've been seeing a lot of client errors when I open the console around borg sprite layers when Im aghosting.
* Hat displacement is incorrect. I'll see floating hats on some borgs and on others, the hat will be in the wrong place.

ShepardToTheStars:
Honestly, the reason why #6334 took days to fix even after I knew the issue is because this feature made borgs so convoluted to work with that I just avoided it lol

SalDragonsNOA:
> * Generic borgs are an accessibility nightmare. We've had people complain about they cannot easily tell borgs apart due to colorblindness making the colors hard to tell apart.
>       
>       * The other ones aren't great sprites either. The best sprite is the 3 mediborgs but even if that isn't enough to keep this whole feature around.

I can understand restating what you have for future people coming in and reading this. Still, I've already  gone over this. Cold blinness issue? Agree. Overall sprites bad? Disagree.
 
>     * The sprite issue that is fixed by [Fix Generic Borg Sprite Issues #6334](https://github.com/DeltaV-Station/Delta-v/pull/6334) is not a fix I like. I'm duplicating sprites due to the new system requiring sprites to be named a certain way and making upstream level things abstract. Not a huge fan of that.

I feel this this doesn't require duplicating sprites. It just requires the YML to get redirected to the right spot. What I can say is, I don't understand glitches or why things like that happen. I can understand that it's annoying though.

>       * Animated moving sprites don't actually work with this feature, and I had to uncomment logic that was "migrated" to this feature, but that logic isn't even working for some reason?
Yeah, that can be annoying. I don't know how much these sprites are animated normally. From what I've played of borg, there's an on and off state. There's not really animations to them moving or the like. I can understand the work needed to actually get things back to functionality, even if some of the logic is there by default.
 
>     * Borg types are a prototype but this feature adds subtypes as a component. That just makes things confusing.

I feel like this is a lore thing. I agree somewhat, but having different borgs for the same job does make sense even if they functionally all act the same.

>     * I've been seeing a lot of client errors when I open the console around borg sprite layers when Im aghosting.

Yeah, that can be annoying. I get that completely.

>     * Hat displacement is incorrect. I'll see floating hats on some borgs and on others, the hat will be in the wrong place.

Here's the thing about this. That was already an issue. Security borgs look weird with the hat and service borgs are usually not wearing hats correctly at all. This can actually be fixed by applying a displacement modifier for the hats so each goes in the correct spot based on the borg type. Those with the more humanoid body could get the code copy and pasted. The more unique ones would have to be tweaked.

Me responding to this is basically to say, I get it. I get it needs work. I get that it can be annoying that something doesn't work right out of the box. I know it's annoying to sit down and fix it. You're busy. You'd rather someone else cared enough to fix this. It's hard finding someone that might care though. (I don't know if the guys who ported it could be able to fix these things either.)
It's not as easy as just changing one line of code. That's not even mentioning tests going wrong.

It just feels like a shame to me just to ditch this.
Maybe one day, when the kinks are worked out on the server it came from we can port it again. Until then, it's a shame things are going this way.

Stxcking:
The discussion for this has died down but I do want to say, as the person who ported them, I wasn't exactly happy with the implementation. I also wasn't happy with the alternatives I have seen either nor have the skill to make it from scratch so I settled for this one. 

CR can approve the concept but maintainers have the final say on implementation, if an implementation is bad then it's bad, I am surprised it got through in the first place with the inheritance bullshitery. I would love for this to come back with better execution and apologies for the headache this might of cause that I was ignorant of

PR #6398, opened 2026-08-06
Fix Antags Being Crazy Smol

ShepardToTheStars:
## About the PR

Fixed a bug where antag profiles would run the scaling again after HumanoidProfileSystem already did it and apply height settings more than once lol.

## Why / Balance

image

## Technical details

Before I was shoving the computed scale into the `Scale` value of `ScaleVisualsComponent` and it was applying the scale more than once for some antags, because it was using the previous scale as a starting point. It no longer does that, and it also separates out each of the scaling modifiers into its own property. `HumanoidProfileSystem.ApplyProfileTo` should be idempotent now.

## Media

https://github.com/user-attachments/assets/6ac77c05-04f8-4e78-bb53-bb3ebe2a3102

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Nuclear operatives and ninjas will now be correctly scaled.

### DIFF
--- a/Content.Shared/Sprite/SharedScaleVisualsSystem.cs
+++ b/Content.Shared/Sprite/SharedScaleVisualsSystem.cs
@@ -1,4 +1,6 @@
 using System.Numerics;
+using Content.Shared.Humanoid; // DeltaV
+using Robust.Shared.Prototypes; // DeltaV
 using Robust.Shared.Serialization;
 
 namespace Content.Shared.Sprite;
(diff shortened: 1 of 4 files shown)


IamVelcroboy:
Feature! Give them vent crawl! 😄

rebe83:
The Inso Effect

Content.Shared/Sprite/SharedScaleVisualsSystem.cs
```diff
@@ -7,6 +9,7 @@ public abstract class SharedScaleVisualsSystem : EntitySystem
 {
     [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
 
+    [Dependency] private readonly IPrototypeManager _proto = default!; // DeltaV
```
SirWarock:
```suggestion
    [Dependency] private readonly IPrototypeManager _proto = default!; // DeltaV

```

SirWarock (dismissed):
Works otherwise

SirWarock (approved):

PR #6399, opened 2026-08-07
Fix parcel wrap transform parenting via SpawnAtPosition (#44679)

### LINKED ISSUE #5256: Parcel Wrap of death?
## Description
I was messing with parcel wrap on evac when I came across a weird bug that would cause you to just die and be perma dead.

**Reproduction**
Possible steps 

Step one: Sit down on evac shuttle and before takeoff wrap yourself while seated. Then unwrap yourself. 
Step two: wait for evac to arrive if you did it correct your screen should become weird for a second and your character. 
Step Three: walk around? (Unsure what fully caused the issue these are just the steps to it.)

**Screenshots**

Image

**When I started doing it I noticed after looking back it registers me as buckled and unbuckled right after finishing wrapping myself. And that was the first thing I did it also flipped my screen weirdly.**

Image

Image

**As we land to midpoint this happens to my character and then I somehow don't get stunned at all which may be a sign it worked.**

Image

**After a bit I just suddenly disappear and then start dying like I am spaced I have this part in screenshots and stories because github did not let me put in the full video.**

**Additional context**
I did not know this would happen I did this completely by mistake while trying to mess around with LO. This was done by accident. Also I can try to make multiple mini videos of each part if needed.

ShepardToTheStars:
ParcelWrap: spawn at position, use PopupEntity

## About the PR

Fix a abuse-able bug that makes you uninteractable/invisible (if you move too far from the chair, since it thinks you're still parented to the chair)

https://github.com/space-wizards/space-station-14/pull/44679

Fixes #5256 

## Why / Balance

Bug was abused in Apo 29075
## Requirements

## Licensing

**Changelog**

:cl:
- fix: Parcel wrapping when buckled to a chair no longer will make you invisible and uninteractable.

### DIFF
--- a/Content.Shared/ParcelWrap/Systems/ParcelWrappingSystem.ParcelWrap.cs
+++ b/Content.Shared/ParcelWrap/Systems/ParcelWrappingSystem.ParcelWrap.cs
@@ -97,7 +97,7 @@ public sealed partial class ParcelWrappingSystem
         if (target == user)
         {
             var selfMsg = Loc.GetString("parcel-wrap-popup-being-wrapped-self");
-            _popup.PopupClient(selfMsg, user, user);
+            _popup.PopupEntity(selfMsg, user, user);
         }
         else
         {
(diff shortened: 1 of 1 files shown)


SirWarock (approved):
works

PR #6407, opened 2026-08-10
Update Configs

ShepardToTheStars:
Some of these things were configured on the host but should have been managed by the publish.

### DIFF
--- a/Resources/ConfigPresets/DeltaV/deltav.toml
+++ b/Resources/ConfigPresets/DeltaV/deltav.toml
@@ -39,6 +39,7 @@ tickrate = 30
 [vote]
 restart_required_ratio = 0.7
 preset_enabled = false
+map_enabled = false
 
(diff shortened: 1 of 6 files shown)


Resources/ConfigPresets/DeltaV/deltav.toml
```diff
@@ -39,6 +39,7 @@ tickrate = 30
 [vote]
 restart_required_ratio = 0.7
 preset_enabled = false
+map_enabled = false
```
JustTNE: Seems to be default off anyway, can probably be omitted.

JustTNE (changes requested):
fallbackpreset on periapsis should also be set to Extended

```
[game]
fallbackpreset = "Extended"
```

Sorry for missing that

JustTNE (approved):
LGTM

PR #6410, opened 2026-08-13
Small Midround Antag Balance Pass

ShepardToTheStars:
## About the PR

I'm going for small changes only to see how this feels. I would rather not change a lot all at once, so I'm just going to target one ability on each antag.

* Added a 5 second cooldown to ninja's teleport
* Increased the charge cooldown of ninja teleport from 20s to 30s
* Reduced Skia regeneration in the dark and increased damage in light a bit.
  * Skia dark regen nerfed from 5 healing per tick (all damage types) to 2 healing per tick (all damage types)
  * Skia light damage increased from 1.5 to 2
* Added 2 seconds of Corporeal when a revenant melee attacks something.

## Why / Balance

Some midround antags are impossible or just very annoying to deal with due to lack of counterplay.

Ninja teleport cooldown changes require the ninja to plan their teleports better, and prevents multiple teleports in a row. If you are caught by security. I play a lot of ninja and I think this will allow security to trap the ninja in some situations because they cannot use their teleport as liberally in succession. Getting cornered as a ninja is how they die.

Skia are just oppressive as a midround antag. They are similar to ninja in a lot of regards but have innate healing and deal shadow damage, which is usually "true" damage and hard to heal a lot of the time (med runs out of salts quick). They have no footsteps so its hard to know if they are around if they haven't screamed. It takes 1 melee to break a light but much longer to replace said light. 

Skia healing makes them VERY strong because they can just walk away and heal up their damage at most 20 seconds. This gives security a bit more time to find the hurt skia and potentially deal with it, instead of it being back to full health and the cycle repeating.

Revenant melee attack is just annoying when all they do is break station lights and there's absolutely no way to deal with it. They can also just focus on one person and deal small amounts of damage (one being soul damage) and there is just no counterplay to that.

## Technical details

YAML

## Media

https://github.com/user-attachments/assets/a6d3383a-8586-471c-ad1a-f98d91839bc0

image

https://github.com/user-attachments/assets/9418aa52-ff89-4faf-8073-4647cf00f6d6

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Ninja's katana dash now has a 5 second cooldown between uses.
- tweak: Ninja's katana dash now takes 30 seconds to recharge each use, up from 20 seconds.
- tweak: Reduced skia's healing in the dark from 5 of each damage type to 2 of each damage type.
- tweak: Increased skia's damage from light from 1.5 brute to 2 brute.
- tweak: Revenant's melee attack will now reveal them for 2 seconds.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
@@ -1086,12 +1086,6 @@
   - type: Item
     heldPrefix: green
   - type: AgentIDCard
-  - type: ActivatableUI # DeltaV - adds the agent ID ui
-    key: enum.AgentIDCardUiKey.Key
-  - type: UserInterface # DeltaV - adds the agent ID ui
-    interfaces:
-      enum.AgentIDCardUiKey.Key:
-        type: AgentIDCardBoundUserInterface
   - type: IdCard
     jobIcon: JobIconNinja
   - type: Access
--- a/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
@@ -1086,6 +1086,12 @@
   - type: Item
     heldPrefix: green
   - type: AgentIDCard
+  - type: ActivatableUI # DeltaV - adds the agent ID ui
+    key: enum.AgentIDCardUiKey.Key
+  - type: UserInterface # DeltaV - adds the agent ID ui
+    interfaces:
+      enum.AgentIDCardUiKey.Key:
+        type: AgentIDCardBoundUserInterface
   - type: IdCard
     jobIcon: JobIconNinja
   - type: Access
--- a/Content.Server/Revenant/EntitySystems/RevenantSystem.cs
+++ b/Content.Server/Revenant/EntitySystems/RevenantSystem.cs
@@ -22,6 +22,7 @@ using Content.Shared.StatusEffect;
 using Content.Shared.Store.Components;
 using Content.Shared.Stunnable;
 using Content.Shared.Tag;
+using Content.Shared.Weapons.Melee.Events; // DeltaV
 using Robust.Server.GameObjects;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Prototypes;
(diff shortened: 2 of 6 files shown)


rebe83:
2 seconds reveal is more than what salt does
The counterplay is not standing around crying, but calling for help from chaplain mystagogue or sec
Going to med or anywhere populated, because it is revealed when it collides with people
Making a salt ring
Mystagogue can dispel it, chaplain can throw the Bible around
These should be pretty obvious solutions for "I'm being haunted"
If it's revealed when melee then it hitting anyone is just going to be a stupid slap fight where a passenger hits it with a toolbox for a minute then dies anyways
It is 2 damage per 3 seconds so very possibly a minute long DPS race more like a ton and Jerry slap fight. This would be a good thing if not for:
The intent of scratch was to make the rev an actual threat 
The person hitting the rev being baited into attacking back will still die so it's literally just bait (maybe unless they have a welding tool) 
The other point of the changes was so that random people are much less able to kill the revenant. Of course this makes it "frustrating" for crew. But it's also frustrating when a nukie or dragon or ninja or traitor kills them too. I guess I can see it being "more fair" if they see what is happening

ShepardToTheStars:
> The counterplay is not standing around crying

Alright, let's say I call the chaplain over. They come over within a minute. The revenant can just see "Oh shit, they are talking about killing the rev" because they can hear and see everything. Then it just fucks off to another department through the wall.

rebe83:
> > The counterplay is not standing around crying
> 
> Alright, let's say I call the chaplain over. They come over within a minute. The revenant can just see "Oh shit, they are talking about killing the rev" because they can hear and see everything. Then it just fucks off to another department through the wall.

Yeah congrats you didn't die and can go get healed. You were successfully annoyed for a minute

Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
```diff
@@ -1086,6 +1086,12 @@
   - type: Item
     heldPrefix: green
   - type: AgentIDCard
+  - type: ActivatableUI # DeltaV - adds the agent ID ui
```
SirWarock: This is not a bug fix, this is a buff.
If an ID only has the `AgentIDCard`, then it's meant to not be able to change its appearance or icon.
Compare these to the nuclear operative IDs where this is the same.

SirWarock: The Ninja Agent ID got added here:

Image

SirWarock (changes requested):
Haven't tested it yet, but the YML and code looks good.

Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
```diff
@@ -1086,6 +1086,12 @@
...
+    key: enum.AgentIDCardUiKey.Key
+  - type: UserInterface # DeltaV - adds the agent ID ui
+    interfaces:
+      enum.AgentIDCardUiKey.Key:
+        type: AgentIDCardBoundUserInterface
```
ShepardToTheStars:
```suggestion
```

Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
ShepardToTheStars: Gotcha. Removed it.

SirWarock (approved):
Works

PR #6416, opened 2026-08-14
Ramping Scheduler Fixes / Convert to Survival Fixes

ShepardToTheStars:
## About the PR

* The ramping scheduler will now look at the time its added as a rule, instead of looking at where the round started to determine event density. 
  * This will result in more time after the major antag fails and the round converts to survival. Basically, the scheduler won't just start firing events because it thinks it was added the whole time.
* Fixed the scheduler converting to survival if a nukie died or went critical before they left their base.
* Fixed the scheduler attempting to add a ramping scheduler after midround zombies are defeated.
* Xenoborg rounds no longer have a ramping scheduler roundstart. Instead, it gets added if the mothership core is defeated.

## Why / Balance

Community feedback, incorrect assumptions made by staff (we said it looked at when it was added instead of looking at round time), and bug fixes.

### Ramping Scheduler Uses Time Added
(Video too big so here's the logs)
image
 

### Xenoborg Change

https://github.com/user-attachments/assets/2519df2b-2960-4208-8fd1-8e220d3028d2

### Nukie Fix

https://github.com/user-attachments/assets/666c5a2b-291e-46ba-b099-5b15864edb55

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: The ramping scheduler will now look at when it was added to determine event density, rather than overall round time. This will result in less events and more time to recover after a major antag is defeated.
- tweak: The Xenoborgs gamemode no longer has a ramping scheduler at round start. Instead, it gets added when the mothership core dies (like other game modes).
- fix: Fixed a bug where the ramping scheduler would get added if a nukie died on nukie planet.
- fix: Fixed a bug where a ramping scheduler would get added if the crew defeated midround zombies.


Content.Server/StationEvents/RampingStationEventSchedulerSystem.cs
```diff
@@ -114,13 +118,29 @@ private void PickNextEventTime(EntityUid uid, RampingStationEventSchedulerCompon
...
+        var ruleActivated = TimeSpan.FromSeconds(0);
+        if (TryComp<GameRuleComponent>(uid, out var gameRule))
+        {
+            ruleActivated = gameRule.ActivatedAt;
+        }
+
+        var ruleTime = _timing.CurTime.Subtract(ruleActivated);
+
+#if DEBUG
+        // Too lazy to localize these since they are only for local debugging
```
ShepardToTheStars: I got motivated.
```suggestion
```

chexo3:
!discord War is over, I’ll finally have enough downtime to ~~kiss boys in maints~~ have meaningful character development

Content.Server/GameTicking/Rules/Components/XenoborgsRuleComponent.cs
```diff
@@ -43,4 +44,17 @@ public sealed partial class XenoborgsRuleComponent : Component
     /// </summary>
     [DataField]
     public bool XenoborgShuttleCalled = false;
+
```
SirWarock, line 47: unnecessary empty space

SirWarock (dismissed):
Works as intended. Just do this one empty space, and you can merge it

SirWarock (approved):

PR #6435, opened 2026-08-20
feat: Unorbitable component

DisposableCrewmember42:
## About the PR
Added the `Unorbitable` component, which prevents ghosts from following an entity. Optionally, admins can be allowed to follow by setting `Unorbitable#AllowAdmins` to true (false by default). Hides the follow verb for those unable to follow.

Prevents warping via the chat links as well.

## Why / Balance
Sometimes I don't want to be followed while aghosted

## Technical details
- Simple comp with boolean for allowing admins to follow
- System forces followers to detach on comp init
- Patched GhostSystem and FollowerSystem to prevent following and hide the Follow verb

## Media
[pr-unorbitable.webm](https://github.com/user-attachments/assets/1d70e1e2-a520-420e-b869-461f1ea10fe8)

## Requirements

## Licensing

**Changelog**

:cl:
DELTAVADMIN:
- add: Added Unorbitable component, which prevents entities from being followed/warped to.

### DIFF
--- a/Content.Server/Administration/Commands/StealthminCommand.cs
+++ b/Content.Server/Administration/Commands/StealthminCommand.cs
@@ -1,4 +1,5 @@
 using Content.Server.Administration.Managers;
+using Content.Shared._DV.Administration.Components; // DeltaV - Unorbitable
 using Content.Shared.Administration;
 using Robust.Shared.Console;
 using Robust.Shared.Utility;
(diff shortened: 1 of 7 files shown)


ShepardToTheStars:
@DisposableCrewmember42 What are your thoughts on adding this functionality to the `stealthmin` command. We could get that added to admins if they do not have access to it.

image

DisposableCrewmember42:
> What are your thoughts on adding this functionality to the `stealthmin` command. We could get that added to admins if they do not have access to it.

I think stealthmin is intended to be used by people to investigate staff reports, it hides you from other staff who might check adminwho to see if they're alone, not from players.

DisposableCrewmember42:
I guess it would still be useful for that though, I don't know if we actually make use of stealthmin, but I'll add it. I don't think everyone should have access to that though, hiding yourself from other staff should really only be used by Heads for investigating staff imo (also if everyone has the perm it does nothing, people with it can see stealthed users).

DisposableCrewmember42:
@ShepardToTheStars added to stealthmin

ShepardToTheStars (approved):

PR #6438, opened 2026-08-22
Buffing the Hushpup

SumofThreeParts:
## About the PR
I changed the magsize of the hushpup from 4 to 5 and reduce the sound it makes when it fires

## Why / Balance
"Essentially it has so little gas all you will do is walk up a guy and blast them, and this is only successful against someone with no armor. 
Which is to say it's not really doing anything any other weapon can't do because you have to point blank them, so why not get a melee weapon where you don't have to reload, or a bomb pen. Or AP. Anything you want this for is done better by something else "
quote from rebe

## Technical details
I will explode if this does not work

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: the hushpup now  makes less sound when fired.


Resources/Prototypes/Entities/Objects/Weapons/Guns/Shotguns/shotguns.yml
```diff
@@ -312,7 +312,7 @@
     shape:
     - 0,0,3,0 # smaller than BaseShotgun by 1 tile
   - type: BallisticAmmoProvider
-    capacity: 4
+    capacity: 6
```
ShepardToTheStars:
```suggestion
    capacity: 6 # DeltaV - was 4
```

SumofThreeParts:
!discord I agree that the main draw of the gun is supposed to be that it is a *stealthy* shotgun, so ill lower the round count to 5 and greatly lower the sound it makes when firing. the bigger issue honestly is that shotgun pellets and shells themselves are just real loud, and im not sure how to make a caseless shotgun shell.

rebe83:
You should wait for them to request changes formally before changing things because often it's a 1guy moment that does not get agreed with by the majority
Especially on a 1 day discussion where not everyone has talked yet 
There's a bunch of different directions people are going so let them argue

SumofThreeParts:
!discord I feel the main issue with the gun is that is is very outclassed in both stealth and combat, and its not good enough at either to be a versatile option for both stealth and combat either.

SumofThreeParts:
!discord alright ill change the count back to 4, the PR only changes the amount of sound it makes when firing now

Radezolid (approved):

PR #6424, opened 2026-08-18
30tc catalogue updates

Proxysseia:
## About the PR
increased cost of various combat essentials

## Why / Balance
was supposed to happen 3 weeks ago woops sorry halo
point of the 30tc change is to let people roleplay more by giving them access to more tc to use on off meta stuff and to make scenes big and memorable. you dont need cheap noslips and emp implants for a big memorable scene.

## Technical details

Item (new base cost/new discount cost) from (old base cost/old discount cost)

Viper 5/3 from 3/2
Python 6/4 from 4/2
Cobra 6/4 from 4/2
Adder 6/4 from 5/3
Hristov 12/6 from 9/4
Hushpup 12/9 from 10/8
SLK93 12/9 from 10/8
C20 18/11 from 15/8
Estoc 18/11 from 15/11
Akurra 18/10 from 15/7
Bulldog 16/9 from 15/8
L6 30/14 from 28/12
Hyperblade 30/26 from 20/16
Harbinger Scythe 16/12 from 12/10

Minibomb 9/6 from 6/3
EMP Grenade 3/1 from 2/1
EMP Kit 7/4 from 5/3
EMP Implant 5/3 from 2/1
Hyperzine 4/2 from 3/1
Hyperzine Bundle 12/6 from 10/5

Hypopen 10/7 from 8/6
NoSlips 4/2 from 2/1
Webvest 6/4 from 5/3
Elite Webvest 6/4 from 5/3
Bloodred 9/4 from 8/4
Raidsuit 9/4 from 8/4

Syndiebaloon 30 from 1 (the whole point is you buy it after doing all objs with 0 tc)

**Changelog**
:cl:
- tweak: most weapons in the traitor arsenal have their had tc cost increased (by 1-10)
- tweak: most armors in the traitor arsenal have had their tc cost increased (by 1-2)
- tweak: combat essential items like the emps, no-slips and auto injectors have had their tc cost increased (by 1-3)
- tweak: the syndie baloon now costs all your tc again

PR #6448, opened 2026-08-23
25TC Hyperblade

ShepardToTheStars:
## About the PR

Hyperblade to 25TC

## Why / Balance

Allows the buyer to also choose no-slips or webvest but not both.

## Technical details

Moved DV listing to the _DV namespace as well

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Hypereutactic Blade TC cost increased from 20 TC to 25TC.


rebe83:
Is it even needed
Was it oppressive or unfun
Or is it just "because it should be"

Radezolid (approved):

ShepardToTheStars:
> Is it even needed
> Was it oppressive or unfun
> Or is it just "because it should be"

It's going to go through CR so maybe. But from what I've seen this weapon just kind of feels like the complete package, especially if gets other combat utility items - web vest, no-slips, hyperzine.
* Previously, it costed 20TC when TC given was 20. It assumed you wouldn't get any utility items.
* It has insane reflect and reach, and the damage very high. You will be winning in melee and no one can touch you at range.
* You can toggle it off when you don't want its downside.
* Oni exist and make this weapon absolutely bonkers.

It's weakness is thrown weapons but unless the Warden is making spears for the armory (lol) but who is going to do that if this weapon is the only weapon that requires that. Throwing batons only works if hyperzine isn't in the picture, so its not reliable.

If you say you can just run away, that's not really the case if one is unarmored (and sometimes even if you are armored because armor slowdowns). First hit is always a surprise, then if you start running, you'll still be in range for another hit because of its reach, and then you are slowed and basically dead. There's not a lot of counterplay other than just throwing a lot of bodies at the problem until its finally overcome through sheer amounts of damage.

The previous PR to increase TC prices by Proxy had this at 30TC, so this is just kind of meeting that one  halfway.

rebe83:
You actually can run away because
Put in hand 1 second cooldown before you can swing after activate
1.5 seconds between swings or whatever number
It slows you down while it's in your hand not just activated, so you have to hotkey out it in your bag and then move, pull it out, then activate it, wait a second, then swing 
I am asking for actual data of it being oppressive because any footage of it you can see the 1 guy who uses it ever getting clowned on usually and the only people that die are running at him
You can genuinely outrun it in armor unless the person is taking (time limited) drugs

SalDragonsNOA:
I think both of you have a point. One's just being more preventative about it.

I get not wanting to do a knee jerk reaction and waiting to see folks complain first. On the other hand, the Hyperblade is still good and you can get additional items to make it better.

My thought is, changing the cost only makes it easier to get a no-slips combo with armor. You can still get no-slips just by stealing the janitor's boots. Armor's trickier, but you can still grab a few defensive options. Heck, if you walk up to sec and ask for an armored coat, there's a chance they'd just give it to you for drip. You can steal stuff from other departments though, like salvager armor.

My point is, is making it slightly more difficult to get this combo going going to be worth it when it can still be done? I think so. Gives them more times to make mistakes, chose what they want now, and breeds resourcefulness. Limitation breeds creativity. You just give someone a combo like that, there's no extra thought going on, and only their plan becomes the next step. Bit boring.

One thing you gotta keep in mind, Traitors have more objectives now. Buying the Hyperblade, even with some extra TC is going to dead-end them unless they know how to get everything else they need without TC. If you have say "Steal Mysta's Digiboard," "Kill the LO and make them stay dead," "Escape to midpoint unrestrained," and "Kill Rando McGee," the Hyperblade only helps 2 of those things, maybe 3, but if you get caught with it, you ain't getting out unrestrained, and if you get caught too soon, you are screwed. Dumping most of what you have into that sword is an investment that can handicap you fast. You're going to need extra stuff to make your run work any ways, and that should be incouraged.

rebe83:
My point is that changing it when the item isn't really being used is making it worse for fear of a problem that doesn't exist so the item is unuseable in most situations. It is not going to ever do as much as an emag + c20 + web vest can. Trying to ask to think "is this a problem" or are we just doing "balance" for the sake of balance on pure hypothetical. Because the apocalypse scenario should have happened but it didnt

SalDragonsNOA:
> My point is that changing it when the item isn't really being used is making it worse for fear of a problem that doesn't exist so the item is unuseable in most situations. It is not going to ever do as much as an emag + c20 + web vest can. Trying to ask to think "is this a problem" or are we just doing "balance" for the sake of balance on pure hypothetical. Because the apocalypse scenario should have happened but it didnt

That's a really good point. I'm not sure I can say much that matters at this point though.
It was already kinda unusable. I'm not sure I'd like to see it more, but yah know, seeing it a tad more often, that would be interesting.

I just kinda see things both ways here. I don't think there is a best option, just options. There's not quiet a right answer to me. I haven't messed around with this stuff that much to really know much in terms of "The meta," so to speak.
I still think the Hyberblade not eating all your TC is good overall.

ShepardToTheStars:
Changing to 30TC changes things, this we cannot deny. If you want me to record X rounds where its DAGD and they buy this and 3 other “i win” utility items, i probably could do this. DAGD is fairly common with 4 objective rolls these days. Not all of them select this 100% of the time but I have been observing rounds quite a bit lately and it is not uncommon, and this is why I made the PR.

I was not as heavy handed as Proxy was in #6424, yet I never saw any criticism on their PR about the price of this weapon. Perhaps it was because they nerfed literally everything and this fell by the wayside, I dont know. This is not the first time suggested that this weapon needs adjustment. I’m not sure why I’m being held to this double standard.Asking for concrete data is a bit much unless I can show data from grafana, which I cannot. And honestly, literally no one can do this. This is about game feel, pure and simple.

if concrete data is what I need to make nerfs, that should be required for *any* nerfs, regardless if Im the one requesting it. And that would prevent us from making any changes to problematic weapons. People complain about ninja all the time. Do I reject ant nerfs unless they show data?

I think we all agree that somethings may need adjustment if TC is set at 30. This just happens to be the first one. I will make targeted nerfed instead of across the board nerfs.

if you think that the nerf is too much, I am willing to listen. If the criticism is “is it a problem, plz show data,” we never nerf or buff anything because no one is in a position to actually show concrete data. Things can feel bad to play against, and thats kinda what this weapon feels like if its given literally every utility open available (webvest+noslips+hyperzine)

rebe83:
I criticized the other or just not in an intellectual way because that one was seemingly arbitrary and on a totally flawed premise imo it was just dumb and if I thought about it to much I'd just get annoyed so I didn't
This one im just questioning the logic
I'm not asking for data im saying is there an event that has happened where it was a big deal
Because in practice meds and armor doesn't do a lot because the guy is already invincible to bullets and slower then everyone else. Even hyperzine doesn't let you run laps around anyone I believe.
I don't think it's being too harsh if I'm just asking if we are on autopilot and making assumptions like we always have about the weapon and I've always pushed back on it

Creb206:
I really don't see a need for this. Honestly as it is, the hyperblade is more of a joke option to me than anything, people only use it for the style points. If you want to kill someone with melee you're better off using an E-sword, the slow down of the hyperblade is extremely bad because movement speed is your life blood when you're a solo antag which you are going to be while using this. If you're moving while shooting someone using the hyperblade most of the reflected bullets are going to miss as well. In my personal experience it's VERY easy for people with armor to just walk away after being hit, like the captain, and most of the kills you're able to get are either by shock of turning it on and attacking someone, something that literally any weapon can do easily, and probably better than the hyperblade, or extremely bad positioning which is just the fault of whoever is fighting them. People using hyperblades might get a few kills but quickly get stuck in a spot where they can't push forward because they are too slow, and they can't run away because they would be easily followed, they also can't turn the sword off and put it away as this would leave them defenseless. Overall I see it as a bad weapon with a lot of style, something that realistically needs a buff rather than a nerf like this.

PR #6450, opened 2026-08-25
Anti-tamper Rebalance

Tmanzxd:
## About the PR
Removed most Anti-tamper from Security/Armor crates allowing crew easier access to Shotguns, Revolvers, Pistols, Armor, Nonlethals, and Laser Rifles.

Kept Anti-tamper for the following Assault weapons: Lecter, Wt550 SMG, Mk.1 Rifle, and the NT-3 Heavy Machine Gun Crates.

## Why / Balance

With WarOps and Ramping gamemodes back on the menu I believe that crew should have more of a fighting chance versus Nukies or other Antags should Security or Command fail to hold the line. As of currently, if Nukies kill all SecOffs and Command it is basically gameover for crew as you need Security ID, Warden ID, or HoS ID to bypass Anti-tamper.

## Technical details

Comments out Generalized Anti-Tamper for all Security/Armory crates, narrows Anti-Tamper down to specific crates.

## Requirements

## Licensing

**Changelog**

:cl:
- remove: Removed Anti-Tamper from most Security/Armory Crates.
- add: Added Anti-Tamper to the following Assault Weapons: Lecter, Wt550, Mk.1 Rifle, and the NT-3 Heavy Machine gun crates.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Crates/armory.yml
+++ b/Resources/Prototypes/Catalog/Fills/Crates/armory.yml
@@ -13,7 +13,7 @@
         - id: MagazinePistolSubMachineGunTopMounted
           amount: 4
   - type: AntiTamper # DeltaV - port impstation's anti-tamper
-  containers: [ entity_storage ] # DeltaV - port impstation's anti-tamper
+    containers: [ entity_storage ] # DeltaV - port impstation's anti-tamper
 
 - type: entity
   parent: [ CrateWeaponSecure, BaseSecurityContraband ]
@@ -117,4 +117,4 @@
         - id: MagazineRifle
           amount: 4
   - type: AntiTamper # DeltaV - port impstation's anti-tamper
-  containers: [ entity_storage ] # DeltaV - port impstation's anti-tamper
+    containers: [ entity_storage ] # DeltaV - port impstation's anti-tamper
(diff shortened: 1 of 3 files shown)


Tmanzxd:
Lemme know if you want me to add anti-tamper to other crates as well.

rebe83:
The theoretical scenario where nukies have killed all of security where none of the 5 options that give armory access are available, and they can't get into sec to get lathes + mats / research 
Then the antag won because they were able to kill vital staff and control assets because it's not easy to pull off
Any fighting scene will have a zillion crew zipping in and stealing things and dragging bodies around, that is the counterplay.
Keeping it on powerful items is mostly an antag nerf so cargo techs don't get WTs

Tmanzxd:
Primary complaint that brought in Anti-tamper is Logistics powergaming via validhunting. It still doesn't really sit right with me that **everything** should be locked away and there should be differing classifications for anti-tamper which is the point of this PR. 

Public CR's primary concern comes to Logi spam ordering NT-3 HMGs and Enforcers, former I put the anti-tamper on, latter I did not but that can be changed.

Personally I have never really agreed with punishing everyone via limiting in game mechanics based on a few bad actors. That should be more of an Administrative issue, but here I am trying to compromise regardless.

Vova567810:
I kinda wanna say to make mk1 rifle not tampered, since it kinda sucks in general besides being one handed 
But that's just my opinion

PR #6455, opened 2026-08-27
Fix/Lower Asakim Weight

kotobdev:
## About the PR
Asakim are allegedly rolling multiple times per day now post-https://github.com/DeltaV-Station/Delta-v/pull/6401, and have the same weight as a loneop- or half the weight of the ever-ubiquitous DisasterVictims. This is extraordinarily *not* what they are supposed to be. Weight has been lowered to one-sixth of current, 3 to 0.5.

## Why / Balance
I mean technically I only needed to lower it to 1, as it's always sat generally around "half the weight of SRN", but Asakim are supposed to be *the* rarest antag in the game and this new scheduler seems to throw way more shuttle antags overall. so shrug

Really, they probably shouldn't be in this table at all, but do you know how long it takes to get a deltav devenv open with rider? ESPECIALLY after contributing to another fork? Takes like 15 minutes! 32gb of DDR4 ram with the repo stored on a HDD is actual torture. Oh, and I always have to swap github accounts and git information whenever I change repos, which is just a nightmare. This doesn't have anything to do with the PR I'm just tired and don't feel like doing anything more then a number change

## Technical details
 if you request changes i will cry just do it through github so i can do the automated signoff ibeg

## Media
no

## Requirements

## Licensing

**Changelog**

:cl:
- fix: Asakim should, once again, be very rare.

### DIFF
--- a/Resources/Prototypes/_Mono/GameRules/asakim.yml
+++ b/Resources/Prototypes/_Mono/GameRules/asakim.yml
@@ -17,7 +17,7 @@
     #startAnnouncement: station-event-asakim-shuttle-detected # DeltaV - no
     #  startAudio:
     #    path: /Audio/Misc/notice1.ogg
-    weight: 0.5 # DeltaV - was 2, got moved to more active scheduler, now 0.5
+    weight: 0.85 # DeltaV - was 2, got moved to more active scheduler, now 0.85
     minimumPlayers: 25 # DeltaV
     maxOccurrences: 1 # DeltaV - was 4
     duration: null # DeltaV
--- a/Resources/Prototypes/_Mono/GameRules/asakim.yml
+++ b/Resources/Prototypes/_Mono/GameRules/asakim.yml
@@ -17,7 +17,7 @@
     #startAnnouncement: station-event-asakim-shuttle-detected # DeltaV - no
     #  startAudio:
     #    path: /Audio/Misc/notice1.ogg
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/_Mono/GameRules/asakim.yml
```diff
@@ -17,7 +17,7 @@
     #startAnnouncement: station-event-asakim-shuttle-detected # DeltaV - no
     #  startAudio:
     #    path: /Audio/Misc/notice1.ogg
-    weight: 3
+    weight: 0.5 # DeltaV - was 2, got moved to more active scheduler, now 0.5
```
ShepardToTheStars, line 20: Are you sure you want it to be *that* low?

kotobdev: Doing a bit more digging- seeing as Asakim are now present in both BasicAntagEventsTableDeltaV *and* UnknownShuttlesTableDeltaV instead of just the latter, the highly-reduced weight is probably warranted. Assuming I'm reading everything correctly.

ShepardToTheStars: Ah, would you rather just remove them from BasicAntagEventsTableDeltaV?

Stop-Signs (approved):
LGTM

PR #6477, opened 2026-09-05
Adds Map Maintainer OOC name color

Bribrooo:
## About the PR
Added AdminFlag.Mapper to auto-set the Discord color for mappers (needs to be set manually by an admin)

## Requirements
- [X] I have tested all added content and changes. (IT LAUNCHES!!!)

## Licensing

**Changelog**
:cl: BriBrooo
- add: Added map maintainer OOC name colors!


Stop-Signs:
@Vapetastic-Gaming @AeraAuling @FieldCommand this one is on y'all

Vapetastic-Gaming:
Whatever happens make sure you do this for Nuclear operative patreons as well 😉

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -294,7 +294,13 @@ private void SendOOC(ICommonSession player, string message)
 
         Color? colorOverride = null;
         var wrappedMessage = Loc.GetString("chat-manager-send-ooc-wrap-message", ("playerName",player.Name), ("message", FormattedMessage.EscapeText(message)));
-        if (_adminManager.HasAdminFlag(player, AdminFlags.Admin))
+        // DeltaV - Mapper color START
+        if (_adminManager.HasAdminFlag(player, AdminFlags.Mapper))
+        {
+            colorOverride = new Color(0x00, 0xCE, 0xCE);
```
DisposableCrewmember42: This should probably be stored in a cvar instead of being hardcoded

FieldCommand:
As a mapper, I don't mind, but overall I think if we allow for this, we kind of have to allow it for every department.

Bribrooo:
> Whatever happens make sure you do this for Nuclear operative patreons as well 😉

When I figure out how to link patreon/discord maybe BLEGH

Bribrooo:
> As a mapper, I don't mind, but overall, I think if we allow for this, we kind of have to allow it for every department.

I thought mapping was the only department left out...

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -1,4 +1,5 @@
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization; // Delta V - OOC color parsing
```
DisposableCrewmember42:
```suggestion
using System.Globalization; // DeltaV - OOC color parsing
```

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -294,7 +295,22 @@ private void SendOOC(ICommonSession player, string message)
 
         Color? colorOverride = null;
         var wrappedMessage = Loc.GetString("chat-manager-send-ooc-wrap-message", ("playerName",player.Name), ("message", FormattedMessage.EscapeText(message)));
-        if (_adminManager.HasAdminFlag(player, AdminFlags.Admin))
+        // DeltaV - Mapper color START
+        if (_adminManager.HasAdminFlag(player, AdminFlags.Mapper))
+        {
+            var mapperColor = _configurationManager.GetCVar(CCVars.MapperOocColor).Trim();
+
+            if (TryParseHexColor(mapperColor, out var parsedMapperColor))
```
DisposableCrewmember42:
```suggestion
            if (Color.TryFromHex(mapperColor, out var parsedMapperColor))
```

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -486,6 +502,32 @@ private bool ShouldShowFollowButton(INetChannel recipient)
...
+            !byte.TryParse(normalized.Substring(2, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var green) ||
+            !byte.TryParse(normalized.Substring(4, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var blue))
+        {
+            return false;
+        }
+
+        color = new Color(red, green, blue);
+        return true;
+    }
+    /// DELTA V - END
```
DisposableCrewmember42:
```suggestion
```

Content.Shared/CCVar/CCVars.Chat.Ooc.cs
```diff
@@ -25,6 +25,12 @@ public static readonly CVarDef<bool>
     public static readonly CVarDef<bool> ShowOocPatronColor =
         CVarDef.Create("ooc.show_ooc_patron_color", true, CVar.ARCHIVE | CVar.REPLICATED | CVar.CLIENT);
 
+    /// <summary>
+    ///     Delta V - The color to use for mapper OOC messages.
```
DisposableCrewmember42:
```suggestion
    ///     DeltaV - The color to use for mapper OOC messages.
```

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -294,7 +295,22 @@ private void SendOOC(ICommonSession player, string message)
...
+            {
+                colorOverride = parsedMapperColor;
+            }
+            else
+            {
+                colorOverride = new Color(0x00, 0xCE, 0xCE);
+            }
+        }
+        // DeltaV - END
+        else if (_adminManager.HasAdminFlag(player, AdminFlags.Admin))
```
DisposableCrewmember42:
```suggestion
        else if (_adminManager.HasAdminFlag(player, AdminFlags.Admin)) // DeltaV - Added else
```

Right now, this will override the custom color staff with the Admin flag set for themselves if they also have Mapper. The custom color should probably take priority.

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -294,7 +295,22 @@ private void SendOOC(ICommonSession player, string message)
 
         Color? colorOverride = null;
         var wrappedMessage = Loc.GetString("chat-manager-send-ooc-wrap-message", ("playerName",player.Name), ("message", FormattedMessage.EscapeText(message)));
-        if (_adminManager.HasAdminFlag(player, AdminFlags.Admin))
+        // DeltaV - Mapper color START
+        if (_adminManager.HasAdminFlag(player, AdminFlags.Mapper))
+        {
+            var mapperColor = _configurationManager.GetCVar(CCVars.MapperOocColor).Trim();
+
+            if (Color.TryFromHex(mapperColor, out var parsedMapperColor))
```
DisposableCrewmember42: Oops, turns out that overload of Color.TryFromHex was added in a more recent engine version. It'll have to be done like this instead:

```suggestion
            if (Color.TryFromHex(mapperColor) is { } parsedMapperColor)
```

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -1,4 +1,5 @@
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization; // DeltaV - OOC color parsing
```
DisposableCrewmember42:
```suggestion
```
no longer needed, we now use RT parsing

DisposableCrewmember42:
Is Mappers being stuck with the role color really desired behavior here? If OOC colors are going to be given to staff without ingame perms aside from Mappers as well, giving them the ability to just set a custom one would be simpler.

If the condition checking for the Admin flag is changed to check for the NameColor flag instead, any rank with just NameColor should be able to set and use a custom color without getting further perms.

Actually, it looks like this has been the case upstream since May 2025, it just wasn't merged here for some reason: https://github.com/space-wizards/space-station-14/pull/36148

Content.Shared/CCVar/CCVars.Chat.Ooc.cs
```diff
@@ -25,6 +25,12 @@ public static readonly CVarDef<bool>
...
+    /// </summary>
+    public static readonly CVarDef<string> MapperOocColor =
+        CVarDef.Create("ooc.mapper_color", "#00CECE", CVar.SERVERONLY);
+
```
DisposableCrewmember42, line 33: This could be added to `Content.Shared/_DV/CCVars/DCCVars.cs` instead to avoid touching the upstream file.

sleepyyapril:
I'm against adding a permission solely for a name color and would prefer using the existing server-based discord bot functionality to use people's discord roles for a similar outcome.

ShepardToTheStars:
Yeah, not to discourage mappers but I agree. This seems like overkill.

Mappers do good work but this isn’t really expandable to other roles that should/would also get the same treatment (e.g. maintainers).

PR #6478, opened 2026-09-05
silvia has negacord venom

keekee38:
## About the PR
changed silvias venom from omni to negacord

## Why / Balance
folks dont like how silvia affects other chems
like 2 med folks kinda were like "thats not a completely terrible idea

## Technical details

yaml

## Media

2 bites
image

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: Silvias antivenom is now negacordrazine! don't ask how.

### DIFF
--- a/Resources/Locale/en-US/_DV/ghost/roles/silvia.ftl
+++ b/Resources/Locale/en-US/_DV/ghost/roles/silvia.ftl
@@ -1,3 +1,3 @@
 ghost-role-information-silvia-name = Silvia
 ghost-role-information-silvia-description = You are Silvia the space snake and the CMO's charming companion.
-ghost-role-information-silvia-rules = Keep the medical team company and help out in emergencies with your omnizine venom. Stick close to the CMO in case they need emergency healing.
+ghost-role-information-silvia-rules = Keep the medical team company and help out in emergencies with your negacordrazine venom. Stick close to the CMO in case they need emergency healing.
--- a/Resources/Prototypes/_DV/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/_DV/Entities/Mobs/NPCs/pets.yml
@@ -220,7 +220,7 @@
         melee:
           maxVol: 30
           reagents:
-          - ReagentId: Omnizine
+          - ReagentId: Negacordrazine
             Quantity: 30 #5 bites to heal 90hp/1 min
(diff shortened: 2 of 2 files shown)


SirWarock:
Not my ambuzol+ source..

keekee38:
ask your local botanist instead grr

keekee38:
oh i gotta fix the yaml comments ill do it later

CMuncey:
If this goes through, might be a good idea to move negacord into the Medicines group so the health analyzer will pick it up and it doesn't look like silvia is poisoning people. Right now it would show up as an unknown reagent since the group is Exotic and not Medicine.

That said, I still don't understand the purpose. From what I can tell, the only recipe omni is used in it ambu+ which I guess could be cheesed if you ask silv to bite someone as you inject normal ambu, but I've never seen that actually happen in a round. If it were an issue though, you could always add a heat requirement to ambu+

keekee38:
ah good point yeah
we'll see how it goes first i suppose

Axianamos:
> If this goes through, might be a good idea to move negacord into the Medicines group so the health analyzer will pick it up and it doesn't look like silvia is poisoning people. Right now it would show up as an unknown reagent since the group is Exotic and not Medicine.
> 
> That said, I still don't understand the purpose. From what I can tell, the only recipe omni is used in it ambu+ which I guess could be cheesed if you ask silv to bite someone as you inject normal ambu, but I've never seen that actually happen in a round. If it were an issue though, you could always add a heat requirement to ambu+

Because omni currently purges any other chems in the bloodstream. So slivia biting people actively hurts doctors trying to treat injuries. Watching 10u of pyra immediately evaporate because silvia wanted to help is frustrating at best.

SirWarock:
> That said, I still don't understand the purpose. From what I can tell, the only recipe omni is used in it ambu+ which I guess could be cheesed if you ask silv to bite someone as you inject normal ambu, but I've never seen that actually happen in a round.

I do it all the time when Silvia is active. I think that's the only time where Silvia feels actively important to an extent where it could have lasting change on the treatment.

Radezolid (changes requested):
The calculations on the comments should be updated or removed, otherwise it's fine.

keekee38:
the interactions i was talking about was purging chems
pyrazine becomes essentially useless etc but sure go off i guess

PR #6479, opened 2026-09-05
Pda evac timer

CMuncey:
## About the PR
Added an evac shuttle status to the PDA's home screen. It will display "Not called" if the evac shuttle is not currently called, "Arriving in <mm:ss>" if the shuttle is on route, "Departing in <mm:ss>" if the shuttle is docked with the station, and "Departed the station" once the shuttle has left.

## Why / Balance
After the new communications console screen changes, the screens are very rarely set to ETA/ETD during an evacuation, making it hard to tell when the evac shuttle will arrive or depart the station. This gives all crew members (who still have a PDA) a way to check on the status of the evac shuttle.

## Technical details
The PDA now subscribes to pre-existing signals from the RoundEndSystem and the EmergencyShuttleSystem upon initialization. When those events are fired, the PDA will grab relevant public member variables from the classes to determine how long the evac shuttle has to arrive/depart.

## Media
https://github.com/user-attachments/assets/f9c5a5ca-c3fd-4406-bb45-19f9fa8b273f

## Requirements

## Licensing

**Changelog**

:cl:
- add: Added evacuation shuttle info to the PDA


Content.Client/PDA/PdaMenu.xaml.cs
```diff
@@ -360,6 +373,34 @@ private void HideAllViews()
...
+                    _evacStatus = Loc.GetString( "comp-pda-ui-evac-gone" );
+                else
+                    _evacStatus = Loc.GetString( "comp-pda-ui-evac-etd", ( "time", diff.ToString( "mm\\:ss" ) ) );
+            }
+            else
+            {
+                if ( _evacArrivalTime is { } evacArrivalTime )
+                {
+                    TimeSpan diff = MathHelper.Max( evacArrivalTime.Subtract( _gameTiming.CurTime ), TimeSpan.Zero );
+                    _evacStatus = Loc.GetString( "comp-pda-ui-evac-eta", ( "time", diff.ToString( "mm\\:ss" ) ) );
```
BarryNorfolk: Same here

Content.Client/PDA/PdaMenu.xaml.cs
```diff
@@ -360,6 +373,34 @@ private void HideAllViews()
...
         }
 
+        // Begin DeltaV - PDA Evac Status
+        private void UpdateEvacStatus()
+        {
+            if ( _evacDepartureTime is { } evacDepartureTime )
+            {
+                TimeSpan diff = MathHelper.Max( evacDepartureTime.Subtract( _gameTiming.CurTime ), TimeSpan.Zero );
+                if ( diff == TimeSpan.Zero )
+                    _evacStatus = Loc.GetString( "comp-pda-ui-evac-gone" );
```
BarryNorfolk: Can use `var` here instead of an explicit type.

BarryNorfolk (changes requested):
Overall I think this is fine. I can't think of a way of neatly threading the round end timer to the client side other than what you've done, maybe something will come to me.
However the formatting needs to be fixed as you have `( foo )` throughout instead of `(foo)`. There shouldn't be spaces between brackets here, I believe we use Riders' formatting rules but VSCode's is also good.

CMuncey:
> Overall I think this is fine. I can't think of a way of neatly threading the round end timer to the client side other than what you've done, maybe something will come to me. However the formatting needs to be fixed as you have `( foo )` throughout instead of `(foo)`. There shouldn't be spaces between brackets here, I believe we use Riders' formatting rules but VSCode's is also good.

The roundend shuttle arrival time is basically just copied from how the station screens get the info. Shuttle departure time might be a little more unique, can't remember if I grabbed that from somewhere or made it myself. Seems to work though. One slight concern I have with this method is that I'm pretty sure if someone joins while evac is called, their PDA will likely read "Not called" since they would have missed the RoundEndChanhedEvent thing. Not sure there's much I can do about it though.

Force of habit on the spacing, that's what we use at work. Not sure what you mean but the formatting rules thing, is that for like auto formatters?

Content.Client/PDA/PdaMenu.xaml.cs
```diff
@@ -360,6 +373,34 @@ private void HideAllViews()
             }
         }
 
+        // Begin DeltaV - PDA Evac Status
+        private void UpdateEvacStatus()
```
sleepyyapril: Partial class pretty please

Content.Client/PDA/PdaMenu.xaml.cs
```diff
@@ -33,7 +33,9 @@ public sealed partial class PdaMenu : PdaWindow
         private string _alertLevel = Loc.GetString("comp-pda-ui-unknown");
         private string _instructions = Loc.GetString("comp-pda-ui-unknown");
         private string _currentDate = Loc.GetString("comp-pda-ui-unknown"); // DeltaV - PDA date
-        
+        private string _evacStatus = Loc.GetString("comp-pda-ui-unknown"); // DeltaV - PDA Evac Status
+        private TimeSpan? _evacArrivalTime = null; // DeltaV - PDA Evac Status
+        private TimeSpan? _evacDepartureTime = null; // DeltaV - PDA Evac Status
```
sleepyyapril: ditto

Content.Server/PDA/PdaSystem.cs
```diff
@@ -69,6 +76,10 @@ public override void Initialize()
             SubscribeLocalEvent<PdaComponent, InventoryRelayedEvent<ChameleonControllerOutfitSelectedEvent>>(OnRelayedEventToIdCard);
             SubscribeLocalEvent<PdaComponent, InventoryRelayedEvent<VoiceMaskNameUpdatedEvent>>(OnRelayedEventToIdCard);
 
+            SubscribeLocalEvent<RoundEndSystemChangedEvent>(OnRoundEndChanged); // DeltaV - PDA Evac Status
+            SubscribeLocalEvent<EvacShuttleDockedEvent>(OnShuttleDockedEvent); // DeltaV - PDA Evac Status
+            SubscribeLocalEvent<EmergencyShuttleAuthorizedEvent>(OnShuttleEarlyLaunch); // DeltaV - PDA Evac Status
```
sleepyyapril: Partial class and make an InitializeDelta method to put these in.

Content.Server/PDA/PdaSystem.cs
```diff
@@ -159,6 +170,42 @@ private void OnAlertLevelChanged(AlertLevelChangedEvent args)
             UpdateAllPdaUisOnStation();
         }
 
+        // Begin DeltaV - PDA Evac Status
```
sleepyyapril: Partial class these 3 methods

sleepyyapril (changes requested):

PR #6486, opened 2026-09-07
Namespace drink metabolites

pixelStrife:
## About the PR
Brought all drinks in line with Nubody metabolism changes (no more adding ethanol AND metabolizing into ethanol at the same time.) Also changed Silencer back to non-alcoholic.

## Why / Balance
Some drinks effectively contained more ethanol than intended, due to the parent class adding the metabolite while the drink still retained the effect of adding ethanol directly. HCV was also parented to BaseDrink despite being alcoholic, and Silencer was parented to BaseAlcohol, giving it the standard 12% ethanol metabolism despite being (formerly) non-alcoholic.

## Technical details
- Replaced AdjustReagent effects in namespace drinks with equivalent metabolites
- Removed SatiateThirst effects with factor equivalent to the inherited one in namespace drinks for cleanliness
- Switched Silencer parent to BaseDrink
- Switched HCV parent to BaseAlcohol

## Media
Before:
SS14 Loader_XD6RiDj7ah
After:
SS14 Loader_dEdV6Im6gB

## Requirements

## Licensing

**Changelog**
:cl:
- fix: Ethanol content of drinks has been re-standardized. No more adding ethanol directly, only metabolizing. (Non-ethanol substances too!)
- fix: Silencer is once again non-alcoholic, for mimes that would rather keep their wits about them.


Stop-Signs (approved):
Very very good. Thanks for the submission!

PR #6489, opened 2026-09-08
Emergency shuttle bug fixes and admin improvements

JustTNE:
## About the PR
1. The shuttle is no longer recallable until the very last second
2. The callshuttle command has received another parameter: `cantRecall` with true/false values
3. The (Re)call shuttle admin UI has received a "Recallable?" checkbox, checked by default, allowing admins to make shuttles unrecallable on demand.
4. Zeta can no longer be called while Zeta is called.

## Why / Balance
Bug fixes, admin/curator tooling

## Technical details
Rewrote the shuttle command to support another parameter and added the `cantRecall` parameter to the `RequestRoundEnd` call.
Added new UI `DVAdminShuttleWindow.xaml` that fully replaces the upstream UI.
Added another update timer that triggers an update to the console components when the shuttle reaches the unrecallable threshold. Previously, the check in `ShuttlesCallable` would never be executed after the initial shuttle call since `AlertLevelChangedEvent` does not trigger again later. (Not entirely accurate, but good enough for this section.)
Added simple "are we zeta" check before calling zeta.

## Media
image
image

## Requirements

## Licensing

**Changelog**

:cl:
- fix: The emergency shuttle is no longer recallable after its "point of no return" threshold.

DeltaVAdmin:
- tweak: Emergency shuttle admin UI now has a "Recallable?" checkbox toggle.


JustTNE:
Test fail seems fake

Content.Client/_DV/Administration/UI/Tabs/AdminTab/DVAdminShuttleWindow.xaml
```diff
@@ -0,0 +1,14 @@
+<DefaultWindow xmlns="https://spacestation14.io"
+            xmlns:cc="clr-namespace:Content.Client.Administration.UI.CustomControls"
+            Title="{Loc 'admin-shuttle-title'}"
+            MinWidth="300">
+    <BoxContainer Orientation="Vertical">
+        <BoxContainer Orientation="Horizontal" HorizontalExpand="True">
+            <LineEdit Name="_callShuttleTime" Text="4:00" PlaceHolder="m:ss" HorizontalExpand="True" SizeFlagsStretchRatio="4"/>
```
sleepyyapril: Generally, UI elements should be named according to what they are or do rather than like a normal variable. In this case, I'd recommend CallShuttleDelayEdit or similar.

Content.Client/_DV/Administration/UI/Tabs/AdminTab/DVAdminShuttleWindow.xaml
```diff
@@ -0,0 +1,14 @@
...
+            <Control HorizontalExpand="True" SizeFlagsStretchRatio="1"/>
+            <CheckBox Name="_callShuttleRecallable" Text="{Loc 'admin-shuttle-recallable'}" HorizontalExpand="True" SizeFlagsStretchRatio="4" Pressed="True"/>
+        </BoxContainer>
+        <cc:CommandButton Command="callshuttle 4:00 False" Name="_callShuttleButton" Text="{Loc 'comms-console-menu-call-shuttle'}" HorizontalAlignment="Center"/>
```
sleepyyapril: At this point, switch this to a normal button and use the LineEdit and CheckBox values.

Content.Client/_DV/Administration/UI/Tabs/AdminTab/DVAdminShuttleWindow.xaml.cs
```diff
@@ -0,0 +1,31 @@
...
+            IoCManager.InjectDependencies(this);
+
+            _callShuttleTime.OnTextChanged += OnChange;
+            _callShuttleRecallable.OnToggled += OnChange;
+        }
+
+        private void OnChange(EventArgs obj) {
+            var loc = IoCManager.Resolve<ILocalizationManager>();
+            _callShuttleButton.Disabled = !TimeSpan.TryParseExact(_callShuttleTime.Text, ContentLocalizationManager.TimeSpanMinutesFormats, loc.DefaultCulture, out _);
+            _callShuttleButton.Command = $"callshuttle {_callShuttleTime.Text} {!_callShuttleRecallable.Pressed}";
```
sleepyyapril: Changing it from a CommandButton would let you get rid of this and just check it in an OnPressed.

Content.Server/Administration/Commands/ShuttleCommands.cs
```diff
@@ -14,15 +14,35 @@ public sealed class CallShuttleCommand : LocalizedEntityCommands
 
         public override void Execute(IConsoleShell shell, string argStr, string[] args)
         {
-            // ReSharper disable once ConvertIfStatementToSwitchStatement
-            if (args.Length == 1 && TimeSpan.TryParseExact(args[0], ContentLocalizationManager.TimeSpanMinutesFormats, LocalizationManager.DefaultCulture, out var timeSpan))
-                _roundEndSystem.RequestRoundEnd(timeSpan, shell.Player?.AttachedEntity, checkCooldown: false);
+            // DeltaV Start - Add support for cantRecall argument to the callshuttle command.
```
sleepyyapril, line 17: The number of changes is a bit extreme. You should probably make a separate command for norecall and then use whether the Recallable checkbox is pressed to either send the normal callshuttle command or the no recall one. I would absolutely love to see a better callshuttle command upstream, though!

Content.Server/_DV/KeycardAuthenticationDevice/DVStationKeycardAuthenticationDeviceSystem.cs
```diff
@@ -28,6 +28,9 @@ protected override void Mayday(Entity<DVStationKeycardAuthenticationDeviceCompon
     {
         base.Mayday(station);
 
+        if (_alertLevel.GetLevel(station) == "zeta")
```
sleepyyapril: Put zeta in a readonly ProtoId at the top of the file just to avoid literals.

Content.Server/_DV/Communications/DVCommunicationsConsoleSystem.cs
```diff
@@ -72,6 +76,23 @@ private void OnRoundEndChanged(RoundEndSystemChangedEvent ev)
...
+            var shuttleCutoffExpected = expected.TotalSeconds * recallThreshold;
+            var shuttleCutoff = left.TotalSeconds - shuttleCutoffExpected;
+            if (shuttleCutoff > 0) {
+                _thresholdUpdateTokenSource = new CancellationTokenSource();
+                Timer.Spawn(TimeSpan.FromSeconds(shuttleCutoff), () =>
+                {
+                    OnRoundEndChanged(null);
+                }, _thresholdUpdateTokenSource.Token);
+            }
+        }
```
sleepyyapril, line 110: One of two things should probably happen here:
1. Find a different way of doing this to avoid a Timer.Spawn, calling the method again, and changing the event method's parameters.
2. Leave this part out and I can fix it in a separate PR. If you'd like to go this route, let me know.

sleepyyapril (changes requested):

PR #4078, opened 2025-07-24
Psionic Eruption

TehFlaminTaco:
## About the PR

Added **Psionic Eruption**, a power which detonates the user.

## Why / Balance

As of writing, all Psionic Powers are purely beneficial, and Psionic Manti rarely have any reason to 'cure' Psions. The added ability gives both the user and station crew a reason to be wearisome of psionics and treat them with the care and respect the lore commands of them.

## Technical details

Adds a single new Psionic Power with 0.1 weight, **Psionic Eruption.**
When used, the Psion begins to shake violently, sparking, before exploding.
They can be stopped during this by standard stun or dispel behaviours (Such as pushing them or hitting them hard enough).
It typically takes 10 seconds for the Psion to detonate.

Possessing the ability causes periodical annoy messages for the Psion, prompting them to not forget about it. Additionally, a small amount of glimmer is generated each time. The frequency of these messages is determined by the Glimmer Tier.

The size of the explosion scales with Glimmer Tier, scaling from Media 3 at minimal, to Media 4 at dangerous.
Should the Glimmer Tier reach Critical (>900) however, the behaviour changes.
 * The detonate time increases to 30 seconds
 * The sparking whilst charging may shoot lightning
 * Explosion size as per Media 5

Glimmer reaching above 900 is rare enough that this should not be disruptive.

## Media

1.  notready
2.  icon
3.  smallboom
4.  biggerboom
5.  image

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

:cl:
- add: Added a new power, Psionic Eruption

PR #6501, opened 2026-09-11
Psionic Eruption 2: DAGD Boogaloo

EmberAstra:
## About the PR
Psionic Eruption (#4078) is back!
This adds a new kit containing an empty implanter and a crystal for Psionic Eruption, available in the uplink for 2TC for DAGD traitors only (mutually exclusive with the other DAGD implanters). It's also now _massively_ more powerful, especially at maximum glimmer.
(inspired by this WYCI thread: https://discord.com/channels/968983104247185448/1545637756712259604 )

## Why / Balance
Eruption is kind of a fun idea but as a random crew power it was... either just go to epi and get it fixed immediately, or self antag bait. So let's just make it _evil_.
The new maximum explosion size is modeled after the prober's explosion, everything below that was kinda eyeballed to be very weak until 500, then pretty strong until 900 where it jumps to maximum power.
It also now increases glimmer on use by a small amount (0 to 50) rather than reducing it.
Otherwise the behavior is pretty much the same as it was before, including slowly ticking up glimmer over time, the long do-after before the explosion, and shooting lightning bolts randomly during the charge up while at max glimmer.

Hopefully it'll create a *serious* psionic threat that the Mantis can hunt down, will give antags a reason to mess with glimmer, and give Epi a reason to not just sit at 900 glimmer all shift every shift.

## Technical details
YAML ops and editing the explosion values in the system itself.
I'm not 100% sure on the explosion values, more than willing to tone them up/down as needed.

## Media
New maximum explosion size.
Untitled
Untitled2

## Requirements

## Licensing

**Changelog**

:cl:
- add: Psionic Eruption returns... as an exclusive implant for DAGD traitors! Spike glimmer up and turn your favorite department (and yourself) into a crater.

### DIFF
--- a/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
+++ b/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
@@ -50,7 +50,7 @@ public sealed class PsionicEruptionSystem : BasePsionicPowerSystem<PsionicErupti
         if (!_player.TryGetSessionByEntity(power, out var session))
             return;
 
-        _eui.OpenEui(new EruptionWarningEui(), session);
+        // _eui.OpenEui(new EruptionWarningEui(), session); // It's DAGD-only now, no need for a warning. You know what will happen.
         power.Comp.NextAnnoy = Timing.CurTime + TimeSpan.FromSeconds(60); // Minute grace period
     }
 
@@ -181,14 +181,14 @@ public sealed class PsionicEruptionSystem : BasePsionicPowerSystem<PsionicErupti
 
         int boom = _glimmer.GetGlimmerTier(_glimmer.Glimmer) switch
         {
-            GlimmerTier.Minimal => 2,
-            GlimmerTier.Low => 3,
-            GlimmerTier.Moderate => 4,
-            GlimmerTier.High => 8,
-            GlimmerTier.Dangerous => 12,
-            GlimmerTier.Critical => 32,
+            GlimmerTier.Minimal => 10,
+            GlimmerTier.Low => 30,
+            GlimmerTier.Moderate => 60,
+            GlimmerTier.High => 120,
+            GlimmerTier.Dangerous => 600,
+            GlimmerTier.Critical => 3000,
             _ => 0
         };
-        _explosion.QueueExplosion(pos, ExplosionSystem.DefaultExplosionPrototypeId, boom, 1, 5, psionic, maxTileBreak: 0);
+        _explosion.QueueExplosion(pos, ExplosionSystem.DefaultExplosionPrototypeId, boom, 4, 75, psionic, maxTileBreak: 5);
     }
 }
--- a/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
+++ b/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
@@ -189,6 +189,6 @@ public sealed class PsionicEruptionSystem : BasePsionicPowerSystem<PsionicErupti
             GlimmerTier.Critical => 3000,
             _ => 0
         };
-        _explosion.QueueExplosion(pos, ExplosionSystem.DefaultExplosionPrototypeId, boom, 4, 75, psionic, maxTileBreak: 5);
+        _explosion.QueueExplosion(pos, ExplosionSystem.DefaultExplosionPrototypeId, boom, 5, 75, psionic, maxTileBreak: 5);
     }
 }
--- a/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
+++ b/Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
@@ -26,7 +26,6 @@ public sealed class PsionicEruptionSystem : BasePsionicPowerSystem<PsionicErupti
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly AudioSystem _audio = default!;
     [Dependency] private readonly DoAfterSystem _doAfter = default!;
-    [Dependency] private readonly EuiManager _eui = default!;
     [Dependency] private readonly ExplosionSystem _explosion = default!;
     [Dependency] private readonly GibbingSystem _gibbing = default!;
     [Dependency] private readonly GlimmerSystem _glimmer = default!;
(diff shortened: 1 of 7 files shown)


EmberAstra:
Test fail seems unrelated?

Creb206:
maybe a bit more than TWO TC for a glorified macro bomb

Creb206:
Also, yes, the test failed because github failed. Nothing related to the PR

EmberAstra:
I kept the price low because it requires a LOT more setup than a macrobomb (full power is only at 900+ glimmer, it jumps down to 1/5th that if below), so kept the same price as the DAGD Microbomb.
Can up it if needed, however.

Creb206:
I don't know, as is it kind of just feels like a more gimmicky toned down macrobomb which I guess is fine, I'm just not sure the benefits would justify going out of your way to get glimmer up, not to mention there's not much you can do to raise glimmer if you're not in epi.

rebe83:
Glimmer objective 2: electric boogaloo
But this at least has a reason for it I guess

rebe83:
What does "bigger radius" and "less damage" mean
5s means anyone is walking away from it? Why increase doafter AND radius?
And this is only something useable if DAHD and high glimmer. "Worst case" epi raises glimmer themself, one guy really can't so much to glimmer besides steal the anomaly generator or collect glimmer mites.
Mostly useless item it's just to explode. If you're afraid of people walking up to a crowd and exploding well
1. They are dagd
2. They are dead and can't do anything else after this so it was a waste
Private discussion for no reason
Bad take CR try again

EmberAstra:
The doafter previously was 10 seconds normally and _30_ at maximum glimmer, but without any stun effect, to clarify. So that did get reduced.

For the rest, uhm, I'm not good with precise number balancing so doing as CR says.
(need to figure out how to do the cancel on crit part)

EmberAstra:
Testing out things. Solved the technical side of things, so mostly just balancing the explosion.

For now tried power 2000, slope 2, tile intensity 100, break tiles 2. Struggles against reinforced walls but otherwise it's very wide and devastates "standard" rooms. Damage is around 300 at point blank from testing.
Damn these numbers are weird to adjust properly.

https://github.com/user-attachments/assets/5e6e0658-5178-4191-8c45-a2965be1831c

Content.Shared/DoAfter/SharedDoAfterSystem.cs
```diff
@@ -143,6 +145,24 @@ private void OnDoAfterHandleState(EntityUid uid, DoAfterComponent comp, ref Comp
...
+            {
+                InternalCancel(doAfter, comp);
+                dirty = true;
+            }
+        }
+
+        if (dirty)
+            Dirty(uid, comp);
+    }
+    // End DeltaV Addition
```
sleepyyapril: Partial class

Content.Shared/DoAfter/DoAfterArgs.cs
```diff
@@ -136,6 +136,14 @@ public sealed partial class DoAfterArgs
...
     public FixedPoint2 DamageThreshold = 1;
 
+    // Begin DeltaV Addition
+    /// <summary>
+    ///     Whether entering critical state will cancel the DoAfter.
+    /// </summary>
+    [DataField]
+    public bool BreakOnCritical;
+    // End DeltaV Addition
+
```
sleepyyapril: Partial class

Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
```diff
@@ -181,14 +190,14 @@ private void OnDoAfter(Entity<PsionicEruptionPowerComponent> psionic, ref Psioni
...
-            GlimmerTier.Dangerous => 12,
-            GlimmerTier.Critical => 32,
+            GlimmerTier.Minimal => 10,
+            GlimmerTier.Low => 30,
+            GlimmerTier.Moderate => 60,
+            GlimmerTier.High => 120,
+            GlimmerTier.Dangerous => 500,
+            GlimmerTier.Critical => 2000,
             _ => 0
         };
```
sleepyyapril: While you're here, you might want to move this to the component so its less hardcoded?

Content.Shared/DoAfter/DoAfterArgs.cs
sleepyyapril: Also you probably should make this generic, such as BreakOnMobStates

sleepyyapril (changes requested):

Content.Shared/DoAfter/DoAfterArgs.cs
EmberAstra: Done!

Content.Shared/DoAfter/SharedDoAfterSystem.cs
EmberAstra: Done!

Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
EmberAstra: Done, tho not 100% sure if there's a more elegant/proper way to do this after some code searching.

Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
```diff
@@ -50,7 +52,7 @@ protected override void OnPowerInit(Entity<PsionicEruptionPowerComponent> power,
         if (!_player.TryGetSessionByEntity(power, out var session))
             return;
 
-        _eui.OpenEui(new EruptionWarningEui(), session);
+        // _eui.OpenEui(new EruptionWarningEui(), session); // It's DAGD-only now, no need for a warning. You know what will happen.
```
ShepardToTheStars: Just remove.
```suggestion
```

ShepardToTheStars (commented):
Do we need the DoAfter partials? Does it not auto-cancel on MobState changed? or does it not cancel if `RequireCanInteract` is set to `false`?

Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
EmberAstra: Done. Also. yeah with RequireCanInteract false it continues even if you go critical. In other places we've seen it used it's usually paired with BreakOnDamage, but that only supports either any damage at all or a single hit over a certain threshold.

Content.Server/_DV/Psionics/Systems/PsionicPowers/PsionicEruptionPowerSystem.cs
sleepyyapril: I would have made it a dictionary in the component itself

EmberAstra: Done!

PR #6518, opened 2026-09-17
Show tip about LOOC usage

CarolCuchara:
## About the PR

Shows a tip about LOOC usage once per session.

## Why / Balance

LOOC abuse is a big problem. 

## Technical details

Added a server side system that shows the tip once per session. Tracks clients who haven seen the hint in a hash set, so it will be only shown once per session. The tip will be force shown, even if the player has hints disabled. The "dont show this again" checkbox is there, but has no effect, it will be shown again the next session.

## Media

Capture

## Requirements

## Licensing

**Changelog**

:cl:
- add: A tip about LOOC etiquette will be shown once per session!


CMuncey:
If the don't show again button does nothing, why have it there in the first place? That will be extremely annoying for users and lead to several false bug reports and ahelps for an intended feature.

I don't understand why you would disable the don't show again button in the first place, if people see the tip with the big red text the first time and then decide to say "yeah it's a xorg round, my buddy told me he's a xorg" the next round, they know what they're doing just get admin intervention.

CarolCuchara:
The button is there because its part of the layout of the tip window. It could be removed, but would require changing the layout of the window. I didnt want to touch that, but i can do that.

I belive the tip should be shown more than once, but i understand that it might be annoying if it is once per session.

I will take a look at the window and a way to show the tip less, maybe once per week.

Woljif:
I'm fine with it popping up once per round because talking in LOOC should be uncommon enough that the little pop up doesn't create too much friction. Also, and I'm sure we all know someone, some people do need reminders, especially when they're mad. An ounce of prevention is better than a pound of cure, and I am certain this popup would help ground people who are _about_ to mald in LOOC but havent said anything yet, as seeing the popup is an instant reminder that what they're about to do is against the rules. So I find the argument of "just ahelp them instead" kind of fundamentally uninterested

IAmNotGray:
Howdy there. I'm the guy whose supposed to handle this as the Head Game Administrator. **I'm still on break and won't be handling the PR in full as such until this weekend,** but I would like that the tip only be shown once for a multitude of reasons. The big main reason is simply because I don't want to treat the community as if they can't read. You guys _can_ and _are_ expected to be familiar with the rules and I really do think that once is enough. It'll get incredibly annoying especially for those familiar with the rules.

I'd also like that the wording of the red text to be changed to point to Rule B1.2, which is our rule on LOOC/OOC usage. There are valid "out-of-context" use cases for LOOC such as roleplay coordination and the like, and that would be better reflected by pointing at the rule. It would also make it easier to maintain this feature going forward, since we would only have to change the rule itself and not this popup as well.

Thank you!

IAmNotGray:
I will discuss this further with Senior Administration and see how everyone else feels about something like this, so feel free to wait until a decision has been made to change what I'm currently requiring from this.

ShepardToTheStars:
So, just so I’m understanding it, this tip will be forcibly shown every time someone uses LOOC if they connect with a new session? Or on server restart?

CarolCuchara:
@ShepardToTheStars its once per connection session to the server. So if you reconnect, its back. 

I have been looking at other ways and been thinking about the frequency of the tip. Once per session is definitly to much. My pitch for the future would be to rewrite it to show a tip every one or two weeks. That would a lot less, while still showing up from time to time as a reminder. I agree that players need no babysitting, but a reminder from time to time is acceptable. 

The text will definitly be reworded, i myself am not happy with its wording and my grammar. If this gets approved, maybe some one from Administration who is a native speaker can provide a better text.

Content.Server/_DV/GameTicking/Rules/LoocTipSystem.cs
```diff
@@ -0,0 +1,53 @@
+using Content.Server._DV.Tips;
+using Content.Shared._DV.Tips;
+using Content.Shared.Chat;
+using Robust.Server.Player;
+using Robust.Shared.Enums;
+using Robust.Shared.Network;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server._DV.GameTicking.Rules;
```
sleepyyapril, line 10: Move to shared

sleepyyapril (changes requested):

PR #6522, opened 2026-09-18
unstable core refractor construction bug patch

Apersos-coll:
## About the PR

Fixes bug that lets you only need core to craft full gun
## Why / Balance

cause bugs are bad
## Technical details

changed the construction graph to resemble more of what the shiv does so it can do both clicky craft and menu craft without breaking.
## Media

bufix
## Requirements

## Licensing

**Changelog**

:cl:
- fix: you now actually need to get all components to craft the unstable core refractor.


SirWarock (changes requested):
Rename the other two rsi folders to follow the naming_convention.rsi

SirWarock (approved):
works

PR #6531, opened 2026-09-20
makin ambu a cryo chem

keekee38:
## About the PR

made ambu a cryo chem

## Why / Balance

zombies is a kinda hit or miss sorta antag, ambu is really easy to get
i wanted to also change the recipe but idunno how would be enjoyable so maybe we can talk about that too idk
was thinking making it more expensive in terms of ratio of ingredients to ambu but maybe also harder to aquire ingredients idk

also like this means cryo would need power but in zombies if power is killed chem usually gets a gen so there's not much different there imho

## Technical details

yml

## Media

https://github.com/user-attachments/assets/fcade611-823a-4139-aa8e-d55b37be8c52

## Requirements

## Licensing

**Changelog**

:cl:
- tweak: ambu is now a cryo chem


Creb206:
The amount of logistical challenge added by needing to put everyone infected into cryo would make this way too cumbersome to be realistic in a round, not to mention II could very easily sabotage cryo or cryo could literally just not be set up and you'd be screwed completely.

keekee38:
im ngl this is half to start a discussion becuase as is ambu is too easy imho

keekee38:
could also make it cold but not quite cryo temp
kitchen freezer stays winnin
or somethin

CMuncey:
> could also make it cold but not quite cryo temp kitchen freezer stays winnin or somethin

The kitchen chest freezer doesn't make things cold if that's what you mean

rebe83:
> > could also make it cold but not quite cryo temp kitchen freezer stays winnin or somethin
> 
> The kitchen chest freezer doesn't make things cold if that's what you mean

The actual walk in freezer for kitchen with a freezer machine that makes it cold

rebe83:
> The amount of logistical challenge added by needing to put everyone infected into cryo would make this way too cumbersome to be realistic in a round, not to mention II could very easily sabotage cryo or cryo could literally just not be set up and you'd be screwed completely.

Oh no zombies could win by attacking medical oh noooooo oh goddd
Can't epi print more of these

Creb206:
Honestly you could take a page from monolith's book with chimera and make it so that you have to centrifuge zombie blood into maybe just normal blood and romerol and then you use romerol instead of zombie blood to make ambuzol, also has the added benefit of letting II do silly things with an actual way to get romerol considering they can use a paper centrifuge.

SalDragonsNOA:
> Honestly you could take a page from monolith's book with chimera and make it so that you have to centrifuge zombie blood into maybe just normal blood and romerol and then you use romerol instead of zombie blood to make ambuzol, also has the added benefit of letting II do silly things with an actual way to get romerol considering they can use a paper centrifuge.

I like this idea. It would slow things up a bit, but that would give a limited amount of the stuff and act as a limiter. You can only make so much ambu at a single time.

The issue I have with it being a Cyro chem is you can only treat one person at a time and have to manage the symptoms of a lot of sick people with Dylo. I'm not even sure all the stations have Cyro. If they do, it might be one pod.
So, cyro could be a bit too limiting in my concern.

EmberAstra:
The whole 'centrifuging zombie blood to get the chem you need' thing is actually in upstream too, now. As well as other ambuzol changes.
https://github.com/space-wizards/space-station-14/pull/45439

